PR #22, opened 2018-10-10
 fix: update to post-v0.9 Transaction wire format

mvines:
cc: https://github.com/solana-labs/solana/issues/1418


mvines:
Excellent.  This PR failed because the fix for https://github.com/solana-labs/solana/pull/1464 has not be published to the edge docker image yet.

PR #30, opened 2018-10-22
feat: Add loader_program factory class

jackcmay:


mvines:
This PR should be to the `dev` branch not `master`.

On master there's already https://github.com/solana-labs/solana-web3.js/blob/dev/src/loader.js though

PR #34, opened 2018-10-23
Add BpfLoader

jackcmay:


flow-typed/file-system.js
```diff
@@ -0,0 +1,4 @@
+declare module 'file-system' {
```
mvines: Definitely don't want this module.   Instead use https://www.npmjs.com/package/mz:
* https://github.com/solana-labs/example-tictactoe/blob/master/package.json#L46
* https://github.com/solana-labs/example-tictactoe/blob/master/flow-typed/npm/mz_vx.x.x.js

src/bpf-loader.js
```diff
@@ -0,0 +1,72 @@
+// @flow
+
+import * as fs from 'file-system';
+import * as elfy from 'elfy';
+
+import { Account, PublicKey, Loader, SystemProgram } from '.';
+import { sendAndConfirmTransaction } from './util/send-and-confirm-transaction';
+import type { Connection } from '.';
```
mvines: nit: less spaces around the import curlies:
```js
import {Account, PublicKey, Loader, SystemProgram} from '.';
import {sendAndConfirmTransaction} from './util/send-and-confirm-transaction';
import type {Connection} from '.';
```

src/bpf-loader.js
```diff
@@ -0,0 +1,72 @@
...
+   * @param programName Name of the BPF program
+   */
+  static async load(
+    connection: Connection,
+    owner: Account,
+    programName: string,
+  ): Promise<PublicKey> {
+    const programAccount = new Account();
+
+    var file = fs.readFileSync(programName);
```
mvines: never use `var` (except for if you *really* mean it, and you very rarely do).  Use `const` by default, or `let` for mutable values (yeah `let` is immutable in Rust, you're welcome!)

mvines: Instead of `fs.readFileSync`:

```js
import fs from 'mz/fs';
const data = await fs.readFile(programName);
```

src/bpf-loader.js
```diff
@@ -0,0 +1,72 @@
...
+    var elf = elfy.parse(file);
+    var section = elf.body.sections.find(function (sect) {
```
mvines: Closure syntax is nicer:
```
const section = elf.body.sections.find(section => section.name === '.text.entrypoint');
```

Also note the `===`.  You want to use `===` generally to ensure the VM isn't doing a type conversion behind your back

src/bpf-loader.js
jackcmay: /Users/jack/sandbox/git/solana-web3.js/src/index.js
  3:25  error  Parse errors in imported module './bpf-loader': Line 23: await is a reserved word

  21 |   static parse(programName: string) {
  22 | 
> 23 |     const data = await fs.readFile(programName);

mvines: `static async parse(programName: string) ...`

functions that use `await` need to be declared as `async` or you get that terrible error message

src/loader.js
```diff
@@ -43,28 +43,30 @@ export class Loader {
...
-    userdataLayout.encode(
-      {
-        instruction: 0, // Load instruction
-        offset,
-        bytes,
-      },
-      userdata,
-    );
+    let i,j;
+    const chunk_size = 256; // Size of chunk, just needs to fit into tx
```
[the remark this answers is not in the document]
mvines: oh no snake case please.  We're camelCase here

src/loader.js
```diff
@@ -43,28 +43,30 @@ export class Loader {
...
+    let userdata = Buffer.alloc(chunk_size + 16);
+    for (i = 0, j = bytes.length; i < j; i += chunk_size) {
```
mvines: I think we can replace ` for (i = 0, j = bytes.length; i < j; i += chunk_size) {` with

```js
const offset = 0;
while (bytes.length > 0) {
  const slice = bytes.slice(0, chunkSize);

  // send `slice`

  offset += chunkSize;
  bytes = bytes.slice(chunkSize);
}
```

test/bpf-loader.test.js
```diff
@@ -0,0 +1,34 @@
...
+  const connection = new Connection(url);
+  const from = await newAccountWithTokens(connection);
+  const noopProgramId = await BpfLoader.load(connection, from, 'test/bin/noop_c.o');
+  const noopTransaction = new Transaction({
+    fee: 0,
+    keys: [from.publicKey],
+    programId: noopProgramId,
+  });
+  const signature = await connection.sendTransaction(from, noopTransaction);
+  await expect(connection.confirmTransaction(signature)).resolves.toBe(true);
```
mvines: Use `sendAndConfirmTransaction` here instead of `connection.confirmTransaction()` for a more robust test

test/bpf-loader.test.js
```diff
@@ -0,0 +1,34 @@
+// @flow
```
mvines: can you add a `test/bin/noop_c.txt` file that describes where this blob came from and how it was generated, to help out the next person that tries to debug why it's failing in the future (like when we make the next entrypoint ABI change)

src/loader.js
```diff
@@ -43,28 +43,30 @@ export class Loader {
       BufferLayout.u32('bytesLengthPadding'),
       BufferLayout.seq(
         BufferLayout.u8('byte'),
-        BufferLayout.offset(BufferLayout.u32(), -8),
-        'bytes'
-      ),
+        BufferLayout.offset(BufferLayout.u32(), -8),'bytes'),
```
mvines: nit: add one space before `'bytes'`

src/loader.js
```diff
@@ -43,28 +43,30 @@ export class Loader {
...
+      );
 
-    const transaction = new Transaction({
-      fee: 0,
-      keys: [program.publicKey],
-      programId: this.programId,
-      userdata,
-    });
-    await sendAndConfirmTransaction(this.connection, program, transaction);
+      const transaction = new Transaction({
```
mvines: Here's how to update to the latest changes:
```diff
-    const transaction = new Transaction({
-      fee: 0,
+    const transaction = new Transaction().add({
```

src/loader.js
```diff
@@ -32,38 +32,45 @@ export class Loader {
...
+        },
+        userdata,
+      );
+
+      const transaction = new Transaction().add({
+        keys: [program.publicKey],
+        programId: this.programId,
+        userdata,
+      });
+      await sendAndConfirmTransaction(this.connection, program, transaction);
```
mvines: oh we should run all these load transactions in parallel.   Here's how:
```js
const transactions = [];
while (array.length > 0) {
  .
  .
  .
  transactions.push(sendAndConfirmTransaction(this.connection, program, transaction));
  .
  .
}
await Promise.all(transactions);
```
(follow-up is fine for this if you prefer)

mvines: (this needs to be moved into the while loop for ^^ to work though: `let userdata = Buffer.alloc(chunkSize + 16);`)

PR #39, opened 2018-10-24
feat: Submit program load transactions in parallel

jackcmay:


src/loader.js
```diff
@@ -49,6 +49,7 @@ export class Loader {
     let userdata = Buffer.alloc(chunkSize + 16);
     let offset = 0;
     let array = data;
+    const transactions= [];
```
mvines: nit: space between `s=`

src/loader.js
```diff
@@ -66,11 +67,12 @@ export class Loader {
         programId: this.programId,
         userdata,
       });
-      await sendAndConfirmTransaction(this.connection, program, transaction);
+      transactions.push(await sendAndConfirmTransaction(this.connection, program, transaction));
```
mvines: don't `await` here.   We want to simply collect all the `Promise`s into the transactions array here, not `await` for them to resolve.

jackcmay: Good point!

jackcmay:
@mvines Looks like token_program failed:

`FAIL test/token-program.test.js (9.145s)`

Please review and I'll squash and merge while trying to keep the sem version.

src/loader.js
```diff
@@ -46,12 +46,12 @@ export class Loader {
...
 
     const chunkSize = 256;
-    let userdata = Buffer.alloc(chunkSize + 16);
     let offset = 0;
     let array = data;
+    const transactions = [];
     while (array.length > 0) {
       const bytes = array.slice(0, chunkSize);
-
+      let userdata = Buffer.alloc(chunkSize + 16);
```
mvines: nit: const userdata = ...

mvines: hey, why no const?

PR #49, opened 2018-10-29
feat: add bpf-c-noop example

mvines:


jackcmay:
Why not leave them in the same directory structure as solana?

mvines:
`examples/bpf-c-noop/...` seems much nicer than `examples/bpf/c/noop/...` to me.  I'd even be happy with `examples-bpf-c-noop/...` actually.   Deep subdirectories are the worse.

PR #58, opened 2018-11-03
fix: adapt to account info RPC format change

mvines:
Will fail until https://github.com/solana-labs/solana/pull/1699 lands


mvines:
Too soon, but this issue is tracked at https://github.com/solana-labs/solana/issues/1781

PR #61, opened 2018-11-07
chore: remove elfy from rollup

jackcmay:
- elfy has moved to `devDependencies`
- add prettier badge for fun


jackcmay:
@mvines, nits, and stuff but CI fails for what seems like an unrelated issue, ok to merge?

PR #62, opened 2018-11-07
Return signature from transfer transactions

CriesofCarrots:


mvines:
lgtm.  We use https://semantic-release.gitbook.io/semantic-release/ in this project though, so your commit title needs to look something like this `fix: return signature from transfer transactions`

PR #82, opened 2018-12-07
code highlight style change

manuel-calavera:
found another issue with code style. you can see it here (black colored code):

https://solana-labs.github.io/solana-web3.js/


mvines:
nice, thanks!

PR #98, opened 2018-12-10
fix: Add network parameter to localnet run

CriesofCarrots:


bin/localnet.sh
```diff
@@ -76,13 +76,18 @@ up)
   fi
   [[ $channel = edge || $channel = beta ]] || usage "Invalid channel: $channel"
 
+  if [[ -n $2 ]]; then
```
mvines: I think I'd prefer this to be a named argument rather than positional, to keep the argument namespace open for future expansion.

Something cheesy like this would be enough for now:
```bash
if [[ $2 = -n ]]; then
  [[ -n $3 ]] || usage "Invalid -n argument" 
  network="$3"
fi
```
Also please document the arg in `usage`

bin/localnet.sh
```diff
@@ -76,13 +77,19 @@ up)
...
+  fi
+
   (
     set -x
     RUST_LOG=${RUST_LOG:-solana=warn,solana_bpf=info,solana_jsonrpc=info,solana::rpc=info,solana_fullnode=info,solana::drone=info,solana::bank=info,solana::banking_stage=info,solana::system_program=info}
 
     docker run \
       --detach \
       --name solana-localnet \
+      --network "$network" \
```
mvines: This won't cause a problem when `$network` is empty?

CriesofCarrots: No, it actually appears to behave the same as if you don’t specify a network

PR #203, opened 2019-03-04
fix: catch up to recent blockhash RPC API changes

mvines:


mvines:
Red due to #3087, but landing anyway since it's forward progress

PR #207, opened 2019-03-05
fix: adapt to slower slot rate

mvines:


src/timing.js
```diff
@@ -0,0 +1,7 @@
+// @flow
+
+// These constants should match the values in
+// https://github.com/solana-labs/solana/blob/master/sdk/src/timing.rs
+export const NUM_TICKS_PER_SECOND = 10;
+export const DEFAULT_TICKS_PER_SLOT = 80;
```
CriesofCarrots: Would it make sense to get these from the cluster via rpc, instead of having to remember to keep these constants updated? Or is that too much i/o overhead?

mvines: Yeah I think I want to get this via RPC eventually.   Not quite ready to build this yet, I'd like to see the fullnode software stack churn less for a while first.

PR #218, opened 2019-03-08
fix: Plumb program-based subscriptions

CriesofCarrots:


module.flow.js
```diff
@@ -76,6 +76,11 @@ declare module '@solana/web3.js' {
       callback: AccountChangeCallback,
     ): number;
     removeAccountChangeListener(id: number): Promise<void>;
+    onProgramAccountChange(
+      programIdL PublicKey,
```
mvines: looks like a typeo here, unfortunately we don't currently test module.flow.js in CI and rely on downstream packages to notify of bustage 🙃

CriesofCarrots: 😝

CriesofCarrots:
@mvines I didn't realize the jest update would be critical for this PR passing CI. If you want us to merge that first and then have me rebase this, just let me know

mvines:
Na just land it all here.

CriesofCarrots:
@mvines , this good to go, then?

PR #290, opened 2019-04-23
feat: add getSlotLeader()/getClusterNodes()

mvines:
cc: https://github.com/solana-labs/solana/pull/3940


mvines:
NB: CI is expected to fail until the solana docker image is updated with the implementation for these new JSON RPC methods

PR #321, opened 2019-05-24
Update transaction wire format, toward credit-only accounts

CriesofCarrots:
Also:
- Add preliminary handling for designating credit-only accounts (currently, solana handles credit-only accounts the same as credit-debit accounts; more tweaking may be needed here when this changes)
- Repair work to get CI passing again


CriesofCarrots:
These test failures only happen when run against localnet (using docker edge image of solana), not when run against a local solana cluster started manually with `.run.sh`.
Will examine more after next solana docker publish...

src/transaction.js
```diff
@@ -109,7 +113,7 @@ export class Transaction {
   /**
    * A recent transaction id.  Must be populated by the caller
    */
-  recentBlockhash: ?Blockhash;
+  recentBlockhash: Blockhash;
```
mvines: `Blockhash|null` perhaps?  There won't always be a value here I think

mvines:
Looks great overall.  Man WASM/Rust would be nice one day :)

CriesofCarrots:
@mvines , okay to merge?
Sorry it ended up to be such a motley collection of commits...
Incidentally, it looks like example-tictactoe and example-messagefeed will both need to run on `dist: xenial` in order to successfully build bpf in CI. Not sure if that will have any negative ramifications.

mvines:
Awesome, thanks!
I don't see any issue with switching the two examples over to xenial

PR #342, opened 2019-06-12
Add binding for getEpochVoteAccounts

mvines:


mvines:
Will fail until https://github.com/solana-labs/solana/pull/4663 lands

PR #366, opened 2019-06-22
Fix rollup dependency conflict

### LINKED ISSUE #327: Debug why rollup 1.12.4 fails to generate lib/index.iife.js in CI
I had to back out rollup 1.12.4 via 3f011bf306aa7008bcf4b48235ab3549944d9427 and a51f041e9a68caf0b8ad716735cd625f2db0174a as it `npm run build` in Travis was not generating `lib/index.iife.js` anymore.  

Debug, fix, then let's move back to the latest rollup

jstarry:
Fixes #327

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,7 +115,7 @@
     "rollup-plugin-json": "^4.0.0",
     "rollup-plugin-node-builtins": "^2.1.2",
     "rollup-plugin-node-globals": "^1.2.1",
-    "rollup-plugin-node-resolve": "4.2.3",
+    "rollup-plugin-node-resolve": "5.0.0",
     "rollup-plugin-replace": "2.2.0",
     "rollup-plugin-terser": "^5.0.0",
     "semantic-release": "^15.9.16",
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
     "rimraf": "2.6.3",
     "rollup": "1.15.2",
     "rollup-plugin-babel": "^4.3.2",
-    "rollup-plugin-commonjs": "9.3.4",
+    "rollup-plugin-commonjs": "^10.0.0",
     "rollup-plugin-json": "^4.0.0",
     "rollup-plugin-node-builtins": "^2.1.2",
     "rollup-plugin-node-globals": "^1.2.1",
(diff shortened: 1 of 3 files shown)


mvines:
Darn!  
```
$ test -r lib/index.iife.js
The command "test -r lib/index.iife.js" exited with 1.
```
(from https://github.com/solana-labs/solana-web3.js/blob/master/.travis.yml#L31)

jstarry:
@mvines found the issue: https://github.com/rollup/rollup/pull/2961

rollup.config.js
```diff
@@ -108,4 +109,4 @@ function generateConfig(configType) {
   return config;
 }
 
-export default [generateConfig('node'), generateConfig('browser')];
+export default [/*generateConfig('node'), */generateConfig('browser')];
```
mvines: Why this change?

package.json
```diff
@@ -109,13 +110,13 @@
     "marked": "^0.6.0",
     "prettier": "^1.14.3",
     "rimraf": "2.6.3",
-    "rollup": "1.15.2",
+    "rollup": "git://github.com/jstarry/rollup.git#fix-resolved-id-result-bug",
```
mvines: I'd prefer to wait for your upstream fix to be merged and released rather than merging this PR with this github dependency.  This isn't a super urgent issue though so we can wait to see what happens upstream

jstarry: @mvines Yeah great call! I added the last commit with the message "Check that rollup fix works" ;) I was planning to remove it once I saw if CI passed

rollup.config.js
jstarry: Inadvertent, it helped me speed up my local testing. Removing

jstarry:
@mvines turns out "bug" that I found in rollup is actually a feature, who knew! I closed that RP and dug a bit deeper.

The real issue is that the `rollup-plugin-commonjs` module calls into `rollup-plugin-node-resolve` and was not updated so the two were incompatible. Now that the commonjs plugin is updated, they are compatible and all is harmonious again 😌

jstarry:
🎉 

```
$ test -r lib/index.iife.js
The command "test -r lib/index.iife.js" exited with 0.
```

mvines:
Hehe awesome, I was hoping it was something dumb like that.  The CI failure here is inrelated, I think I know what’s up there

PR #380, opened 2019-06-26
chore: update BPF SDK portions of README

mvines:


src/budget-program.js
```diff
@@ -277,9 +277,7 @@ export class BudgetProgram {
 
       default:
         throw new Error(
-          `A maximum of two conditions are support: ${
-            conditions.length
-          } provided`,
+          `A maximum of two conditions are support: ${conditions.length} provided`,
```
CriesofCarrots: "supported" ... might as well fix now?

CriesofCarrots: Whoops, too slow

mvines: Better luck next time :)

PR #381, opened 2019-06-26
fix: add burnPercent field to FeeCalculator

mvines:
cc: https://github.com/solana-labs/solana/pull/4818


mvines:
cc: @rob-solana

PR #403, opened 2019-07-15
Add validator info deserialization

jstarry:


src/validator-info.js
```diff
@@ -0,0 +1,81 @@
...
+   * Deserialize ValidatorInfo from the config account data.
+   *
+   * @return ValidatorInfo or null if info was not found
+   */
+  static fromConfigData(buffer: Buffer): ?ValidatorInfo {
+    const PUBKEY_LENGTH = 32;
+
+    let byteArray = [...buffer];
+    const configKeyCount = shortvec.decodeLength(byteArray);
+    let configKeys: Array<ConfigKey> = [];
```
mvines: nit: `const configKeys`

src/validator-info.js
```diff
@@ -0,0 +1,81 @@
...
+    for (let i = 0; i < configKeyCount; i++) {
+      const publicKey = new PublicKey(byteArray.slice(0, PUBKEY_LENGTH));
+      byteArray = byteArray.slice(PUBKEY_LENGTH);
+      const isSigner = byteArray.slice(0, 1)[0] === 1;
+      byteArray = byteArray.slice(1);
+      configKeys.push({publicKey, isSigner});
+    }
+
+    if (configKeys.length >= 2 && configKeys[0].publicKey.equals(VALIDATOR_INFO_KEY)) {
+      if (configKeys[1].isSigner) {
```
mvines: I think we'd want to return `configKeys[1]` as well

src/validator-info.js
```diff
@@ -0,0 +1,81 @@
...
+   *
+   * @return ValidatorInfo or null if info was not found
+   */
+  static fromConfigData(buffer: Buffer): ?ValidatorInfo {
+    const PUBKEY_LENGTH = 32;
+
+    let byteArray = [...buffer];
+    const configKeyCount = shortvec.decodeLength(byteArray);
+    let configKeys: Array<ConfigKey> = [];
+    for (let i = 0; i < configKeyCount; i++) {
```
mvines: How about we change this loop to only look for the first 2 keys, we don't really care about if any other keys are present.

src/validator-info.js
jstarry: Yup, it is returned below

mvines: Duh, thanks :)

CriesofCarrots: ? I think it does, as part of ValidatorInfo?

src/validator-info.js
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * @private
+ */
+type ConfigKey = {|
+  publicKey: PublicKey,
+  isSigner: boolean,
+|};
+
+/**
+ * @private
```
mvines: oh wait this isn't private.  Let's add esdoc comments so we get info in https://solana-labs.github.io/solana-web3.js/

src/validator-info.js
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * ValidatorInfo class
+ */
+export class ValidatorInfo {
+  key: PublicKey;
+  info: Info;
+
+  /**
+   * Construct a valid ValidatorInfo
+   */
```
mvines: esdoc comments for the params too please

src/validator-info.js
CriesofCarrots: Whoops, sorry, refresh needed

mvines:
We use semantic-release commit titles in this repo too please

CriesofCarrots:
> I wish there was a test for the decoding, I don't see nice way to do that right now though :-/

My thinking as well. We could use a Config account golden, generated from the CLI as a stop-gap...
I suppose we could have solana's `run.sh` include a ValidatorInfo step. Then web3.js could impl a live test at least

jstarry:
@CriesofCarrots @mvines I added a test for deserializing the validator info, let me know what you think!

src/validator-info.js
```diff
@@ -0,0 +1,100 @@
...
+   *
+   * @param buffer config account data
+   * @return null if info was not found
+   */
+  static fromConfigData(buffer: Buffer): ?ValidatorInfo {
+    const PUBKEY_LENGTH = 32;
+
+    let byteArray = [...buffer];
+    const configKeyCount = shortvec.decodeLength(byteArray);
+    if (configKeyCount < 2) return null;
```
mvines: I'd bail if `configKeyCount !== 2`, since it looks like we'll go off the rails below if there are more than 2 keys (`byteArray` won't be in the right place)

test/validator-info.test.js
```diff
@@ -0,0 +1,27 @@
...
+  const expectedValidatorInfo = new ValidatorInfo(
+    new PublicKey(keypair.publicKey),
+    {
+      name: 'Validator',
+      keybaseId: 'validator_id',
+    },
+  );
+
+  const configData = Buffer.from(
+    'AgdRlwF0SPKsXcI8nrx6x4wKJyV6xhRFjeCk8W+AAAAAABOY9ixtGkV8UbpqS189vS9p/KkyFiGNyJl+QWvRfZPKAS8AAAAAAAAAeyJrZXliYXNlSWQiOiJ2YWxpZGF0b3JfaWQiLCJuYW1lIjoiVmFsaWRhdG9yIn0',
```
mvines: Cool, can you sneak in a comment that documents how you generated this *random* string for the poor soul that needs to update it if/when the format changes :)

src/validator-info.js
jstarry: oops, good catch

test/validator-info.test.js
jstarry: hehe, will do :)

PR #404, opened 2019-07-16
fix: end of life native loader

jstarry:
- NativeLoader has been removed in favor of BpfLoader


mvines:
Let's rip out this too: https://github.com/solana-labs/solana-web3.js/blob/master/bin/localnet.sh#L32

PR #405, opened 2019-07-16
chore: update testnet default to stable

jstarry:


jstarry:
We prefer to keep the testnet default as edge to catch breaking changes from the main repo as fast as possible

PR #418, opened 2019-07-23
feat: add vote account decoder

mvines:


mvines:
@sunnygleason - FYI, using this'll allow you to remove the spawning of `solana-wallet`

PR #484, opened 2019-09-06
RFC/WIP: TypeScript example(s)

sunnygleason:
Problem:

* we don't currently have a solid set of examples for interested parties to integrate using TypeScript

Solution:

* build out some TypeScript examples/demos

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4830,6 +4830,12 @@
         }
       }
     },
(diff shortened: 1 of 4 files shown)


jstarry:
No longer needed, added a ts declaration file here: https://github.com/solana-labs/solana-web3.js/pull/630

PR #485, opened 2019-09-13
fix: Update examples

jackcmay:
Update the examples to the use libstd, the new Solana SDK, and new program return values


jackcmay:
@jstarry FYI

jstarry:
@jackcmay wow, 72% size decrease in the rust noop shared object file? Is that right?

jackcmay:
Yeah, I'm stripping the ELF, going to add that to the main solana repo soon too

PR #486, opened 2019-09-14
fix: add transaction instruction data-parsing functionality for system-instructions

CriesofCarrots:
cc: @sunnygleason 
Feedback definitely wanted, if there's a more elegant way to do this.

### DIFF
--- a/src/system-program.js
+++ b/src/system-program.js
@@ -2,9 +2,154 @@
 
 import * as BufferLayout from 'buffer-layout';
 
-import {Transaction} from './transaction';
+import {Transaction, TransactionInstruction} from './transaction';
 import {PublicKey} from './publickey';
 import * as Layout from './layout';
+import type {TransactionInstructionCtorFields} from './transaction';
+
+/**
+ * System Instruction class
+ */
+export class SystemInstruction extends TransactionInstruction {
+  /**
+   * Type of SystemInstruction
+   */
+  type: SystemInstructionType;
+
+  constructor(
+    opts?: TransactionInstructionCtorFields,
+    type?: SystemInstructionType,
+  ) {
+    if (
+      opts &&
+      opts.programId &&
+      !opts.programId.equals(SystemProgram.programId)
+    ) {
+      throw new Error('programId incorrect; not a SystemInstruction');
+    }
+    super(opts);
+    if (type) {
+      this.type = type;
+    }
+  }
+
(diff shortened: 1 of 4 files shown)


src/system-program.js
```diff
@@ -2,9 +2,154 @@
...
+      },
+      type,
+    );
+  }
+
+  /**
+   * The `from` public key of the instruction;
+   * returns null if SystemInstructionType does not support this field
+   */
+  get From(): PublicKey | null {
```
mvines: Can we rename `From` to maybe fromPubkey to avoid confusion with the the `from` static method.  Other names work too.   Also the accessor methods `From`/`To`/`Amount` should start with a lowercase letter

src/system-program.js
```diff
@@ -2,9 +2,154 @@
...
+ */
+type SystemInstructionType = {|
+  index: number,
+  layout: typeof BufferLayout,
+|};
+
+/**
+ * An enumeration of valid SystemInstructionTypes
+ */
+const SystemInstructionEnum = Object.freeze({
```
mvines: `SystemInstructionEnum` -> `SystemInstructionLayout`?

src/system-program.js
```diff
@@ -2,9 +2,154 @@
...
+  CREATE: {
```
mvines: nit: IMO `Create` instead of `CRATE`

src/system-program.js
mvines: (or rename the `from` static method to something else)

CriesofCarrots:
@mvines , look okay?

test/system-program.test.js
```diff
@@ -43,3 +49,95 @@ test('assign', () => {
...
+    from.publicKey,
+    to.publicKey,
+    amount,
+    space,
+    program.publicKey,
+  );
+  const transaction = new Transaction({recentBlockhash}).add(create);
+
+  const systemInstruction = SystemInstruction.from(transaction.instructions[0]);
+  expect(systemInstruction.fromPubkey).toEqual(from.publicKey);
```
mvines: I just realized we're further mixing `pubkey` and `publicKey`.   `fromPublicKey` and `toPublicKey` is probably most consistent with the rest of solana-web3.js at this point.  Sorry to bike shed.

CriesofCarrots: meh 👌

PR #489, opened 2019-09-20
fix: remove grant/grant_term from inflation

sunnygleason:
Problem:

* grant/grant_term were removed from RPC by https://github.com/solana-labs/solana/pull/5936

Solution:

* update web3js to not include those attributes
* extend tests so we could have caught this pre-release w/ integration tests


sunnygleason:
@mvines @CriesofCarrots just FYI, merging this so I can get a jump on patching network explorer!

mvines:
thanks @sunnygleason

PR #491, opened 2019-09-26
fix: add api for getMinimumBalanceForRentExemption

CriesofCarrots:


CriesofCarrots:
CI will fail until https://github.com/solana-labs/solana-web3.js/issues/490 is resolved.

mvines:
Let's hook this into https://github.com/solana-labs/solana-web3.js/blob/171ab6b9b6ba6d0fae459af91466cc4e2147053f/src/loader.js#L50 as well, so all example programs hopefully continue to work once https://github.com/solana-labs/solana/pull/6118 lands

CriesofCarrots:
@mvines , does this look okay?

src/connection.js
```diff
@@ -683,6 +688,22 @@ export class Connection {
     return GetInflationResult(res.result);
   }
 
+  /**
+   * Fetch the cluster Inflation parameters (TODO - book link/terminology?)
```
mvines: This comment looks wrong

src/loader.js
```diff
@@ -44,10 +44,11 @@ export class Loader {
     data: Array<number>,
   ): Promise<PublicKey> {
     {
+      const balanceNeeded = await connection.getMinimumBalanceForRentExemption(data.length);
       const transaction = SystemProgram.createAccount(
         payer.publicKey,
         program.publicKey,
-        1,
+        balanceNeeded > 0 ? balanceNeeded : 1,
```
mvines: when do you see `balanceNeeded < 1`?

CriesofCarrots: On a cluster not charging rent, as those in our current solana docker images.
Perhaps that will never exist on any cluster going forward, but not knowing whether that might be configurable per cluster, I thought this seemed less brittle.

src/connection.js
CriesofCarrots: 🤦‍♀

PR #494, opened 2019-09-27
fix: Update BPF Rust example ELF

jackcmay:


mvines:
heh our commit linter is super picky
```
⧗   input: fix: Update BPF Rust example ELF
✖   subject must not be sentence-case, start-case, pascal-case, upper-case [subject-case]
✖   found 1 problems, 0 warnings
ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint
```

PR #497, opened 2019-09-29
fix: graceful fallback when minimum balance request fails

jstarry:


jstarry:
FYI: @CriesofCarrots I'm making this change so that this doesn't prevent loading programs on the stable testnet which doesn't have the min balance endpoint yet

PR #527, opened 2019-10-15
feat: rfc - show how to use partial for extendable objects

sunnygleason:
PROBLEM:

* we frequently break APIs unnecessarily by extending objects with non-required fields

SOLUTION:

* use `struct.partial` instead of `struct` for extendable objects
* see https://github.com/ianstormtaylor/superstruct/blob/master/src/structs/partial.ts for impl

TODO:

* aggregate a list of extensible structs here (feedback welcome!)


mvines:
Are the non-specified fields stripped out during the validation too?   This looks to be what we want though for all the RPC responses

sunnygleason:
> Are the non-specified fields stripped out during the validation too? This looks to be what we want though for all the RPC responses

@mvines yes, I think the behavior will be:

* for forward compatibility, we use partial (`struct.partial`)
* when an unspecified field is added, it is ignored (TODO: verify this behavior with a test)
* for backward compatibility, we can use partial *and* use an optional field (`'number?'`, `'string?'`, etc)

I'll extend this PR with that behavior w/in the next couple days!

PR #528, opened 2019-10-15
feat: add optional rootSlot parameter

sunnygleason:
PROBLEM:

* the new `rootSlot` field is breaking stuff

SOLUTION:

* add an optional field definition until we figure out something like #527


sunnygleason:
@CriesofCarrots merging proactively to get a fix out, let me know if you see any issues!

PR #533, opened 2019-10-21
fix: wrap program invocation

jackcmay:
`LoaderInstruction` program invocation `InvokeMaon` is being treated as a special case, differently than Write and Finalize.

Instead, encapsulate those details within the loader JS API.  This change is in conjunction with changing Solana's runtime to be agnostic to the caller <--> program instruction data contract

https://github.com/solana-labs/solana/pull/6470

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,11 +42,11 @@ script:
   - npm run codecov
   - make -C examples/bpf-c-noop/
   - examples/bpf-rust-noop/do.sh build
-  - npm run localnet:update
-  - npm run localnet:up
-  - npm run examples || true
-  - npm run test:live
-  - npm run localnet:down
+  # - npm run localnet:update
+  # - npm run localnet:up
+  # - npm run examples || true
+  # - npm run test:live
+  # - npm run localnet:down
 
 before_deploy:
   - rm -rf deploy
--- a/src/loader.js
+++ b/src/loader.js
@@ -5,8 +5,12 @@ import * as BufferLayout from 'buffer-layout';
 import {Account} from './account';
 import {PublicKey} from './publickey';
 import {NUM_TICKS_PER_SECOND} from './timing';
-import {Transaction, PACKET_DATA_SIZE} from './transaction';
-import type {TransactionSignature} from './transaction';
+import {
+  Transaction,
+  TransactionInstruction,
+  PACKET_DATA_SIZE,
+} from './transaction';
+import type {TransactionInstructionCtorFields} from './transaction';
 import {SYSVAR_RENT_PUBKEY} from './sysvar-rent';
 import {sendAndConfirmTransaction} from './util/send-and-confirm-transaction';
 import {sleep} from './util/sleep';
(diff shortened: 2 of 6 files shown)


.travis.yml
```diff
@@ -42,11 +42,11 @@ script:
   - npm run codecov
   - make -C examples/bpf-c-noop/
   - examples/bpf-rust-noop/do.sh build
-  - npm run localnet:update
-  - npm run localnet:up
-  - npm run examples || true
-  - npm run test:live
-  - npm run localnet:down
+  # - npm run localnet:update
```
mvines: Temp change only I assume?

src/loader.js
```diff
@@ -147,4 +152,42 @@ export class Loader {
...
+      BufferLayout.u32('bytesLength'),
+      BufferLayout.u32('bytesLengthPadding'),
+      BufferLayout.seq(
+        BufferLayout.u8('byte'),
+        BufferLayout.offset(BufferLayout.u32(), -8),
+        'bytes',
+      ),
+    ]);
+
+    var instruction;
```
mvines: Use let, not var please

test/bpf-loader.test.js
```diff
@@ -26,20 +23,23 @@ test('load BPF C program', async () => {
...
   const connection = new Connection(url);
   const [, feeCalculator] = await connection.getRecentBlockhash();
   const fees =
     feeCalculator.lamportsPerSignature *
-    (BpfLoader.getMinNumSignatures(data.length) + NUM_RETRIES);
-  const from = await newAccountWithLamports(connection, fees);
+    (BpfLoader.getMinNumSignatures(data.length) + NUM_RETRIES + 1);
+  const payer = await newAccountWithLamports(connection, fees);
+  const programId = await BpfLoader.load(connection, payer, data);
+  const program_data = Buffer.from([1, 2, 3, 4, 5, 6, 7, 8]);
```
mvines: nit: programData

mvines: or just `data

.travis.yml
jackcmay: Yeah, sucks: https://github.com/solana-labs/solana/issues/6545

test/bpf-loader.test.js
```diff
@@ -51,18 +51,21 @@ test('load BPF Rust program', async () => {
...
 
-  const programId = await BpfLoader.load(connection, from, data);
-  const transaction = new Transaction().add({
-    keys: [{pubkey: from.publicKey, isSigner: true, isDebitable: true}],
-    programId,
-  });
-  await sendAndConfirmTransaction(connection, transaction, from);
+  const transaction = new Transaction().add(
+    Loader.invokeMainInstruction({
```
mvines: I don’t think I grok what you’re doing here.  Where does the 1,2,3,4,5,6,7,8 input go exactly? The programId is of the rust_noop.so and not the BPFLoader right?

jackcmay: Yes, the programId returned from load is rust_noop.so's and the `programData` is the instruction data sent to noop's process instruction

jackcmay: `Loader` encapsulates `LoaderInstruction` (Write, Finalize, InvokeMain).  Yes, the examples change so they don't have to construct the `InvokeMain` variant of `LoaderInstruction` themselves.  We could put `invokeMainInstruction` someplace else but either way, the examples change.  This PR coincide with https://github.com/solana-labs/solana/pull/6470 where the runtime no longer assumes `InvokeMain`.   This allows loaders to extend or replace `LoaderInstruction` at the client/loader level without forcing a Solana runtime change.

jackcmay: Here is an example of how the examples invoke a program: https://github.com/solana-labs/example-messagefeed/pull/201

mvines:
stale!

PR #538, opened 2019-10-29
fix: wait for airdrop

jackcmay:
Airdrop's tx may take a bit to clear even if balance reports the correct account balance, wait a bit.


mvines:
Let's back this out once https://github.com/solana-labs/solana/pull/6538 lands

PR #540, opened 2019-10-29
feat: getEpochInfo rpc

sunnygleason:
PROBLEM:

* we need current epoch for computing uptime in network explorer 🕚 

SOLUTION:

* add the already-existing getEpochInfo() call to the JS library 🔨 

OPEN QUESTION (assume it's ok):

* the existing RPC has camel-case response keys instead of underscore-separated, is that ok? (I figure now's not a good time to break/fix that) 🌊


sunnygleason:
@CriesofCarrots this is a small one (hopefully)...

PR #549, opened 2019-11-10
feat: add commitment param to JSON RPC API requests

jstarry:
Fixes: https://github.com/solana-labs/solana-web3.js/issues/546

Other related changes:
* Allow specifying a default commitment level for a `Connection` instance
* Add new `sendAndConfirmRecentTransaction` convenience method
* Fix broken examples
* Fix live tests
* Add test for max commitment

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -20,16 +20,22 @@ type RpcRequest = (methodName: string, args: Array<any>) => any;
 
 /**
  * The node will query the most recent block which has reached max voter lockout
+ *
+ * @typedef {'max'} MAX_COMMITMENT
  */
 export type MAX_COMMITMENT = 'max';
 
 /**
  * The node will query its most recent block
+ *
+ * @typedef {'recent'} RECENT_COMMITMENT
  */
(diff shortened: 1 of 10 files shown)


jstarry:
Tests are still failing when `to` is unsigned, I think those can be fixed in a follow up.

src/connection.js
```diff
@@ -18,6 +18,27 @@ import type {TransactionSignature} from './transaction';
...
+ * @typedef {'recent'} RECENT_COMMITMENT
+ */
+export type RECENT_COMMITMENT = 'recent';
+
+/**
+ * The level of commitment desired when querying state
+ *
+ * @typedef {MAX_COMMITMENT | RECENT_COMMITMENT} Commitment
+ */
+export type Commitment = MAX_COMMITMENT | RECENT_COMMITMENT;
```
mvines: I prefer just a simple: 
```js
export type Commitment = 'max' | 'recent';
```

flow will yell if any other strings are used

jstarry: yeah much better!

PR #552, opened 2019-11-11
fix: add getBlocksSince Connection method

CriesofCarrots:
Also sneaks in a naming fix for isCreditDebit() -> isWriteable()


src/connection.js
```diff
@@ -853,6 +858,19 @@ export class Connection {
     return res.result;
   }
 
+  /**
+   * Fetch a tree of rooted blocks from the cluster
```
mvines: nit: it's not a tree, it's a list

CriesofCarrots:
CI failure unrelated 😬

PR #555, opened 2019-11-12
fix: add getBlock Connection method

CriesofCarrots:
Will add issue to update getBlock test on upstream updates

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -391,6 +391,18 @@ const GetMinimumBalanceForRentExemptionRpcResult = jsonRpcResult('number');
 const GetBlocksSinceRpcResult = jsonRpcResult(struct.list(['number']));
 
 /**
+ * Expected JSON RPC response for the "getBlock" message
+ */
+const GetBlockRpcResult = jsonRpcResult(
+  struct.list([
+    struct.tuple([
+      struct.list(['number']),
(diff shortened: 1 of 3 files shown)


CriesofCarrots:
CI failure is because this hasn't propagated to our docker image yet: https://github.com/solana-labs/solana/pull/6889

src/connection.js
```diff
@@ -904,6 +916,23 @@ export class Connection {
     return res.result;
   }
 
+  /**
+   * Fetch a list of Transactions and transaction statuses from the cluster
+   */
+  async getBlock(
+    slot: number,
+  ): Promise<Array<[Transaction, SignatureSuccess | TransactionError | null]>> {
```
jstarry: esdoc is complaining about this type ^

It looks like it's supposed to be: `[Transaction, SignatureSuccess] | ...` just missing a closing bracket

CriesofCarrots: The closing bracket is after the `null`...
I tried putting the whole OR expression (the second value in the tuple) in parens, but prettier removed them

CriesofCarrots: Maybe I have to be explicit?
`[Transaction, SignatureSuccess] | [Transaction, TransactionError] ...`

CriesofCarrots: Yes, the explicit tuples approach works.
Thanks, @jstarry !

Issue #556, opened 2019-11-12
adjust API for new RPC response context


sunnygleason:
Problem:

* we just changed 4 methods to have RpcResponseContext results

- [x] getBalance
- [x] getRecentBlockhash
- [x] getAccount
- [x] confirmTransaction

Solution:

* introduce the outer struct containing `context` and `value` elements
* retrofit API calls to return the `value` field directly
* add new `xxxWithCommitment()` versions that return `RpcResponse` objects with `context`
* (this is similar to the rust client approach in https://github.com/solana-labs/solana/ client)

PR #557, opened 2019-11-13
feat: implement new response format for __WithCommitment() methods

sunnygleason:
Problem:

* need to update API for new RPC response format 🔧 

Solution:

* update the API 🌈 

Fixes: #556

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -18,6 +18,50 @@ import type {TransactionSignature} from './transaction';
 
 type RpcRequest = (methodName: string, args: Array<any>) => any;
 
+type RpcResponseAndContext<T> = {
+  context: {
+    slot: number,
+  },
+  value: T,
+};
+
+/**
+ * @private
+ */
+function jsonRpcResultAndContext(resultDescription: any) {
+  return struct.union([
+    // those same methods return results with context in v0.21+ servers
+    jsonRpcResult({
+      context: struct({
+        slot: 'number',
+      }),
+      value: resultDescription,
+    }),
+    // selected methods return "bare" results in pre-v0.21 servers
+    jsonRpcResult(resultDescription),
+  ]);
+}
+
+/**
+ * @private
+ */
+function jsonRpcResult(resultDescription: any) {
+  const jsonRpcVersion = struct.literal('2.0');
+  return struct.union([
+    struct({
+      jsonrpc: jsonRpcVersion,
+      id: 'string',
+      error: 'any',
+    }),
+    struct({
+      jsonrpc: jsonRpcVersion,
+      id: 'string',
+      error: 'null?',
+      result: resultDescription,
+    }),
+  ]);
+}
+
 /**
  * The level of commitment desired when querying state
  *   'max':    Query the most recent block which has reached max voter lockout
@@ -207,12 +251,7 @@ const GetEpochScheduleRpcResult = struct({
 /**
  * Expected JSON RPC response for the "getBalance" message
  */
-const GetBalanceRpcResult = struct({
-  jsonrpc: struct.literal('2.0'),
-  id: 'string',
-  error: 'any?',
-  result: 'number?',
-});
+const GetBalanceAndContextRpcResult = jsonRpcResultAndContext('number?');
 
 /**
  * Expected JSON RPC response for the "getVersion" message
@@ -223,26 +262,6 @@ const GetVersionRpcResult = struct({
   error: 'any?',
   result: Version,
 });
-
-/**
- * @private
- */
-function jsonRpcResult(resultDescription: any) {
-  const jsonRpcVersion = struct.literal('2.0');
-  return struct.union([
-    struct({
-      jsonrpc: jsonRpcVersion,
-      id: 'string',
-      error: 'any',
-    }),
-    struct({
-      jsonrpc: jsonRpcVersion,
-      id: 'string',
-      error: 'null?',
-      result: resultDescription,
-    }),
-  ]);
-}
 
 /**
  * @private
@@ -258,7 +277,9 @@ const AccountInfoResult = struct({
 /**
  * Expected JSON RPC response for the "getAccountInfo" message
  */
-const GetAccountInfoRpcResult = jsonRpcResult(AccountInfoResult);
+const GetAccountInfoAndContextRpcResult = jsonRpcResultAndContext(
+  struct.union(['null', AccountInfoResult]),
+);
 
 /***
  * Expected JSON RPC response for the "accountNotification" message
@@ -291,7 +312,9 @@ const GetProgramAccountsRpcResult = jsonRpcResult(
 /**
  * Expected JSON RPC response for the "confirmTransaction" message
  */
-const ConfirmTransactionRpcResult = jsonRpcResult('boolean');
+const ConfirmTransactionAndContextRpcResult = jsonRpcResultAndContext(
+  'boolean',
+);
 
 /**
  * Expected JSON RPC response for the "getSlot" message
@@ -405,7 +428,7 @@ const GetBlockRpcResult = jsonRpcResult(
 /**
  * Expected JSON RPC response for the "getRecentBlockhash" message
  */
-const GetRecentBlockhash = jsonRpcResult([
+const GetRecentBlockhashAndContextRpcResult = jsonRpcResultAndContext([
   'string',
   struct({
     burnPercent: 'number',
@@ -416,6 +439,7 @@ const GetRecentBlockhash = jsonRpcResult([
     targetSignaturesPerSlot: 'number',
   }),
 ]);
+
 /**
  * @ignore
  */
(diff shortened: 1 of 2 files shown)


sunnygleason:
fixes #556

CriesofCarrots:
`WithCommitment` doesn't seem quite right to me on these methods, since there are others that take the optional commitment param, and the difference is actually in the data returned.
How about `WithContext`?

sunnygleason:
@CriesofCarrots that's a great point, maybe @mvines has thoughts on that?

(this PR lines up with the rust client API changes previously merged)

src/connection.js
```diff
@@ -207,12 +214,14 @@ const GetEpochScheduleRpcResult = struct({
 /**
  * Expected JSON RPC response for the "getBalance" message
  */
-const GetBalanceRpcResult = struct({
-  jsonrpc: struct.literal('2.0'),
-  id: 'string',
-  error: 'any?',
-  result: 'number?',
-});
+const GetBalanceWithCommitmentRpcResult = jsonRpcResult(
```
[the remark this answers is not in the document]
jstarry: It would also be nice if it was implemented in a backwards compatible way with an issue / comment to remove when your rpc changes get released in 0.21.

Something like (untested)
```
function jsonRpcResultWithContext(value: any) {
  return jsonRpcResult(struct.union([
    value,
    struct({
      context: struct({
        slot: 'number',
      }),
      value,
    }),
  ]);
}
```

CriesofCarrots:
> (this PR lines up with the rust client API changes previously merged)

It sort of does, in that the methods to which you added the RpcResponse type in rpc_client are all `_with_commitment` methods. But they are only a subset of a broader group of `_with_commitment` methods, which all take a commitment parameter.

Thinking about that more, we seem to be using "with" to mean "with argument". So I amend my suggestion to `AndContext`; "and" meaning the thing that follows is part of the stuff we are getting.

src/connection.js
```diff
@@ -882,12 +952,19 @@ export class Connection {
...
-    const res = GetRecentBlockhash(unsafeRes);
+    const res = GetRecentBlockhashWithCommitmentRpcResult(unsafeRes);
     if (res.error) {
       throw new Error(res.error.message);
     }
     assert(typeof res.result !== 'undefined');
-    return res.result;
+    return RpcResponse(res.result);
+  }
+  async getRecentBlockhash(
```
jstarry: Do we really need to have two separate methods? `x` and `xWithCommitment`? I think it would be better to stick with one and break the previous api

sunnygleason:
> It sort of does

Wondering, are you referring to your original PR?

https://github.com/solana-labs/solana/pull/6538/files

And I'm referring to my subsequent PR?

https://github.com/solana-labs/solana/pull/6716/files

I'm trying to sort this out -- it seems like we're at a point where rust
client implements X, we have a waiting JS PR that matches X here.

I'm wondering, is X ok? Or can y'all (@mvines @jstarry @CriesofCarrots)
 tell me the Y that I should implement (I'm assuming we want it the
same in both JS and Rust client)?

I'm just slightly hesitant to chase a Y because implementing it in both
places would easily eat up a couple days. But up for anything!

CriesofCarrots:
@sunnygleason , I'm referring to both PRs.
1. https://github.com/solana-labs/solana/pull/6759 adds a number of `_with_commitment` methods to rpc_client, so named because they add a `commitment_config` parameter
2. https://github.com/solana-labs/solana/pull/6716 updates a subset of those methods to return the RpcResponse struct

1. #549 is the analog to 6759, except it adds the optional `?Commitment` parameter to that set of methods, instead duplicating into `WithCommitment` methods.
2. ...and this PR is the analog to 6716, updating that subset of methods to return the RpcResponse struct

It's true that the js sdk does not perfectly mirror the rust client, but I think that is acceptable, and somewhat inevitable, given the language differences. (Also, it's quite possible that I will eliminate the duplicate `_with_commitment` methods from rpc_client at some point in the future, when commitment is integrated more fully into downstream rust client usage.)

I don't think I'm suggesting anything that requires changes in rpc_client, just renaming to describe what we're actually doing in this repository, which is adding methods with a different return type. (Alternately, I'm also on board with @jstarry 's suggestion to stick with non-duplicated methods and break the previous api)

src/connection.js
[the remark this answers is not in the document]
sunnygleason: this is neat and I love the backwards compatibility idea, one issue I see is that Envelope<T> isn't a T. So I think we need to have a different method name, dispatch based on the presence of commitment in the input argument, or something like that. Let me code the dispatch idea and see what you all think.

src/connection.js
sunnygleason: I like this sentiment, will go for the dispatch idea

sunnygleason:
@jstarry @CriesofCarrots ok I think this should do it! let me know if you have any other suggestions!

src/connection.js
```diff
@@ -594,15 +613,25 @@ export class Connection {
...
+    assert(typeof res.result.value !== 'undefined');
+
+    const rpcResponseAndContext = {
+      context: {
+        slot: res.result.context.slot,
+      },
+      value: res.result.value,
+    };
+
+    return commitment ? rpcResponseAndContext : rpcResponseAndContext.value;
```
jstarry: This conditional isn't right. Someone using this API could be relying on the default behaviour when no commitment is passed but still want to receive the context. I think we need a check above to see if the returned response has a context or not and use that to determine the return value

jstarry: There will be a few weeks until your RPC api changes are live on the next Solana release. In the meantime, we will want to use web3 sdk for examples and handle responses from both edge testnet (with your api changes) and the stable testnet (without your api changes)

sunnygleason: I'm having trouble seeing how that works. In this code, the client "opts in" to the new response object by providing the commitment as input. I don't see how this revision can be a drop-in replacement if we do what you're saying?

jstarry: @sunnygleason I don't see this "opt in" that you're referring to. RPC api returns the context whether you specify a commitment or not: https://github.com/solana-labs/solana/blob/master/core/src/rpc.rs#L107

CriesofCarrots: I was thinking the backward compatibility aspect would look more like what was implemented for recentBlockhash:
https://github.com/solana-labs/solana-web3.js/blob/31ecd913a4a85c20bc23a0ad8c531891870ac8c3/src/connection.js#L872
We need to be trying both result formats, right?

src/connection.js
```diff
@@ -591,43 +615,104 @@ export class Connection {
...
     const unsafeRes = await this._rpcRequest('getBalance', args);
-    const res = GetBalanceRpcResult(unsafeRes);
+    const res = GetBalanceAndContextRpcResult(unsafeRes);
     if (res.error) {
       throw new Error(res.error.message);
     }
     assert(typeof res.result !== 'undefined');
-    return res.result;
+
+    const isV21 =
```
jstarry: nit: isV021

src/connection.js
sunnygleason: @CriesofCarrots @jstarry I think I have something workable! The issue earlier was that I didn't know we needed JS API changes to be backwards-compatible with respect to server responses since my earlier PR to solana rust client wasn't, I think I got lost in the notions of client vs server-side backwards compatibility across multiple conversations.

* I needed to add the `xyzAndContext()` methods and keep because the JS tests were too gnarly to port to the new value format *quickly*
* If someone has opinions about those methods, I don't object to anyone diving in
* I've incorporated all your other feedback (backwards compatibility etc)
* this should unblock @jackcmay and give him a drop-in version of this package to work with

src/connection.js
```diff
@@ -591,43 +615,106 @@ export class Connection {
...
+      };
+    } else if (!isV21 && res.result) {
+      const {executable, owner, lamports, data} = res.result;
+      value = {
+        executable,
+        owner: new PublicKey(owner),
+        lamports,
+        data: Buffer.from(data),
+      };
+    }
```
jstarry: How about:
```js
    let resultValue = isV021 ? res.result.value : res.result;
    assert(typeof resultValue !== 'undefined');
    const {executable, owner, lamports, data} = resultValue;
    let value = {
        executable,
        owner: new PublicKey(owner),
        lamports,
        data: Buffer.from(data),
    };
```

sunnygleason: resultValue can be null, would that destructure ok? is there a slight tweak for that? 😕

sunnygleason: ah ok just realized the reorder caught that! thanks!

src/connection.js
```diff
@@ -591,43 +615,106 @@ export class Connection {
...
       throw new Error(res.error.message);
     }
     assert(typeof res.result !== 'undefined');
-    return res.result;
+
+    const isV21 =
+      typeof res.result.context !== 'undefined' &&
+      typeof res.result.value !== 'undefined';
+
+    const slot = isV21 ? res.result.context.slot : NaN;
```
CriesofCarrots: This might be a little simpler as:
```
if (isV21) {
  return res.result;
} else {
  return {
    context: {
      NaN,
    },
    res.result,
  };
}
```
...assuming flow is okay with that. (I've been deep in Rustland for a while)

src/connection.js
CriesofCarrots: Sweet, looks close

src/connection.js
sunnygleason: ooh nice idea!

jstarry:
Don't forget to update module.flow.js :)

sunnygleason:
> I would like to remove the duplication of getX + getXAndContext but this is fine, we can do it later

I agree, and tried, just ran out of time sifting through broken tests! Looking for a /torch/ emoji to pass... 😂

sunnygleason:
> Don't forget to update module.flow.js :)

Ah! good catch, thank you!

sunnygleason:
> Actually I think we can leave `module.flow.js` for now since it sounds like we will soon un-duplicate these methods.

think I got it, feel free to patch if I got anything wrong...

PR #559, opened 2019-11-16
IGNORE_ME: why is CI sad (reverting last commit)

sunnygleason:
trying to see what changed to make CI fail recently.


sunnygleason:
closing: this is not it -- I think the docker entrypoint is the culprit

PR #566, opened 2019-11-25
feat: slot change callback (provides real-time fork information)

sunnygleason:
PROBLEM:

* clients would like realtime slot notifications (e.g. for fork visualization)
* see solana-labs/solana#7114 for latest problem description

SOLUTION:

* hook up the websocket slot notification API

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -319,6 +319,23 @@ const ProgramAccountNotificationResult = struct({
 });
 
 /**
+ * @private
+ */
+const SlotInfo = struct({
+  parent: 'number',
+  slot: 'number',
+  root: 'number',
+});
+
+/***
+ * Expected JSON RPC response for the "slotNotification" message
+ */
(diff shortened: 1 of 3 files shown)


sunnygleason:
see https://github.com/solana-labs/solana/pull/7114

src/connection.js
```diff
@@ -618,6 +640,10 @@ export class Connection {
     [number]: ProgramAccountSubscriptionInfo,
   } = {};
   _programAccountChangeSubscriptionCounter: number = 0;
+  _slotSubscriptions: {
+    [number]: number,
```
CriesofCarrots: Looks like this should be `[number]: SlotSubscriptionInfo`, or similar (analog of `AccountSubscriptionInfo`, will need defining.)
This should appease flow in ci...

sunnygleason: (should be all set, thank you!)

PR #571, opened 2019-12-10
add createAccountWithSeed

rob-solana:


rob-solana:
untested

where's the document that tells me how to run the tests locally?

src/layout.js
```diff
@@ -43,5 +43,25 @@ export const rustString = (property: string = 'string') => {
     return _encode(data, buffer, offset);
   };
 
+  rsl.alloc = str => {
```
rob-solana: here's the magic!

src/system-program.js
```diff
@@ -139,13 +144,27 @@ const SystemInstructionLayout = Object.freeze({
...
+    ]),
+  },
 });
 
 /**
  * Populate a buffer of instruction data using the SystemInstructionType
  */
 function encodeData(type: SystemInstructionType, fields: Object): Buffer {
-  const data = Buffer.alloc(type.layout.span);
+  const allocLength =
```
rob-solana: more magic?

rob-solana:
thank you @CriesofCarrots

src/system-program.js
CriesofCarrots: Yes. I'm not entirely sure this is the best way to handle the the rust string, but it seemed reasonably extensible for other value-specific layouts, and I didn't see a more straightforward way using the `buffer-layout` package.

mvines: Can we just use a simple `char *` string?  Requiring utf8 is unnecessarily restrictive and is annoying to handle -- ie, all this code

CriesofCarrots: I'm not sure I see what you're envisioning, @mvines . The gymnastics in this pr are largely around handling dynamically sized layouts in buffer-layout, which would still be an issue even if not utf8, right?

mvines: I guess it seems wasteful to require a length of 8 bytes for a seed with a length of 32 bytes or less

CriesofCarrots: Ah, yeah. I expect we could use shortvec easily enough here, if @rob-solana likes that option upstream.

rob-solana: I've no preferences for how this side comes out...

for the system instruction, my goals are:
1. looks pretty for rust clients and for the implementation
2. supports a reasonable-sized mnemonic (32 is a nice length if one stays in ascii)
3. limit the space the seed can cover to something smaller than 256-bits (ascii and utf-8 have those restrictions)

PR #576, opened 2019-12-13
fix: account change test broken on beta testnet

jstarry:


src/connection.js
```diff
@@ -399,9 +399,10 @@ const GetVoteAccounts = jsonRpcResult(
         nodePubkey: 'string',
         activatedStake: 'number',
         epochVoteAccount: 'boolean',
-        epochCredits: struct.union(['null', struct.list([
-          struct.tuple(['number', 'number', 'number']),
-        ])]),
+        epochCredits: struct.union([
+          'null',
```
mvines: epochCredits landed in beta, https://github.com/solana-labs/solana/pull/7445, so why do we need this?

jstarry: I thought @jackcmay needed v0.21 compatibility but maybe not?

mvines: Yes he certainly does.  But epochCredits is now in v0.21

jackcmay: that fix also fixed one of the two failures I was seeing with the beta localnet

jstarry: > But epochCredits is now in v0.21


Ah ok, good point, I'll remove.

jstarry: removed here: https://github.com/solana-labs/solana-web3.js/pull/577

PR #591, opened 2019-12-24
Add stake program

CriesofCarrots:
@mvines , fyi, and one question:
I wrote `delegate`, `authorize`, `split`, `withdraw`, and `deactivate` on the assumption that the client would use a non-stake-account authorized keypair to sign/execute the transaction. This seems like the most common downstream use case to me, so it is a convenient short-cut. Should I rework now to offer something like the solana-stake-program's `with_signer()`?

- [x] I will add a live integration test to bump this up to ready-for-review.

Toward https://github.com/solana-labs/solana/issues/7583

### DIFF
--- a/src/instruction.js
+++ b/src/instruction.js
@@ -1,0 +1,27 @@
+// @flow
+
+import * as BufferLayout from 'buffer-layout';
+
+import * as Layout from './layout';
+
+/**
+ * @typedef {Object} InstructionType
+ * @property (index} The Instruction index (from solana upstream program)
+ * @property (BufferLayout} The BufferLayout to use to build data
+ */
+export type InstructionType = {|
+  index: number,
+  layout: typeof BufferLayout,
+|};
+
+/**
+ * Populate a buffer of instruction data using an InstructionType
+ */
+export function encodeData(type: InstructionType, fields: Object): Buffer {
+  const allocLength =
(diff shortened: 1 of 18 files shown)


src/instruction.js
```diff
@@ -0,0 +1,29 @@
...
+
+import * as BufferLayout from 'buffer-layout';
+
+import * as Layout from './layout';
+// import type {StakeInstructionType} from './stake-program';
+// import type {SystemInstructionType} from './system-program';
+
+/**
+ * @typedef {Object} InstructionType
+ * @property (index} The Instruction index (from solana upstrem program)
```
mvines:
```suggestion
 * @property (index} The Instruction index (from solana upstream program)
```

mvines:
> I wrote `delegate`, `authorize`, `split`, `withdraw`, and `deactivate` on the assumption that the client would use a non-stake-account authorized keypair to sign/execute the transaction. This seems like the most common downstream use case to me, so it is a convenient short-cut.

sounds good to me, let's start with this and if needed we can add more bells and whistles later

CriesofCarrots:
This PR needs to be rebased on #593 , once it lands

CriesofCarrots:
@mvines , I'm going to go ahead and land this, but happy to fix up any concerns you have post-merge. (Will also monitor for more upstream stake program changes.)

mvines:
Thanks, I just scrolled through and looks 💯

PR #595, opened 2020-01-03
Add Nonce functionality to solana-web3

CriesofCarrots:
Depends on https://github.com/solana-labs/solana/pull/7645 

Closes https://github.com/solana-labs/solana/issues/7583
and addresses web3 concern of https://github.com/solana-labs/solana/issues/7700

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -1224,59 +1224,63 @@ export class Connection {
     transaction: Transaction,
     ...signers: Array<Account>
   ): Promise<TransactionSignature> {
-    for (;;) {
-      // Attempt to use a recent blockhash for up to 30 seconds
-      const seconds = new Date().getSeconds();
-      if (
-        this._blockhashInfo.recentBlockhash != null &&
-        this._blockhashInfo.seconds < seconds + 30
-      ) {
-        transaction.recentBlockhash = this._blockhashInfo.recentBlockhash;
-        transaction.sign(...signers);
-        if (!transaction.signature) {
-          throw new Error('!signature'); // should never happen
-        }
-
-        // If the signature of this transaction has not been seen before with the
-        // current recentBlockhash, all done.
-        const signature = transaction.signature.toString();
-        if (!this._blockhashInfo.transactionSignatures.includes(signature)) {
-          this._blockhashInfo.transactionSignatures.push(signature);
-          if (this._disableBlockhashCaching) {
-            this._blockhashInfo.seconds = -1;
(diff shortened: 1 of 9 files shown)


CriesofCarrots:
@mvines , the most controversial part of this is commit 0dd4c89 that adds a `nonceInfo` field to Transaction, and trumps `recentBlockhash` in `Transaction._getSignData()` and `Connection.sendTransaction()`. Are you okay with this approach?

I would have liked to only store the Nonce account pubkey and the Nonce hash in that object, but building the NonceAdvance instruction dynamically in `_getSignData()` caused a circular dependency issue between Transaction and SystemProgram that did not seem simpler when untangled.

src/connection.js
```diff
@@ -1149,6 +1150,56 @@ export class Connection {
     };
   }
 
+  /**
+   * Fetch a list of Transactions and transaction statuses from the cluster
+   * for a confirmed block
+   */
```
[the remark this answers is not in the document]
CriesofCarrots: 🤦‍♀

src/connection.js
```diff
@@ -1149,6 +1150,55 @@ export class Connection {
...
+    if (res.error) {
+      throw new Error(res.error.message);
+    }
+    assert(typeof res.result !== 'undefined');
+
+    const isV021 =
+      typeof res.result.context !== 'undefined' &&
+      typeof res.result.value !== 'undefined';
+
+    const slot = isV021 ? res.result.context.slot : NaN;
```
mvines: I think we can drop v0.21 support anytime now

src/connection.js
```diff
@@ -1149,6 +1150,55 @@ export class Connection {
...
+    const value = NonceAccount.fromAccountData(Buffer.from(resultValue.data));
+
+    return {
+      context: {
+        slot,
+      },
+      value,
+    };
+  }
+  async getNonce(
```
mvines: whitespace before new function, and esdoc header please

mvines:
> Are you okay with this approach?

Yep seems reasonable!

src/connection.js
CriesofCarrots: Sounds good. I'll do that in a separate PR.

src/connection.js
CriesofCarrots: I was following the pattern of all the other `AndContext` method pairs on this.
I'll fix this up for all methods it that aforementioned separate PR.

PR #596, opened 2020-01-03
fix: add getMinimumBalanceForRentExemption and getBlock to flow

jackcmay:
module.flow.rs was missing:
`getMinimumBalanceForRentException`
`getBlock`


CriesofCarrots:
Hey, @jackcmay, `getBlock` wasn't in flow because there is no `Connection.getBlock()` method. There is a `getConfirmedBlock`...

jackcmay:
Ok, I'll remove

PR #630, opened 2020-02-03
fix: add typescript type declaration to npm bundle

jstarry:
#### Problem
Developers using TypeScript need a type declaration file to get static typing when using our web3 sdk. This is important because [TypeScript seems to be a lot more popular than flow](https://insights.stackoverflow.com/trends?tags=flowtype%2Ctypescript)

#### Changes
* Add a `module.d.ts` which mirrors the `module.flow.js` file
* Copy the `module.d.ts` file into npm lib package
* Configure prettier and eslint for the typescript file


jstarry:
@mvines @CriesofCarrots thoughts on adding this? Seems nice to have but hard (easy to forget) to maintain. My vote is for including it and I hope that when it inevitably gets out of date, a helpful dev notices and upstreams a fix.

Hopefully we don't live in this world for long, the real solution here is to generate type declarations automatically. Possibly with [`wasm-bindgen`](https://rustwasm.github.io/docs/wasm-bindgen/reference/cli.html#--typescript)

mvines:
lgtm

jstarry:
@CriesofCarrots nah, I think we can close that. The declaration file should be enough, and also the "Break Solana" is using TypeScript if people really want to see an example

PR #632, opened 2020-02-04
fix: remove references to decommissioned testnets

jstarry:
#### Problem
The edge and beta testnets have been turned off but the sdk still references them. Removing them will cause an error to be thrown in the testnet utility method

This also impacts the `bpf-sdk-install` and `localnet` binaries. They read `"testnetDefaultChannel"` to determine which sdk and docker image to install/download/run. But both work fine with "stable" as the new default.


jstarry:
Hmm, might be better to have separate release trains for edge / stable. Ideally we always have a web3 sdk that works for stable and one for cutting edge.

PR #707, opened 2020-03-03
feat: add API for decoding system instructions

jstarry:
#### Problem
* The sdk doesn't provide a way to decode the finer details of system instruction data (especially nonce instructions)
* System instruction API methods have lots of parameters and can be easily misused

#### Changes
* Remove `SystemInstruction.from` method in favor of `SystemInstruction.decode`... methods.
* Refactor `SystemProgram` methods to accept a typed param object rather than a list of params
* Add methods to decode system instructions into the typed param objects they were created with
* Add more tests for peace of mind (and @CriesofCarrots's peace of mind!)

### DIFF
--- a/module.d.ts
+++ b/module.d.ts
@@ -475,6 +475,65 @@ declare module '@solana/web3.js' {
   }
 
   // === src/system-program.js ===
+  export type CreateAccountParams = {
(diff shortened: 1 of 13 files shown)


jstarry:
Thanks for updating for the new test!

PR #712, opened 2020-03-05
fix: update NonceAccount to upstream changes

CriesofCarrots:
#### Problem
A new structure for Nonce accounts just landed on Solana master and v1.0

#### Changes
Fix js apis to accommodate new format. Also, make nonceSpace definition less magical.


CriesofCarrots:
@jstarry , fyi. Also, if you could take a peek at the flow/ts exports to make sure I've handled properly, it would be much appreciated!

PR #742, opened 2020-03-20
feat: add getSignatureConfirmation endpoint

jstarry:
#### Problem
* getSignatureConfirmation RPC method is missing from the SDK

#### Changes
* Rename `SignatureStatusResult` type to `SignatureStatus` since it conflicted with the struct def of the same name
* Add `getSignatureConfirmation` method to `Connection`


jstarry:
@CriesofCarrots I thought I needed this for the explorer but now I don't think I do so no rush on the review!

PR #755, opened 2020-03-23
Address latest RPC API breaking changes

jstarry:
Changes:
* Added context to pubsub notification callbacks
* Updated `getSignatureStatus` return type to include slot
* Added `getSignatureStatusBatch` method

### DIFF
--- a/module.flow.js
+++ b/module.flow.js
@@ -50,10 +50,12 @@ declare module '@solana/web3.js' {
   /* TODO */
 
   // === src/connection.js ===
+  declare export type Context = {
+    slot: number,
+  };
+
   declare export type RpcResponseAndContext<T> = {
-    context: {
-      slot: number,
-    },
+    context: Context,
     value: T,
   };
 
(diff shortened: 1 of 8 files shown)


jstarry:
@CriesofCarrots I decided to merge to speed up fixing the examples for v1.0.9. I think it would still be nice to get another set of eyes on these changes though when you have time

module.flow.js
```diff
@@ -63,6 +65,11 @@ declare module '@solana/web3.js' {
     | SignatureSuccess
     | TransactionError;
 
+  declare export type SignatureStatus = {
+    slot: number,
+    status: SignatureSuccess | TransactionError,
```
CriesofCarrots: `SignatureStatusResult`?

test/connection.test.js
```diff
@@ -1011,18 +1025,64 @@ test('transaction', async () => {
...
-  await expect(connection.getSignatureStatus(signature)).resolves.toEqual({
-    Ok: null,
-  });
+
+  const response = await connection.getSignatureStatus(signature);
+  if (response !== null) {
+    expect(typeof response.slot).toEqual('number');
+    expect(response.status).toEqual({Ok: null});
+  } else {
+    expect(response).not.toBeNull();
```
CriesofCarrots: Is this pattern right?
```
if (reponse === null) {
  expect(response).not.toBeNull();
}
```

jstarry: Yeah, unfortunately needed to add the if statement to make flow happy

CriesofCarrots: Oh, flow... 😕

module.flow.js
jstarry: Will incorporate this here when I rebase: https://github.com/solana-labs/solana-web3.js/pull/742

PR #763, opened 2020-03-26
feat: update getSignatureStatus api

jstarry:
Depends on https://github.com/solana-labs/solana/pull/9069

### DIFF
--- a/test/connection.test.js
+++ b/test/connection.test.js
@@ -877,9 +877,52 @@ test('request airdrop - max commitment', async () => {
     },
   ]);
 
-  await connection.requestAirdrop(account.publicKey, minimumAmount + 40);
+  const signature = await connection.requestAirdrop(
+    account.publicKey,
+    minimumAmount + 40,
+  );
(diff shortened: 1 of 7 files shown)


test/connection.test.js
```diff
@@ -1066,24 +1095,47 @@ test('transaction', async () => {
...
+
+  if (typeof firstResponse.confirmations !== 'number') {
+    expect(typeof firstResponse.confirmations).toEqual('number');
+    return;
+  }
+
+  expect(firstResponse.confirmations).toBeGreaterThanOrEqual(
+    responseConfirmations,
+  );
+
```
CriesofCarrots: What do you think about adding a test case here for `confirmations: null`? Should be able to loop on getSignatureStatus with commitment: 'max' to know when the transaction is rooted for the live test.

Seems like a pretty easy add, and it'd be nice to have that tested in an integration test somewhere, at least.

jstarry: Yes, great call!

PR #774, opened 2020-03-30
feat: rename testnet util methods

jstarry:
#### Problem
* Edge / Beta devnets don't exist
* Not possible to select `testnet` or `mainnet-beta`
* We refer to testnets as "clusters" now

#### Changes
* Rename `testnetChannelEndpoint` function to `clusterApiUrl`
* No longer pull the default cluster from `package.json` key `"testnetDefaultChannel"`
* Set the default cluster to `"devnet"`
* Add `Cluster` type to restrict key passed to `clusterApiUrl` util method


src/util/cluster.js
```diff
@@ -0,0 +1,36 @@
+//@flow
+
+/**
+ * @private
+ */
+const endpoint = {
+  http: {
+    devnet: 'http://devnet.solana.com',
+    testnet: 'http://tds.solana.com',
```
mvines:
```suggestion
    testnet: 'http://testnet.solana.com',
```

src/util/cluster.js
```diff
@@ -0,0 +1,36 @@
...
+    'mainnet-beta': 'http://api.mainnet-beta.solana.com',
+  },
+  https: {
+    devnet: 'https://devnet.solana.com',
+    testnet: 'https://tds.solana.com',
```
mvines:
```suggestion
    testnet: 'https://testnet.solana.com',
```

jstarry:
> It's time to rename tds.solana.com to testnet.solana.com! I'll do that in DNS/TLS today, so just assume it exists for now :)

woot!

PR #831, opened 2020-04-21
feat: add getConfirmedTransaction and getConfirmedSignaturesForAddress

jstarry:
#### Changes
* Add support and tests for `getConfirmedTransaction` and `getConfirmedSignaturesForAddress`

### DIFF
--- a/module.d.ts
+++ b/module.d.ts
@@ -98,6 +98,17 @@ declare module '@solana/web3.js' {
     }>;
   };
 
+  export type ConfirmedTransaction = {
+    slot: number;
+    transaction: Transaction;
+    meta: {
+      fee: number;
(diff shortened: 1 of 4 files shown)


module.d.ts
```diff
@@ -98,6 +98,17 @@ declare module '@solana/web3.js' {
     }>;
   };
 
+  export type ConfirmedTransaction = {
+    slot: number;
+    transaction: Transaction;
+    meta: {
```
CriesofCarrots, line 104: Worth creating a type for this, since it's a dupe of meta in ConfirmedBlock?

jstarry: @CriesofCarrots good call, done!

PR #856, opened 2020-04-30
feat: expose transaction message publicly

jstarry:
#### Problem
- No way to map account ids from confirmed transaction metadata balances to transaction accounts because `Message` was not exposed
- Not easy to determine which accounts in a transaction are readonly

#### Changes
- Create `Message` class and move some serialization logic from `Transaction`
- Add `MessageHeader` and `CompiledInstruction` types
- Add `Transaction.compileMessage()` method to expose message data
- Add `Message.isAccountWritable()` convenience method

### DIFF
--- a/src/message.js
+++ b/src/message.js
@@ -1,0 +1,170 @@
+// @flow
+
+import bs58 from 'bs58';
+import * as BufferLayout from 'buffer-layout';
+
+import {PublicKey} from './publickey';
+import type {Blockhash} from './blockhash';
+import * as Layout from './layout';
+import {PACKET_DATA_SIZE} from './transaction';
+import * as shortvec from './util/shortvec-encoding';
+
+/**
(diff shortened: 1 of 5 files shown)


src/message.js
```diff
@@ -0,0 +1,170 @@
...
+ * @property {number} numReadonlyUnsignedAccounts The last `numReadonlySignedAccounts` of the unsigned keys are read-only accounts
+ */
+export type MessageHeader = {
+  numRequiredSignatures: number,
+  numReadonlySignedAccounts: number,
+  numReadonlyUnsignedAccounts: number,
+};
+
+/**
+ * An instruction to execute a program
```
mvines, line 27:
```suggestion
 * An instruction to execute by a program
```

jstarry:
man that commit message linter is so annoying

PR #883, opened 2020-05-19
feat: add root and single commitment levels

jstarry:
#### Problem
`root` and `single` commitment levels are not supported.

---
Unfortunately `sendAndConfirmTransaction` and `sendAndConfirmRawTransaction` are currently incorrect (before and after this change) and are not fixable without RPC changes.

This is because `getSignatureStatuses` does not accept a commitment level param and always returns the confirmation count as seen by the connected node. It doesn't support fetching confirmation count as seen by the _cluster_. This broke when we separated the notion of `root` and `max` commitment level and is also an issue for accurately returning status for `single` commitment. Issue here: https://github.com/solana-labs/solana/issues/10108

### DIFF
--- a/module.d.ts
+++ b/module.d.ts
@@ -46,7 +46,7 @@ declare module '@solana/web3.js' {
     value: T;
   };
 
-  export type Commitment = 'max' | 'recent';
+  export type Commitment = 'max' | 'recent' | 'root' | 'single';
 
   export type SignatureStatusConfig = {
     searchTransactionHistory: boolean;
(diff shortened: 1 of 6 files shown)


jstarry:
> This is because getSignatureStatuses does not accept a commitment level param and always returns the confirmation count as seen by the connected node.

This was stated incorrectly. We will only return `confirmations: null` if the cluster has finalized the transaction.

jstarry:
I misunderstood how `getSignaturesStatuses` works so I'll need to rethink this

PR #886, opened 2020-05-19
chore: fix flaky test and stop running broken examples in ci

jstarry:


jstarry:
Found these flaky tests when looping test runs in an attempt to repro the CI failures on latest node. Unfortunately didn't find the cause of CI failure

PR #897, opened 2020-05-20
Add new commitment levels and update tx confirmation apis

jstarry:
Fixes: https://github.com/solana-labs/solana-web3.js/issues/792
Fixes: https://github.com/solana-labs/solana-web3.js/issues/793
Fixes: https://github.com/solana-labs/solana-web3.js/issues/892

#### Problem
RPC has new commitment levels and the behaviour of the `getSignatureStatuses` endpoint has shifted to represent cluster confirmations rather than the incorrect notion of node votes as "confirmations".

#### Changes
This felt like a good time to clean up our apis around transaction confirmation
* Removed `Connection.confirmTransactionAndContext` API
* Updated `Connection.confirmTransaction` to poll the `getSignatureStatuses` RPC API until a specified confirmation count has been reached or the transaction fails
* Added `confirmations` param to both `sendAndConfirmTransaction` and `sendAndConfirmRawTransaction` to allow waiting for a particular number of confirmations (or finalization in the default case)
* Removed the `sendAndConfirmRecentTransaction` API since it's ripe for misuse. Its behaviour can be achieved by passing `0` confirmations to `sendAndConfirmTransaction`. 
* Removed the `AccountInUse` error check from `sendAndConfirmTransaction` since that error never gets committed to the status cache
* In order to allow passing a `confirmations` param to `sendAndConfirmTransaction` I changed the `...signers` varargs-style param to a single array param and updated `Connection.sendTransaction` to have a consistent API.

### DIFF
--- a/module.d.ts
+++ b/module.d.ts
@@ -46,7 +46,7 @@ declare module '@solana/web3.js' {
     value: T;
   };
 
-  export type Commitment = 'max' | 'recent';
+  export type Commitment = 'max' | 'recent' | 'root' | 'single';
 
   export type SignatureStatusConfig = {
     searchTransactionHistory: boolean;
--- a/module.d.ts
+++ b/module.d.ts
@@ -210,14 +210,10 @@ declare module '@solana/web3.js' {
       endSlot: number,
     ): Promise<Array<TransactionSignature>>;
     getVoteAccounts(commitment?: Commitment): Promise<VoteAccountStatus>;
-    confirmTransactionAndContext(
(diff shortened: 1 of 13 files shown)


jstarry:
This change updates the API signature of some of the most commonly used API's in the SDK so I would appreciate some extra eyes on this. But I mostly just wanted everyone to be aware of these pending changes because of their impact downstream

jackcmay:
lgtm, nice cleanup!

PR #905, opened 2020-05-21
feat: add commitment param to subscription apis

jstarry:
Fixes: https://github.com/solana-labs/solana-web3.js/issues/904

#### Problem
RPC API supports commitment param but the sdk doesn't provide a way to pass it


src/connection.js
```diff
@@ -1962,16 +1983,19 @@ export class Connection {
    *
    * @param signature Transaction signature string in base 58
    * @param callback Function to invoke on signature notifications
+   * @param commitment Specify the commitment level account changes must reach before notification
```
CriesofCarrots, line 1986:
```suggestion
   * @param commitment Specify the commitment level signature must reach before notification
```

PR #963, opened 2020-06-10
feat: remove commitment param from requestAirdrop method

jstarry:
#### Problem
The `requestAirdrop` method on the RPC API service used to wait until a status was available at the specified `commitment` level before returning a response. But after https://github.com/solana-labs/solana/pull/10446, it only uses the `commitment` param to determine which blockhash to use for the airdrop transaction.

There's no reason I can think of that a developer would want to use a more recent blockhash so I think it's best to remove this parameter from the `Connection.requestAirdrop` API to avoid confusion and encourage best practices.

#### Changes
- Remove `commitment` param from `requestAirdop`
- Fix failing CI

### DIFF
--- a/module.d.ts
+++ b/module.d.ts
@@ -281,7 +281,6 @@ declare module '@solana/web3.js' {
     requestAirdrop(
       to: PublicKey,
       amount: number,
-      commitment?: Commitment,
     ): Promise<TransactionSignature>;
     sendTransaction(
       transaction: Transaction,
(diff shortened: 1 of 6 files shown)


jstarry:
@mvines I need some more web3 changes to go in for explorer so pushed this through FYI

mvines:
LMK, commitment level doesn't make much sense in this context

PR #968, opened 2020-06-12
fix: avoid double spend in sendAndConfirmTransaction

jstarry:
#### Problem
Now that RPC retries transactions, it's more likely that `sendAndConfirmTransaction` inadvertently does a double spend. This is because `sendAndConfirmTransaction` will send a new transaction if `connection.confirmTransaction` times out (https://github.com/solana-labs/solana-web3.js/issues/967) and `sendTransaction` will always fetch a new blockhash for a duplicate transaction.

Fixes: https://github.com/solana-labs/solana-web3.js/issues/967

#### Changes
- Remove retry loop from `sendAndConfirmTransaction`
- Update stale timing constants
- Increase wait timeout in `confirmTransaction`
- Allow options to be undefined in `ConfirmOptions` and `SendOptions` for better dev UX
- Fix up docs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: required
 language: node_js
 node_js:
   - "lts/*"
-  - "node"
+  # - "node" disabled due to flaky test failures
 
 services:
   - docker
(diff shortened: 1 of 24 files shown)


mvines:
Sorry @jstarry, I didn't mean to close this.  It was a side-effect of the history rewrite

jstarry:
No worries, moved it over

PR #977, opened 2020-08-09
Fix createProgramAdress for bundled applications

hendrikhofstadt:
The way bn.js checks if Buffers are supported does not work with bundled applications that don't allow `require` at runtime.
Using toArray works nicely because most libraries (including the curve point check) do accept `Uint8Array`

https://github.com/indutny/bn.js/blob/5707aed66c02ae10f8c7d59f5244fcce1c24cdb6/lib/bn.js#L51-L55

### DIFF
--- a/src/publickey.js
+++ b/src/publickey.js
@@ -110,7 +110,7 @@ export class PublicKey {
       Buffer.from('ProgramDerivedAddress'),
     ]);
     let hash = await sha256(new Uint8Array(buffer));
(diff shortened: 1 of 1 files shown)


mvines:
Hey, confusingly the source for web3.js is now at https://github.com/solana-labs/solana/tree/master/web3.js.  This repo is just a mirror that exists for the purpose of shipping releases.  

Would you be able to move your PR here?

hendrikhofstadt:
https://github.com/solana-labs/solana/pull/11498

PR #980, opened 2020-08-24
Double the wait time between each 429 induced retry.

dafyddd:


mvines:
Would you mind sending this PR to https://github.com/solana-labs/solana/tree/master/web3.js instead?  This repo is actually read-only. 

CI is going to make you change your commit title to something like `fix: double the wait time between each 429 induced retry` as well.  

Thanks!

dafyddd:
> Would you mind sending this PR to https://github.com/solana-labs/solana/tree/master/web3.js instead? This repo is actually read-only.
> 
> CI is going to make you change your commit title to something like `fix: double the wait time between each 429 induced retry` as well.
> 
> Thanks!

Ok I think I've done it correctly this time.

PR #996, opened 2020-10-27
Add inner instructions to ConfirmedTransactionMeta

vtsurka:


src/connection.js
```diff
@@ -367,18 +367,25 @@ const SimulatedTransactionResponseValidator = jsonRpcResultAndContext(
...
  *
  * @typedef {Object} ConfirmedTransactionMeta
  * @property {number} fee The fee charged for processing the transaction
+ * @property {Array<InnerInstruction>} An array of cross program invoked instructions
  * @property {Array<number>} preBalances The balances of the transaction accounts before processing
  * @property {Array<number>} postBalances The balances of the transaction accounts after processing
  * @property {Array<string>} logMessages An array of program log messages emitted during a transaction
  * @property {object|null} err The error result of transaction processing
  */
 type ConfirmedTransactionMeta = {
```
jstarry, line 386: I think we'll need both a `ParsedConfirmedTransactionMeta` and a `ConfirmedTransactionMeta`. `ParsedConfirmedTransactionMeta` can be used for `ParsedConfirmedTransaction` and `ConfirmedTransactionMeta` can be used for `ConfirmedTransaction`

And then we can have a `ParsedInnerInstruction` for parsed instructions and `InnerInstruction` for hex encoded instruction data

src/connection.js
```diff
@@ -1126,6 +1133,18 @@ const ConfirmedTransactionMetaResult = struct.union([
...
+    innerInstructions: struct.array([
+      struct({
+        index: 'number',
+        instructions: struct.array([
+          struct({
+            accounts: struct.array(['number']),
+            data: 'string',
+            programIdIndex: 'number',
+          }),
+        ])}
```
jstarry, line 1145: We'll need the following struct for parsed inner instructions, what you have is fine for non-parsed

```rust
        instructions: struct.array([
          struct.union([
            struct({
              accounts: struct.array(['string']),
              data: 'string',
              programId: 'string',
            }),
            struct({
              parsed: 'any',
              program: 'string',
              programId: 'string',
            }),
          ]),
        ]),
```

src/connection.js
```diff
@@ -1126,6 +1158,25 @@ const ConfirmedTransactionMetaResult = struct.union([
   struct.pick({
     err: TransactionErrorResult,
     fee: 'number',
+    innerInstructions: struct.array([
```
jstarry, line 1161: We'll need both a `ConfirmedTransactionMetaResult` and a `ParsedConfirmedTransactionMetaResult`. Can you please add `ParsedConfirmedTransactionMetaResult` and look at `ConfirmedTransactionResult` and `ParsedConfirmedTransactionResult` for the difference between how instructions are shaped in those validator structs?

jstarry: Also, since `innerInstructions` is a relatively new feature, a developer could use a local cluster which doesn't return this information so we need to use `struct.union(['undefined', ...])`. And also `innerInstructions` can be `null` for older transactions that were processed before we started tracking inner instructions so you'll want to use `struct.union(['undefined', 'null', ...])`

jstarry:
@vtsurka I just realized this PR isn't on the monorepo. Could you copy your changes over to a branch on the https://github.com/solana-labs/solana repo? This web3 repo is just a mirror

jstarry:
Also, don't forget to run `npm run lint:fix` to avoid linter issues

PR #1049, opened 2021-08-10
feat: support arbitrary instruction index for secp instructions

piazzatron:
Adds the ability to include multiple secp recovery instructions in a single transaction, and to include those instructions in places other than the 0th index. Tested and working as expected.


oJshua:
Applied patch + moved PR to main solana repo. 

https://github.com/solana-labs/solana/pull/19162

PR #1050, opened 2021-09-14
getBlocks API is added

kamaci:
getBlocks API is added

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -3072,6 +3072,26 @@ export class Connection {
   }
 
   /**
+   * Fetch confirmed blocks between two slots
+   */
+  async getBlocks(
+    startSlot: number,
+    endSlot?: number,
+    commitment?: Finality,
+  ): Promise<Array<number>> {
+    const args = this._buildArgsAtLeastConfirmed(
+      [startSlot, endSlot],
+      commitment,
+    );
+    const unsafeRes = await this._rpcRequest('getBlocks', args);
+    const res = create(unsafeRes, jsonRpcResult(array(number())));
+    if ('error' in res) {
+      throw new Error('failed to get blocks: ' + res.error.message);
+    }
+    return res.result;
+  }
+
+  /**
    * Fetch a list of Signatures from the cluster for a confirmed block, excluding rewards
    */
   async getConfirmedBlockSignatures(
--- a/test/connection.test.ts
+++ b/test/connection.test.ts
@@ -2077,6 +2077,17 @@ describe('Connection', () => {
     );
   });
 
+  it('get blocks', async () => {
+    await mockRpcResponse({
+      method: 'getBlocks',
+      params: [0, 1],
+      value: [1, 2, 3, 4, 5],
+    });
+
+    const slotLeaders = await connection.getBlocks(0, 1);
+    expect(slotLeaders).to.have.length(5);
+  });
+
   it('get confirmed block signatures', async () => {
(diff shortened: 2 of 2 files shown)


src/connection.ts
```diff
@@ -3071,6 +3071,23 @@ export class Connection {
...
 
+  /**
+   * Fetch confirmed blocks between two slots
+   */
+  async getBlocks(
+    startSlot: number,
+    endSlot?: number,
+    commitment?: Finality
+  ): Promise<Array<number>> {
+    const args = this._buildArgsAtLeastConfirmed([startSlot, endSlot], commitment);
```
jstarry, line 3082: I don't think this will work when `endSlot` is not defined. Can you fix and add a test please?

test/connection.test.ts
```diff
@@ -2077,6 +2077,17 @@ describe('Connection', () => {
     );
   });
 
+  it('get blocks', async () => {
+    await mockRpcResponse({
+      method: 'getBlocks',
+      params: [0, 1],
+      value: [1, 2, 3, 4, 5],
```
jstarry, line 2084: I would prefer if this value is realistic.. we shouldn't return 5 blocks for a 2 slot range

test/connection.test.ts
```diff
@@ -2077,6 +2077,17 @@ describe('Connection', () => {
...
+    });
+
+    const slotLeaders = await connection.getBlocks(0, 1);
+    expect(slotLeaders).to.have.length(5);
```
jstarry, line 2088:
```suggestion
    const blocks = await connection.getBlocks(0, 1);
    expect(blocks).to.have.length(5);
```

jstarry (commented):
Thanks for adding this, just a few comments

jstarry:
Oh actually do you mind opening this PR over in the monorepo? This repo is just a mirror, thanks!

kamaci:
> Oh actually do you mind opening this PR over in the monorepo? This repo is just a mirror, thanks!

Sure!

test/connection.test.ts
kamaci: I have put something similar to that one: https://docs.solana.com/developing/clients/jsonrpc-api#getblocks

However, I can replace it with current real values:

```
curl https://solana-api.projectserum.com -X POST -H "Content-Type: application/json" -d '
  {"jsonrpc": "2.0","id":1,"method":"getBlocks","params":[96460288, 96460330]}
'
{"jsonrpc":"2.0","result":[96460288,96460289,96460290,96460291,96460292,96460293,96460294,96460295,96460296,96460297,96460298,96460299,96460300,96460301,96460302,96460303,96460304,96460305,96460306,96460307,96460308,96460309,96460310,96460311,96460316,96460317,96460318,96460319,96460320,96460321,96460322,96460323,96460324,96460325,96460326,96460327,96460328,96460329,96460330],"id":1}
```

jstarry: > I have put something similar to that one: https://docs.solana.com/developing/clients/jsonrpc-api#getblocks

Yeah 5-10 sounds great!

src/connection.ts
kamaci: Do you mean that:

`endSlot ? [startSlot, endSlot] : [startSlot]`

jstarry: correct

kamaci:
@jstarry I've created https://github.com/solana-labs/solana/pull/19854

PR #1057, opened 2021-11-18
add BpfUpgradeableLoader

fanatid:
https://docs.rs/solana-program/1.8.3/solana_program/bpf_loader_upgradeable/index.html


jstarry:
Hey thanks @fanatid! What problem are you solving with this change? I was thinking we could deprecate the bpf loader with maybe the exception of the upgrade instruction which is used by multisig and gov contracts

fanatid:
I'd like to use it in tests for multisig contracts :)

fanatid:
@jstarry is there anything that I need to change in this PR?

src/bpf-loader-upgradeable.ts
```diff
@@ -0,0 +1,699 @@
...
+      programId: this.programId,
+      data,
+    });
+  }
+}
+
+/**
+ * BpfLoaderUpgradeable program interface
+ */
+export class BpfLoaderUpgradeable {
```
jstarry, line 400: I would prefer to not include this class and all of its methods inside web3. I think they could live inside a separate package if needed but I don't think it's too valuable to deploy programs using JS right now. It's a complex process to ensure that all transactions get processed properly and I think that it's fine to keep that process only in the CLI for now. It's too complex to maintain in web3.js for now.

fanatid: :confused: is it possible to store and maintain a separated package with `BpfLoaderUpgradeable` under `solana-labs` organization or I would need to store and distribute under my own account?

jstarry: Separate account for now until we setup a place for solana community web3 packages. Sorry about that! Maintenance can be a burden 😉

jstarry:
@fanatid can you please reopen the PR in the main monorepo?

axleiro:
Sorry @fanatid, it was a mistake

fanatid:
@axleiro it's ok, move PR is not hard. I was more interested in why repo history was rewritten... :disappointed:

PR #1059, opened 2021-12-23
feat: add getMultipleAccountsInfoAndContext method to Connection

kklas:
Similar to `getAccountInfoAndContext`.


mvines:
This repo is just used for release management.  The web3.js source is located here: https://github.com/solana-labs/solana/tree/master/web3.js

Thanks!

PR #1065, opened 2022-02-27
Allow for custom recentBlockhash to be supplied

Totenfluch:
This enables us to supply a custom recentBlockhash. The use-case is for a transaction to be able to be re-broadcasted with the same blockhash when it is dropped from the mempool.

One question along this PR: Why do we use for(;;) = while(true) in this piece of code? In the end this only checks for one signature doesn't it?


jstarry:
Hey, thanks for contributing. Please make your changes in the monorepo (https://github.com/solana-labs/solana), this is a read-only mirror repo

PR #1068, opened 2022-04-05
feat: Add Memo Program

Anurag5086:
Added the typescript implementation of SPL Memo Program.


joncinque:
Thanks for the contribution! This repo is just a mirror of the monorepo, so please make the PR to https://github.com/solana-labs/solana instead

CriesofCarrots:
I am of the opinion that this should live in the SPL repo, similar to the spl-token typescript library.
I have a draft up here: https://github.com/solana-labs/solana-program-library/pull/2583 But I was hoping for someone who actually knows about JS/TS packaging to get it over the line. We might also want to add an instruction parser.

joncinque:
That's fair.  Since memos are exposed in `getSignaturesForAddress`, and it only comes down to one exported function, it seemed overkill to have a separate package, and that memos could live happily alongside the system programs.  I can appreciate that it's still BPF, and not a system program.

And I wasn't aware of your PR, thanks for flagging that!  We'll move over towards SPL in that case.

PR #1070, opened 2022-04-28
Add bigint support to system transfers

konytech:
For big transfers we currently only rely on 'number', which is dangerous when sending huge amounts of lamports.

Changes:
- Add bigint support to system transfers
- Install buffer-layout-utils with `npm install @solana/buffer-layout-utils` from `npm 8.5.5`

This follows the same encoding/decoding layout used by the spl-token library:
https://github.com/solana-labs/solana-program-library/blob/master/token/js/src/instructions/transfer.ts#L56

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,7 @@
         "@babel/runtime": "^7.12.5",
         "@ethersproject/sha2": "^5.5.0",
         "@solana/buffer-layout": "^4.0.0",
+        "@solana/buffer-layout-utils": "^0.2.0",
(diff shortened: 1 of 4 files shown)


konytech:
@steveluscher 
@yihau 
Sorry to randomly ping you on this one, is there a way someone can help me out with this?

konytech:
image

jstarry:
Hey, thanks for contributing. Please make your changes in the monorepo (https://github.com/solana-labs/solana), this is a read-only mirror repo

PR #1075, opened 2022-07-11
Safeguard for executing withdraws that will clean out a vote account

aschonfeld:
In our time working with the solana validator we've run into an issue where we try to withdraw rewards, but the withdraw results in a balance which is lower than the  minimum requirement for rent.  Thus causing our vote account to be closed.

I've added a check for VoteProgram 'withdraw' transaction instructions in the `sendAndConfirmTransaction` function which will throw an exception in this case.


FarhadHashimov:
#1075

PR #1077, opened 2022-09-23
Fixed the error display

GDdark:
Fixed the error display


steveluscher:
Hi there! Thanks for this. First of all, this repo is only a mirror. Can you reopen this on https://github.com/solana-labs/solana? Second of all, when you do, I'd love to hear more. What was the bad display, and what does it look like after this fix?

PR #1086, opened 2023-02-01
Update Mergify configuration

steveluscher:
Copying over useful rules from the monorepo.


.mergify.yml
```diff
@@ -1,23 +1,61 @@
...
+    conditions:
+      - and:
+          - status-success=all-web3-checks
+          - status-success=web3-commit-lint
+          - label=automerge
+          - label!=no-automerge
     actions:
       merge:
-        method: rebase
-        strict: true
```
steveluscher, line 9: This was a syntax error.

Issue #1103, opened 2022-08-03
web3.js: replace dependencies to lower supply-chain risk, improve tree-shakeability, and reduce bundle size


paulmillr:
You have quite a few dependencies in web3. All of those can be replaced by 4 deps:

```
"@ethersproject/sha2": "^5.5.0",
"bigint-buffer": "^1.1.5",
"bn.js": "^5.0.0",
"bs58": "^4.0.1",
"buffer": "6.0.1",
"js-sha3": "^0.8.0",
"secp256k1": "^4.0.2",
"tweetnacl": "^1.0.0"
```

I suggest switching to [audited libraries](https://paulmillr.com/noble/) that are faster, use less dependencies.

- `@noble/hashes` for sha2, sha3. It's faster than ethersproject/sha2, js-sha3, and was audited
- `@noble/secp256k1` for secp256k1. It's faster than js implementations of secp, and was audited
- `@noble/ed25519` for ed25519 — which is 7x faster than tweetnacl, and was audited
- `@scure/base` could be used for `base58` — it was audited and has 0 deps
- `bn.js`, `bigint-buffer`, `buffer` can be removed altogether

Supply chain attacks are very dangerous. Breaking any sub-dependency of your dependencies could leak user keys. So, the best option is to use projects that don't have sub-dependencies.

---
Individual tracking tasks

- [x] solana-labs/solana#26933
- [x] solana-labs/solana#27390
- [ ] solana-labs/solana-web3.js#1100
- [x] solana-labs/solana-web3.js#1101
- [x] solana-labs/solana#27629

paulmillr:
@steveluscher heads up, much smaller v2 of noble-secp, noble-ed is out. Many features have been removed, but should be fine for your case. The libraries are just 4kb gzipped now.

Another option is to use noble-curves which have improved security, are larger; but tree shaking is supported so all is cool.

steveluscher:
Thanks, @paulmillr!

steveluscher:
Closed by 5a20ed0649fd0fa28d4b6a96ba042c9cdb47630e.

PR #1142, opened 2023-02-02
fix: eliminate bn.js and replace public key storage with JavaScript `bigint`

steveluscher:
#### Problem

JavaScript `bigint` has been available in browsers since September 2020. Despite that we've implemented large value storage in `PublicKey` using the `bn.js` library. This library presently makes up ~13% of `@solana/web3.js`, or around 35K uncompressed.

image

Replacing this system with pure `bigint` storage would cut the size of this library down considerably.

#### Summary of Changes

* Patch `borsh` to serialize to and from `bigint` rather than `BN` instances. A [pull request](https://github.com/near/borsh-js/pull/60) has been sent to the upstream project.
* Replace `BN` storage in `PublicKey` with `bigint`

#### Bundle size

Using `package-build-stats`:

```
const p = require('package-build-stats');
p.getPackageStats('~/solana/web3.js/').then(r => console.log(r))
```

**Size before**: 75299 gzipped bytes
**Size after**: 64191 gzipped bytes

Reduction of ~14%.

#### Notes

* Anyone who used to rely on accessing the private variable `PublicKey::_bn` will find that its type has changed from `BN` to `bigint`. Such is the risk when you reach into private variables.
* This change relies on patching `borsh-js` in a way that represents a breaking change. For this reason we modify the bundle script to bundle the modified version _into_ the `web3.js` bundle. This has two implications:
  * An app that uses `borsh` outside of `web3.js` will end up bundling a second copy of it (presumably one that includes 'bn.js` which negates the benefits of this change)
  * Our patched version is accessible only within the confines of `web3.js` and will not cause a breaking change to the rest of the app.

Addresses #1103.


wireless-anon:
will this break the old `PublicKey.equals` when you pass in a new `PublicKey` object (i.e. a `bigint`-based one)?

concern is that calling a function from a library that uses an older version of `@solana/web3.js` and passing in a `bigint`-based `PublicKey` may fail

steveluscher:
> will this break the old PublicKey.equals when you pass in a new PublicKey object (i.e. a bigint-based one)?

Most certainly; good call-out.

Options:

1. Breaking version bump.
2. Add `@solana/web3.js` as a `peerDependency` of itself, meaning that in order to install it installing an older version of it somewhere else makes it fail the peer dependency check.

I'm not sure if 2 would actually ‘work,’ but it does in my head. Thoughts, @joncinque, @jordansexton?

PR #1179, opened 2023-02-21
fix: locale independent sort

### LINKED ISSUE #1094: [web3.js] Figure out a better way to sort pubkeys than using `localeCompare`
#### Problem

This immediately smells, because sorting should _never_ depend on locale. Indeed, it appears, through submissions like solana-labs/solana#28234, that different environments behave differently, which is _extremely_ bad news and will lead to transaction failures.

#### Proposed Solution

Either:

1. Implement sorting in a way that replicates the canonical `localCompare` result in a way that _doesn't_ rely on locale, or
2. Be explicit in our call to `localeCompare` (ie. supply both a hard-coded locale, and a static set of `Intl.Collator` [options](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator#options))

bw-solana:
### Problem

Need deterministic pubkey sorting that is not dependent on browser locale

### Summary of Changes
Supply the default options listed on [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator) and specify a static locale (`'en'`).

Fixes #1094.

### DIFF
--- a/packages/library-legacy/src/transaction/legacy.ts
+++ b/packages/library-legacy/src/transaction/legacy.ts
@@ -457,7 +457,17 @@ export class Transaction {
         return x.isWritable ? -1 : 1;
       }
       // Otherwise, sort by pubkey, stringwise.
-      return x.pubkey.toBase58().localeCompare(y.pubkey.toBase58());
+      const options = {
+        localeMatcher: 'best fit',
+        usage: 'sort',
+        sensitivity: 'variant',
+        ignorePunctuation: false,
+        numeric: false,
+        caseFirst: 'lower',
+      };
+      return x.pubkey
+        .toBase58()
+        .localeCompare(y.pubkey.toBase58(), 'en', options);
     });
 
     // Move fee payer to the front


steveluscher:
Wow. I missed this. All that's missing is a 100% airtight test for this.

steveluscher:
Currently running this test: 

```js
const {PublicKey} = require('./lib/index.cjs.js');

// Defaults from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator#options
const DEFAULT_COLLATOR_OPTIONS = {
  caseFirst: 'lower',
  ignorePunctuation: false,
  localeMatcher: 'best fit',
  numeric: false,
  sensitivity: 'variant',
  usage: 'sort',
};

function pubkeyCompareForSort(a, b) {
  const pubkeyStringA = a.toBase58();
  const pubkeyStringB = b.toBase58();
  return pubkeyStringA.localeCompare(
    pubkeyStringB,
    'en', // Choose a static locale so the sort is always stable.
    DEFAULT_COLLATOR_OPTIONS,
  );
}

for (
  let ii = 0n;
  ii <
  BigInt('0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff');
  ii++
) {
  for (
    let jj = 0n;
    jj <
    BigInt(
      '0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff',
    );
    jj++
  ) {
    const a = new PublicKey(ii);
    const b = new PublicKey(jj);
    if (
      a.toBase58().localeCompare(b.toBase58()) !== pubkeyCompareForSort(a, b)
    ) {
      throw new Error(`${ii.toString()} ${jj.toString()}`);
    }
  }
}
```

steveluscher:
That test should be complete sometime next week.

steveluscher:
Yeah, I'm pretty confident this is gtg. Thanks!

PR #1188, opened 2023-03-04
Using signersOrOptions for VersionedTransaction

febo:
For `VersionedTransaction`, the method `sendTransaction` currently passes to `sendRawTransaction` the third parameter `options`, which is always undefined – `SendOptions` is defined as the the second parameter `signersOrOptions` for `VersionedTransaction`.

This PR changes the parameter of `sendRawTransaction` to `signersOrOptions` instead of `options` when handling `VersionedTransaction`.


steveluscher:
thank_you___gif__by_flavia_elric-dbe01nw

PR #1189, opened 2023-03-07
Add import documentation to `sign()`

Henry-E:
This seems like a pretty important thing to know about the new `sign` method and is only mentioned as a footnote on another random solana page unrelated to the web3.js docs.


Henry-E:
It seems like maybe it's the tutorial that's out of date since the behaviour of the `sign` function does allow for partial signatures. 🤷

PR #1190, opened 2023-03-09
[experimental] A types-only RPC implementation using JavaScript proxies

steveluscher:
This PR introduces an implementation of the client RPC that does as little work as possible. It essentially passes calls to the server RPC unmodified, and enforces correctness through nothing but Typescript types.

# Problem

The `Connection` class today contains _concrete implementations_ for every RPC method, past, present, and future. This implies that the bundle size of this library may grow unbounded as the number of RPC methods grows.

# Solution

The RPC interface is simple and conforms to JSON RPC 2.0. The only things that are required to make an RPC call are a method name and a list of parameters.

The solution herein creates an rpc object on which you can call _any_ method. That method name and the params get passed through, unmodified.

```ts
// Example: this call...
rpc.doSomething(transport, 1, 'foo', {bar: 'baz'});

// ...gets put on the wire as a JSON RPC call of the format:
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "doSomething",
  "params": [1, "foo", {"bar": "baz"}]
}
```

In this way, the set of RPC methods can grow, unbounded, without increasing the size of the client library at all.

Correctness, in this world, is enforced only by Typescript. The input params are completely specified on the JavaScript `Proxy`, and the return types change with the inputs given. See the complicated `getAccountInfo` commit on this PR for an example of how this is achieved using function overloads.

Crucially, this library will never concern itself with how to _send_ RPC calls – only with how you should format your calls and reason about the return values. The _sending_ is left up to the transport, which you can read about in https://github.com/solana-labs/solana-web3.js/pull/1187.

# How to review this PR

Click the ‘commits’ tab and treat each commit as though it were its own PR.


packages/rpc-core/src/types/rpc-methods/getBlockHeight.d.ts
```diff
@@ -0,0 +1,20 @@
...
+     */
+    getBlockHeight(
+        transport: IJsonRpcTransport,
+        config?: readonly {
+            // Defaults to `finalized`
+            commitment?: Finality;
+            // The minimum slot that the request can be evaluated at
+            minContextSlot?: Slot;
+        }
+    ): Promise<GetBlockHeightApiResponse>;
```
steveluscher, line 19: See https://docs.solana.com/api/http#getblockheight.

packages/rpc-core/src/types/rpc-methods/getBlocks.d.ts
```diff
@@ -0,0 +1,18 @@
...
+     */
+    getBlocks(
+        transport: IJsonRpcTransport,
+        startSlot: Slot,
+        endSlotInclusive?: Slot,
+        config?: readonly {
+            // Defaults to `finalized`
+            commitment?: Exclude<Finality, 'processed'>;
+        }
+    ): Promise<GetBlocksApiResponse>;
```
steveluscher, line 17: See https://docs.solana.com/api/http#getblocks.

packages/rpc-core/src/types/rpc-methods/getAccountInfo.d.ts
```diff
@@ -0,0 +1,92 @@
...
+            GetAccountInfoApiBase64EncodingCommonConfig
+    ): Promise<GetAccountInfoApiResponseBase & GetAccountInfoApiResponseWithEncodedZStdCompressedData>;
+    getAccountInfo(
+        transport: IJsonRPCTransport,
+        address: Base58EncodedAddress,
+        config?: readonly {
+            encoding: 'jsonParsed';
+        } &
+            GetAccountInfoApiCommonConfig
+    ): Promise<GetAccountInfoApiResponseBase & GetAccountInfoApiResponseWithJsonData>;
```
steveluscher, line 91: This is a much more complicated example, where you can see how modulating the inputs changes the output type (eg. specifying different `encoding` results in the output being different.

Things to pay attention to:

* `jsonParsed` encoding results in `data` being a parsed data structure _or_ `[bytes, encoding]` as a fallback if no parser is available.
* only `base64` encoding accepts `dataSlice` as an input

See https://docs.solana.com/api/http#getaccountinfo.

packages/rpc-core/src/rpc.ts
```diff
@@ -0,0 +1,22 @@
+import { IJsonRpcTransport } from '@solana/rpc-transport';
+import { JsonRpcApi } from './types/jsonRpcApi';
+
+export const rpc = /* #__PURE__ */ new Proxy<JsonRpcApi>({} as JsonRpcApi, {
```
joncinque, line 4: TIL what a Proxy is, very cool!

packages/rpc-core/src/rpc.ts
```diff
@@ -0,0 +1,22 @@
...
+        return false;
+    },
+    get<TMethodName extends keyof JsonRpcApi>(target: JsonRpcApi, p: TMethodName) {
+        if (target[p] == null) {
+            const method = p.toString();
+            target[p] = async function (transport: IJsonRpcTransport, ...params: Parameters<JsonRpcApi[TMethodName]>) {
+                const normalizedParams = params.length ? params : undefined;
+                const result = await transport.send(method, normalizedParams);
+                return result;
+            } as unknown as JsonRpcApi[TMethodName];
```
joncinque, line 18: Learning a lot today with `... as unknown as ...` -- does this cause potential footguns for transport implementers?

packages/rpc-core/src/types/rpc-methods/getAccountInfo.d.ts
```diff
@@ -0,0 +1,92 @@
...
+type GetAccountInfoApiResponseBase = Readonly<{
+    context: Readonly<{
+        slot: Slot;
+    }>;
+    value: Readonly<{
+        executable: boolean;
+        lamports: number; // TODO(solana-labs/solana/issues/30341) Represent as bigint
+        owner: Base64EncodedAddress;
+        rentEpoch: number; // TODO(solana-labs/solana/issues/30341) Represent as bigint
+        space: number; // TODO(solana-labs/solana/issues/30341) Represent as bigint
```
joncinque, line 19: Maybe I'm missing something, but in testing I didn't see `space` returned here. Here's a token account:

```
{
  "jsonrpc": "2.0",
  "result": {
    "context": {
      "apiVersion": "1.13.6",
      "slot": 181671824
    },
    "value": {
      "data": [
        "xvp6877brTo9ZfNqq8l0MbG75MLS9uDkfKYCA0UvXWEs07wIp7mh6Iz0YahgZDSe/Psf+oXDeVKNALXKNdFOFAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
        "base64"
      ],
      "executable": false,
      "lamports": 2039280,
      "owner": "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA",
      "rentEpoch": 0
    }
  },
  "id": 1
}
```

packages/rpc-core/src/types/rpc-methods/getAccountInfo.d.ts
joncinque: This is extremely cool, and not nearly as complicated to read as I feared when I first opened the file :sweat_smile:

packages/rpc-core/src/types/rpc-methods/common.d.ts
```diff
@@ -0,0 +1,11 @@
+declare type DataSlice = readonly {
+    offset: number;
+    length: number;
+};
+
+// TODO: Eventually move this into whatever package implements transactions
+declare type Finality = 'confirmed' | 'finalized' | 'processed';
```
joncinque, line 7: nit: Is `Finality` the new term? If so, we should probably update the docs and all that. I'd prefer to keep this `Commitment` to make mental mapping easier, but I won't put up a big fight about it.

joncinque (commented):
Only one tiny substantial point -- looks really great overall!

packages/rpc-core/src/types/rpc-methods/common.d.ts
```diff
@@ -0,0 +1,6 @@
...
+
+declare type Slot =
+    // TODO(solana-labs/solana/issues/30341) Represent as bigint
+    number;
```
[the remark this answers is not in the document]
steveluscher: So, fast forward to when everything is `bigint` (or rewind, and have represented all `u64s` as JavaScript strings in the first place) and all that the transport has to do is to serialize `bigints` as `string` over the wire.

We can't do this today, because the Solana RPC doesn't accept them.

```shell
# params is normally [5, 10] but here I'm representing them as ["5", "10"]
curl https://api.mainnet-beta.solana.com -X POST -H "Content-Type: application/json" -d '
  {
    "jsonrpc": "2.0", "id": 1,
    "method": "getBlocks",
    "params": ["5", "10"]
  }
'
{"jsonrpc":"2.0","error":{"code":-32602,"message":"Invalid params: invalid type: string \"5\", expected u64."},"id":1}
```

We need to change the server to parse strings as `u64` before we can do that on the client.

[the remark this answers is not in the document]
steveluscher: But I see what you're saying. You're saying: make it a `bigint` now at the outer part of the interface, and have the transport (temporarily) downcast that to a JavaScript `number`, maybe optionally warning when it overflows. Is that right?

packages/rpc-core/src/types/rpc-methods/getBlocks.d.ts
steveluscher: I fixed it! https://github.com/solana-labs/solana/pull/30351/files

steveluscher: Ah… https://discord.com/channels/428295358100013066/966782389839876096/1083467864650039347

packages/rpc-core/src/rpc.ts
steveluscher: At that point, it's our job to make sure that the implementation of `rpc.*` is perfect. Essentially what we're saying is that the dynamic `get` returns a method that will definitely conform to `JsonRpcApi[TMethodName]` but the implementation on lines 14-18 does not, from Typescript's perspective.

```shell
src/rpc.ts:14:13 - error TS2322: Type '(transport: IJsonRpcTransport, ...params: Parameters<JsonRpcApi[TMethodName]>) => Promise<unknown>' is not assignable to type 'JsonRpcApi[TMethodName]'.
  Type '(transport: IJsonRpcTransport, ...params: Parameters<JsonRpcApi[TMethodName]>) => Promise<unknown>' is not assignable to type '{ (transport: IJsonRPCTransport, address: Base58EncodedAddress, config?: ({ encoding: "base64"; } & GetAccountInfoApiCommonConfig & GetAccountInfoApiBase64EncodingCommonConfig) | undefined): Promise<...>; (transport: IJsonRPCTransport, address: Base58EncodedAddress, config?: ({ ...; } & ... 1 more ... & GetAccountIn...'.
    Type '(transport: IJsonRpcTransport, ...params: Parameters<JsonRpcApi[TMethodName]>) => Promise<unknown>' is not assignable to type '{ (transport: IJsonRPCTransport, address: Base58EncodedAddress, config?: ({ encoding: "base64"; } & GetAccountInfoApiCommonConfig & GetAccountInfoApiBase64EncodingCommonConfig) | undefined): Promise<...>; (transport: IJsonRPCTransport, address: Base58EncodedAddress, config?: ({ ...; } & ... 1 more ... & GetAccountIn...'.
      Types of parameters 'params' and 'address' are incompatible.
        Type '[address: Base58EncodedAddress, config?: ({ encoding: "base64"; } & GetAccountInfoApiCommonConfig & GetAccountInfoApiBase64EncodingCommonConfig) | undefined]' is not assignable to type 'Parameters<JsonRpcApi[TMethodName]>'.

14             target[p] = async function (transport: IJsonRpcTransport, ...params: Parameters<JsonRpcApi[TMethodName]>) {
               ~~~~~~~~~
```

steveluscher: It _seems_ like these two should be compatible:

* `[address: Base58EncodedAddress, config?: ({ encoding: "base64"; } & GetAccountInfoApiCommonConfig & GetAccountInfoApiBase64EncodingCommonConfig) | undefined]`
* `Parameters<JsonRpcApi[TMethodName]>`

…but they aren't.

packages/rpc-core/src/types/rpc-methods/common.d.ts
steveluscher: Ooh. I thought these three were what we're calling finality, as opposed to the olde deprecated suite of commitments.

From web3.js today:

https://github.com/solana-labs/solana-web3.js/blob/ad23683e8a42c726995cf0c1f0f903b20152854f/packages/library-legacy/src/connection.ts#L476-L501

So yeah, I have this wrong, somehow. I'll rename it.

packages/rpc-core/src/types/rpc-methods/getAccountInfo.d.ts
steveluscher: Haha. In the docs, this is called `size`, in the docs' example it's called `space`, in the current web3.js implementation it's absent, and in the server implementation it's `space` and it's an `Option<u64>`.

steveluscher: 🤔 https://github.com/solana-labs/solana/pull/28421

steveluscher: Looks like this is only in v1.15.

packages/rpc-core/src/types/rpc-methods/common.d.ts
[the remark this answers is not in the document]
steveluscher: I love this!

packages/rpc-core/src/types/rpc-methods/getBlocks.d.ts
joncinque: ah yes. Anyway, thanks for fixing!

packages/rpc-core/src/types/rpc-methods/common.d.ts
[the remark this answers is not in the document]
joncinque: I was wondering it if it would be possible at the transport level to use `bigint` everywhere, but I forgot that the problem is at the RPC level, and not web3.js. But yeah, since all of these interfaces are new, it might be worth doing everything right and use `bigint`s everywhere.

packages/rpc-core/src/rpc.ts
joncinque: Ah I see, that's really weird, thanks for explaining. Is that a typescript bug? Or does `Parameters` need to have some extra declaration to make that clear?

packages/rpc-core/src/types/rpc-methods/getAccountInfo.d.ts
joncinque: Ah gotcha

packages/rpc-core/src/rpc.ts
steveluscher: I'm not sure, but I think I made a mistake including `transport` in all the method signatures. I wonder what would happen if I took that out…

steveluscher: #1196.

PR #1194, opened 2023-03-10
[experimental] a browser-fork of the `fetch` implementation

steveluscher:
# Summary
This is a separate module so that we can typecheck it using the `DOM` lib, without polluting any of the upstream packages with `DOM` libdefs. This ensures that downstream packages don't unintentionally use `DOM` types that won't work in Node/React Native..

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/solana-labs/solana-web3.js/pull/1194).
* #1195
* __->__ #1194


joncinque (commented):
Looks good!

PR #1196, opened 2023-03-11
[experimental] Remove the concept of the transport from the types of the RPC

steveluscher:
# Summary

@joncinque pointed out in #1190 that the Typescript types of `@solana/rpc-core` relied on a cast through `unknown` to work. The more I thought about it the more I realized that the `transport` shouldn't form part of the typespec for the RPC.

In this PR we remove it from the source types, and map it back in where it's needed in the implementation..


joncinque (commented):
Nice job fixing up the types! Looks good

PR #1204, opened 2023-03-16
[experimental] A function to patch up JSON RPC params that are destined for the Solana JSON RPC

steveluscher:
[experimental] A function to patch up JSON RPC params that are destined for the Solana JSON RPC
## Summary

`u64` values, in the Solana JSON RPC, are implemented as JavaScript numbers. This means that any number greater than `Number.MAX_SAFE_INTEGER` (9007199254740991) can't be trusted. This is tracked in https://github.com/solana-labs/solana/issues/30741.

In this PR we create a serializer that converts `bigint` values coming in from `@solana/rpc-core` to JavaScript `number`, as a concession.

## Test Plan

```shell
pnpm test:unit:node
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/solana-labs/solana-web3.js/pull/1204).
* #1210
* #1205
* __->__ #1204
* #1209

### DIFF
--- a/packages/rpc-transport/src/__tests__/params-patcher-test.ts
+++ b/packages/rpc-transport/src/__tests__/params-patcher-test.ts
@@ -1,0 +1,46 @@
+import { patchParamsForSolanaLabsRpc } from '../params-patcher';
+
+describe('patchParamsForSolanaLabsRpc', () => {
+    [10, '10', null, undefined, Symbol()].forEach(input => {
+        describe(`given \`${String(input)}\` (${typeof input}) as input`, () => {
+            it('returns the value as-is', () => {
+                expect(patchParamsForSolanaLabsRpc(input)).toBe(input);
+            });
+        });
+    });
+    describe('given a `bigint` as input', () => {
+        const input = 10n;
+        it('casts the input to a `number`', () => {
+            expect(patchParamsForSolanaLabsRpc(input)).toBe(Number(input));
+        });
+    });
+    describe('given a `bigint` two larger than `Number.MAX_SAFE_INTEGER` as input', () => {
+        const input = BigInt(Number.MAX_SAFE_INTEGER) + 2n; // 9007199254740993
+        it('casts the input to a `number`', () => {
+            expect(input - BigInt(patchParamsForSolanaLabsRpc(input)))
+                // Incorrectly rounds to 9007199254740992, leaving a difference of 1.
+                .toBe(1n);
+        });
+    });
+    describe('given an array as input', () => {
+        const input = [10n, 10, '10', ['10', [10, 10n], 10n]] as const;
+        it('casts the bigints in the array to a `number`, recursively', () => {
+            expect(patchParamsForSolanaLabsRpc(input)).toStrictEqual([
+                Number(input[0]),
+                input[1],
+                input[2],
+                [input[3][0], [input[3][1][0], Number(input[3][1][0])], Number(input[3][2])],
+            ]);
+        });
+    });
+    describe('given an object as input', () => {
+        const input = { a: 10n, b: 10, c: { c1: '10', c2: 10n } } as const;
+        it('casts the bigints in the array to a `number`, recursively', () => {
+            expect(patchParamsForSolanaLabsRpc(input)).toStrictEqual({
+                a: Number(input.a),
+                b: input.b,
+                c: { c1: input.c.c1, c2: Number(input.c.c2) },
+            });
+        });
+    });
+});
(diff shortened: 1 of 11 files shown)


packages/rpc-transport/src/params-patcher.ts
```diff
@@ -0,0 +1,26 @@
...
+            if (Object.prototype.hasOwnProperty.call(params, propName)) {
+                out[propName] = patchParamsForSolanaLabsRpc(params[propName]);
+            }
+        }
+        return out as TypescriptBug33014;
+    } else if (typeof params === 'bigint') {
+        // TODO(solana-labs/solana/issues/30341) Create a data type to represent u64 in the Solana
+        // JSON RPC implementation so that we can throw away this entire patcher instead of unsafely
+        // downcasting `bigints` to `numbers`.
+        return Number(params) as TypescriptBug33014;
```
joncinque, line 22: If it's outside of the number range, how about throwing an error?

joncinque (commented):
Oops sorry, I only reviewed #1205 -- I've moved the comments over to here where they make more sense

packages/rpc-transport/src/params-patcher.ts
steveluscher: Oooh. Interesting. Yeah, there are three ways to handle numbers above `Number.MAX_SAFE_INTEGER`.

1. Let the cast happen and pass it on.
2. Warn on the console, cast, and pass it on.
3. Throw.

The only options that are API compatible with the current `@solana/web3.js` are the first two. The third would be a breaking change.

There's a fourth, more arduous option: creating an opaque datatype for safe numbers.

```ts
type SafeNumber = bigint & {__safenumber: unique symbol}

function assertIsSafeNumber(num: bigint): asserts num is SafeNumber {
  if (num > Number.MAX_SAFE_INTEGER) {
    throw new Error('...');
  }
}
function asSafeNumber(num: bigint): SafeNumber {
  assertIsSafeNumber(num);
  return num as SafeNumber;
}
```

…and then everywhere you ever wanted to use a number with the RPC…

```ts
rpc.getBlock(asSafeNumber(slot));
```

I think the best solution is the one that requires people to do _nothing_ when we eventually fix the RPC, other than to ask people to update `web3.js`.

Thoughts?

packages/rpc-transport/src/__tests__/params-patcher-test.ts
```diff
@@ -0,0 +1,38 @@
...
+        const input = { a: 10n, b: 10, c: { c1: '10', c2: 10n } } as const;
+        it('casts the bigints in the array to a `number`, recursively', () => {
+            expect(patchParamsForSolanaLabsRpc(input)).toStrictEqual({
+                a: Number(input.a),
+                b: input.b,
+                c: { c1: input.c.c1, c2: Number(input.c.c2) },
+            });
+        });
+    });
+});
```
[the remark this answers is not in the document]
steveluscher: Added!

packages/rpc-transport/src/params-patcher.ts
joncinque: If this is meant to be a new library that people consciously switch to, or a new major version of web3.js, then I think it's fine to break and go the route of number 3.  As it stands, people will get garbage anyway, right?

On the flipside, I see what you mean. it would really stink to get an error because you happen to query an account with a lot of lamports, for example. Maybe number 2 isn't so bad then?

Number 4 does seem quite painful, and I think I'm missing something -- how is it much different from number 3 in the end?

steveluscher: Number 3 and 4 both kill your program at runtime, but number 4 pushes the assertion out of the framework and into userspace.

This _is_ meant to be a new library, but I _would_ like to be able to rebuild the existing `web3.js` on top of the modern core in the meantime. If the RPC started to throw here, there's nothing that legacy `web3.js` could do to say ‘no no, forget it, just send the cast'd value through.’

Ha. I could literally have the transport take in a handler that's like `onIntegerOverflow`. The legacy one could be a  no-op, and the modern one could throw.

🤔

joncinque: > number 4 pushes the assertion out of the framework and into userspace.

Ah I see, so you're thinking that by forcing people to make a code change like `asSafeNumber` in front of everything, people will think twice? I unfortunately think it'll just cause churn

> Ha. I could literally have the transport take in a handler that's like onIntegerOverflow. The legacy one could be a no-op, and the modern one could throw.

Oooh I really like this!

joncinque (commented):
Since I can't approve post-merge, "Approved"!

PR #1208, opened 2023-03-16
feat: coalesce calls to `getBlockHeight`

### LINKED ISSUE #1182: getBlockHeight RPC method leaks long after confirmTransaction ends
**Problem:**

We noticed millions of getBlockHeight RPC calls after deploying a long-lived service to broadcast transactions for our customers. However, we only broadcast thousands of transactions a week. 

image
 

Putting this wrapper on the connection showed that long after some transactions were confirmed getBlockHeight was being called.

```

      /// @ts-ignore
      const oldRequest = connection._rpcRequest;

      /// @ts-ignore
      connection._rpcRequest = function (...args) {
        const url = new URL(connection.rpcEndpoint);
        console.log(`rpc request: ${url.hostname} `, args);
        return oldRequest(...args);
      };

```

**Proposal:** 

Cache and throttle calls to getBlockHeight so multiple calls to confirmTransaction share the cached blockheight.

and/or

Find source of leak.

steveluscher:
## Summary

There is no sense having more than one of these calls in flight at any one time; once there's a single call in flight, all callers should share the result.

For much more on why this is necessary, see #1182.

Fixes #1182.

## Test Plan

```shell
pnpm test:unit:browser
pnpm test:unit:node
```

### DIFF
--- a/packages/library-legacy/src/connection.ts
+++ b/packages/library-legacy/src/connection.ts
@@ -4739,28 +4739,40 @@ export class Connection {
   /*
    * Returns the current block height of the node
    */
-  async getBlockHeight(
-    commitmentOrConfig?: Commitment | GetBlockHeightConfig,
-  ): Promise<number> {
-    const {commitment, config} =
-      extractCommitmentFromConfig(commitmentOrConfig);
-    const args = this._buildArgs(
-      [],
-      commitment,
-      undefined /* encoding */,
-      config,
-    );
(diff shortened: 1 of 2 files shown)


steveluscher:
What do you think, @Jac0xb?

joncinque (approved):
Looks good to me! very slick

steveluscher:
I should probably take a look to see what _other_ API methods this should apply to.

PR #1210, opened 2023-03-17
[experimental] A handler you can supply to transports in the event of integer overflows

steveluscher:
[experimental] A handler you can supply to transports in the event of integer overflows
## Summary

The JSON RPC was designed to communicate JavaScript `Numbers` over the wire, which puts values over `Number.MAX_SAFE_INTEGER` at risk of rounding errors. In this PR we add a hook that fires in the event of such an overflow. Implementers of that hook can decide to do whatever they want: throw, warn, or allow.
## Test Plan

```shell
pnpm test:unit:browser
pnpm test:unit:node
```


joncinque (commented):
Looks great!

PR #1223, opened 2023-03-27
[experimental] Invert the responsibilities of the transport and rpc

steveluscher:
[experimental] Invert the responsibilities of the transport and rpc
## Summary

Previously, the RPC API was responsible for:

* exposing RPC methods using a JavaScript proxy
* formatting params for delivery to an RPC server
* calling `send()` on a supplied transport

What I discovered was that this made a whole bunch of things _really_ hard.

* supplying a per-call cancellation signal
* bundling calls in a batch

In this PR, we inside out this design.

* the RPC package is now exclusively made up of types
* the transport features a generic that accepts that type, exposing all of its methods
* the transport now ‘collects’ calls until you call `send()` or `sendBatch()`
* the ‘send’ methods are a logical place to configure a cancellation signal, and more.

Look at the tests to see how this works in practice, but in the meantime here's an example:

```ts
const devnet = createJsonRpcTransport({ url:'https://api.devnet.solana.com' });

const blockHeight = await devnet
  .getBlockHeight({ commitment: 'finalized' })
  .send();

const [blockHeight, balance] = await devnet
  .getBlockHeight({ commitment: 'finalized' })
  .getBalance('M2mx93ekt1fmXSVkTrUL9xVFHkmME8HTUi5Cyc5aF7K' as Base58EncodedAddress)
  .sendBatch();
```

## Test Plan

```shell
pnpm test:unit:browser
pnpm test:unit:node
```

### DIFF
--- a/packages/rpc-transport/src/json-rpc-transport-types.ts
+++ b/packages/rpc-transport/src/json-rpc-transport-types.ts
@@ -1,0 +1,80 @@
+/**
+ * Public API.
+ */
+export type Transport<TApi> = TransportMethods<TApi>;
+export type ArmedTransport<Api, TResponse> = ArmedTransportMethods<Api, TResponse> & {
+    send(): Promise<TResponse>;
+};
+export type ArmedBatchTransport<Api, TResponses extends unknown[]> = ArmedBatchTransportMethods<Api, TResponses> & {
+    sendBatch(): Promise<TResponses>;
+};
+
+/**
+ * Private transport-building types.
+ */
+type TransportMethods<TApi> = {
+    [TMethodName in keyof TApi]: ArmedTransportReturner<TApi, ApiMethodImplementations<TApi, TMethodName>>;
+};
+type ArmedTransportMethods<TApi, TResponse> = {
+    [TMethodName in keyof TApi]: ArmedBatchTransportReturner<
+        TApi,
+        ApiMethodImplementations<TApi, TMethodName>,
+        [TResponse]
+    >;
+};
+type ArmedBatchTransportMethods<TApi, TResponses extends unknown[]> = {
+    [TMethodName in keyof TApi]: ArmedBatchTransportReturner<
+        TApi,
+        ApiMethodImplementations<TApi, TMethodName>,
+        TResponses
+    >;
+};
+type ApiMethodImplementations<TApi, TMethod extends keyof TApi> = Overloads<TApi[TMethod]>;
+type ArmedTransportReturner<TApi, TMethodImplementations> = UnionToIntersection<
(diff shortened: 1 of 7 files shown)


packages/rpc-transport/src/json-rpc-transport-types.ts
```diff
@@ -0,0 +1,80 @@
...
+type TransportMethods<TApi> = {
+    [TMethodName in keyof TApi]: ArmedTransportReturner<TApi, ApiMethodImplementations<TApi, TMethodName>>;
+};
+type ArmedTransportMethods<TApi, TResponse> = {
+    [TMethodName in keyof TApi]: ArmedBatchTransportReturner<
+        TApi,
+        ApiMethodImplementations<TApi, TMethodName>,
+        [TResponse]
+    >;
+};
```
joncinque, line 24: Could this just be simplified to `type ArmedTransportMethods<TApi, TResponse> = ArmedBatchTransportMethods<TApi, [TResponse]>`?

packages/rpc-transport/src/json-rpc-transport-types.ts
```diff
@@ -0,0 +1,80 @@
...
+type ArmedBatchTransportMethods<TApi, TResponses extends unknown[]> = {
+    [TMethodName in keyof TApi]: ArmedBatchTransportReturner<
+        TApi,
+        ApiMethodImplementations<TApi, TMethodName>,
+        TResponses
+    >;
+};
+type ApiMethodImplementations<TApi, TMethod extends keyof TApi> = Overloads<TApi[TMethod]>;
+type ArmedTransportReturner<TApi, TMethodImplementations> = UnionToIntersection<
```
joncinque, line 33: Is there something to the `ArmedTransport*` names that I'm missing, as in, other than their funniness?

joncinque (commented):
Looks really great overall! Just a few little nits

packages/rpc-transport/src/json-rpc-transport-types.ts
steveluscher: I think so! #1242.

packages/rpc-transport/src/json-rpc-transport-types.ts
steveluscher: A regular transport hasn't accumulated any method calls yet. You can't call `send()` on it.

As soon as it has accumulated one, it's ‘armed.’ You can call `send()` on it.

joncinque: Ah, that makes sense, thanks!

PR #1234, opened 2023-04-01
test: repair compute budget test by adding at least one instruction before `requestHeapFrame`

steveluscher:
test: repair compute budget test by adding at least one instruction before `requestHeapFrame`
## Summary
I believe that what's going on here is that you can no longer `requestHeapFrame` if there are no instructions in the transaction. This is new since the 1.14 line.

## Test Plan

```
```


steveluscher:
@taozhu-chicago, does this seem about right to you?

PR #1235, opened 2023-04-01
test: remove test for priority fee that's ‘too high’

steveluscher:
test: remove test for priority fee that's ‘too high’
## Summary
My best guess here is that setting the priority fee high has no effect unless there's a bidder in the same block to outbid you. My solution here is to just remove the test.

## Test Plan
```
```


steveluscher:
@taozhu-chicago, can I also get your eyes on this one? I think that setting the fee ‘too high’ no longer has any effect because there's no bidder in the same block to force you to pay that fee. Can you think of a way to rewrite this test in a way that exercises something, or should we just delete it as I've done here?

PR #1237, opened 2023-04-02
test: repair compute budget test by adding at least one instruction before `requestHeapFrame`

steveluscher:
test: repair compute budget test by adding at least one instruction before `requestHeapFrame`
## Summary
I believe that what's going on here is that you can no longer `requestHeapFrame` if the transaction consumes no compute units.

## Test Plan

```
```

### DIFF
--- a/packages/library-legacy/test/program-tests/compute-budget.test.ts
+++ b/packages/library-legacy/test/program-tests/compute-budget.test.ts
@@ -72,7 +72,7 @@ describe('ComputeBudgetProgram', () => {
   });
 
   if (process.env.TEST_LIVE) {
-    it.skip('send live request heap ix', async () => {
+    it('send live request heap ix', async () => {
       const connection = new Connection(url, 'confirmed');
       const STARTING_AMOUNT = 2 * LAMPORTS_PER_SOL;
       const baseAccount = Keypair.generate();
@@ -85,6 +85,7 @@ describe('ComputeBudgetProgram', () => {
 
       async function expectRequestHeapFailure(bytes: number) {
         const requestHeapFrameTransaction = new Transaction().add(
+          ComputeBudgetProgram.setComputeUnitLimit({units: 1_000_000}),
           ComputeBudgetProgram.requestHeapFrame({bytes}),
         );
         await expect(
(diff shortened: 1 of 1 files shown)


steveluscher:
@taozhu-chicago, does this seem about right to you?

taozhu-chicago (changes requested):
i am not confident the cause mentioned in pr is accurate. Even if that is the case, then we need to look into reasoning why "request heap limit" requires _any_ random instruction. 

If the error was, as desccribed in discord, `computationalBudgetExceeded`, then it is likely due to test transaction not requesting compue unit limits. 

Can we double check the actual cause of test faliure?

steveluscher:
> Can we double check the actual cause of test faliure?

Yeah! That's totally my goal, having reached out to you. I'm 0% familiar with this system. Are you the best person to ask how and why this is all supposed to work, or is there someone more familiar with it?

steveluscher:
> I believe that what's going on here is that you can no longer requestHeapFrame if there are no instructions in the transaction.

I based this hypothesis on the Rust tests, which do exactly that.

https://github.com/solana-labs/solana/blob/8c860e98949efafee878acfac9da67b0bf927cac/rpc/src/rpc.rs#L8636-L8642

taozhu-chicago:
> > I believe that what's going on here is that you can no longer requestHeapFrame if there are no instructions in the transaction.
> 
> I based this hypothesis on the Rust tests, which do exactly that.
> 
> https://github.com/solana-labs/solana/blob/8c860e98949efafee878acfac9da67b0bf927cac/rpc/src/rpc.rs#L8636-L8642

Not sure why, but the link point to `SetComputeUnitPrice` instead of `requestHeapFrame`. Let's try to add `SetComputeUnitLimit` to the transaction without adding memo ix, see if that works. I am happy to look into it as soon as I return to keyboard in few days.

steveluscher:
Yeah, sorry, here's the right link!

https://github.com/solana-labs/solana/blob/8c860e98949efafee878acfac9da67b0bf927cac/runtime/src/transaction_priority_details.rs#L68-L74

steveluscher:
Alright! Here's another crack at it!

1. Added a compute limit instruction instead of a memo instruction.
2. Added a test that asserts it _fails_ when no such CU-consuming instruction is present.

taozhu-chicago (approved):
lgtm

PR #1238, opened 2023-04-02
test: remove test for priority fee that's ‘too high’

steveluscher:
test: remove test for priority fee that's ‘too high’
## Summary
My best guess here is that setting the priority fee high has no effect unless there's a bidder in the same block to outbid you. My solution here is to just remove the test.

## Test Plan
```
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/solana-labs/solana-web3.js/pull/1238).
* __->__ #1238
* #1236

### DIFF
--- a/packages/library-legacy/test/program-tests/compute-budget.test.ts
+++ b/packages/library-legacy/test/program-tests/compute-budget.test.ts
@@ -72,7 +72,7 @@ describe('ComputeBudgetProgram', () => {
   });
 
   if (process.env.TEST_LIVE) {
-    it('send live request heap ix', async () => {
+    it.skip('send live request heap ix', async () => {
       const connection = new Connection(url, 'confirmed');
       const STARTING_AMOUNT = 2 * LAMPORTS_PER_SOL;
       const baseAccount = Keypair.generate();
@@ -115,7 +115,7 @@ describe('ComputeBudgetProgram', () => {
       );
     });
 
-    it('send live compute unit ixs', async () => {
+    it.skip('send live compute unit ixs', async () => {
       const connection = new Connection(url, 'confirmed');
       const FEE_AMOUNT = LAMPORTS_PER_SOL;
       const STARTING_AMOUNT = 2 * LAMPORTS_PER_SOL;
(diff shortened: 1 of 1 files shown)


steveluscher:
@taozhu-chicago, can I also get your eyes on this one? I think that setting the fee ‘too high’ no longer has any effect because there's no bidder in the same block to force you to pay that fee. Can you think of a way to rewrite this test in a way that exercises something, or should we just delete it as I've done here?

taozhu-chicago:
> @taozhu-chicago, can I also get your eyes on this one? I think that setting the fee ‘too high’ no longer has any effect because there's no bidder in the same block to force you to pay that fee. Can you think of a way to rewrite this test in a way that exercises something, or should we just delete it as I've done here?

The test being deleted seems to test a scenario that payer account should have enough balance to allow high prioritization fee, which sounds to me a valid test. Even there is no bidder to force, one is free to set prioritization fee to as high as payer account can pay for it -- i am thinking in a scenario that one *really* wants transaction to land in block without knowing if there are other bidders there (which should be a common case, is it?).

taozhu-chicago (changes requested):
i think this test is valid, should not be removed.

steveluscher:
Well, with 1.16 the test doesn't pass. Do you have any ideas of what needs changing, @taozhu-chicago?

steveluscher:
> Even there is no bidder to force, one is free to set prioritization fee to as high as payer account can pay for it…

On 1.16, the test fails because this transaction _lands_. I'm guessing that means that the prioritization fee _isn't_ being charged, so the payer _isn't_ running out of funds to pay for it.

taozhu-chicago:
> Well, with 1.16 the test doesn't pass. Do you have any ideas of what needs changing, @taozhu-chicago?

I am not aware of any changes to how prioritization fee being charged - the test calculates p.fee correctly, as far as I can tell. I think the test is valid, if you agree too, then something must be wrong. We should fine out why. I am currently OOO. I'll be happy to look into it when I am back. Please tag me if you find more about it.

steveluscher:
This just started working again.

PR #1239, opened 2023-04-02
refactor(experimental): add a numeric value allowlist to the response patcher

steveluscher:
refactor(experimental): add a numeric value allowlist to the response patcher

## Summary
Most of the values that come from the RPC are supposed to be `u64` (ie. must be represented as JavaScript `bigint`, eventually). Some are legitimately `usize` or `u8` though.

In this PR we create an allowlist for those method/path pairs. Any value at that path for that method will not be upcast.
## Test Plan

```
cd packages/rpc-transport
pnpm test:unit:browser
pnpm test:unit:node
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/solana-labs/solana-web3.js/pull/1239).
* #1241
* #1232
* __->__ #1239

### DIFF
--- a/packages/test-config/jest-dev.config.ts
+++ b/packages/test-config/jest-dev.config.ts
@@ -14,6 +14,7 @@ const config: Config = {
         'jest-watch-typeahead/filename',
         'jest-watch-typeahead/testname',
     ],
+    workerThreads: true,
 };
 
 export default config;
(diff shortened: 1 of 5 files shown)


packages/test-config/jest-dev.config.ts
```diff
@@ -14,6 +14,7 @@ const config: Config = {
         'jest-watch-typeahead/filename',
         'jest-watch-typeahead/testname',
     ],
+    workerThreads: true,
```
steveluscher, line 17: For posterity: https://github.com/facebook/jest/issues/11617#issuecomment-1458155552

PR #1250, opened 2023-04-07
refactor(experimental): ultra super mega refactor of transport/rpc-core responsibilities

steveluscher:
refactor(experimental): ultra super mega refactor of transport/rpc-core responsibilities
## Summary

In this massive, barely-reviewable PR, we reorganize the same code and tests that you know and love so as to better separate the responsibilities between the transport and the rpc-core package. Now, if the Solana RPC were to go away, introduce a new version, or be replaced with something else, the `rpc-transport` package would not need to change _at all_.

* The `rpc-transport` now accepts a _concrete_ implementation of an API. This is a set of methods that take params as input, and output the method name and params that should be forwarded to the RPC. This concrete implementation may also supply a function that the transport should use to post-process the response. Those two functions alone are enough to implement _all_ of the params patching that the Solana RPC needs to do, and it can be done somewhere _other_ than `@solana/rpc-transport`
* The `rpc-core` package is now a concrete implementation of particularly the Solana JSON RPC as it is today. It's an implementation of the interface described in the previous bullet point.
* The `library` package combines these two things. It instantiates and configures an instance of `@solana/rpc-core`, and uses that to create an instance of `@solana/rpc-transport`.

## Test Plan

All the same tests you know and love, running in CI.


packages/rpc-transport/src/json-rpc-transport/json-rpc-transport-types.ts
```diff
@@ -1,46 +1,74 @@
...
-    sendBatch(): Promise<TResponses>;
+export type Transport<TRpcMethods> = TransportMethods<TRpcMethods>;
+export type TransportConfig<TRpcMethods> = Readonly<{
+    api: IRpcApi<TRpcMethods>;
+    url: string;
+}>;
+export type TransportRequest<TResponse> = {
+    methodName: string;
+    params: unknown[];
+    responseProcessor?: (response: unknown) => TResponse;
```
steveluscher, line 17: The whole universe of this PR starts from here. This is a brand new interface that you must comply with in order to supply RPC methods to a `@solana/rpc-transport`. Your concrete implementation of the RPC API is a bunch of methods that take in params, and return these things.

PR #1256, opened 2023-04-11
Make build work better on macOS

mcintyre94:
- Use /tmp/lock for locks, avoiding macOS lockdown on /var/lock
- Replace sleep infinity with a loop

Misc other changes:

- Run prettier:fix in packages/library-legacy
- Fix doc directory in packages/library-legacy clean


packages/library-legacy/package.json
```diff
@@ -41,7 +41,7 @@
     "compile:js": "cross-env NODE_ENV=production rollup -c",
     "compile:typedefs": "./scripts/typegen.sh",
     "build:fixtures": "set -ex; ./test/fixtures/noop-program/build.sh",
-    "clean": "rimraf ./docs ./declarations ./lib",
+    "clean": "rimraf ./doc ./declarations ./lib",
```
steveluscher, line 44: LOL.

packages/library-legacy/package.json
```diff
@@ -128,4 +128,4 @@
     "typedoc": "^0.23.28",
     "typescript": "^5.0.3"
   }
-}
+}
```
steveluscher, line 131: giphy

packages/test-config/test-validator-setup.js
```diff
@@ -7,7 +7,7 @@ module.exports = async function globalSetup() {
         { command: '../../scripts/start-shared-test-validator.sh' },
         // This 'server' is a noop; we only use it to run the 'wait for server' logic.
         {
-            command: 'sleep infinity',
+            command: 'while true; do sleep 86400; done',
```
steveluscher, line 10: Really Mac OS? Ugh.

scripts/start-shared-test-validator.sh
```diff
@@ -22,7 +24,7 @@ TEST_VALIDATOR_LEDGER="$( cd "$(dirname "${BASH_SOURCE[0]}")/.." ; pwd -P )/test
       wait
     else
       echo "Sharing lock on already running test validator (PID $(pidof $TEST_VALIDATOR))"
-      sleep infinity
+      while true; do sleep 86400; done
```
steveluscher, line 27: I'm concerned. The idea here is that `sleep infinity` emits a file descriptor that forms part of the lock. If `sleep 86400` finishes, does it temporarily release the lock before the loop restarts?

Yes, I know 86400 seconds is a whole day :)

steveluscher (approved):
Looks great! Just leaving some comments for your consideration. And that EOL.

scripts/start-shared-test-validator.sh
mcintyre94: Ah, got you. How about we just do like `sleep 86400000` here? I don't know what the macOS actual max is but it doesn't complain about that

PR #1257, opened 2023-04-11
test: re-enable compute budget test

steveluscher:
Previously: #1236.


steveluscher:
I have no idea what happened to break this test, nor what happened to fix it, but here we are. @taozhu-chicago is going to dig in and bisect, but for now, let's just move on.

PR #1258, opened 2023-04-11
refactor(experimental): add the `getBalance` RPC method

mcintyre94:
## Summary
Add the `getBalance` API and unit tests
Note that we currently only test for a zero balance, as we haven't implemented any RPC methods that will give a non-zero balance

## Test Plan
```
```

Note that since we don't have keygen yet, I've just hardcoded an address which is assumed to have a zero balance in the local validator. See also https://xkcd.com/221/

### DIFF
--- a/packages/rpc-core/src/rpc-methods/__tests__/get-balance-test.ts
+++ b/packages/rpc-core/src/rpc-methods/__tests__/get-balance-test.ts
@@ -29,11 +29,11 @@ describe('getBalance', () => {
         });
     });
 
-    describe('returns a non-zero balance for an address with lamports', () => {
-        it('returns the correct balance for a fixture account', async () => {
+    describe('given an account with a non-zero balance', () => {
+        // See scripts/fixtures/4nTLDQiSTRHbngKZWPMfYnZdWTbKiNeuuPcX7yFUpSAc.json
+        const publicKey = '4nTLDQiSTRHbngKZWPMfYnZdWTbKiNeuuPcX7yFUpSAc' as Base58EncodedAddress;
+        it('returns the correct balance', async () => {
             expect.assertions(1);
-            // See scripts/fixtures/4nTLDQiSTRHbngKZWPMfYnZdWTbKiNeuuPcX7yFUpSAc.json
-            const publicKey = '4nTLDQiSTRHbngKZWPMfYnZdWTbKiNeuuPcX7yFUpSAc' as Base58EncodedAddress;
             const balancePromise = transport.getBalance(publicKey).send();
             await expect(balancePromise).resolves.toHaveProperty('value', BigInt(5_000_000));
         });
--- a/packages/rpc-core/src/rpc-methods/__tests__/get-balance-test.ts
+++ b/packages/rpc-core/src/rpc-methods/__tests__/get-balance-test.ts
@@ -4,7 +4,7 @@ import type { Transport } from '@solana/rpc-transport/dist/types/json-rpc-transp
 import fetchMock from 'jest-fetch-mock';
 import { createSolanaRpcApi, SolanaRpcMethods } from '../../index';
 import { Commitment } from '../common';
-import { assertIsBase58EncodedAddress } from '@solana/keys';
+import { Base58EncodedAddress } from '@solana/keys';
 
 describe('getBalance', () => {
     let transport: Transport<SolanaRpcMethods>;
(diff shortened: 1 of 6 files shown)


packages/rpc-core/src/rpc-methods/getBalance.ts
```diff
@@ -0,0 +1,22 @@
+import { Base58EncodedAddress } from '@solana/keys';
+import { Commitment, Slot, U64UnsafeBeyond2Pow53Minus1 } from './common';
+
+type GetBalanceApiResponse = Readonly<{
```
steveluscher, line 4: I guess it's probably worth lifting this up to `common.ts` now.

```ts
// common.ts
export type RpcResponse<TValue> = Readonly<{
  context: Readonly<{
    slot: Slot,
  }>,
  value: TValue,
}>;
```

```ts
// getBalance.ts
type GetBalanceApiResponse = RpcResponse<U64UnsafeBeyond2Pow53Minus1>;
```

packages/rpc-core/src/rpc-methods/__tests__/get-balance-test.ts
```diff
@@ -0,0 +1,56 @@
...
+        // TODO: requires requestAirdrop or something else that can give an address lamports
+        it.todo('returns the correct balance after an airdrop');
+    });
+
+    describe('when called with a `minContextSlot` higher than the highest slot available', () => {
+        it('throws an error', async () => {
+            expect.assertions(1);
+            // This key is random, don't re-use in any tests that affect balance
+            const publicKey = '4BfxgLzn6pEuVB2ynBMqckHFdYD8VNcrheDFFCB6U5TH';
+            assertIsBase58EncodedAddress(publicKey);
```
steveluscher, line 43: Nice.

You can _also_:

```ts
const publicKey = '4BfxgLzn6pEuVB2ynBMqckHFdYD8VNcrheDFFCB6U5TH' as Base58EncodedAddress;
```

packages/rpc-core/src/rpc-methods/__tests__/get-balance-test.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    (['confirmed', 'finalized', 'processed'] as Commitment[]).forEach(commitment => {
+        describe(`when called with \`${commitment}\` commitment`, () => {
+            it('returns a balance of zero for a new address', async () => {
+                expect.assertions(1);
+                // This key is random, don't re-use in any tests that affect balance
+                const publicKey = '4BfxgLzn6pEuVB2ynBMqckHFdYD8VNcrheDFFCB6U5TH';
+                assertIsBase58EncodedAddress(publicKey);
+                const balance = await transport.getBalance(publicKey).send();
+                expect(balance.value).toEqual(BigInt(0));
```
steveluscher, line 28: I know I did this in the other tests, but it actually probably preferable to do:

```ts
await expect(transport.getBalance(publicKey).send()).resolves.toEqual(0n);
```

…or…

```ts
const balancePromise = transport.getBalance(publicKey).send();
await expect(balancePromise).resolves.toEqual(0n);
```

That way, if it rejects, we get it _in the assertion_ rather than just in the body of the test.

steveluscher (approved):
So, by all means land this as is, _but_ now would be the perfect time to figure out how to load up the validator with test accounts.

https://solana.stackexchange.com/a/4598/75

Imagine if starting the validator blew away the ledger, and primed the newly created ledger with fixtures. We'd need to use the _same_ fixtures for the entire repository, and hook the fixture loader into `scripts/start-shared-test-validator.sh`.

4fb0397d-5097-4544-8f24-f79ceece216b_text

packages/rpc-core/src/rpc-methods/__tests__/get-balance-test.ts
mcintyre94: Nice that makes sense

Because it's an RPC response and I just want to check the value I went with:

```ts
const balancePromise = transport.getBalance(publicKey).send();
await expect(balancePromise).resolves.toHaveProperty('value', BigInt(0));
```

mcintyre94:
> So, by all means land this as is, _but_ now would be the perfect time to figure out how to load up the validator with test accounts.

Nice thanks, I think I have this working. I've added a directory `scripts/fixtures/` and all accounts there are loaded into the validator when it starts. For now I've just added one account there to test getBalance with a non-zero balance

PR #1263, opened 2023-04-14
refactor(experimental): Add `getBlockProduction` API

mcintyre94:
## Summary

- Add the `getBlockProduction` API + unit tests
- Add jest-extended which provides some extra jest matchers

## Test Plan

```
```

Misc notes:

- The number matchers in `jest-extended` don't support BigInt yet (https://github.com/jest-community/jest-extended/issues/591). Didn't affect my final tests here but worth being aware of
- I'm not sure exactly where the 32602 error code comes from on the `lastSlot` too high test. Grepping the `solana` codebase doesn't show the error I'm seeing and the code isn't in https://github.com/solana-labs/solana/blob/master/rpc-client-api/src/custom_error.rs.  The promise rejects with `Rejected to value: [SolanaJsonRpcError: JSON-RPC 2.0 error (-32602): lastSlot, 9223372036854776000, is too large; max 0]`


steveluscher:
That error code is a JSON-RPC reserved one that means ‘invalid params.’ https://www.jsonrpc.org/specification#error_object

Basically, that's too large a number to jam into a JSON-RPC call, because it's over `Number.MAX_SAFE_INTEGER`.

packages/rpc-core/src/rpc-methods/getBlockProduction.ts
```diff
@@ -0,0 +1,30 @@
+import { Base58EncodedAddress } from '@solana/keys';
+import { Commitment, RpcResponse, U64UnsafeBeyond2Pow53Minus1 } from './common';
+
+type NumberOfLeaderSlots = number;
+type NumberOfBlocksProduced = number;
```
steveluscher, line 5: (Y). Looked at source code. Saw these were both `usize`.

packages/rpc-core/src/rpc-methods/getBlockProduction.ts
```diff
@@ -0,0 +1,30 @@
...
+
+type Range = Readonly<{
+    firstSlot: U64UnsafeBeyond2Pow53Minus1;
+    lastSlot: U64UnsafeBeyond2Pow53Minus1;
+}>;
+
+type GetBlockProductionApiResponse = RpcResponse<{
+    byIdentity: Readonly<{
+        [address: string]: [NumberOfLeaderSlots, NumberOfBlocksProduced];
```
steveluscher, line 14:
```suggestion
        [address: Base58EncodedAddress]: [NumberOfLeaderSlots, NumberOfBlocksProduced];
```

packages/rpc-core/src/rpc-methods/getBlockProduction.ts
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * Returns recent block production information from the current or previous epoch.
+     */
+    getBlockProduction(
+        config?: Readonly<{
+            commitment?: Commitment;
+            identity?: Base58EncodedAddress;
+            range?: Range;
+        }>
+    ): GetBlockProductionApiResponse;
```
steveluscher, line 29: Make overloads! You have two scenarios here which will have different return types.

* Has an identity => `byIdentity` is _definitely_ `{[TAddress in TIdentity]?: [NumberOfLeaderSlots, NumberOfBlocksProduced]}`
* Omits the identity => `byIdentity` is `Record<Base58EncodedAddress, [NumberOfLeaderSlots, NumberOfBlocksProduced]>`

```suggestion
    getBlockProduction(
        config?: Readonly<{
            commitment?: Commitment;
            identity: undefined
            range?: Range;
        }>
    ): GetBlockProductionApiResponseBase & GetBlockProductionApiResponseDefinitelyByIdentityOrNothing;
    getBlockProduction(
        config?: Readonly<{
            commitment?: Commitment;
            identity: Base58EncodedAddress;
            range?: Range;
        }>
    ): GetBlockProductionApiResponseBase;
```

…or whatever actually works, and think of names that aren't stupid. Look at `getAccountInfo` for some intense overrides.

packages/rpc-core/src/rpc-methods/__tests__/get-block-production-test.ts
```diff
@@ -0,0 +1,81 @@
+import { createJsonRpcTransport } from '@solana/rpc-transport';
+import type { SolanaJsonRpcErrorCode } from '@solana/rpc-transport/dist/types/json-rpc-transport/json-rpc-errors';
+import type { Transport } from '@solana/rpc-transport/dist/types/json-rpc-transport/json-rpc-transport-types';
+import fetchMock from 'jest-fetch-mock-fork';
+import { createSolanaRpcApi, SolanaRpcMethods } from '../../index';
+import { Commitment } from '../common';
+import { Base58EncodedAddress } from '@solana/keys';
+
+describe('getBalance', () => {
```
steveluscher, line 9:
```suggestion
describe('getBlockProduction', () => {
```

packages/rpc-core/src/rpc-methods/__tests__/get-block-production-test.ts
```diff
@@ -0,0 +1,81 @@
...
+    let transport: Transport<SolanaRpcMethods>;
+    beforeEach(() => {
+        fetchMock.resetMocks();
+        fetchMock.dontMock();
+        transport = createJsonRpcTransport({
+            api: createSolanaRpcApi(),
+            url: 'http://127.0.0.1:8899',
+        });
```
steveluscher, line 17: I think I changed this. Rebase!

steveluscher (changes requested):
image

packages/rpc-core/src/rpc-methods/getBlockProduction.ts
steveluscher: Uh, oh damn. Yes on 64 bit machines it could be. I got this wrong elsewhere; I’m going to have to go fix that now!

packages/rpc-core/src/rpc-methods/__tests__/get-block-production-test.ts
```diff
@@ -0,0 +1,81 @@
...
+                    }),
+                });
+            });
+
+            it('has the latest context slot as the last slot', async () => {
+                expect.assertions(1);
+                const blockProductionPromise = transport.getBlockProduction({ commitment }).send();
+                await expect(blockProductionPromise).resolves.toSatisfy(
+                    rpcResponse => rpcResponse.context.slot === rpcResponse.value.range.lastSlot
+                );
```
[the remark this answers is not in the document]
mcintyre94: Agreed, I've changed this to the semantic matcher you suggested

packages/rpc-core/src/rpc-methods/getBlockProduction.ts
mcintyre94: Nice! I've updated this to use overloads, definitely seems better!

packages/rpc-core/src/rpc-methods/__tests__/get-block-production-test.ts
```diff
@@ -0,0 +1,81 @@
...
+    });
+
+    describe('when called with a `lastSlot` higher than the highest slot available', () => {
+        it('throws an error', async () => {
+            expect.assertions(1);
+            const blockProductionPromise = transport
+                .getBlockProduction({
+                    range: {
+                        firstSlot: 0n,
+                        lastSlot: 2n ** 63n - 1n, // u64:MAX; safe bet it'll be too high.
```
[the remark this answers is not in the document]
steveluscher: Yes! Totally. That RPC method could do with throwing a more precise error.

steveluscher:
Dang it, the goal with that return type for the `identity` case was to make this a type error:

```ts
const bp = await rpc
    .getBlockProduction({
        commitment,
        identity: 'abc' as Base58EncodedAddress,
    })
    .send();
bp.value.byIdentity['def' as Base58EncodedAddress];
```

…but that's not happening yet. Essentially the `string` type is flowing through that type, rather than the string _literal_.

Think this is possible to type correctly?

mcintyre94:
I think this should work with:

```
    getBlockProduction<TIdentity extends Base58EncodedAddress>(
        config: Readonly<{
            commitment?: Commitment;
            identity: TIdentity;
            range?: SlotRange;
        }>
    ): GetBlockProductionApiResponseBase & GetBlockProductionApiResponseWithSingleIdentity<TIdentity>;
```

[See example playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAIglgY2AHgCoEkAmEB2w6hQQAewumAzlBcAE5w4DmAfFALxQDeAUFHwNqoAgpky0IFKgygZseAiAC6AfgBcUHAFcAtgCMItbgF8A3N24AzTTiRwA9jiiMIKWbnyESZHJWp0GLAAUvHxQcHIeIOpu8qDcAJTq8EhoWO4KrDyhUAgONFDiFJoANsBJiK5psSCsHJymIXyFJcD84emgiuxQAMyNBS6atI7NpWZG5rk4+aPA3c7AgQDk2BZL8eYA9Js5drTiSMUgUACGxcV2AO4QmNyzAHSrZrP8KxBrimbc27v7EIfHTBwChnC7XW4PRgACzgzwkLVe0LgS0+3CAA)

My local environment has gone screwy so I need to fix that and test this a bit more locally, but I think this should be right!

packages/rpc-core/src/rpc-methods/getBlockProduction.ts
```diff
@@ -0,0 +1,47 @@
...
+    ): GetBlockProductionApiResponseBase & GetBlockProductionApiResponseWithSingleIdentity<TIdentity>;
+
+    getBlockProduction(
+        config?: Readonly<{
+            commitment?: Commitment;
+            identity: undefined;
+            range?: SlotRange;
+        }>
+    ): GetBlockProductionApiResponseBase & GetBlockProductionApiResponseWithAllIdentities;
+}
```
steveluscher, line 47:
```suggestion
import { Base58EncodedAddress } from '@solana/keys';
import { Commitment, RpcResponse, U64UnsafeBeyond2Pow53Minus1 } from './common';

type NumberOfLeaderSlots = U64UnsafeBeyond2Pow53Minus1;
type NumberOfBlocksProduced = U64UnsafeBeyond2Pow53Minus1;

type SlotRange = Readonly<{
    firstSlot: U64UnsafeBeyond2Pow53Minus1;
    lastSlot: U64UnsafeBeyond2Pow53Minus1;
}>;

type GetBlockProductionApiConfigBase = Readonly<{
    commitment?: Commitment;
    range?: SlotRange;
}>;

type GetBlockProductionApiResponseBase = RpcResponse<{
    range: SlotRange;
}>;

type GetBlockProductionApiResponseWithAllIdentities = Readonly<{
    value: Readonly<{
        byIdentity: Record<Base58EncodedAddress, [NumberOfLeaderSlots, NumberOfBlocksProduced]>;
    }>;
}>;

type GetBlockProductionApiResponseWithSingleIdentity<TIdentity extends string> = Readonly<{
    value: Readonly<{
        byIdentity: Readonly<{ [TAddress in TIdentity]?: [NumberOfLeaderSlots, NumberOfBlocksProduced] }>;
    }>;
}>;

export interface GetBlockProductionApi {
    /**
     * Returns recent block production information from the current or previous epoch.
     */
    getBlockProduction<TIdentity extends Base58EncodedAddress>(
        config: GetBlockProductionApiConfigBase &
            Readonly<{
                identity: TIdentity;
            }>
    ): GetBlockProductionApiResponseBase & GetBlockProductionApiResponseWithSingleIdentity<TIdentity>;
    getBlockProduction(
        config?: GetBlockProductionApiConfigBase
    ): GetBlockProductionApiResponseBase & GetBlockProductionApiResponseWithAllIdentities;
}
```

steveluscher:
This has type errors now when you just supply `commitment` (see the tests file for the error) and still doesn't constrain the type of `byIdentity` to having only the one key. If that second thing isn't possible with Typescript, then that's cool ¯\_(°ペ)_/¯

mcintyre94:
Ah, thanks, fixed those up! On the second thing, it definitely **should** work, but also.. isn't. I don't know if some layer of indirection is throwing it here

[A more realistic playground using all the same types here does work!](https://www.typescriptlang.org/play?#code/C4TwDgpgBAcgrgWwEYQE4HkBmAZCBDAEzQGUAbAe2AGcoBeKAO0RVQG4AoUSWZtLAIQoBjANZUACqnIE4QiATqNebdp3DQylAEp4GAc2j0t+AuQakQAHgDe7KPaiYAlqirBNwAFxLkaDg6hSPDcPbyZfFQBfAD4ONW4AYXIEBCdgBAgGYEUAciSGZ1QMghyoAB8oHMlyOSoqeRy4rmgtMCFjKjAzessAFQA1PFI4CGjFY0IzCxs7BwA3IZHvAcWIDhim9Sh+YIgAVgAOAFEGIWl5AEECAlQIOsU3VCd9Te4AcQhgQRqRapkhYBOMwXMBOfLOPQ7erjExTKy2AJnFJpDJZAD83iSyPSmWA-gcqF0BgxUA8On0a3YG1UzSgHy+wl+Un+gOBoI6XQY9ShhigrXad05PQRBKJEG8ZLF61iNK29O+oj+slZDBBTg53QgAHU0gALC6kUgASSIWTSTjuMMm5nhs3sC2G4r5sJtMwCASQIBNuLSIG8xjOqAIlh5hxOZyIBCuNzuVAANFAANrwCJYXCEEgUagJlMsASMiTM2TyAC6MoC1Op8Wg8sZSoBQNV7MFmp1wF1xGeelIEG9ZtAfT7gNAUAgAA9gJkCDRHl2xkYXdMRfbVv7F7b3Q5PUPfWvrUuk71o7d7s8oL0d6ASyTk8o0yZM5R4zxU5gFWJ66WoNSKzKq5g4FOFUoAMBkfnrFVB1NYcQFHCcpxoUNjlOc4o2uE8qGiAAKO13TOAonD0bxa3AosGzZMEzAhHkoAAMlwzd7AmUxXWXRjNycaDd3PS8QHxdj7BiXCAEpiM+d8IMbNUNS5CAaNoulxLrMiVWkltZLbDsux7XioJ9UAxjYgIe2yTj9ItKhvGsQ9j1jKAzwvLirxvXM+BwB9UA8Z9XIwN8C0-AgS2-RRrEifj3TM-sLMTfCIQAOkimCgvoRMAEY4wAJhLcKAgYglPjgVAGCgIzGIdJYSrygSoG3Jy-XspyLLjKqBMJCkrMKEIs28AAGBMgi6yhesiFqK2a9iRorVR8LcGrlOkZVG0UUCJJUxssLYxLuJyPAkCEHIqWE1QkHmllGzi8qIDi2r9JAOLdqEDgTtIhbyIYC7Vmur06riohMFYIA)

I think it's probably worth parking for now, but it would be really nice to have that type safety!

PR #1267, opened 2023-04-18
refactor(experimental): add a `solana-client` version header to the default RPC

steveluscher:
refactor(experimental): add a `solana-client` version header to the default RPC
## Summary

In this PR we implement the build step that defines `__VERSION__` as the NPM package version, and config that adds the `solana-client` header to each request. RPC providers, at their option, can use this header to slice their metrics.

## Test Plan

```
```


mcintyre94 (approved):
Looks great! Aside: That disallowed headers type stuff is very slick!

steveluscher:
Slick-ish. It's an egregious Typescript hack that forms the stuff of nightmares.

PR #1270, opened 2023-04-19
refactor(experimental): coalesce requests with identical in-flight requests

steveluscher:
refactor(experimental): coalesce requests with identical in-flight requests
## Background

Throwback to #1208. There, we implemented something I call ‘request coalescing’ at the request of @Jac0xb. He was doing a _ton_ of transaction confirmations in a long-running process, all of them using the blockheight-based confirmation strategy. This resulted in a _ton_ of `getBlockHeight` requests to their RPC infra.

We decided that because the block height depends on nothing but time, essentially, that there was no point in firing off multiple of these requests at the same time. Instead we coalesced all calls into a single request; we either started a new request or associated our request with the currently-in-fight request. This ensured that there was only one request for the same arguments in flight at any given time.

## Summary

In this PR we decide how to bring this into the rewrite of web3.js in a general way. One of this PR's central questions is: can we apply request coalescing to _every_ method?

### Request-in-flight based coalescing

One option is to allow an infinite number of requests to ‘get behind’ one in-flight network fetch. If no fetch exists, the first call creates it. While a fetch is in flight, the next calls add themselves as subscribers to it. The network fetch doesn't have its `AbortSignal` triggered until the last subscriber has called `abort()`.

* If a request is _not_ in flight for a call's parameters, create the request
* If a request _is_ in flight for a call's parameters, add yourself as a subscriber

*Cons*: If a request takes a long time, this might yield unexpected behavior. Imagine that a `getBlockHeight` request takes 5 _seconds_ to respond, and there are calls made at t0, t1, t2, t3, and t4. The value returned will be a different amount of stale from the perspective of each one of those calls. If the call at t4 was allowed to make its own network request, then it _might_ get fresher data.

### Time-windowed coalescing

To combat the effect of slow responses described above, we could add a rule to the coalescing algorithm:

* If a request is _not_ in flight for a call's parameters, create the request
* If a request has been in flight for a call's parameters for more than X ms, create a new request
* If a request has been in flight for a call's parameters for X or fewer ms, add yourself as a subscriber

*Cons*: While this _minimizes_ the effect of slow responses, X still represents a magic number, and each method might require a different tuning of X given its characteristics.

### JavaScript runloop-based coalescing (✅ chose this one)

A variation of the time-windowed algorithm might be to make the ‘time window’ exactly the length of one JavaScript runloop. This is to say that only requests made in the _same_ runloop get coalesced together. You can think of the size of this window as ‘zero’ and the kinds of requests that it coalesces together as being the same as those typically made in a `for` loop.

*Cons*: While this will certainly reduce fetches, without the staleness problems of the last two approaches, ‘identical’ requests made very close together in time but in different runloops will not benefit from coalescing.

## Test Plan

```ts
const rpc = createSolanaRpc({ url: 'https://api.devnet.solana.com' });
for (let ii = 0; ii < 10; ii++) {
  rpc.getBlockHeight.send().then(console.log);
}

```

Observe that only one network request gets made, but all logs report the same value.

### DIFF
--- a/packages/library/src/rpc-request-coalescer.ts
+++ b/packages/library/src/rpc-request-coalescer.ts
@@ -1,0 +1,65 @@
+import { IRpcTransport } from '@solana/rpc-transport/dist/types/transports/transport-types';
+
+type CoalescedRequest = {
+    readonly abortController: AbortController;
+    numConsumers: number;
+    readonly responsePromise: Promise<unknown>;
+};
+
+type GetDeduplicationKeyFn = (payload: unknown) => string | undefined;
+
+export function getRpcTransportWithRequestCoalescing(
+    transport: IRpcTransport,
+    getDeduplicationKey: GetDeduplicationKeyFn
+): IRpcTransport {
+    let coalescedRequestsByDeduplicationKey: Record<string, CoalescedRequest> | undefined;
+    return async function makeCoalescedHttpRequest<TResponse>(
+        config: Parameters<IRpcTransport>[0]
+    ): Promise<TResponse> {
+        const { payload, signal } = config;
+        const deduplicationKey = getDeduplicationKey(payload);
+        if (deduplicationKey === undefined) {
+            return await transport(config);
+        }
+        if (!coalescedRequestsByDeduplicationKey) {
+            Promise.resolve().then(() => {
+                coalescedRequestsByDeduplicationKey = undefined;
+            });
+            coalescedRequestsByDeduplicationKey = {};
+        }
+        if (coalescedRequestsByDeduplicationKey[deduplicationKey] == null) {
+            const abortController = new AbortController();
+            coalescedRequestsByDeduplicationKey[deduplicationKey] = {
+                abortController,
+                numConsumers: 0,
+                responsePromise: transport<TResponse>({
+                    ...config,
+                    signal: abortController.signal,
+                }),
+            };
+        }
+        const coalescedRequest = coalescedRequestsByDeduplicationKey[deduplicationKey];
+        coalescedRequest.numConsumers++;
+        if (signal) {
+            const responsePromise = coalescedRequest.responsePromise as Promise<TResponse>;
+            return await new Promise<TResponse>((resolve, reject) => {
+                const handleAbort = (e: AbortSignalEventMap['abort']) => {
+                    signal.removeEventListener('abort', handleAbort);
+                    coalescedRequest.numConsumers -= 1;
+                    if (coalescedRequest.numConsumers === 0) {
+                        const abortController = coalescedRequest.abortController;
+                        abortController.abort();
+                    }
+                    const abortError = new DOMException((e.target as AbortSignal).reason, 'AbortError');
+                    reject(abortError);
+                };
+                signal.addEventListener('abort', handleAbort);
+                responsePromise.then(resolve).finally(() => {
+                    signal.removeEventListener('abort', handleAbort);
+                });
+            });
+        } else {
+            return (await coalescedRequest.responsePromise) as TResponse;
+        }
+    };
+}
--- a/packages/library/package.json
+++ b/packages/library/package.json
@@ -66,7 +66,8 @@
     "dependencies": {
         "@solana/keys": "workspace:*",
         "@solana/rpc-core": "workspace:*",
-        "@solana/rpc-transport": "workspace:*"
+        "@solana/rpc-transport": "workspace:*",
+        "fast-stable-stringify": "^1.0.0"
     },
     "devDependencies": {
         "@solana/eslint-config-solana": "^1.0.0",
--- a/packages/library/src/__tests__/rpc-request-coalescer-test.ts
+++ b/packages/library/src/__tests__/rpc-request-coalescer-test.ts
@@ -1,0 +1,191 @@
+import { IRpcTransport } from '@solana/rpc-transport/dist/types/transports/transport-types';
+import { getRpcTransportWithRequestCoalescing } from '../rpc-request-coalescer';
+
+describe('RPC request coalescer', () => {
+    let coalescedTransport: IRpcTransport;
+    let hashFn: jest.MockedFunction<() => string | undefined>;
+    let mockTransport: jest.MockedFunction<IRpcTransport>;
+    beforeEach(() => {
+        jest.useFakeTimers();
+        hashFn = jest.fn();
+        mockTransport = jest.fn();
+        coalescedTransport = getRpcTransportWithRequestCoalescing(mockTransport, hashFn);
+    });
+    describe('when requests produce the same hash', () => {
+        beforeEach(() => {
+            hashFn.mockReturnValue('samehash');
+        });
+        it('multiple requests in the same tick produce a single transport request', () => {
+            coalescedTransport({ payload: null });
+            coalescedTransport({ payload: null });
+            expect(mockTransport).toHaveBeenCalledTimes(1);
+        });
+        it('multiple requests in different ticks each produce their own transport request', async () => {
+            expect.assertions(1);
+            coalescedTransport({ payload: null });
+            await jest.runOnlyPendingTimersAsync();
+            coalescedTransport({ payload: null });
+            expect(mockTransport).toHaveBeenCalledTimes(2);
+        });
+        it('multiple requests in the same tick receive the same response', async () => {
+            expect.assertions(2);
+            const mockResponse = { response: 'ok' };
+            mockTransport.mockResolvedValueOnce(mockResponse);
+            const responsePromiseA = coalescedTransport({ payload: null });
+            const responsePromiseB = coalescedTransport({ payload: null });
+            await Promise.all([
+                expect(responsePromiseA).resolves.toBe(mockResponse),
+                expect(responsePromiseB).resolves.toBe(mockResponse),
+            ]);
+        });
+        it('multiple requests in different ticks receive different responses', async () => {
+            expect.assertions(2);
+            const mockResponseA = { response: 'okA' };
+            const mockResponseB = { response: 'okB' };
+            mockTransport.mockResolvedValueOnce(mockResponseA);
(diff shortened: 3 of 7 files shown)


steveluscher:
@joncinque, @jordansexton, what do you think of this? Imagine 20 simultaneous calls to `getBlockHeight()` (very doable – just try to confirm 20 transactions at once). This code, like #1208, makes it so that if there's _already_ a request in flight for some exact flavor of `getBlockHeight()`, then all interested callers will wait on the same network request.

My question is: should we apply this technique to _every_ RPC method? Can you think of an instance where coalescing requests behind a single fetch is _not_ the right thing to do? Even in the case of `sendTransaction` I'm not sure it would be _bad_ to coalesce requests. Transactions use a blockhash as an idempotency key, so there's no reason not to coalesce identical `sendTransactions()` right?

Thoughts, in general, on this PR?

packages/library/src/rpc-request-coalescer.ts
```diff
@@ -0,0 +1,63 @@
...
+export function getRpcTransportWithRequestCoalescing(transport: IRpcTransport): IRpcTransport {
+    const coalescedRequestsByHash: Record<string, CoalescedRequest> = {};
+    return async function makeCoalescedHttpRequest<TResponse>(
+        config: Parameters<IRpcTransport>[0]
+    ): Promise<TResponse> {
+        const { payload, signal } = config;
+        if (!shouldCoalesceRequest(payload)) {
+            return await transport(config);
+        }
+        const requestHash = fastStableStringify([payload.method, payload.params]);
```
steveluscher, line 22: Uniquely identify a request by its RPC method and params.

packages/library/src/rpc-request-coalescer.ts
```diff
@@ -0,0 +1,63 @@
...
+                    try {
+                        return await transport<TResponse>({
+                            ...config,
+                            signal: abortController.signal,
+                        });
+                    } finally {
+                        delete coalescedRequestsByHash[requestHash];
+                    }
+                })(),
+            };
```
steveluscher: Do you not already have an in-flight request for this method/params combo? Make one!

packages/library/src/rpc-request-coalescer.ts
```diff
@@ -0,0 +1,63 @@
...
+        }
+        const coalescedRequest = coalescedRequestsByHash[requestHash];
+        coalescedRequest.numConsumers++;
```
steveluscher, line 41: Otherwise, just grab the existing one and add yourself as a consumer.

packages/library/src/rpc-request-coalescer.ts
```diff
@@ -0,0 +1,63 @@
...
+        if (signal) {
+            const handleAbort = (e: AbortSignalEventMap['abort']) => {
+                signal.removeEventListener('abort', handleAbort);
+                coalescedRequest.numConsumers -= 1;
+                if (coalescedRequest.numConsumers === 0) {
+                    const { reason } = e.target as AbortSignal;
+                    const abortController = coalescedRequest.abortController;
+                    delete coalescedRequestsByHash[requestHash];
+                    abortController.abort(reason);
+                }
```
steveluscher, line 51: Even if you trigger the `AbortSignal` on your request, don't nuke the fetch until you're the _last_ interested party.

steveluscher:
I just added a bunch more detail to this PR summary, describing three approaches to request coalescing. I'd be interested in your thoughts here, @Jac0xb, @mertimus, @t-nelson, @CriesofCarrots, or anyone you can think to tag in here!

packages/library/src/rpc-request-coalescer-gating.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export function shouldCoalesceRequest(payload: unknown): payload is IHasMethod & IHasParams {
+    if (payload == null || typeof payload !== 'object' || Array.isArray(payload)) {
+        return false;
+    }
+    if (!('jsonrpc' in payload) || payload.jsonrpc !== '2.0' || !('method' in payload)) {
+        return false;
+    }
+    switch (payload.method) {
+        case 'sendTransaction':
```
joncinque, line 16: Let's coalesce `sendTransaction` too -- there's no reason to send the same exact transaction if it's already in-flight.

packages/library/src/rpc-request-coalescer-gating.ts
```diff
@@ -0,0 +1,22 @@
+interface IHasParams {
+    params: unknown;
+}
+interface IHasMethod {
+    method: string;
+}
+
+export function shouldCoalesceRequest(payload: unknown): payload is IHasMethod & IHasParams {
+    if (payload == null || typeof payload !== 'object' || Array.isArray(payload)) {
```
joncinque, line 9: Sorry, this is my silly lack of knowledge of TypeScript. If you're already specifying `payload is IHasMethod & IHasParams`, how could it not be an object or be an array?

joncinque (approved):
Just one little point, the rest looks good

packages/library/src/rpc-request-coalescer-gating.ts
steveluscher: That `payload is IHasMethod & IHasParams` assertion is actually on the _return_ type. This function I've written is what's called a ‘type guard.’ We're basically saying that if you assert, anywhere in your program, that the return value of this function is `true` then `payload` is `IHasMethod & IHasParams`;

```ts
function isFish(pet: Pet): pet is Fish {
  return pet.swims === true;
}

function addToFishtank(fish: Fish) {
  /* ... */
}

const pet = getPet();
addToFishtank(pet); // ERROR: Expected `Fish`
if (isFish(pet)) {
  addToFishtank(pet); // OK
}
```

steveluscher: So tl;dr `payload` enters that function a pile of who knows what, and we run assertions on it to the point where if the function returns `true`, Typescript will trust that `payload` is a `IHasMethod & IHasParams`.

steveluscher:
I'm still not convinced that this is the right implementation to land on. I think that I should either:

1. Close this PR and forget about it; let people write their own if they want coalescing and are willing to deal with the tradeoffs, or
2. Implement the design called ‘Time-windowed coalescing,’ but only for an allowlist of methods whose return values can be thought of as changing at a predictable cadence (ie. every slot), or
3. Implement the design called ‘JavaScript runloop-based coalescing,’ for every method.

I'm conflicted.

The question is not which of these approaches to take – all of them should be made available – the question is which one to bake into the library for everyone, by default.

mschneider:
this seems like functionality that should rather be implemented on a load balancer / haproxy then a client library

steveluscher:
> this seems like functionality that should rather be implemented on a load balancer / haproxy then a client library…

How though? HAProxy can't see inside the body of a POST request to compute the stable hash of the method/params, right?

I'm kinda of the mind that the best kind of request is the one you never make.

thesoftwarejedi:
Code looks good.  Given that this is just for inflight calls, I think this is a great idea to be **able to toggle on**.

mschneider:
> > this seems like functionality that should rather be implemented on a load balancer / haproxy then a client library…
> 
> How though? HAProxy can't see inside the body of a POST request to compute the stable hash of the method/params, right?

You would need to ask the triton team about that. It definitely works. Don't think the implementation detail is relevant

> I'm kinda of the mind that the best kind of request is the one you never make.

I wouldn't use a client library that just magically let's requests disappear. That's really not what I expect of any "Connection" style library. Most people I work with like javascript because you can quickly try out a new idea, the more abstraction you layer on top of each other the harder it becomes to reason about the result and hence actually learn. If something needs to be optimized we anyways rewrite it in rust.

steveluscher:
> Don't think the implementation detail is relevant

If the proxy is going to build per-method-and-params caches, doesn't it need to be able to:

1. Read the method/params from the `POST` body
2. Convert them into a stable hash (ie. `getFoo:{"commitment":"confirmed","minSlotHeight":123}` has to produce the same hash as `getFoo:{"minSlotHeight":123,"commitment":"confirmed"}`)
3. Use those hashes to route the request through the correct read-through cache

Is there a proxy that can do that?

> I wouldn't use a client library that just magically let's requests disappear.

100% true. The new primitives I'm building will have unopinionated cores (eg. `@solana/rpc-core`). Also true though is that public JavaScript infrastructure like `@solana/web3.js` is a big lever. The things we do at that point of leverage can make huge numbers of apps work better or worse. Coalescing _in general_ was an [idea from the ecosystem](https://github.com/solana-labs/solana-web3.js/issues/1182), who was struggling with floods of wasted requests. If there's something I can do from here to make their app and everyone else's app better by default, I should probably do it.

steveluscher:
> Configurable caching (we use like 2000ms - too much for some but works well for us as a dash) so gAI calls return instantly with no network call.

You'll be able to do that with the new web3.js _super_ easy. The transport interface is:

```ts
<TResponse>(config: {payload: unknown, signal?: AbortSignal}) => Promise<TResponse>;
```

You will be able to wrap that to your heart's content:

```ts
function withStepFinanceCustomCachingRules(transport: IRpcTransport): IRpcTransport {
  return async function(config) {
    const shouldCache = config.payload.method === 'getAccountInfo';
    const cacheKey = shouldCache ? hash(config.payload) : null;
    if (shouldCache && cacheWithTTL.has(cacheKey)) {
      return cacheWithTTL.get(cacheKey);
    }
    const response = await transport(config);
    if (shouldCache) {
      cacheWithTTL.set(cacheKey, response);
    }
    return response;
  }
}
```

…then to use it…

```ts
const customTransport = withStepFinanceCustomCachingRules(
  createDefaultRpcTransport({ url: '...' }),
);

const rpc = createSolanaRpc({ transport: customTransport });
```

steveluscher:
Alright, I ended up rewriting this to coalesce all identical requests in the same JavaScript event loop. This almost certainly captures most (pathological) cases where someone is firing off network requests in a loop.

Wrote a load of tests. Ready to land, unless anyone has any feedback!

PR #1271, opened 2023-04-19
refactor(experimental): add RPC methods returning just a number

joncinque:
#### Summary

While doing one of the simple RPC calls, I noticed there's a bunch of simple RPC calls that take just a commitment with optional minContextSlot, or nothing at all, and return either a number, or a number wrapped in an rpc response.  The calls covered here are:

* getFirstAvailableBlock (take nothing, just give number)
* getMaxRetransmitSlot (ditto)
* getMaxShredInsertSlot (ditto)
* getSlot (takes commitment and context, just gives number)
* getStakeMinimumDelegation (takes ONLY commitment, wraps the number)
* getTransactionCount (takes commitment and context, just gives number)
* minimumLedgerSlot (takes nothing, just gives a number)

Since so many of the tests just ended up doing the same as the `getBlockHeight` test, I moved them all into the same file. Then I noticed that some of them take parameters and some don't, and still kept them in the same file. Then I noticed some return just a number, and others wrap it, and still I kept them in the same file.

So if you want these broken up differently, I'm totally game! Also, I used the `each` construction, which I'm not sure if that's best or if you prefer something else.

#### Test Plan

```
pnpm test
```

Also, I ran prettier by hand with `pnpm prettier --write '{,{src,test}/**/}*.{j,t}s'`, I didn't find any other scripts or configuration, and it seemed to pass the prettier test.


joncinque:
Changing back to draft while I figure out the build error, sorry about that

joncinque:
Ok all good, just had to use `Reflect.get` in the tests rather than `obj[func]`

packages/rpc-core/src/rpc-methods/getFirstAvailableBlock.ts
```diff
@@ -0,0 +1,10 @@
+import { RpcResponse, Slot } from './common';
+
+type GetFirstAvailableBlockApiResponse = RpcResponse<Slot>;
```
steveluscher, line 3: This doesn't return an `RpcResponse` in fact. It's just the raw integer!

```ts
// Not an `RpcResponse<TValue>`...
{context: {...}, value: 1234}
// Actually just `TValue`
1234
```
```suggestion
type GetFirstAvailableBlockApiResponse = Slot;
```

packages/rpc-core/src/rpc-methods/getMaxRetransmitSlot.ts
```diff
@@ -0,0 +1,10 @@
+import { RpcResponse, Slot } from './common';
+
+type GetMaxRetransmitSlotApiResponse = RpcResponse<Slot>;
```
steveluscher, line 3:
```suggestion
type GetMaxRetransmitSlotApiResponse = Slot;
```

packages/rpc-core/src/rpc-methods/getMaxShredInsertSlot.ts
```diff
@@ -0,0 +1,10 @@
+import { RpcResponse, Slot } from './common';
+
+type GetMaxShredInsertSlotApiResponse = RpcResponse<Slot>;
```
steveluscher, line 3:
```suggestion
type GetMaxShredInsertSlotApiResponse = Slot;
```

packages/rpc-core/src/rpc-methods/getSlot.ts
```diff
@@ -0,0 +1,15 @@
+import { Commitment, RpcResponse, Slot } from './common';
+
+type GetSlotApiResponse = RpcResponse<Slot>;
```
steveluscher, line 3:
```suggestion
type GetSlotApiResponse = Slot;
```

packages/rpc-core/src/rpc-methods/getStakeMinimumDelegation.ts
```diff
@@ -0,0 +1,14 @@
+import { RpcResponse, U64UnsafeBeyond2Pow53Minus1, Commitment } from './common';
+
+type GetStakeMinimumDelegationApiResponse = RpcResponse<U64UnsafeBeyond2Pow53Minus1>;
```
steveluscher, line 3: 👍🏻

steveluscher: I think at this point it's worth creating a branded `Lamports` type. See `Base58EncodedAddress` for how to create a branded type.

This will imply that you can create a function that only accepts `Lamports` as a type and if you try to pass it a non-branded `BigInt` you'll get a Typescript error. Should save a couple million in mistake fees.

packages/rpc-core/src/rpc-methods/getTransactionCount.ts
```diff
@@ -0,0 +1,17 @@
+import { U64UnsafeBeyond2Pow53Minus1, Commitment, Slot } from './common';
+
+type GetTransactionCountApiResponse = U64UnsafeBeyond2Pow53Minus1;
```
steveluscher, line 3: 👍🏻

packages/rpc-core/src/rpc-methods/minimumLedgerSlot.ts
```diff
@@ -0,0 +1,10 @@
+import { Slot } from './common';
+
+type MinimumLedgerSlotApiResponse = Slot;
```
steveluscher, line 3: 👍🏻

packages/rpc-core/src/rpc-methods/minimumLedgerSlot.ts
```diff
@@ -0,0 +1,10 @@
...
+
+export interface MinimumLedgerSlotApi {
+    /**
+     * Returns the lowest slot that the node has information about in its ledger.
```
steveluscher, line 7:
```suggestion
     * Returns the lowest slot that the node has information about in its ledger.
     * This value may increase over time if the node is configured to purge older ledger data.
```

steveluscher (changes requested):
giphy

I _do_ indeed prefer the ‘repetitive but separate’ approach to tests, because:

* reading through all the interpolation and metaprogramming in this PR's test makes me think too hard
* if (when) someone breaks the RPC implementation for _one_ of these suckers it's going to show up in a test called ‘getBigInt’ which will again require thinking
* some of these methods are actually different (some return context, and others a raw value) so you have logic in this file that doesn't actually apply to all.
* You could imagine adding more special cases here which would make these tests even more dissimilar. Could you actually start the validator with a custom minimum stake delegation and then add a test for that? Might you write a test that actually lands a transaction and asserts that the value of `getTransactionCount()` increased?
* when we add a new RPC method that only returns an integer, it will be really obvious that the author didn't add a test if there's literally no file for it, but less obvious if it's just missed here. A 1:1 correspondence between a method definition and a test file makes it easy to spot.

I was going to say ‘despite all that I don't feel _super_ strongly about this,’ but the more bullet points I added the more I think… yeah maybe I do :)

packages/rpc-core/src/rpc-methods/getFirstAvailableBlock.ts
joncinque: Well at least I got it right in my PR description :sweat: any idea about why this didn't get picked up by the compiler or the tests?

packages/rpc-core/src/rpc-methods/getStakeMinimumDelegation.ts
joncinque: Yeah sure, where do you want it go?

packages/rpc-core/src/rpc-methods/getFirstAvailableBlock.ts
steveluscher: Yeah! I mean we could write _Typescript_ tests, but that’s pretty next level.

packages/rpc-core/src/rpc-methods/getStakeMinimumDelegation.ts
steveluscher: I have a `@solana/keys` package which is a nice place to put the address types, but for everything else I’ve just been hucking ‘primitive’ types into global typedefs.

Throw it anywhere for now, just try to avoid circular references between modules.

joncinque: Ok, put it under `common.ts`

joncinque:
Ok, this should be ready for another look!

packages/rpc-core/src/rpc-methods/getFirstAvailableBlock.ts
steveluscher: In case it wasn't clear what I meant by this, some people _actually_ have written tests that assert whether source code raises type errors or not. Pretty extreme.

packages/rpc-core/src/rpc-methods/__tests__/get-first-available-block-test.ts
```diff
@@ -0,0 +1,23 @@
...
+    beforeEach(() => {
+        fetchMock.resetMocks();
+        fetchMock.dontMock();
+        rpc = createJsonRpc<SolanaRpcMethods>({
+            api: createSolanaRpcApi(),
+            transport: createHttpTransport({ url: 'http://127.0.0.1:8899' }),
+        });
+    });
+    describe('when called with no parameters', () => {
+        it('returns a bigint', async () => {
```
steveluscher, line 17: I guess there are probably reasons why it's not safe to presume that the test validator will always return `0n` here? I seem to remember that being a problem before: https://github.com/solana-labs/solana/issues/23853

packages/rpc-core/src/rpc-methods/getFirstAvailableBlock.ts
joncinque: Haha don't worry, it was *very* clear. I didn't want to go there under any circumstances :sweat_smile:

packages/rpc-core/src/rpc-methods/__tests__/get-stake-minimum-delegation-test.ts
```diff
@@ -0,0 +1,26 @@
...
+            api: createSolanaRpcApi(),
+            transport: createHttpTransport({ url: 'http://127.0.0.1:8899' }),
+        });
+    });
+    (['confirmed', 'finalized', 'processed'] as Commitment[]).forEach(commitment => {
+        describe(`when called with \`${commitment}\` commitment`, () => {
+            it('returns the result as a bigint wrapped in an RpcResponse', async () => {
+                expect.assertions(1);
+                const result = await rpc.getStakeMinimumDelegation({ commitment }).send();
+                expect(result.value).toEqual(expect.any(BigInt));
```
steveluscher, line 22: Should we drop a `TODO` here to actually fetch the minimum delegation value from the stake program?

Probably not, now that I think about it. This test shouldn't really be in the business of _testing_ the underlying method. It should stop short of testing the type of the responses, as you've done here.

packages/rpc-core/src/rpc-methods/common.ts
```diff
@@ -19,3 +19,5 @@ export type RpcResponse<TValue> = Readonly<{
     }>;
     value: TValue;
 }>;
+
+export type Lamports = U64UnsafeBeyond2Pow53Minus1 & { readonly __lamports: unique symbol };
```
steveluscher, line 23: Mmmm… I think this is more like…

```suggestion
export type LamportsUnsafeBeyond2Pow53Minus1 = bigint & { readonly __lamports: unique symbol };
```

packages/rpc-core/src/rpc-methods/minimumLedgerSlot.ts
```diff
@@ -0,0 +1,11 @@
...
+     * This value may increase over time if the node is configured to purge older ledger data.
+     */
+    minimumLedgerSlot(): MinimumLedgerSlotApiResponse;
```
steveluscher, line 10: LOL @ no `get`.

steveluscher (approved):
Thank you!!! I'll fix up my own nits and land this.

packages/rpc-core/src/rpc-methods/__tests__/get-stake-minimum-delegation-test.ts
joncinque: Yeah exactly, that was my intention

packages/rpc-core/src/rpc-methods/minimumLedgerSlot.ts
joncinque: Same :smile:

packages/rpc-core/src/rpc-methods/common.ts
joncinque: As you wish! In that case, do you also want to do `SlotUnsafeBeyond2Pow53Minus1`?

steveluscher: I asked how long it would take to get there and @willhickey [told me](https://discord.com/channels/428295358100013066/439194979856809985/1075222932936216647) ~114M years, so I think I decided against it. 

image

steveluscher: Not opposed to it, for completeness, though.

joncinque: ah gotcha, that definitely makes sense

PR #1272, opened 2023-04-20
refactor(experimental): delete the batch API

steveluscher:
refactor(experimental): delete the batch API
## Summary

I had a long discussion with RPC providers about the JSON-RPC batch API.

In summary,

* it causes problems with resource exhaustion,
* makes it difficult to balance load,

…on top of having the user-facing drawbacks of,

* blowing the entire request up if even one method fails,
* making the entire batch take as long as the slowest request

For all of these reasons, and the detail that I go into [here](https://twitter.com/steveluscher/status/1532443998488363008), we're going to eliminate the batch API as designed here.

*This doesn't mean that you can't still batch in Solana apps*.

1. In fact, you can use the API that remains to do _proper_ batching using JavaScript primitives,
2. We're going to commit to delivering HTTP/2 pipelining in web3.js in _both_ Node and the browser.

## Test Plan

**DON'T DO THIS:**

```ts
// These execute in series. Each network request will wait
// for the one before it to finish before even starting.
const latestBlockhash = await rpc.getLatestBlockhash().send();
const accountInfo = await rpc.getAccountInfo(ownerAddress).send();
const epoch = await rpc.getEpoch().send();
```

Do this instead:

```ts
// These execute in parallel and will all share the same
// connection wherever HTTP/2 is available.
const [
  latestBlockhash,
  accountInfo,
  epoch,
] = await Promise.allSettled([
  rpc.getLatestBlockhash().send(),
  rpc.getAccountInfo(ownerAddress).send(),
  rpc.getEpoch().send(),
]);
// Example result:
//   { status: 'fulfilled', value: 'QaMAF1AaapDdocMdZgXqKX2ixd7YYSxTpKHMcsbcF318' },
//   { status: 'rejected', reason: Error },
//   { status: 'fulfilled', value: 420 },
```

Using `Promise.allSettled` you get _all_ of the benefits of batching, _plus_ the benefits of being able to properly load balance your requests (fewer rate limits and blocks), _plus_ you get to salvage successful requests despite _some_ of the requests in your batch having failed.

### DIFF
--- a/packages/rpc-transport/src/__tests__/json-rpc-test.ts
+++ b/packages/rpc-transport/src/__tests__/json-rpc-test.ts
@@ -36,45 +36,11 @@ describe('JSON-RPC 2.0', () => {
             payload: { ...createJsonRpcMessage('someMethod', [123]), id: expect.any(Number) },
         });
     });
-    it('sends a batch of requests to the transport', () => {
-        rpc.someMethod(123).someMethod(456).sendBatch();
-        expect(makeHttpRequest).toHaveBeenCalledWith({
-            payload: [
-                { ...createJsonRpcMessage('someMethod', [123]), id: expect.any(Number) },
-                { ...createJsonRpcMessage('someMethod', [456]), id: expect.any(Number) },
-            ],
-        });
-    });
     it('returns results from the transport', async () => {
         expect.assertions(1);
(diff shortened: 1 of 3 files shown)


joncinque (approved):
So long, `Armed*` types. To be honest, I do find the `Pending*` name to be clearer. Looks good to me!

PR #1281, opened 2023-04-26
refactor(experimental): add the `getLatestBlockhash` RPC method

mcintyre94:
- I've added a branded `Blockhash` type, since we'll use blockhash as input for some RPC methods
- I haven't added the deprecated `getRecentBlockhash` method, assuming we're only targeting 1.9+ here

---


packages/rpc-core/src/index.ts
```diff
@@ -14,6 +14,7 @@ import { GetSlotApi } from './rpc-methods/getSlot';
 import { GetStakeMinimumDelegationApi } from './rpc-methods/getStakeMinimumDelegation';
 import { GetTransactionCountApi } from './rpc-methods/getTransactionCount';
 import { MinimumLedgerSlotApi } from './rpc-methods/minimumLedgerSlot';
+import { GetLatestBlockhashApi } from './rpc-methods/getLatestBlockhash';
```
steveluscher, line 17: image

This is on me. I need to add the `sort-imports` linter upstream.

packages/rpc-core/src/rpc-methods/common.ts
```diff
@@ -24,3 +24,5 @@ export type RpcResponse<TValue> = Readonly<{
     }>;
     value: TValue;
 }>;
+
+export type Blockhash = string & { readonly __blockhash: unique symbol };
```
steveluscher, line 28: Smart. I guess this will need an assertion function too!

steveluscher (approved):
thank-you-cute-jumping-baby-ray-t1eqae9qyrudb38v

PR #1283, opened 2023-04-27
refactor(experimental): add `getSignaturesForAddress` API

mcintyre94:
- Tests don't include any accounts with transactions yet. It seems that the fixture account is seeded without a transaction, and even the vote account doesn't reliably have transactions in unit tests. I think we'll need to have transaction sending implemented to test this
- I've only typed this in as much detail as the RPC docs, specifically the `err` field on each transaction is just an object. Do we want to try to create a `TransactionError` type or is this the right level of granularity? 
- I'm just returning `[GetSignaturesForAddressTransaction]`, but this RPC call does guarantee an ordering of that array. Would it be useful to type this in a way that describes its sort order, so that callers can have their own functions specify that type when they care about ordering?
- I've typed `blockTime` as u64. The RPC docs say it's i64 but it's a unix epoch so I don't think it's worth adding a distinct i64 type for it here

----


packages/rpc-core/src/rpc-methods/__tests__/get-signatures-for-address-test.ts
```diff
@@ -0,0 +1,57 @@
...
+            });
+        });
+    });
+
+    describe('given an account with transactions', () => {
+        // TODO Need to be able to send transactions
+        it.todo('returns the transactions for that account');
+        it.todo('returns only the first transactions when called with a `limit`');
+        it.todo('returns transactions from `since` if defined');
+        it.todo('returns transactions only to `until` if defined');
```
steveluscher, line 37: Thanks for leaving these here. I spent a few minutes digging around. What do you think the solution to this is?

* Make it so that you can start up the test validator with an array of transactions to play into it (eg. `--transaction-fixtures` or something)
* Literally play transactions into the validator when we start it up (eg. using cURL in `start-shared-test-validator.sh`)
* Create a one-time script that plays a bunch of transactions into the validator, then commit the `test-ledger/` directory that results.
* Something else?

packages/rpc-core/src/rpc-methods/common.ts
```diff
@@ -24,3 +24,6 @@ export type RpcResponse<TValue> = Readonly<{
     }>;
     value: TValue;
 }>;
+
+// TODO: Eventually move this into whatever package implements transactions
```
steveluscher, line 28: (Y)

packages/rpc-core/src/rpc-methods/getSignaturesForAddress.ts
```diff
@@ -0,0 +1,44 @@
...
+    memo: string | null;
+    /** estimated production time, as Unix timestamp (seconds since the Unix epoch) of when transaction was processed. null if not available. */
+    blockTime: U64UnsafeBeyond2Pow53Minus1 | null;
+    /** The transaction's cluster confirmation status */
+    confirmationStatus: Commitment | null;
+}>;
+
+type GetSignaturesForAddressApiResponse = [GetSignaturesForAddressTransaction];
+
+export type AllowedCommitmentForGetSignaturesForAddress = Exclude<Commitment, 'processed'>;
```
steveluscher, line 21: <3

steveluscher: I'd err on the side of not exporting stuff. I see all you've needed this for is a test. You can do this instead of having to export it.

```ts
['confirmed', 'finalized'] as Parameters<typeof rpc.getSignaturesForAddress>[1]['commitment'][]
```

packages/rpc-core/src/rpc-methods/common.ts
```diff
@@ -24,3 +24,6 @@ export type RpcResponse<TValue> = Readonly<{
...
+export type TransactionSignature = string & { readonly __sig: unique symbol };
```
steveluscher, line 29: Want to throw an assertion function into this PR? Sort of like I did in: https://github.com/solana-labs/solana-web3.js/commit/dfc6d588b4b7c880d74dc727399ee8a84979b480#diff-f50f62f903effcf4d511f3f216959cd18570bfd499b12a824d8fc29effee6172R5-R27

packages/rpc-core/src/rpc-methods/getSignaturesForAddress.ts
```diff
@@ -0,0 +1,44 @@
...
+type GetSignaturesForAddressConfig = Readonly<{
+    commitment?: AllowedCommitmentForGetSignaturesForAddress;
+    /** The minimum slot that the request can be evaluated at */
+    minContextSlot?: Slot;
+    /** maximum transaction signatures to return (between 1 and 1,000). Default: 1000 */
+    limit?: number;
+    /** start searching backwards from this transaction signature. If not provided the search starts from the top of the highest max confirmed block. */
+    before?: TransactionSignature;
+    /** search until this transaction signature, if found before limit reached */
+    until?: TransactionSignature;
```
steveluscher, line 32: Presumably you can supply `until` _without_ `before`? If you _can't_ then just turn this into multiple overloads, each enforcing a valid combination.

steveluscher (approved):
Accept, with nits!

love-thank-you

packages/rpc-core/src/rpc-methods/getSignaturesForAddress.ts
```diff
@@ -0,0 +1,44 @@
...
+    slot: Slot;
+    /** Error if transaction failed, null if transaction succeeded. */
+    err: object | null;
+    /** Memo associated with the transaction, null if no memo is present */
+    memo: string | null;
+    /** estimated production time, as Unix timestamp (seconds since the Unix epoch) of when transaction was processed. null if not available. */
+    blockTime: U64UnsafeBeyond2Pow53Minus1 | null;
+    /** The transaction's cluster confirmation status */
+    confirmationStatus: Commitment | null;
+}>;
```
steveluscher, line 17: Oh, actually, this is totally worth making a union type for, because if someone asserts on `err` being `null`, Typescript will narrow it to the right type.

https://www.typescriptlang.org/play?ssl=16&ssc=1&pln=16&pc=2#code/C4TwDgpgBAKgTgQwHYGcEGNgEsD2SCiccOcUAvFAN4BQUUERAXFDgEYBWEm1AvtdaEixEqDNjzkqtekyhIArgBtF0gLYRVOZimBwsSAOZQAPnKWKA3L37o8OqMAAezeMjSZcSE8LdjPhYlIKSh4oBBQwpBAraiwAMygACicAOgYgigVlAEopOjpU9U0LKAB6UqgAeQBpXnpFFGgafIdHFKKcEvKofAAlXt4gA

mcintyre94: Do you know if `memo` is guaranteed not to be present if there's an error? It looks like everything else is still there:

```json
        {
            "blockTime": 1671619698,
            "confirmationStatus": "finalized",
            "err": {
                "InstructionError": [
                    0,
                    "IllegalOwner"
                ]
            },
            "memo": null,
            "signature": "66CXkJDxKr52jxxGec4k3cK4h5R5edWZXsfxH9w3CcbtwnrAK4awK8S1LiSE7h768M4rvH332TAs3MQZ4sGwvSGF",
            "slot": 168164198
        },
```

But I don't know of any examples of failing transactions with memo instructions to check

packages/rpc-core/src/rpc-methods/getSignaturesForAddress.ts
mcintyre94: Yep, I can't see any restrictions on these combinations that we can capture in these types. Neither, just `until`, just `before`, and both are all valid.

packages/rpc-core/src/rpc-methods/__tests__/get-signatures-for-address-test.ts
mcintyre94: - I think the ideal here would be that first option, having the equivalent of `--account-dir` for transactions and us just having a directory of JSON-encoded transactions in this repo. I'm not sure what that'd look like in the validator though, and I'm not sure how we'd be able to deal with signatures. Feels like it needs a lot of design work and might need too much change in the `solana` codebase to be accepted though (thinking about the JSON -> transaction conversion mostly there, which I don't think currently exists in that codebase?)

- I'm a bit worried that doing it at the level of `sendTransaction` with curl or similar will be really tricky to debug, because we'll have just the encoded + signed transaction bytes being sent. Would be the same if we added fixtures to the validator at that level too. We'd be really reliant on documentation that nobody can easily review the correctness of, and if it was incorrect it could be really tricky to debug what's going on. Also would we be able to pre-generate transactions with valid blockhashes?

- I think I'd prefer to have transactions played on startup than commit the `test-ledger/` directory, just because it feels like a bit less indirection and would be easier to debug by just updating them and restarting. I'm guessing we won't end up needing enough transactions that we need to worry about the performance of playing them in CI, and I guess that'd be the main benefit of just checking in the ledger instead. 

- Would using a different Node transaction builder like `umi` for now be too circular? We'd have the benefit of being able to write our transactions in a Node script and easily modify/debug them, but we'd be relying on the correctness of that conversion to the transaction format.

packages/rpc-core/src/rpc-methods/getSignaturesForAddress.ts
mcintyre94: Nice, makes sense! TypeScript doesn't seem to like it though ([playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAggNnA9gdwgEwMKILbYJbDYQB2wAYogE4DiEwAyngObECGwArpRAM4WUw0abjx5QAvFADkAY0TEAZnkpE0UqAB9pStnDwAvdFIDcAKFOhIUWg2ZtOI-oOG8eWRcwlQAShFZp5OBAAHgBvUyhIqDlcAiJSAH4ALlgEFHQsWMISciobRhZ2Ll4nIREeMyioAHoAKlqoABUAC2h8YjxsDmwoHiRgKGBm9kHWqG4ARw5eAZlWYigAI2gIADdWOA52dCgR2uqIqPb3YAgADwZ+5KhibuXKSqi6huxWM87uwcp5nlYZYDw8l6diKIkGiHGdC4CwAFMtgKgSFAAIy7YhoFEAGgADLiAJQAOigABEIApWBw4MAUsjcdioPtDpE9PhgNdbth7o9Is9esBWJQBjw-JQZM08MQmEs-gBrZACtBiBSUHCjPBiYDfYi-f6AhY8EEOCBEgCSChuiAGYBVqzwaB2Q2gwoFYr5AuASpVPUd4LAUEQ5p94qYrR4A1eZ2i8iUKh2iyQMhlRMZVWWCioEGuYcoEqY3Jq9V6ItdHFIeDgao1Wp1AKBBsKRsxUDw5vTpYxaYzUBZBEhf1aGJTUVLALgWc1ubMAF8AHxmUwKUu6oFMOgFezFPhUZzlGGHfwuUQpbO5zGHOQeJjXfKGzelQ9uaPMUx4lKIRYAKwg-yg4Ui3AcBZQinadzEsaBMlZeJgBNYgwA4YBGnAaBJAABQFVgiFOSgeGCcCAygVdbAbO9tzKVwZwAbWRABdSjZBwKCcikGjjCAA))

Property 'commitment' does not exist on type 'Readonly<{ commitment?: AllowedCommitmentForGetSignaturesForAddress | undefined; minContextSlot?: number | undefined; limit?: number | undefined; before?: string | undefined; until?: string | undefined; }> | undefined'.(2339)

I think it's because the `config` object is optional, do you know if there's a way to unwrap that in this type definition?

packages/rpc-core/src/rpc-methods/common.ts
mcintyre94: Thanks! Added in 1111ae5a0e472350181438cdf0a6c30bd6e1d165

mcintyre94:
> Yeah, I think this PR is the right time to enumerate and create _all_ of the `TransactionError` shapes. If not now, when, right?

Cool! Added in 3b2b8c1350cdc48194f1255bfe3d7cc884c80f16. I'm not sure how to deal with keeping them in sync though

packages/rpc-core/src/rpc-methods/getSignaturesForAddress.ts
steveluscher: Yeah! Add `NonNullable` ([playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAggNnA9gdwgEwMKILbYJbDYQB2wAYogE4DiEwAyngObECGwArpRAM4WUw0abjx5QAvFADkAY0TEAZnkpE0UqAB9pStnDwAvdFIDcAKFOhIUWg2ZtOI-oOG8eWRcwlQAShFZp5OBAAHgBvUyhIqDlcAiJSAH4ALlgEFHQsWMISciobRhZ2Ll4nIREeMyioAHoAKlqoABUAC2h8YjxsDmwoHiRgKGBm9kHWqG4ARw5eAZlWYigAI2gIADdWOA52dCgR2uqIqPb3YAgADwZ+5KhibuXKSqi6huxWM87uwcp5nlYZYDw8l6diKIkGiHGdC4CwAFMtgKgSFAAIy7YhoFEAGgADLiAJQAOigABEIApWBw4MAUsjcdioPtDpE9PhgNdbth7o9Is9esBWJQBjw-JQZM08MQmEs-gBrZACtBiBSUHCjPBiYDfYi-f6AhY8EEOCBEgCSChuiAGYBVqzwaB2Q2gwoFYr5AuASpVPUd4LAUEQ5p94qYrR4A1eZ2i8iUKh2iyQMhlRMZVWWCioEGuYcoEqY3Jq9V6ItdHFIeDgao1Wp1AKBBsKRsxUDw5vTpYxaYzUBZBEhf1aGJTUVLALgWc1ubMAF8AHxmUwKUu6oFMOgFezFPhUZzlGGHfwuUQpbO5zGHOQeJjXfKGzelQ9uaPMUx4lKIRYAKwg-yg4Ui3AcBZQinadzEsaBMlZeJgBNYgwA4YBGnAaBJAAOXkVDKTgVh4wgYIAAUBVYIhTkoHhgnAgMoFXWwGzvbcylcGcAG1kQAXRY2QcCgnIpDY4wgA)).

packages/rpc-core/src/rpc-methods/getSignaturesForAddress.ts
steveluscher: Ah. Point taken. All the properties are there all the time then. Got it.

```
curl https://api.mainnet-beta.solana.com -X POST -H "Content-Type: application/json" -d '
  {
    "jsonrpc": "2.0",
    "id": 1,
    "method": "getSignaturesForAddress",
    "params": [
      "MemoSq4gqABAXKb96qnH8TysNcWxMyWCqXgDLGmfcHr",
      {
        "before":"4aBLTpNXCtugZUfyNenEoAA7tjHnJ8j4AJ8jpDZQotCvbR2A7AXX3KfwfKkfXQsp83yYNQdW1TQCyrh6njK1ZPtd","limit":1
      }
    ]
  }
'
{"jsonrpc":"2.0","result":[{"blockTime":1684351582,"confirmationStatus":"finalized","err":{"InstructionError":[2,{"Custom":6001}]},"memo":"[27] \"Sending tx from 0 1.6.3 1\"","signature":"3ddcFK6Kf8y3r8TMF1PJV8V2Cp6TkyWV2Y3odS47JW9zWPfb8NMFxVnG3GSoGzPJ6JCGdhJmurwQMnpL2LoBdtCT","slot":194464790}],"id":1}
```

Contains the memo, even for a failed tx.

packages/rpc-core/src/__tests__/transaction-signature-test.ts
```diff
@@ -0,0 +1,78 @@
+import bs58 from 'bs58';
+import { assertIsTransactionSignature } from '../transaction-signature';
+
+describe('base58', () => {
```
steveluscher, line 4: Remove.

packages/rpc-core/src/__tests__/transaction-signature-test.ts
```diff
@@ -0,0 +1,78 @@
...
+                );
+                // eslint-disable-next-line no-empty
+            } catch {}
+            expect(decodeMethod).not.toHaveBeenCalled();
+        });
+        it('does not attempt to decode too-long input strings', () => {
+            const decodeMethod = jest.spyOn(bs58, 'decode');
+            try {
+                assertIsTransactionSignature(
+                    // 89 bytes [0, ..., 0]
```
steveluscher, line 70: The point is not actually that it's 89 bytes, it's that a 64-byte array that's full of `255` encodes (in base58) to a string 88 characters long.

An array 65 bytes long where every byte is `255` except for the first one would produce a base58 string 89 characters long.

packages/rpc-core/src/__tests__/transaction-signature-test.ts
```diff
@@ -0,0 +1,78 @@
...
+                    '1'.repeat(89)
```
steveluscher, line 71:
```suggestion
                    // 65 bytes [0, 255, ..., 255]
                    '167rpwLCuS5DGA8KGZXKsVQ7dnPb9goRLoKfgGbLfQg9WoLUgNY77E2jT11fem3coV9nAkguBACzrU1iyZM4B8roQ'
```

packages/rpc-core/src/rpc-methods/__tests__/get-signatures-for-address-test.ts
```diff
@@ -0,0 +1,57 @@
...
+            transport: createHttpTransport({ url: 'http://127.0.0.1:8899' }),
+        });
+    });
+
+    (['confirmed', 'finalized'] as AllowedCommitmentForGetSignaturesForAddress[]).forEach(commitment => {
+        describe(`when called with \`${commitment}\` commitment`, () => {
+            it('returns no transactions for a new address', async () => {
+                expect.assertions(1);
+                // This key is random, don't re-use in any tests that perform transactions
+                const publicKey = '3F6rba4VRgdGeYzgCNWQaEJUerUEQVVuwKrETigvHhJP' as Base58EncodedAddress;
```
steveluscher, line 25:
```suggestion
                const publicKey = '3F6rba4VRgdGeYzgCNWQaEJUerUEQVVuwKrETigvHhJP' as Base58EncodedAddress<'3F6rba4VRgdGeYzgCNWQaEJUerUEQVVuwKrETigvHhJP'>;
```

packages/rpc-core/src/rpc-methods/__tests__/get-signatures-for-address-test.ts
steveluscher: > I think I'd prefer to have transactions played on startup than commit the test-ledger/ directory…

I imagine that as the validator upgrades, the old `test-ledger/` directory will break and need to be regenerated?

packages/rpc-core/src/rpc-methods/getSignaturesForAddress.ts
```diff
@@ -0,0 +1,46 @@
...
+    err: TransactionError | null;
+    /** Memo associated with the transaction, null if no memo is present */
+    memo: string | null;
+    /** estimated production time, as Unix timestamp (seconds since the Unix epoch) of when transaction was processed. null if not available. */
+    blockTime: U64UnsafeBeyond2Pow53Minus1 | null;
+    /** The transaction's cluster confirmation status */
+    confirmationStatus: Commitment | null;
+}>;
+
+type GetSignaturesForAddressApiResponse = [GetSignaturesForAddressTransaction];
```
steveluscher, line 21:
```suggestion
type GetSignaturesForAddressApiResponse = readonly GetSignaturesForAddressTransaction[];
```

steveluscher (changes requested):
were-very-close-very-close

PR #1286, opened 2023-05-03
Fixed the type for lamport

### LINKED ISSUE #999: Make PublicKey.createWithSeed (js) same as create_with_seed in runtime to avoid confusion
SystemProgram CreateAccountWithSeed instruction was failing sometimes, because seed too long.

Here ./sdk/src/pubkey.rs which is used here ./runtime/src/system_instruction_processor.rs

Javascript func doesn't care about seed length https://solana-labs.github.io/solana-web3.js/file/src/publickey.js.html#lineNumber82

So you think it's good, but then the runtime does nothing (nothing reported in logs either)

Suggest: make js function same as runtime, with error if seed too long.

### LINKED ISSUE #1116: JSON RPC API using (u)int64 whereas JS can only handle int53
#### Problem

The JSON API uses int64 for fields like lamports. However, the maximum safe integer in browsers is 2^53-1

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

#### Proposed Solution

Make lamports a string in the HTTP/RPC API and parse with BigInt on the web3.js client side?

kunal-gg:
Fixes #1116
- As far as I understood we need to change the type of Lamports to String everywhere. If I understand this right, I can make further PRs, to help in the rewrite of the code.

### DIFF
--- a/packages/library-legacy/src/fee-calculator.ts
+++ b/packages/library-legacy/src/fee-calculator.ts
@@ -14,5 +14,5 @@ export const FeeCalculatorLayout = BufferLayout.nu64('lamportsPerSignature');
  */
 export interface FeeCalculator {
   /** Cost in lamports to validate a signature. */
-  lamportsPerSignature: number;
+  lamportsPerSignature: string;
 }
--- a/packages/library-legacy/test/connection.test.ts
+++ b/packages/library-legacy/test/connection.test.ts
@@ -97,7 +97,7 @@ async function mockNonceAccountResponse(
     params: [nonceAccountPubkey, {encoding: 'base64'}],
     value: {
       owner: SystemProgram.programId.toBase58(),
-      lamports: LAMPORTS_PER_SOL,
+      lamports: LAMPORTS_PER_SOL.toString(),
       data: [mockNonceAccountData.toString('base64'), 'base64'],
       executable: false,
       rentEpoch: 20,


steveluscher (changes requested):
I'm looking at the RPC responses in [the docs](https://docs.solana.com/api/http) and they appear to be u64 everywhere.

```
> curl https://api.mainnet-beta.solana.com -X POST -H "Content-Type: application/json" -d '
  {"jsonrpc":"2.0","id":1, "method":"getRecentBlockhash"}
'

{"jsonrpc":"2.0","result":{"context":{"apiVersion":"1.13.7","slot":192403900},"value":{"blockhash":"D59Fvi8JpnEMBdbeRX2PbfZreqrd2R5kxn3FAYkfWHNf","feeCalculator":{"lamportsPerSignature":5000}}},"id":1}
```

While it would be great for these to be actual `bigint` values, you can't serialize `bigint` as JSON. Serializing the integer as a string over the wire is a good strategy, doing so as you've done in this PR would break old clients.

kunal-gg:
> I'm looking at the RPC responses in [the docs](https://docs.solana.com/api/http) and they appear to be u64 everywhere.
> 
> ```
> > curl https://api.mainnet-beta.solana.com -X POST -H "Content-Type: application/json" -d '
>   {"jsonrpc":"2.0","id":1, "method":"getRecentBlockhash"}
> '
> 
> {"jsonrpc":"2.0","result":{"context":{"apiVersion":"1.13.7","slot":192403900},"value":{"blockhash":"D59Fvi8JpnEMBdbeRX2PbfZreqrd2R5kxn3FAYkfWHNf","feeCalculator":{"lamportsPerSignature":5000}}},"id":1}
> ```
> 
> While it would be great for these to be actual `bigint` values, you can't serialize `bigint` as JSON. Serializing the integer as a string over the wire is a good strategy, doing so as you've done in this PR would break old clients.

So shall I just make a js function throw an error if the seed is too long.

steveluscher:
This is actually how the new web3.js rewrite will behave ([link](https://github.com/solana-labs/solana-web3.js/pull/1226)). I think it's too late at this point to change the behaviour of the existing library, because it would break _existing_ apps.

I'm sorry that you picked this issue up with the ‘good first issue’ tag. I think it might be [more complicated](https://github.com/solana-labs/solana-web3.js/issues/1116#issuecomment-1537570357) than a quick fix.

kunal-gg:
Oh my gosh. I am so sorry for bugging you. I was just trying to play around with this library and solve some simple issues to get involved in the community. I am closing the PR right away.

steveluscher:
No need to be sorry! _Thank you_ for contributing to Solana!

PR #1294, opened 2023-05-10
Refactor (experimental): add getSupply RPC method

mcintyre94:
Adds the `getSupply` RPC method which returns the circulating/total SOL. It also returns `nonCirculatingAccounts`, which is a list of non-circulating account addresses. When called with `excludeNonCirculatingAccountsList: true`, this is an empty list.

I've found that in the test validator we start with no non-circulating accounts, and then quickly get a lot of them. But for now I've skipped the tests that require a non-empty list to be returned.

- [x] Wherever my method's return value changes based on its inputs, I've implemented that as a [Typescript function overload](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads) ([example](https://github.com/solana-labs/solana-web3.js/blob/b69daf278045bc96740450e123a322e0dd95f60e/packages/rpc-core/src/rpc-methods/getAccountInfo.ts#L80-L114))

### DIFF
--- a/packages/rpc-core/src/rpc-methods/getSupply.ts
+++ b/packages/rpc-core/src/rpc-methods/getSupply.ts
@@ -1,0 +1,44 @@
+import { Base58EncodedAddress } from '@solana/keys';
+import { Commitment, LamportsUnsafeBeyond2Pow53Minus1, RpcResponse } from './common';
+
+type GetSupplyConfig = Readonly<{
+    commitment?: Commitment;
+    excludeNonCirculatingAccountsList?: boolean;
+}>;
+
+type GetSupplyApiResponseBase = RpcResponse<{
+    /** Total supply in lamports */
+    total: LamportsUnsafeBeyond2Pow53Minus1;
+    /** Circulating supply in lamports */
+    circulating: LamportsUnsafeBeyond2Pow53Minus1;
+    /** Non-circulating supply in lamports */
+    nonCirculating: LamportsUnsafeBeyond2Pow53Minus1;
+}>;
+
+type GetSupplyApiResponseWithNonCirculatingAccounts = GetSupplyApiResponseBase &
+    Readonly<{
+        value: Readonly<{
+            /** an array of account addresses of non-circulating accounts */
+            nonCirculatingAccounts: [Base58EncodedAddress];
+        }>;
+    }>;
+
+type GetSupplyApiResponseWithoutNonCirculatingAccounts = GetSupplyApiResponseBase &
+    Readonly<{
+        value: Readonly<{
+            nonCirculatingAccounts: [];
+        }>;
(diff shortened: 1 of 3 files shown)


packages/rpc-core/src/rpc-methods/getSupply.ts
```diff
@@ -0,0 +1,49 @@
...
+                excludeNonCirculatingAccountsList: true;
+            }>
+    ): GetSupplyApiResponseWithoutNonCirculatingAccounts;
+    getSupply(
+        config: GetSupplyCommonConfig &
+            Readonly<{
+                excludeNonCirculatingAccountsList: false;
+            }>
+    ): GetSupplyApiResponseWithNonCirculatingAccounts;
+    getSupply(config?: GetSupplyCommonConfig): GetSupplyApiResponseWithNonCirculatingAccounts;
```
steveluscher, line 48: Pretty sure you only have to supply an overload for the case where `excludeNonCirculatingAccountsList` is `true`.

```suggestion
type GetSupplyConfig = Readonly<{
    commitment?: Commitment;
    excludeNonCirculatingAccountsList?: boolean;
}>;

type GetSupplyApiResponseBase = RpcResponse<{
    /** Total supply in lamports */
    total: LamportsUnsafeBeyond2Pow53Minus1;
    /** Circulating supply in lamports */
    circulating: LamportsUnsafeBeyond2Pow53Minus1;
    /** Non-circulating supply in lamports */
    nonCirculating: LamportsUnsafeBeyond2Pow53Minus1;
}>;

type GetSupplyApiResponseWithNonCirculatingAccounts = GetSupplyApiResponseBase &
    Readonly<{
        value: Readonly<{
            /** an array of account addresses of non-circulating accounts */
            nonCirculatingAccounts: [Base58EncodedAddress];
        }>;
    }>;

type GetSupplyApiResponseWithoutNonCirculatingAccounts = GetSupplyApiResponseBase &
    Readonly<{
        value: Readonly<{
            nonCirculatingAccounts: [];
        }>;
    }>;

export interface GetSupplyApi {
    /**
     * Returns information about the current supply.
     */
    getSupply(
        config: GetSupplyConfig &
            Readonly<{
                excludeNonCirculatingAccountsList: true;
            }>
    ): GetSupplyApiResponseWithoutNonCirculatingAccounts;
    getSupply(config?: GetSupplyConfig): GetSupplyApiResponseWithNonCirculatingAccounts;
```

steveluscher (approved):
Accept with nits!

giphy-1

packages/rpc-core/src/rpc-methods/getSupply.ts
mcintyre94: Thanks, that's way cleaner!

PR #1296, opened 2023-05-12
refactor(experimental): add the `getVoteAccounts` RPC method

mcintyre94:
Adds the `getVoteAccounts` RPC method which returns the current + delinquent voting accounts

I've found that in the test validator we start with no vote accounts, and then get 1 current one. To test this in more depth we'll need a way to create active + delinquent ones and modify their stakes.

Note that when called with `votePubkey` the response is narrowed a bit, but I haven't modeled this in the types. This is because there are 3 possible responses:

- `current` with length 1 and `delinquent` with length 0, if the `votePubkey` is active
- `current` with length 0 and `delinquent` with length 1, if the `votePubkey` is not active
- `current` with length 0 and `delinquent` with length 0, if the pubkey is not a vote account, or if it's filtered out by `keepUnstakedDelinquents`

I don't think in practice this is likely to be a useful narrowing, so I've just kept it as `{current: [VoteAccount], delinquent: [VoteAccount]}` for now. 

### DIFF
--- a/packages/rpc-core/src/rpc-methods/__tests__/get-vote-accounts.ts
+++ b/packages/rpc-core/src/rpc-methods/__tests__/get-vote-accounts.ts
@@ -1,0 +1,56 @@
+import { createHttpTransport, createJsonRpc } from '@solana/rpc-transport';
+import type { Rpc } from '@solana/rpc-transport/dist/types/json-rpc-types';
+import fetchMock from 'jest-fetch-mock-fork';
+import { createSolanaRpcApi, SolanaRpcMethods } from '../index';
+import { Commitment } from '../common';
+import { Base58EncodedAddress } from '@solana/keys';
+
+describe('getVoteAccounts', () => {
+    let rpc: Rpc<SolanaRpcMethods>;
+    beforeEach(() => {
+        fetchMock.resetMocks();
+        fetchMock.dontMock();
+        rpc = createJsonRpc<SolanaRpcMethods>({
+            api: createSolanaRpcApi(),
+            transport: createHttpTransport({ url: 'http://127.0.0.1:8899' }),
+        });
+    });
+
+    (['confirmed', 'finalized', 'processed'] as Commitment[]).forEach(commitment => {
+        describe(`when called with \`${commitment}\` commitment`, () => {
+            it('returns current and delinquent vote accounts', async () => {
+                expect.assertions(1);
+                const voteAccountsPromise = rpc.getVoteAccounts().send();
+                await expect(voteAccountsPromise).resolves.toMatchObject({
+                    current: expect.any(Array),
+                    delinquent: expect.any(Array),
+                });
+            });
+
+            // TODO: don't have a way to create vote accounts yet
+            it.todo('returns vote accounts with the correct shape');
+        });
+    });
+
+    describe('when called with a `votePubkey` of a single vote account', () => {
+        it.todo('returns only that vote account');
+    });
+
+    describe('when called with a `votePubkey` of a vote account that does not exist', () => {
+        it('returns empty `current` and `delinquent` fields', async () => {
+            expect.assertions(1);
+            // randomly generated address, don't re-use in anything that creates a vote account
+            const address =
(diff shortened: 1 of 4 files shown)


packages/rpc-core/src/rpc-methods/__tests__/get-vote-accounts.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    describe('when called with a `votePubkey` of a single vote account', () => {
+        it.todo('returns only that vote account');
+    });
+
+    describe('when called with a `votePubkey` of a vote account that does not exist', () => {
+        it('returns empty `current` and `delinquent` fields', async () => {
+            expect.assertions(1);
+            // randomly generated address, don't re-use in anything that creates a vote account
+            const address = '2eTCZxWZkU5h3Mo162gLRTECzuJhPgC1McB9rCcoqNm2' as Base58EncodedAddress;
```
steveluscher, line 43: Just for fun.

```suggestion
            const address = '2eTCZxWZkU5h3Mo162gLRTECzuJhPgC1McB9rCcoqNm2' as Base58EncodedAddress<'2eTCZxWZkU5h3Mo162gLRTECzuJhPgC1McB9rCcoqNm2'>;
```

packages/rpc-core/src/rpc-methods/getVoteAccounts.ts
```diff
@@ -0,0 +1,47 @@
...
+    /** the stake, in lamports, delegated to this vote account and active in this epoch */
+    activatedStake: U64UnsafeBeyond2Pow53Minus1;
+    /** whether the vote account is staked for this epoch */
+    epochVoteAccount: boolean;
+    /** percentage (0-100) of rewards payout owed to the vote account */
+    commission: number;
+    /** Most recent slot voted on by this vote account */
+    lastVote: U64UnsafeBeyond2Pow53Minus1;
+    /** Latest history of earned credits for up to five epochs */
+    epochCredits: [EpochCredit];
```
steveluscher, line 24:
```suggestion
    epochCredits: readonly EpochCredit[];
```

packages/rpc-core/src/rpc-methods/getVoteAccounts.ts
```diff
@@ -0,0 +1,47 @@
...
+    /** Current root slot for this vote account */
+    rootSlot: U64UnsafeBeyond2Pow53Minus1;
+}>;
+
+type GetVoteAccountsApiResponse = Readonly<{
+    current: [VoteAccount];
+    delinquent: [VoteAccount];
```
steveluscher, line 31:
```suggestion
    current: readonly VoteAccount[];
    delinquent: readonly VoteAccount[];
```

packages/rpc-core/src/rpc-methods/getVoteAccounts.ts
```diff
@@ -0,0 +1,47 @@
...
+    votePubkey?: Base58EncodedAddress;
+    /** Do not filter out delinquent validators with no stake */
+    keepUnstakedDelinquents?: boolean;
+    /** Specify the number of slots behind the tip that a validator must fall to be considered delinquent. **NOTE:** For the sake of consistency between ecosystem products, _it is **not** recommended that this argument be specified._ */
+    delinquentSlotDistance?: U64UnsafeBeyond2Pow53Minus1;
+}>;
+
+export interface GetVoteAccountsApi {
+    /** Returns the account info and associated stake for all the voting accounts in the current bank. */
+    getVoteAccounts(config?: GetVoteAccountsConfig): GetVoteAccountsApiResponse;
```
steveluscher, line 46: Do you think we can get the type of the `votePubkey` to flow through now?

```suggestion
    getVoteAccounts<TVoteAccount>(
      config?: GetVoteAccountsConfig<TVoteAccount>,
    ): GetVoteAccountsApiResponse<TVoteAccount>;
```

Then add `<TVoteAccount>` everywhere and also `votePubkey?: Base58EncodedAddress<TVoteAccount>`?

steveluscher (approved):
Accept with (important) nits!

thumbs-up-double-thumbs-up

PR #1297, opened 2023-05-15
Refactor (experimental): add the `getRecentPerformanceSamples`  RPC method

mcintyre94:
Adds the `getRecentPerformanceSamples` RPC method which returns the latest performance samples

Currently the performance sample rate is 60s, so we don't have any data for the first 60s. This makes it impossible to meaningfully unit test this for now. We'll need to find a way to create performance samples without waiting 60s. 

- [x] I have written a test in [`packages/rpc-core/src/rpc-methods/__tests__`](https://github.com/solana-labs/solana-web3.js/tree/master/packages/rpc-core/src/rpc-methods/__tests__) that exercises as much of my RPC method's functionality as is practical (Todos for now)

### DIFF
--- a/packages/rpc-core/src/rpc-methods/__tests__/get-recent-performance-samples-test.ts
+++ b/packages/rpc-core/src/rpc-methods/__tests__/get-recent-performance-samples-test.ts
@@ -1,0 +1,10 @@
+describe('getRecentPerformanceSamples', () => {
+    // TODO: need a way to create performance samples without waiting 60s
+    describe('when called without a limit', () => {
+        it.todo('returns the latest performance samples');
+    });
+
+    describe('when called with a limit', () => {
+        it.todo('only returns the requested number of performance samples');
+    });
+});
--- a/packages/rpc-core/src/rpc-methods/getRecentPerformanceSamples.ts
+++ b/packages/rpc-core/src/rpc-methods/getRecentPerformanceSamples.ts
@@ -1,0 +1,26 @@
+import { Slot, U64UnsafeBeyond2Pow53Minus1 } from './common';
+
+type PerformanceSample = Readonly<{
+    /** Slot in which sample was taken at */
+    slot: Slot;
+    /** Number of transactions in sample */
+    numTransactions: U64UnsafeBeyond2Pow53Minus1;
+    /** Number of slots in sample */
+    numSlots: U64UnsafeBeyond2Pow53Minus1;
+    /** Number of seconds in a sample window */
+    samplePeriodSecs: number;
+    /** Number of non-vote transactions in sample. */
+    numNonVoteTransaction: U64UnsafeBeyond2Pow53Minus1;
+}>;
+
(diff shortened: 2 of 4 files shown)


steveluscher:
> This makes it impossible to meaningfully unit test this for now. We'll need to find a way to create performance samples without waiting 60s.

Oof. Is this something we can ask for, or just do? Make the test validator boot up with _at least_ all zeroes?

packages/rpc-core/src/rpc-methods/getRecentPerformanceSamples.ts
```diff
@@ -0,0 +1,26 @@
...
+
+type GetRecentPerformanceSamplesApiResponse = [PerformanceSample];
+
+export interface GetRecentPerformanceSamplesApi {
+    /**
+     * Returns a list of recent performance samples, in reverse slot order. Performance samples are taken every 60 seconds and include the number of transactions and slots that occur in a given time window.
+     */
+    getRecentPerformanceSamples(
+        /** number of samples to return (maximum 720) */
+        limit?: number
```
steveluscher, line 24: Nice.

packages/rpc-core/src/rpc-methods/getRecentPerformanceSamples.ts
```diff
@@ -0,0 +1,26 @@
...
+    numTransactions: U64UnsafeBeyond2Pow53Minus1;
+    /** Number of slots in sample */
+    numSlots: U64UnsafeBeyond2Pow53Minus1;
+    /** Number of seconds in a sample window */
+    samplePeriodSecs: number;
+    /** Number of non-vote transactions in sample. */
+    numNonVoteTransaction: U64UnsafeBeyond2Pow53Minus1;
+}>;
+
+type GetRecentPerformanceSamplesApiResponse = [PerformanceSample];
```
steveluscher, line 16: I don't think this is right. This means ‘a one-element tuple that contains a `PerformanceSample`’.
```suggestion
type GetRecentPerformanceSamplesApiResponse = readonly PerformanceSample[];
```

packages/rpc-core/src/rpc-methods/__tests__/get-recent-performance-samples-test.ts
```diff
@@ -0,0 +1,25 @@
...
+describe('getRecentPerformanceSamples', () => {
+    // let rpc: Rpc<SolanaRpcMethods>;
+    beforeEach(() => {
+        fetchMock.resetMocks();
+        fetchMock.dontMock();
+        // rpc = createJsonRpc<SolanaRpcMethods>({
+        //   api: createSolanaRpcApi(),
+        //   transport: createHttpTransport({ url: 'http://127.0.0.1:8899' }),
+        // });
+    });
```
steveluscher, line 15: If we can't do it, just delete all of this. By the time we can, the test infra might have changed.

mcintyre94:
> Is this something we can ask for, or just do?

Opened https://github.com/solana-labs/solana/issues/31695 for this

PR #1300, opened 2023-05-16
refactor(experimental): add the `getBlockTime` RPC method

mcintyre94:
Adds the `getBlockTime` RPC method

Note that I think [the current docs](https://docs.solana.com/api/http#getblocktime) are slightly incorrect, this method doesn't return null. It either returns the block time or an error. Fix in https://github.com/solana-labs/solana/pull/31653

Note that tests here are limited. I think there are 3 responses:

- Returns the block time for an available block. For this to be reliable in tests we need to either get an available block (eg using `getFirstAvailableBlock` when implemented), or be able to control the ledger so that a certain block is guaranteed to be available
- Returns an error when a block has been cleaned up. For this to be performant in tests we'll need a way to control the ledger so that a block is cleaned up, without waiting however long that would take to happen on its own
- Returns an error when a block is not available up. I've included a test using the usual maxint for this. 

---

### DIFF
--- a/packages/rpc-core/src/__tests__/unix-timestamp-test.ts
+++ b/packages/rpc-core/src/__tests__/unix-timestamp-test.ts
@@ -1,0 +1,35 @@
+import { assertIsUnixTimestamp } from '../unix-timestamp';
+
+describe('assertIsUnixTimestamp()', () => {
+    it('throws when supplied a too large number', () => {
+        expect(() => {
+            assertIsUnixTimestamp(Number.MAX_SAFE_INTEGER);
+        }).toThrow();
+        expect(() => {
+            assertIsUnixTimestamp(8.64e15 + 1);
+        }).toThrow();
+    });
+    it('throws when supplied a too small number', () => {
+        expect(() => {
+            assertIsUnixTimestamp(Number.MIN_SAFE_INTEGER);
+        }).toThrow();
+        expect(() => {
+            assertIsUnixTimestamp(-8.64e15 - 1);
+        }).toThrow();
+    });
+    it('does not throw when supplied a zero timestamp', () => {
+        expect(() => {
+            assertIsUnixTimestamp(0);
+        }).not.toThrow();
+    });
+    it('does not throw when supplied a valid non-zero timestamp', () => {
+        expect(() => {
+            assertIsUnixTimestamp(1_000_000_000);
+        }).not.toThrow();
+    });
+    it('does not throw when supplied the max valid timestamp', () => {
+        expect(() => {
+            assertIsUnixTimestamp(8.64e15);
+        }).not.toThrow();
+    });
+});
--- a/packages/rpc-core/src/rpc-methods/getBlockTime.ts
+++ b/packages/rpc-core/src/rpc-methods/getBlockTime.ts
@@ -1,4 +1,5 @@
-import { Slot, UnixTimestamp } from './common';
+import { UnixTimestamp } from '../unix-timestamp';
+import { Slot } from './common';
 
 /** Estimated production time, as Unix timestamp (seconds since the Unix epoch) */
 type GetBlockTimeApiResponse = UnixTimestamp;
--- a/packages/rpc-core/src/rpc-methods/getBlockTime.ts
+++ b/packages/rpc-core/src/rpc-methods/getBlockTime.ts
@@ -1,0 +1,14 @@
+import { Slot, UnixTimestamp } from './common';
+
+/** Estimated production time, as Unix timestamp (seconds since the Unix epoch) */
+type GetBlockTimeApiResponse = UnixTimestamp;
+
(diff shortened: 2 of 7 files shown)


packages/rpc-core/src/rpc-methods/getBlockTime.ts
```diff
@@ -0,0 +1,14 @@
...
+/** Estimated production time, as Unix timestamp (seconds since the Unix epoch) */
+type GetBlockTimeApiResponse = U64UnsafeBeyond2Pow53Minus1;
+
+export interface GetBlockTimeApi {
+    /**
+     * Returns the estimated production time of a block.
+     */
+    getBlockTime(
+        /** block number, identified by Slot */
+        blockNumber: U64UnsafeBeyond2Pow53Minus1
```
steveluscher, line 12: Shall we just use the `Slot` type?
```suggestion
        blockNumber: Slot,
```

packages/rpc-core/src/rpc-methods/getBlockTime.ts
```diff
@@ -0,0 +1,14 @@
+import { U64UnsafeBeyond2Pow53Minus1 } from './common';
+
+/** Estimated production time, as Unix timestamp (seconds since the Unix epoch) */
+type GetBlockTimeApiResponse = U64UnsafeBeyond2Pow53Minus1;
```
steveluscher, line 4: I'm _sort_ of inclined to make a new opaque type that represents timestamps. What do you think?

Something like this:

https://github.com/solana-labs/solana-web3.js/blob/d16acbdd1d451120ddac983e076545ce35934785/packages/rpc-core/src/rpc-methods/common.ts#L12

…but like this:

```ts
type UnixTimestamp = number & {readonly __unixTimestamp: unique symbol};
```

steveluscher (approved):
giphy

packages/rpc-core/src/rpc-methods/getBlockTime.ts
mcintyre94: I like this idea! The only downside is that we patch a `u64` response from the RPC to number, which is maybe a bit confusing despite being safe (until 2255). I think it makes sense though, updated to use this

steveluscher: It's actually more complicated that, even. JavaScript date maxes out at:

```ts
new Date(8.64e15)
> Fri Sep 12 275760 17:00:00 GMT-0700 (Pacific Daylight Time)
```

Which is lower even than `Number.MAX_SAFE_INTEGER`.

Anyway, see you in the year 275760.

steveluscher (approved):
Ship it!

You might consider adding an `assertIsUnixTimestamp` method (copy and paste `assertIsBase58EncodedAddress`) that asserts the time is positive and less than or equal to `8.64e15`, but whatever!

packages/rpc-core/src/__tests__/unix-timestamp-test.ts
```diff
@@ -0,0 +1,31 @@
+import { assertIsUnixTimestamp } from '../unix-timestamp';
+
+describe('number', () => {
```
steveluscher, line 3: I don't think this describe block is necessary. You're just describing what the assert function does.

packages/rpc-core/src/__tests__/unix-timestamp-test.ts
```diff
@@ -0,0 +1,31 @@
...
+            }).not.toThrow();
+        });
+        it('does not throw when supplied a valid non-zero timestamp', () => {
+            expect(() => {
+                assertIsUnixTimestamp(1_000_000_000);
+            }).not.toThrow();
+        });
+        it('does not throw when supplied the max valid timestamp', () => {
+            expect(() => {
+                assertIsUnixTimestamp(8.64e15);
```
steveluscher, line 27: I'd write one extra test that does `8.64e15 + 1`.

mcintyre94:
Cool updated to add the `8.64e15 + 1` test and drop the negative check for now. Thanks!

packages/rpc-core/src/unix-timestamp.ts
```diff
@@ -0,0 +1,14 @@
+export type UnixTimestamp = number & { readonly __unixTimestamp: unique symbol };
+
+export function assertIsUnixTimestamp(putativeTimestamp: number): asserts putativeTimestamp is UnixTimestamp {
+    // see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#the_epoch_timestamps_and_invalid_date
+    try {
+        if (putativeTimestamp > 8.64e15) {
```
steveluscher, line 6: Tiny nit: both directions. And add a test.

`if (putativeTimestamp > 8.64e15 || putativeTimestamp < -8.64e15) {`

PR #1301, opened 2023-05-16
refactor(experimental): allow literal `Base58EncodedAddress` types

steveluscher:
refactor(experimental): allow literal `Base58EncodedAddress` types
## Summary

If you want to not only assert that something is a base58 address, but a _particular_ base58 address you can make use of this new optional type parameter.

## Test Plan

```ts
const memoProgramV2Address = 'MemoSq4gqABAXKb96qnH8TysNcWxMyWCqXgDLGmfcHr';
assertIsBase58EncodedAddress(memoProgramV2Address);
memoProgramV2Address as Base58EncodedAddress; // OK
memoProgramV2Address as Base58EncodedAddress<'MemoSq4gqABAXKb96qnH8TysNcWxMyWCqXgDLGmfcHr'>; // OK
memoProgramV2Address as Base58EncodedAddress<'Memo1UhkJRfHyvLMcVucJwxXeuD728EqVDDwQDxFMNo'>; // ERROR
```

### DIFF
--- a/packages/keys/src/base58.ts
+++ b/packages/keys/src/base58.ts
@@ -1,10 +1,12 @@
 import bs58 from 'bs58';
 
-export type Base58EncodedAddress = string & { readonly __base58EncodedAddress: unique symbol };
+export type Base58EncodedAddress<TAddress extends string = string> = TAddress & {
+    readonly __base58EncodedAddress: unique symbol;
+};
 
 export function assertIsBase58EncodedAddress(
     putativeBase58EncodedAddress: string
-): asserts putativeBase58EncodedAddress is Base58EncodedAddress {
+): asserts putativeBase58EncodedAddress is Base58EncodedAddress<typeof putativeBase58EncodedAddress> {
     try {
         // Fast-path; see if the input string is of an acceptable length.
         if (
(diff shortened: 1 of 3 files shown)


steveluscher:
@mcintyre94 why can't I add you as a reviewer?

packages/keys/src/base58.ts
```diff
@@ -1,10 +1,12 @@
 import bs58 from 'bs58';
 
-export type Base58EncodedAddress = string & { readonly __base58EncodedAddress: unique symbol };
+export type Base58EncodedAddress<TAddress extends string = string> = TAddress & {
+    readonly __base58EncodedAddress: unique symbol;
+};
 
 export function assertIsBase58EncodedAddress(
     putativeBase58EncodedAddress: string
```
joncinque, line 8: Any reason to not also make this generic over `<TAddress extends string = string>`? Or does that encompass potentially too many types?

joncinque (commented):
Looks good! Just the one question

packages/keys/src/base58.ts
steveluscher: It actually already is; it's sort of magic.

image

anna-kendrick-mind-blown

joncinque: dang, that's really cool

PR #1306, opened 2023-05-23
fix: update the size of vote accounts now that the network has moved on past 1.14.11

steveluscher:
fix: update the size of vote accounts now that the network has moved on past 1.14.11
## Summary

The new size of the vote account is `3762`. This update will make `VoteProgram.initializeAccount()` work with validators newer than 1.14.11 (which is now the majority of the network)

## Test Plan

```ts
pnpm test:live-with-test-validator
```

PR #1307, opened 2023-05-23
fix: update the size of vote accounts now that the network has moved on past 1.14.11

steveluscher:
fix: update the size of vote accounts now that the network has moved on past 1.14.11
## Summary

The new size of the vote account is `3762`. This update will make `VoteProgram.initializeAccount()` work with validators newer than 1.14.11 (which is now the majority of the network)

## Test Plan

```ts
```


steveluscher:
Duplicate of #1306.

steveluscher:
Duplicate of #1306.

PR #1326, opened 2023-05-31
fix: Fix getPA type when config is undefined

Arrowana:
when using `connection.getProgramAccounts(<public-key>)`, without a config argument, the return type is `RpcResponseAndContext<GetProgramAccountsResponse>` while configOrCommitment is false so it does not run the function returning the context. This result into an error at runtime for anyone trying to access `value`.


steveluscher (approved):
giphy

Thanks for the catch!

Broken by #1311.

PR #1340, opened 2023-06-06
refactor(experimental): add the getEpochSchedule RPC method

mcintyre94:
Nice straightforward RPC method for once! Used in the `ClusterProvider` in explorer, which would be good to migrate to new web3js!


steveluscher (approved):
easy-button-that-was-easy

PR #1341, opened 2023-06-06
refactor(experimental): add the getTokenLargestAccounts RPC method

mcintyre94:
This PR adds the `getTokenLargestAccounts` RPC method. This is used by the shared `MintsProvider` in explorer, which we should migrate to new web3js

We can't test this on the local validator yet, since we have no way to create a token mint + accounts. But I have tested on devnet using this test:

```ts
it('manual test with devnet', async () => {
    const tokenMint = '4HZCNvobxtDA3uezTGmDAEqVLp7oo73UrnbxNeUMszd4' as Base58EncodedAddress;
    const tokenAccounts = await rpc.getTokenLargestAccounts(tokenMint).send().then(r => r.value);

    expect(tokenAccounts.length).toEqual(20);
    expect(tokenAccounts[0]).toMatchObject({
        address: expect.any(String),
        amount: expect.any(String),
        decimals: expect.any(Number),
        uiAmount: expect.any(Number),
        uiAmountString: expect.any(String)
    });
});
```

### DIFF
--- a/packages/rpc-core/src/rpc-methods/getTokenLargestAccounts.ts
+++ b/packages/rpc-core/src/rpc-methods/getTokenLargestAccounts.ts
@@ -1,22 +1,25 @@
 import { Base58EncodedAddress } from '@solana/keys';
 
+import { StringifiedBigInt } from '../stringified-bigint';
 import { Commitment, RpcResponse } from './common';
 
-type GetTokenLargestAccountsApiResponse = RpcResponse<{
-    /** the address of the token account */
-    address: Base58EncodedAddress;
-    /** the raw token account balance without decimals, a string representation of u64 */
-    amount: string;
-    /** number of base 10 digits to the right of the decimal place */
-    decimals: number;
-    /**
-     * the token account balance, using mint-prescribed decimals
-     * @deprecated
-     */
-    uiAmount: number | null;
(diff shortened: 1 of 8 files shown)


packages/rpc-core/src/rpc-methods/getTokenLargestAccounts.ts
```diff
@@ -0,0 +1,31 @@
+import { Base58EncodedAddress } from '@solana/keys';
+
+import { Commitment, RpcResponse } from './common';
+
+type GetTokenLargestAccountsApiResponse = RpcResponse<{
+    /** the address of the token account */
+    address: Base58EncodedAddress;
+    /** the raw token account balance without decimals, a string representation of u64 */
+    amount: string;
```
steveluscher, line 9: Sigh. I wonder if it's worth creating an opaque type for this now. Basically, this is ‘bigint as a string.’ It would be worth knowing when you have one of these types so that you know for sure it's parseable by `BigInt(...)`.

steveluscher (approved):
OMG, I'm sorry I missed this PR.
giphy

mcintyre94:
Used `pnpm update @solana/eslint-config-solana @solana/prettier-config-solana eslint-config-turbo eslint prettier` to update the lint/prettier dependencies

PR #1342, opened 2023-06-09
refactor(experimental): add a comparator you can use to sort base58 addresses

steveluscher:
refactor(experimental): add a comparator you can use to sort base58 addresses
## Summary

We sort addresses peculiarly. Not based on the value of the underlying key. Not using the default `Array.sort()`. Rather, using `localeCompare` with a specific set of options. This PR encodes this into the new experimental library, through the use of a getter that returns you the compare functino of an `Intl.Collator`.

## Test Plan

```
pnpm test:unit:browser
pnpm test:unit:node
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/solana-labs/solana-web3.js/pull/1342).
* #1302
* #1345
* #1344
* #1343
* __->__ #1342

### DIFF
--- a/packages/keys/src/__tests__/base58-test.ts
+++ b/packages/keys/src/__tests__/base58-test.ts
@@ -1,6 +1,6 @@
 import bs58 from 'bs58';
 
-import { assertIsBase58EncodedAddress } from '../base58';
+import { assertIsBase58EncodedAddress, getBase58EncodedAddressComparator } from '../base58';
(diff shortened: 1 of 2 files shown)


steveluscher:
Shipped as 5013dc68e6d13b733cc6eec5b07afd3958c3d458.

PR #1343, opened 2023-06-09
refactor(experimental): shim in `TextEncoder` into tests

steveluscher:
refactor(experimental): shim in `TextEncoder` into tests
## Summary

This is apparently not available with JSDOM.

## Test Plan

`TextEncoder` is now defined in tests. Used in later PRs.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/solana-labs/solana-web3.js/pull/1343).
* #1302
* #1345
* #1344
* __->__ #1343
* #1342

PR #1344, opened 2023-06-09
refactor(experimental): set `__DEV__` earlier in the test environment

steveluscher:
refactor(experimental): set `__DEV__` earlier in the test environment
## Summary

This code used to define `__DEV__` before the first test ran, but if `__DEV__` was needed _before_ that (eg. in a module factory of something under test) it would not be defined yet. This PR defines it right off the bat, in addition to resetting it before each test.

## Test Plan

Add `__DEV__` into a module factory of something under test. No more runtime fatal.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/solana-labs/solana-web3.js/pull/1344).
* #1302
* #1345
* __->__ #1344
* #1343
* #1342

### DIFF
--- a/packages/keys/src/__tests__/base58-test.ts
+++ b/packages/keys/src/__tests__/base58-test.ts
@@ -1,6 +1,6 @@
 import bs58 from 'bs58';
 
-import { assertIsBase58EncodedAddress } from '../base58';
+import { assertIsBase58EncodedAddress, getBase58EncodedAddressComparator } from '../base58';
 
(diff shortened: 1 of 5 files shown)


steveluscher:
Shipped as 22e0010576a4645f805ed329005f0da413da71bd.

steveluscher:
Shipped as 461cafe1391ccf378918f9cbfba7c473999e7211.

PR #1349, opened 2023-06-09
refactor(experimental): a helper you can use to set the fee payer on a transaction

steveluscher:
refactor(experimental): a helper you can use to set the fee payer on a transaction
## Summary
This helper both sets the fee payer on a transaction and clears the signatures if the fee payer has changed. If the fee payer has _not_ changed it returns the identical object without mutation.

## Test Plan

```
cd packages/transactions
pnpm test:unit:browser
pnpm test:unit:node
```


packages/transactions/src/fee-payer.ts
```diff
@@ -0,0 +1,44 @@
...
+    if ('signatures' in transaction) {
+        // The implication of the fee payer changing is that any existing signatures are invalid.
+        const {
+            signatures: _, // eslint-disable-line @typescript-eslint/no-unused-vars
+            ...unsignedTransaction
+        } = transaction;
+        out = {
+            ...unsignedTransaction,
+            feePayer,
+        };
```
mcintyre94, line 33: Would it be clearer and equivalent to just do:

```ts
out = {
    ...transaction,
    feePayer,
    signatures: undefined
}
```

Or does that mess up the typing?

mcintyre94 (approved):
Nice! It's great to get this sort of test coverage too, pure data wrangling FTW!

PR #1350, opened 2023-06-09
refactor(experimental): a helper to add instructions to a transaction

steveluscher:
refactor(experimental): a helper to add instructions to a transaction
## Summary

This helper takes care of appending instructions to the list, and unsetting the signatures each time a new instruction is added.
## Test Plan

```
cd packages/transactions
pnpm test:unit:browser
pnpm test:unit:node
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/solana-labs/solana-web3.js/pull/1350).
* #1351
* __->__ #1350

### DIFF
--- a/packages/transactions/src/__tests__/instructions-test.ts
+++ b/packages/transactions/src/__tests__/instructions-test.ts
@@ -1,0 +1,79 @@
+import 'test-matchers/toBeFrozenObject';
+
+import { IInstruction } from '@solana/instructions';
+import { Base58EncodedAddress } from '@solana/keys';
+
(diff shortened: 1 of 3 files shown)


mcintyre94 (approved):
Looks good! Do we want to add a `prepend` too? Metaplex's builder API has it and IIRC some programs of theirs are sensitive to instruction order. Not sure how widespread that is more generally. I think anchor has both append and prepend too.

steveluscher:
Oh, sure. I can do that. How long before someone asks for an `insert(ix, index, tx)` too?

PR #1351, opened 2023-06-09
refactor(experimental): a serializer for the transaction version header

steveluscher:
refactor(experimental): a serializer for the transaction version header
## Summary

This is the very first transaction serializer. It demonstrates how we'll use the serializer from Metaplex's Umi library to define custom serializers for transactions.

## Test Plan

```
cd packages/transactions
pnpm test:unit:browser
pnpm test:unit:node
```

### DIFF
--- a/packages/transactions/src/serializers/__tests__/transaction-version-test.ts
+++ b/packages/transactions/src/serializers/__tests__/transaction-version-test.ts
@@ -1,0 +1,31 @@
+import { TransactionVersion } from '../../types';
+import { transactionVersionHeader } from '../transaction-version';
+
+const VERSION_FLAG_MASK = 0x80;
+const VERSION_TEST_CASES = // Versions 0–127
+    [...Array(128).keys()].map(version => [version | VERSION_FLAG_MASK, version as TransactionVersion] as const);
+describe('Transaction version serializer', () => {
+    it('serializes no data when the version is `legacy`', () => {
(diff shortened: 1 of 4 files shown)


packages/transactions/src/serializers/__tests__/transaction-version-test.ts
```diff
@@ -0,0 +1,22 @@
...
+        [...Array(128).keys()].map(version => [version | VERSION_FLAG_MASK, version as TransactionVersion])
+    )('serializes to `%s` when the version is `%s`', (expected, version) => {
+        expect(transactionVersionHeader.serialize(version)).toEqual(new Uint8Array([expected]));
+    });
+    it.each([-1 as TransactionVersion, 128 as TransactionVersion])(
+        'throws when passed the out-of-range version `%s`',
+        version => {
+            expect(() => transactionVersionHeader.serialize(version)).toThrow();
+        }
+    );
```
mcintyre94, line 21: Should we add tests for deserialize too? To make sure value -> serialize -> deserialize -> value works correctly

mcintyre94 (approved):
This looks good, but how come we need to do the bit wrangling?

- in the case of legacy, why isn't `firstByte === undefined` sufficient, since we're serializing to an empty array?

- in the case of versioned, why do we need to mask the value when we know it's in the range [0,127]? It looks like we serialize `1` to `[129]`, then deserialize that back to 1. Why can't we just serialize to `[1]`?

PR #1358, opened 2023-06-26
refactor(experimental): convert signer-writability flags on accounts into `role` enum

steveluscher:
refactor(experimental): convert signer-writability flags on accounts into `role` enum
## Summary

Instead of tweaking `isSigner` and `isWritable` flags on account metadata, let's create one enum to describe all four combinations. This enables us to finally treat account metadata as a discriminated union in TypeScript. That is to say that checking an account's `role` will now refine its type down to the type containing that role.

* Makes code less verbose, particularly tests
* Makes it less tempting to mutate an object in-place by twiddling the boolean of `isSigner` or `isWritable`

## Test Plan

See next PRs.

### DIFF
--- a/packages/instructions/src/roles.ts
+++ b/packages/instructions/src/roles.ts
@@ -1,0 +1,64 @@
+/**
+ * Quick primer on bitwise operations: https://stackoverflow.com/a/1436448/802047
+ */
+
+export enum AccountRole {
+    // Bitflag guide: is signer ⌄⌄ is writable
+    WRITABLE_SIGNER = /* 3 */ 0b11, // prettier-ignore
+    READONLY_SIGNER = /* 2 */ 0b10, // prettier-ignore
+    WRITABLE =        /* 1 */ 0b01, // prettier-ignore
+    READONLY =        /* 0 */ 0b00, // prettier-ignore
+}
+
+const IS_SIGNER_BITMASK = 0b10;
+const IS_WRITABLE_BITMASK = 0b01;
+
+export function downgradeRoleToNonSigner(role: AccountRole.READONLY_SIGNER): AccountRole.READONLY;
+export function downgradeRoleToNonSigner(role: AccountRole.WRITABLE_SIGNER): AccountRole.WRITABLE;
+export function downgradeRoleToNonSigner(role: AccountRole): AccountRole;
+export function downgradeRoleToNonSigner(role: AccountRole): AccountRole {
+    return role & ~IS_SIGNER_BITMASK;
+}
+
+export function downgradeRoleToReadonly(role: AccountRole.WRITABLE): AccountRole.READONLY;
+export function downgradeRoleToReadonly(role: AccountRole.WRITABLE_SIGNER): AccountRole.READONLY_SIGNER;
+export function downgradeRoleToReadonly(role: AccountRole): AccountRole;
+export function downgradeRoleToReadonly(role: AccountRole): AccountRole {
+    return role & ~IS_WRITABLE_BITMASK;
+}
+
+export function isSignerRole(role: AccountRole): role is AccountRole.READONLY_SIGNER | AccountRole.WRITABLE_SIGNER {
+    return role >= AccountRole.READONLY_SIGNER;
+}
+
+export function isWritableRole(role: AccountRole): role is AccountRole.WRITABLE | AccountRole.WRITABLE_SIGNER {
+    return (role & IS_WRITABLE_BITMASK) !== 0;
+}
+
+export function mergeRoles(roleA: AccountRole.WRITABLE, roleB: AccountRole.READONLY_SIGNER): AccountRole.WRITABLE_SIGNER; // prettier-ignore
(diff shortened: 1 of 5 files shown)


mcintyre94 (approved):
LGTM 🚀 JS/TS really need proper pattern matching!

packages/instructions/src/roles.ts
```diff
@@ -0,0 +1,59 @@
...
+export function mergeRoles(roleA: AccountRole.READONLY_SIGNER, roleB: AccountRole.WRITABLE): AccountRole.WRITABLE_SIGNER; // prettier-ignore
+export function mergeRoles(roleA: AccountRole, roleB: AccountRole.WRITABLE_SIGNER): AccountRole.WRITABLE_SIGNER; // prettier-ignore
+export function mergeRoles(roleA: AccountRole.WRITABLE_SIGNER, roleB: AccountRole): AccountRole.WRITABLE_SIGNER; // prettier-ignore
+export function mergeRoles(roleA: AccountRole, roleB: AccountRole.READONLY_SIGNER): AccountRole.READONLY_SIGNER; // prettier-ignore
+export function mergeRoles(roleA: AccountRole.READONLY_SIGNER, roleB: AccountRole): AccountRole.READONLY_SIGNER; // prettier-ignore
+export function mergeRoles(roleA: AccountRole, roleB: AccountRole.WRITABLE): AccountRole.WRITABLE; // prettier-ignore
+export function mergeRoles(roleA: AccountRole.WRITABLE, roleB: AccountRole): AccountRole.WRITABLE; // prettier-ignore
+export function mergeRoles(roleA: AccountRole.READONLY, roleB: AccountRole.READONLY): AccountRole.READONLY; // prettier-ignore
+export function mergeRoles(roleA: AccountRole, roleB: AccountRole): AccountRole; // prettier-ignore
```
joncinque, line 39: What does this line do? It seems like it should be obviated by the next line, but I'm probably missing a fun typescript thing

packages/instructions/src/roles.ts
```diff
@@ -0,0 +1,59 @@
+export enum AccountRole {
+    WRITABLE_SIGNER = 0b1, // 1
+    READONLY_SIGNER = 0b10, // 2
+    WRITABLE = 0b100, // 4
+    READONLY = 0b1000, // 8
+}
```
joncinque, line 6: I'm probably missing something, but is there a reason for defining each role as a separate bit rather than as just a signer bit and a writable bit? ie:

```
export enum AccountRole {
    READONLY = 0b00, // 0
    WRITABLE = 0b01, // 1
    READONLY_SIGNER = 0b10, // 2
    WRITABLE_SIGNER = 0b11, // 3
}
```

Then you don't need to do any of the `<` and `>` checks or bit shifts and just use `&` and `|` everywhere

joncinque (approved):
Nothing stopping this from going in, just a nitty implementation-level question

packages/instructions/src/roles.ts
steveluscher: Yeah, super annoying. The final method signature _doesn't actually participate_ when you have overloads like this. It's the ‘sum total’ of all the possbilities.

Here I wanted to express ‘if you don't know what the `AccountRole` is then your return is ‘some `AccountRole`’,’ so I needed to actually say that.

https://www.typescriptlang.org/play?#code/KYOwrgtgBAggxnA9mEAXASogNsKBvAKCmKnQFEYARAeQDkAZATQBoiSB1dASQBUYAhemQIBfAgWAAPAA6IATqigAzFHFQBLRCCgRgcgObBMOAM4AKOdmAwAXLATI0x4MyiWc-O-CQoMVgHScvAJCAJReDr7Ogdx8gmQA3BIy8ooqIGqa2roGRlbm7tYRPk4BQXFCroWe9iV+OOG1jvXAMcHxSVKyCsqqGlo6eobOBVa2TVEB5FR0TFVWNd7N0dM0DIyNS5M4-quzjJ0pPemZAznD+RZjxctW8x4328CbkaU4CVAA9J9QPHIAnlAkBBdGh1CB9FBUAALdQmKDIVD+ZLdNJ9LKDXIjK44cZbN4uNwLR4El51Zz4NhyYCoMBybT4lq7ChrJhJMQEJAgEyKGBQAC8mIupjMjJWLP2rjFUwl61CnK0PKg-AFQryIulOz26ylryZ5RCZHlXKVAGFVed1cBzJrWgb4rryTKZnKFdzFJQLUMrTa9dF7ZUJgS2hUjW6lSgANYgRAAdxAtr5grMsfBABM41AAIbwrMgf6hfyGVC2syhbPw21JE2KKMx+O2lXJ1MgDOxivZ-OF4ul8s5oMtauKxRkL1Yy51uMJv1jVyThszjyhIA

packages/instructions/src/roles.ts
steveluscher: That seems less dumb. Let me refactor that and if I don't discover a reason why I did it this way and not the way you've described, I'll ship it.

Hooray for test coverage that lets me do this with reckless abandon.

packages/instructions/src/roles.ts
joncinque: Gotcha, I had a feeling it was a weird unspoken requirement.

PR #1362, opened 2023-06-26
refactor(experimental): a function to compile a transaction's header into a message

steveluscher:
refactor(experimental): a function to compile a transaction's header into a message
## Summary

Now that we can compile a list of ordered accounts, we can iterate over that list to produce the transaction's header. This is made up of:

* The number of signers
* The number of readonly signers
* The number of readonly non-signers

## Test Plan

```
cd packages/transactions
pnpm test:unit:browser
pnpm test:unit:node
```


joncinque (commented):
Post-merge review: looks good!

PR #1363, opened 2023-06-26
refactor(experimental): a function to compile a transaction's instructions into a message

steveluscher:
refactor(experimental): a function to compile a transaction's instructions into a message
## Summary

Over the wire, transaction instructions don't carry the addresses that they refer to, but rather the _indices_ of those addresses in a sidecar address table. This function takes in the ordered addresses and the instructions, and returns the instructions with the addresses replaced with indices.

## Test Plan

```
cd packages/transactions/
pnpm test:unit:browser
pnpm test:unit:node
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/solana-labs/solana-web3.js/pull/1363).
* #1372
* #1371
* #1370
* #1369
* #1368
* #1365
* #1364
* __->__ #1363

### DIFF
--- a/packages/transactions/src/compile-instructions.ts
+++ b/packages/transactions/src/compile-instructions.ts
@@ -1,0 +1,32 @@
+import { IInstruction } from '@solana/instructions';
+import { Base58EncodedAddress } from '@solana/keys';
+
+import { OrderedAccounts } from './accounts';
+
+type CompiledInstruction = Readonly<{
+    addressIndices?: number[];
+    data?: Uint8Array;
+    programAddressIndex: number;
+}>;
+
+function getAccountIndex(orderedAccounts: OrderedAccounts) {
+    const out: Record<Base58EncodedAddress, number> = {};
+    for (const [index, account] of orderedAccounts.entries()) {
+        out[account.address] = index;
+    }
+    return out;
+}
+
+export function getCompiledInstructions(
+    instructions: readonly IInstruction[],
+    orderedAccounts: OrderedAccounts
+): CompiledInstruction[] {
+    const accountIndex = getAccountIndex(orderedAccounts);
+    return instructions.map(({ accounts, data, programAddress }) => {
+        return {
+            programAddressIndex: accountIndex[programAddress],
+            ...(accounts ? { accountIndices: accounts.map(({ address }) => accountIndex[address]) } : null),
+            ...(data ? { data } : null),
+        };
+    });
+}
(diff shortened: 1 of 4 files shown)


packages/transactions/src/compile-instructions.ts
```diff
@@ -0,0 +1,32 @@
+import { IInstruction } from '@solana/instructions';
+import { Base58EncodedAddress } from '@solana/keys';
+
+import { OrderedAccounts } from './accounts';
+
+type CompiledInstruction = Readonly<{
+    addressIndices?: number[];
+    data?: Uint8Array;
```
joncinque, line 8: Why are these nullable exactly? The `CompiledInstruction` Rust type has these as `Vec`s, which makes me think these would be their non-nullable types. Is there some neater mapping later on that you can do if they're nullable?

joncinque (approved):
Just a question, no reason to stop this from going forward though, looks great!

packages/transactions/src/compile-instructions.ts
steveluscher: Yeah, you're not wrong. In general, I try to not make garbage (in the heap sense) when it's not used (ie. don't create an empty array, just supply nothing).

The thing I do later is to add defaulting in the serializer, here: https://reviewstack.dev/solana-labs/solana-web3.js/pull/1372. That is to say if `data` or `addressIndices` are missing, they get serialized to/from zero-length compact arrays.

The idea is that it's better to create the empty garbage at the very last step, but avoid it up until then.

steveluscher: There is _still_ the possibility that someone could go and supply an _empty_ array instead of omitting the property. I wish there was a good way of typing ‘a non-empty array’ in TypeScript.

```ts
type Foo = {
  addressIndices: NonEmptyArray<number>;
}
```

joncinque: Gotcha, that makes sense, thanks for clarifying

steveluscher: 😢  https://www.typescriptlang.org/play?#code/C4TwDgpgBAYg9nKBeKBvAUFKBDAJrgJwgGdiBJAO1wEsBjEgfgC4oBRAD1oBsBXXCADwUeAWwBGEAgG0AugBo0XCBQDmwABYsADAF8AfOh3p0tOBWLAoAMxbxEKVHkIlyVOiRaydQA

joncinque: Stack overflow gave this https://www.typescriptlang.org/play?#code/C4TwDgpgBAcg9gOwKIFsygIICcsEMQA8AKgHxQC8UA2kQDRQB0TRVAuqwNwBQokUAYnDgUoAby5QouACbSsEAM4KAkgmkBLAMaKA-AC5YiVOhDY8hBAFcUAIwhYSXAL5cumxAuBQAZgcHDKURk5RRU1LUUDNicgA

PR #1368, opened 2023-06-28
refactor(experimental): set up a pattern for splitting the serializers into co/dec/codec slices

steveluscher:
refactor(experimental): set up a pattern for splitting the serializers into co/dec/codec slices
## Summary

Most use cases for these serializers will only ever need to encode or decode, never both. This PR sets up a standard for splitting serializers into a version that encodes, decodes, or does both. Import the one you need, and let your build system tree-shake the other one away.

## Test Plan

```
cd packages/transactions/
pnpm test:unit:browser
pnpm test:unit:node
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/solana-labs/solana-web3.js/pull/1368).
* #1374
* #1373
* #1372
* #1370
* #1369
* __->__ #1368

### DIFF
--- a/packages/transactions/package.json
+++ b/packages/transactions/package.json
@@ -61,8 +61,7 @@
         "maintained node versions"
     ],
     "dependencies": {
-        "@metaplex-foundation/umi": "link:/home/sol/src/umi-git/packages/umi/",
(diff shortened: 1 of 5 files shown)


joncinque (commented):
Looks good!

PR #1371, opened 2023-06-28
refactor(experimental): bye bye `bs58`

steveluscher:
refactor(experimental): bye bye `bs58`
## Summary

This library is very frustrating in that it's large, not tree-shakable, and esm-only. Let's switch it out for Umi's slimmed down implementation.

## Test Plan

```
pnpm test:unit:browser
pnpm test:unit:node
```

### DIFF
--- a/packages/keys/package.json
+++ b/packages/keys/package.json
@@ -60,6 +60,9 @@
         "supports bigint and not dead",
         "maintained node versions"
     ],
+    "dependencies": {
+        "@metaplex-foundation/umi-serializers-encodings": "^0.8.2"
+    },
     "devDependencies": {
         "@solana/eslint-config-solana": "^1.0.1",
         "@swc/core": "^1.3.18",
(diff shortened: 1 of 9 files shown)


paulmillr:
@steveluscher why bring a bunch of dependencies for this simple task? I've glanced over https://github.com/metaplex-foundation/umi/blob/main/packages/umi-serializers-encodings/src/base64.ts and it's just atob/btoa. You would be better just using it natively.

Remember, every dependency and sub-dep is a supply chain risk.

paulmillr:
BTW it could be that the global variables atob/btoa are not supported in all engines, something to keep in mind.

steveluscher:
This PR does not concern itself with base64 encoding, nor `atob`/`btoa`.

paulmillr:
Crap, i've mistaken base64 for base58. 😐

PR #1378, opened 2023-06-30
Export some more required types from rpc-core

mcintyre94:
These types are intended to be exported, so that apps can create/pass them around. They follow the same pattern as `blockhash` which is exported.

### DIFF
--- a/packages/rpc-core/src/index.ts
+++ b/packages/rpc-core/src/index.ts
@@ -1,2 +1,5 @@
 export * from './blockhash';
 export * from './rpc-methods';
+export * from './stringified-bigint';
+export * from './transaction-signature';
+export * from './unix-timestamp';


steveluscher (approved):
So, we _definitely_ overexported everything in the existing web3.js, and I want to try to avoid that. Before we release this sucker I want to lock everything down as much as possible.

Before you ship this, can I ask what you need these types for? If importing them can be replaced with:

* just letting them flow through, or
* `ReturnType<typeof theFunctionThatProducesThem>`

…then maybe consider doing that.

Once we release these types into the wild, we can't change them without a complicated and lengthy deprecation process.

mcintyre94:
I think StringifiedBigInt, UnixTimestamp and TransactionError are currently only used as returns, so I'll remove them

I think TransactionSignature needs to be exported because it is used in the config input for eg `getSignaturesForAddress`. As far as I can tell there's no easy way to currently call that function, codesandbox: https://codesandbox.io/p/sandbox/find-ref-exports-9xp923 

You can't pass `{until: '4sVoc..'}` because it expects `TransactionSignature`, not string
You can't do `{until: '4sVoc...' as TransactionSignature}` because `TransactionSignature` isn't exported
You can't do `assertIsTransactionSignature(x)` because `assertIsTransactionSignature` isn't exported 

There's probably a way to extract it from `Parameters<rpc.getSignaturesForAddress>` but it'll probably be horrible because it'll be pulling it from the config object. And you wouldn't get the assert function which is probably the best way for a caller to use it.

steveluscher:
Oh, the intent is _definitely_ to export all of the `assertIsTransactionSignature` style methods. People need those helper methods to refine untyped values (eg. ones that come from some webservice over the network).

PR #1379, opened 2023-07-01
refactor(experimental): a function for generating secret keys

steveluscher:
refactor(experimental): a function for generating secret keys

## Summary

This PR introduces `generateSecretKey()`. You might need to use this when you need to sign for the creation of an account, for instance.

Instead of vending the _bytes_ of a secret key, however, we use JS-native `CryptoKey` instances. These are opaque tokens that you can return at a later time to perform some action, like deriving the public key for the secret they represent, or signing a message.

The idea is that you can freely pass these `CryptoKey` instances around your application without worrying about accidentally logging the key material itself – ie. to Sentry or to the browser console.

The only environments that support Ed25519 key generation at the moment:

* Node >=17.4
* Safari 17

For other environments, we'll supply a polyfill that implements key generation, signing, encryption, decryption, and verification in userspace.

Spec: https://wicg.github.io/webcrypto-secure-curves/#ed25519
Proposal repo: https://github.com/WICG/webcrypto-secure-curves
Implementation status: https://github.com/WICG/webcrypto-secure-curves/issues/20

## Test Plan

```
cd packages/keys/
pnpm test:unit:browser
pnpm test:unit:node
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/solana-labs/solana-web3.js/pull/1379).
* #1400
* #1399
* #1398
* #1397
* #1396
* #1395
* #1394
* __->__ #1379
* #1393
* #1392

### DIFF
--- a/packages/keys/src/__tests__/key-pair-test.ts
+++ b/packages/keys/src/__tests__/key-pair-test.ts
@@ -1,0 +1,51 @@
+import { generateKeyPair } from '../key-pair';
+
+describe('generateKeyPair', () => {
+    let oldIsSecureContext: boolean;
+    beforeEach(() => {
+        if (__BROWSER__) {
+            // FIXME: JSDOM does not set `isSecureContext` or otherwise allow you to configure it.
+            // Some discussion: https://github.com/jsdom/jsdom/issues/2751#issuecomment-846613392
+            if (globalThis.isSecureContext !== undefined) {
+                oldIsSecureContext = globalThis.isSecureContext;
+            }
+            globalThis.isSecureContext = true;
+        }
+    });
+    afterEach(() => {
+        if (oldIsSecureContext !== undefined) {
+            globalThis.isSecureContext = oldIsSecureContext;
+        }
+    });
+    it.each(['private', 'public'])('generates an ed25519 %s `CryptoKey`', async type => {
+        expect.assertions(1);
+        const keyPair = await generateKeyPair();
+        expect(keyPair).toMatchObject({
+            [`${type}Key`]: expect.objectContaining({
+                [Symbol.toStringTag]: 'CryptoKey',
+                algorithm: { name: 'Ed25519' },
+                type,
+            }),
+        });
+    });
+    it('generates a non-extractable private key', async () => {
+        expect.assertions(1);
+        const { privateKey } = await generateKeyPair();
+        expect(privateKey).toHaveProperty('extractable', false);
+    });
+    it('generates a private key usable for signing operations', async () => {
+        expect.assertions(1);
+        const { privateKey } = await generateKeyPair();
+        expect(privateKey).toHaveProperty('usages', ['sign']);
+    });
+    it('generates an extractable public key', async () => {
+        expect.assertions(1);
+        const { publicKey } = await generateKeyPair();
+        expect(publicKey).toHaveProperty('extractable', true);
+    });
+    it('generates a public key usable for verifying signatures', async () => {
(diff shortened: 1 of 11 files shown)


mcintyre94 (approved):
Nice, this looks great! Really like the ability to keep the generated keys from being logged etc

Have you thought about if/how we should support use cases where you do need the raw private keys? One example that comes to mind would be Metaplex's `keypairIdentity`

packages/keys/src/secrets.ts
```diff
@@ -0,0 +1,109 @@
...
+ *   |＿＿＿＿＿＿＿＿＿＿＿|
+ *   (\__/) ||
+ *   (•ㅅ•) ||
+ *   /    づ
+ */
+function getSecretKeyBytes_INTERNAL_ONLY_DO_NOT_EXPORT(secretKey: SecretKey): Uint8Array {
+    const bytes = storageKeyBySecretKey_INTERNAL_ONLY_DO_NOT_EXPORT?.get(secretKey);
+    if (!bytes) {
+        // TODO: Coded error.
+        throw new Error(
```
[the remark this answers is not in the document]
steveluscher: Yeah, they're all over. We can do them all at once.

steveluscher:
> Have you thought about if/how we should support use cases where you do need the raw private keys? One example that comes to mind would be Metaplex's `keypairIdentity`

Say more!

My rough plan if/when someone asks for this is:

1. Ask them to generate the key themselves
2. _If_ they need interop with web3.js at that point, export `createSecretKeyFromBytes` for them to use

steveluscher:
For the Metaplex case I'm pretty certain we just write a new `IdentityDriver` that takes in a `SecretKey`, implement all of these methods, and we're done.

https://github.com/ledamint-IO/js/blob/e10eba4c136d0ded2e495e671a2c8d76ac3e6270/src/drivers/identity/IdentityDriver.ts#L5-L23

steveluscher:
I'm going to massively rework this.

I just learned that ED25519 keys are coming to `SubtleCrypto`. It's already in Node 18.16.1 LTS and Safari 17.

1. Awesome.
2. `CryptoKey` instances can _actually_ be made un-exfiltratable, which is _great_.
3. All of these APIs are going to have to become async.

mcintyre94:
I think you're right about `IdentityDriver` for Metaplex, and anything using keypairs/secrets should ideally be structured like that so we can just write drivers without having to pass raw secret keys around.

packages/keys/src/__tests__/key-pair-test.ts
```diff
@@ -0,0 +1,51 @@
...
+    it('generates a non-extractable private key', async () => {
+        expect.assertions(1);
+        const { privateKey } = await generateKeyPair();
+        expect(privateKey).toHaveProperty('extractable', false);
+    });
+    it('generates a private key usable for signing operations', async () => {
+        expect.assertions(1);
+        const { privateKey } = await generateKeyPair();
+        expect(privateKey).toHaveProperty('usages', ['sign']);
+    });
```
mcintyre94, line 40: Nice, this API looks great! 🔥

steveluscher: It's SO good. Except for the part where it's all stupid magic combinations you have to know (ie. you can generate `Ed25519` keys with `"sign"` and `"verify"` usages, but supply the `"deriveKey"` usage and it will _fatal at runtime_. That part I hate.

PR #1380, opened 2023-07-04
refactor(experimental): add the getSlotLeaders RPC method

mcintyre94:
This PR adds the `getSlotLeaders` RPC method

I've used `minimumLedgerSlot` which we already have implemented to find a valid start slot for the test

Note that the docs mistakenly say these params are optional but they're required, see docs PR: https://github.com/solana-labs/solana/pull/32376 


packages/rpc-core/src/rpc-methods/__tests__/get-slot-leaders-test.ts
```diff
@@ -0,0 +1,60 @@
...
+            expect(typeof result[0]).toBe('string');
+        });
+    });
+
+    describe('when called with a `startSlot` higher than the highest slot available', () => {
+        it('throws an error', async () => {
+            expect.assertions(1);
+            const sendPromise = rpc
+                .getSlotLeaders(
+                    2n ** 63n - 1n, // u64:MAX; safe bet it'll be too high.
```
buffalojoec, line 35: 😂

buffalojoec (commented):
lgtm but I just wandered in here so will leave the check-mark for @steveluscher

PR #1381, opened 2023-07-05
refactor(experimental): add unit tests for getAccountInfo RPC method

mcintyre94:
This PR adds unit tests for the `getAccountInfo` RPC call. I'm probably going to end up refactoring this a bit to add `getMultipleAccounts` and `getProgramAccounts` so I want some tests first

Note that I've been unable to test the `jsonParsed` fully. I tried copying the data from a mainnet stake account (where `jsonParsed` works) into a fixture, but requesting `jsonParsed` encoding just returns the base64 encoding. I'm not sure if the test validator is missing the parsers?

### DIFF
--- a/packages/rpc-core/src/rpc-methods/__tests__/get-account-info-test.ts
+++ b/packages/rpc-core/src/rpc-methods/__tests__/get-account-info-test.ts
@@ -1,0 +1,207 @@
+import { Base58EncodedAddress } from '@solana/keys';
+import { createHttpTransport, createJsonRpc } from '@solana/rpc-transport';
+import type { SolanaJsonRpcErrorCode } from '@solana/rpc-transport/dist/types/json-rpc-errors';
+import type { Rpc } from '@solana/rpc-transport/dist/types/json-rpc-types';
+import fetchMock from 'jest-fetch-mock-fork';
+
+import { Commitment } from '../common';
+import { createSolanaRpcApi, SolanaRpcMethods } from '../index';
+
+describe('getAccountInfo', () => {
+    let rpc: Rpc<SolanaRpcMethods>;
+    beforeEach(() => {
+        fetchMock.resetMocks();
+        fetchMock.dontMock();
+        rpc = createJsonRpc<SolanaRpcMethods>({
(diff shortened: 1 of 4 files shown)


packages/rpc-core/src/rpc-methods/__tests__/get-account-info-test.ts
```diff
@@ -0,0 +1,182 @@
...
+                    .send();
+
+                expect(accountInfo.value?.data).toStrictEqual(['dGVzdCBkYXRh', 'base64']);
+            });
+        });
+
+        describe('for an account with parse-able JSON data', () => {
+            // TODO: not sure how to do this in the test validator
+            // Creating an account with data from a mainnet stake account doesn't work
+            // I think the testnet validator might not have the parsers?
```
steveluscher, line 142: I just tried this and it worked.

1. Find an account that definitely parses on mainnet (eg. `CSg2vQGbnwWdSyJpwK4i3qGfB6FebaV3xQTx4U1MbixN`)
2. `solana-test-validator -u m --clone CSg2vQGbnwWdSyJpwK4i3qGfB6FebaV3xQTx4U1MbixN`

```
curl http://localhost:8899 -X POST -H "Content-Type: application/json" -d '
  {
    "jsonrpc": "2.0",
    "id": 1,
    "method": "getAccountInfo",
    "params": [
      "CSg2vQGbnwWdSyJpwK4i3qGfB6FebaV3xQTx4U1MbixN",
      { "encoding": "jsonParsed" }
    ]
  }
'
```

steveluscher (changes requested):
I love this! Take another crack at the `jsonParsed` encoding, maybe?

packages/rpc-core/src/rpc-methods/__tests__/get-account-info-test.ts
mcintyre94: Thanks for the pointer! Figured it out, I was using the wrong `owner` in the fixture and I guess it uses that to figure out how to parse the account. Makes sense!

PR #1386, opened 2023-07-06
feat: Add isBlockhashValid

Arrowana:


steveluscher:
Awesome! Do you have time to add it here too?

https://github.com/solana-labs/solana-web3.js/tree/master/packages/rpc-core/src/rpc-methods

Contributor's guide:

https://github.com/solana-labs/solana-web3.js/issues/1278

packages/library-legacy/src/connection.ts
```diff
@@ -4601,6 +4616,32 @@ export class Connection {
     return res.result;
   }
 
+  /**
+   * Returns whether a blockhash is still valid or not
+   */
+  async isBlockhashValid(
+    blockhash: Blockhash,
+    commitmentOrConfig?: Commitment | IsBlockhashValidConfig,
```
steveluscher, line 4624: We never wanted these polymorphic inputs in the first place. The only reason we have them is because some methods used to be `foo(commitment)` and then we discovered we needed room for more config and made them `foo(commitment | config)`.

When starting fresh, for _sure_ just make it take a config.

packages/library-legacy/src/connection.ts
```diff
@@ -430,7 +430,7 @@ function jsonRpcResultAndContext<T, U>(value: Struct<T, U>) {
       context: pick({
         slot: number(),
       }),
-      value,
+      value: value,
```
steveluscher, line 433: giphy-2

packages/library-legacy/src/connection.ts
```diff
@@ -4601,6 +4616,32 @@ export class Connection {
...
+      commitment,
+      undefined /* encoding */,
+      config,
+    );
+    const unsafeRes = await this._rpcRequest('isBlockhashValid', args);
+    const res = create(unsafeRes, IsBlockhashValidRpcResult);
+    if ('error' in res) {
+      throw new SolanaJSONRPCError(
+        res.error,
+        'failed to fetch is blockhash valid',
```
steveluscher, line 4639:
```suggestion
        'failed to determine if the blockhash `' + blockhash + '`is valid',
```

steveluscher (changes requested):

steveluscher (approved):
Thank you!

Arrowana:
I'll also do the new one but in #1388

steveluscher:
Just have a TypeScript error in `test/connection.test.ts`. Other than that ready to ship!

PR #1388, opened 2023-07-07
refactor(experimental): Add IsBlockhashValid API

Arrowana:
# Summary
Add the IsBlockhashValid API + unit tests

# Checklist

- [x] Wherever my method's return value changes based on its inputs, I've implemented that as a [Typescript function overload](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads) ([example](https://github.com/solana-labs/solana-web3.js/blob/b69daf278045bc96740450e123a322e0dd95f60e/packages/rpc-core/src/rpc-methods/getAccountInfo.ts#L80-L114))
- [x] Wherever my test relies on reading account data I have created an account fixture in [`scripts/fixtures`](https://github.com/solana-labs/solana-web3.js/tree/master/scripts/fixtures) that gets loaded into the test validator ([example](https://github.com/solana-labs/solana-web3.js/blob/master/scripts/fixtures/4nTLDQiSTRHbngKZWPMfYnZdWTbKiNeuuPcX7yFUpSAc.json))

Testing wise i can see other methods sometimes only expect the type, do we want a true/false for this API? It will have to depend on read from another API if we want to exercise `true`

### DIFF
--- a/packages/rpc-core/src/rpc-methods/__tests__/is-blockhash-valid-test.ts
+++ b/packages/rpc-core/src/rpc-methods/__tests__/is-blockhash-valid-test.ts
@@ -1,0 +1,29 @@
+import { createHttpTransport, createJsonRpc } from '@solana/rpc-transport';
+import type { Rpc } from '@solana/rpc-transport/dist/types/json-rpc-types';
+import fetchMock from 'jest-fetch-mock-fork';
+
+import { Blockhash } from '../../blockhash';
+import { Commitment } from '../common';
+import { createSolanaRpcApi, SolanaRpcMethods } from '../index';
+
+describe('isBlockhashValid', () => {
+    let rpc: Rpc<SolanaRpcMethods>;
+    beforeEach(() => {
+        fetchMock.resetMocks();
+        fetchMock.dontMock();
+        rpc = createJsonRpc<SolanaRpcMethods>({
+            api: createSolanaRpcApi(),
+            transport: createHttpTransport({ url: 'http://127.0.0.1:8899' }),
+        });
+    });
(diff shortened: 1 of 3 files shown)


steveluscher:
I think a file got missed in the last commit!

Arrowana:
> I think a file got missed in the last commit!

Yep sorry, i have been quite sloppy.

it looks like multiple live test are sort of broken, is this a known issue?

It looks like they depend on node runtime behaviour which has changed
```
 2) Connection
@solana/web3.js:test:live-with-test-validator:        simulateTransaction:
@solana/web3.js:test:live-with-test-validator: 
@solana/web3.js:test:live-with-test-validator:       AssertionError: expected [ { data: [ '', 'base64' ], …(4) } ] to deeply equal [ { data: [ '', 'base64' ], …(5) } ]
@solana/web3.js:test:live-with-test-validator:       + expected - actual
@solana/web3.js:test:live-with-test-validator: 
@solana/web3.js:test:live-with-test-validator:            ]
@solana/web3.js:test:live-with-test-validator:            "executable": false
@solana/web3.js:test:live-with-test-validator:            "lamports": 999995000
@solana/web3.js:test:live-with-test-validator:            "owner": "11111111111111111111111111111111"
@solana/web3.js:test:live-with-test-validator:       -    "rentEpoch": 0
@solana/web3.js:test:live-with-test-validator:       +    "rentEpoch": 18446744073709552000
@solana/web3.js:test:live-with-test-validator:       +    "space": 0
@solana/web3.js:test:live-with-test-validator:          }
@solana/web3.js:test:live-with-test-validator:        ]
```

steveluscher:
Push safety is so so hard when you don't control both ends. We just test against the edge validator, because I don't know what else to do.

Your PR passes CI, so 🚢

PR #1390, opened 2023-07-11
refactor(experimental): add tests for all jsonParsed parseable accounts

mcintyre94:
While looking at using our new `getAccountInfo` with `jsonParsed` encoding in Explorer, I've realised that we need to go further with our parsing. There are many different programs which have a parser available, and many of these have non-BigInt numbers in their parsed representation of accounts, including floats. We need to include all of these in `response-patcher-allowed-numeric-values.ts` to avoid errors from trying to parse them as BigInt incorrectly.

This PR adds fixtures for all the currently parseable program types, and a unit test to parse each one.


steveluscher:
Nice sleuthing!

Just before we go forward with this:

1. this is better than just opting _all_ parsed account data out of the bigint conversion, right?
2. should we update the response patcher to detect floats and exempt them from conversion?

steveluscher (commented):
Nice sleuthing!

Just before we go forward with this:

1. this is better than just opting _all_ parsed account data out of the bigint conversion, right?
2. should we update the response patcher to detect floats and exempt them from conversion?

packages/rpc-core/src/response-patcher-allowed-numeric-values.ts
```diff
@@ -9,7 +9,55 @@ import { createSolanaRpcApi } from './rpc-methods';
...
+        ['value', 'data', 'parsed', 'info', 'numValidSigners'],
+        // parsed Stake account
+        ['value', 'data', 'parsed', 'info', 'stake', 'delegation', 'warmupCooldownRate'],
+        // parsed Sysvar rent account
+        ['value', 'data', 'parsed', 'info', 'exemptionThreshold'],
+        ['value', 'data', 'parsed', 'info', 'burnPercent'],
+        // parsed Vote account
+        ['value', 'data', 'parsed', 'info', 'commission'],
+        ['value', 'data', 'parsed', 'info', 'votes', KEYPATH_WILDCARD, 'confirmationCount'],
+    ],
```
steveluscher, line 60: If we decided to bail altogether we could make something like `['value', 'data', 'parsed', GREEDY_WILDCARD]` that matches everything from `value.data.parsed` on down.

mcintyre94:
@steveluscher Thanks for the review!

1. I don't think we should opt all the parsed account data out because there are a bunch of `u64` fields in the mix too. I expect we'd be incorrectly parsing a lot of account fields if we just parsed them all as Number

2. I think we probably should do this. It would avoid bugs where a field being missed from this list causes the function call to error with no useful response data, which is where we are now. But I think the result would be that a field is sometimes returned as a BigInt and sometimes a Number, because often something that can be a float will just be an integer. So I think we'd still want to add these fields as exceptions for that consistency. But it's much better to return something with slightly inconsistent types than to error out. And I'm sure we are going to sometimes miss stuff from this list, especially with floats that are only occasionally not integers.

steveluscher (approved):
Nice. Shipit.

giphy

PR #1395, opened 2023-07-13
refactor(experimental): a polyfill for `generateKey()` that implements Ed25519 key generation in userspace

steveluscher:
refactor(experimental): a polyfill for `generateKey()` that implements Ed25519 key generation in userspace

## Summary

For environments where Ed25519 key generation is not supported, this polyfill injects a suitable implementation that uses `@noble/curves/ed25519` behind the scenes.

Here's the list of environments that do support it: https://github.com/WICG/webcrypto-secure-curves/issues/20

## Test Plan

```
cd packages/webcrypto-ed25519-polyfill
pnpm test:unit:browser
pnpm test:unit:node
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/solana-labs/solana-web3.js/pull/1395).
* #1400
* #1399
* #1398
* #1397
* #1396
* __->__ #1395
* #1394


packages/webcrypto-ed25519-polyfill/src/__tests__/index-test.ts
```diff
@@ -0,0 +1,185 @@
...
+                globalThis.crypto.subtle.generateKey('Ed25519', /* extractable */ false, ['sign', 'verify'])
+            ).resolves.toBe(mockKeyPair);
+        });
+        it('calls the original `generateKey` once per call to `generateKey` when the algorithm is "Ed25519" but never again (parallel version)', async () => {
+            expect.assertions(1);
+            await Promise.all([
+                globalThis.crypto.subtle.generateKey('Ed25519', /* extractable */ false, ['sign', 'verify']),
+                globalThis.crypto.subtle.generateKey('Ed25519', /* extractable */ false, ['sign', 'verify']),
+            ]);
+            expect(originalGenerateKey).toHaveBeenCalledTimes(2);
```
mcintyre94, line 154: Is the "but never again" supposed to be in the name of this test? It looks like this will call the original function once per call, always? (Same question for the serial version below)

mcintyre94 (approved):

packages/webcrypto-ed25519-polyfill/src/__tests__/index-test.ts
steveluscher: giphy

PR #1401, opened 2023-07-14
Add empty optional config object to RPC methods with no params

### LINKED ISSUE #1389: Experimental web3js: RPC methods with no inputs are typed `unknown`
## Overview

I'm trying to use `getFirstAvailableBlock`, `getEpochInfo` and `getEpochSchedule` in a refactor of Explorer. I've found that the return of `getFirstAvailableBlock` and `getEpochSchedule` are typed as `unknown` by TypeScript. However `getEpochInfo` is typed correctly even when called with no parameters, I think because it has an optional parameter. 

## Steps to reproduce

See replit: https://replit.com/@mcintyre94/web3js-unknown#index.ts 

See how TypeScript types these incorrectly:
Screenshot 2023-07-10 at 17 39 02
Screenshot 2023-07-10 at 17 39 15
Screenshot 2023-07-10 at 17 39 25

We also see this in the autocomplete:
Screenshot 2023-07-10 at 17 39 59
Screenshot 2023-07-10 at 17 40 05

I've also seen this locally using VSCode so it's not a Replit issue 

## Description of bug

As far as I can tell this only happens with RPC methods that take no inputs, either optional or required

I'm guessing we need to tighten up the proxy a bit to correctly type this case

The full message when autocompleting `getEpochSchedule` is:

```
(method) getEpochSchedule(...args: unknown[]): PendingRpcRequest<unknown> (+1 overload)
```

mcintyre94:
This fixes an issue where TS incorrectly types their returns as `unknown`

We now get the correct response type: 
Screenshot 2023-07-14 at 14 43 55

Closes #1389 (with a workaround)


packages/rpc-core/src/rpc-methods/getEpochSchedule.ts
```diff
@@ -16,6 +16,10 @@ type GetEpochScheduleApiResponse = Readonly<{
 export interface GetEpochScheduleApi {
     /**
      * Returns the epoch schedule information from this cluster's genesis config
+     * Note that the optional config object is ignored. See https://github.com/solana-labs/solana-web3.js/issues/1389
      */
-    getEpochSchedule(): GetEpochScheduleApiResponse;
+    getEpochSchedule(
+        // FIXME: https://github.com/solana-labs/solana-web3.js/issues/1389
+        config?: Record<string, never>
```
steveluscher, line 23: _Maybe_ consider naming this argument something else? Think of what you'd want to show up in autocomplete.

```ts
// Is this the worst idea?
getEpochSchedule(NO_CONFIG?: Record<string, never>)
```

steveluscher (approved):
giphy-1

PR #1408, opened 2023-07-17
ci: add concurrency limit to turbo in CI (default was 10)

mcintyre94:
This seems to resolve flakiness in CI

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -52,7 +52,7 @@ jobs:
       - name: Publish NPM
         run: |
           pnpm config set '//registry.npmjs.org/:_authToken' "${NPM_TOKEN}"
-          pnpm publish-packages
+          pnpm publish-packages --concurrency=100%
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/pull-requests.yml
+++ b/.github/workflows/pull-requests.yml
@@ -99,7 +99,7 @@ jobs:
         run: scripts/setup-test-validator.sh
 
       - name: Build & Test
-        run: pnpm build --concurrency=5
+        run: pnpm build --concurrency=100%
 
       - name: Upload Experimental library build artifacts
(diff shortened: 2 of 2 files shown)


steveluscher:
OMG. I remember doing this somewhere else now, I just can’t remember where.

We for sure need to target this just to CI though.

This probably means the PRs that succeed most are the ones with the most cached Turborepo steps, which explains a lot about which ones succeed the most.

package.json
```diff
@@ -4,7 +4,7 @@
         "packages/*"
     ],
     "scripts": {
-        "build": "turbo run build",
+        "build": "turbo run build --concurrency=5",
```
steveluscher, line 7: Let’s make this a part of cicd.yaml instead.

steveluscher (changes requested):

package.json
mcintyre94: I've added it to pull-requests.yaml. It looks like cicd.yaml doesn't call the build/test and from a quick look I don't think any other workflows do either

steveluscher: Related: https://github.com/solana-labs/wallet-adapter/commit/6f2dfe515224f4c16ebf9803432f5b7b1fc0d93a

steveluscher: > It looks like cicd.yaml doesn't call the build/test and from a quick look I don't think any other workflows do either

It does, as a dependency of `publish-packages`.

https://github.com/solana-labs/solana-web3.js/blob/0e113c0e26358bf0ace7b5833f03f84e7a5a38f0/.github/workflows/cicd.yml#L55C11-L55C32

https://github.com/solana-labs/solana-web3.js/blob/0e113c0e26358bf0ace7b5833f03f84e7a5a38f0/package.json#L9

I think the right way to do this is to change `cicd.yml` to read `pnpm publish-packages -- --concurrency=100%` or something. You can test whether that works with `pnpm build -- --concurrency=100%` maybe?

jordansexton: Yeah, 100% is best rather than a fixed amount. We do this @ https://github.com/solana-labs/wallet-adapter/blob/ba57f756ec76349c6e46420d387cd706eecd5ad2/package.json#L17

PR #1411, opened 2023-07-18
Update JSDoc of the legacy keypair, transaction, and clusterApiUrl API

Rowaaaaan:
This pull request adds some minor JSDoc to the legacy keypair, transaction, and clusterApiUrl API so that they actually show up in the dev's LSP hover instead of having to dig through the documentation themselves.


steveluscher:
giphy-1

PR #1421, opened 2023-07-21
refactor(experimental): the `sendTransaction` RPC method

steveluscher:
refactor(experimental): the `sendTransaction` RPC method

## Summary

This is the method that you can use to send a fully signed wire transaction to the network.

## Test Plan

```
cd packages/rpc-core/
pnpm test:unit:browser
pnpm test:unit:node
```

explorer solana com_tx_5cZkueHfsXxZ9C7JEK4QG6MyEPMxWBgGkDdnxqLNQCrSHJ9iAr2cXRm5NDBjhtfTzDAE6i6PkcW5xHj66LDdADJV_cluster=custom customUrl=http%3A%2F%2F127 0 0 1%3A8899 signatures=%255B%252235yVDM4AzXiQfsWSyyrYoDDvoiBKyHh6sRFn2G7xz6nYLWADVxK2

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/solana-labs/solana-web3.js/pull/1421).
* #1422
* __->__ #1421
* #1420

PR #1423, opened 2023-07-21
chore: wait for the test validator to form first root before running tests

steveluscher:
chore: wait for the test validator to form first root before running tests

## Summary

This change makes it so that you never have to think about whether you can run a test at `"finalized"` commitment or if the validator was started too recently. This PR makes the tests wait (~13s) until the test validator has formed its first root (ie. the slot is >1)

## Test Plan

```
pnpm turbo test:unit:browser test:unit:node
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/solana-labs/solana-web3.js/pull/1423).
* #1422
* #1421
* __->__ #1423
* #1420
* #1419
* #1416
* #1415
* #1414
* #1413
* #1412
* #1400


steveluscher:
Ugh. We can't exactly do this because it makes `--watch` mode unbearable.

steveluscher:
I'm going to abandon this until there's a better solution than blocking the tests for 13 seconds. We just can't test against `finalized` in tests until then.

PR #1425, opened 2023-07-22
First Pass at Wiring in Experimental Web3.js

buffalojoec:
Here's what I've wired in so far. Works on local validator tests, but the unit tests seem busted right now.


packages/library-legacy/src/connection.ts
```diff
@@ -3188,6 +3196,458 @@ export class Connection {
...
+  private castToEpochInfo(value: any): EpochInfo {
+    return {
+      epoch: Number(value.epoch),
+      slotIndex: Number(value.slotIndex),
+      slotsInEpoch: Number(value.slotsInEpoch),
+      absoluteSlot: Number(value.absoluteSlot),
+      blockHeight: Number(value.blockHeight),
+      transactionCount: value.transactionCount
+        ? Number(value.transactionCount)
+        : undefined,
```
steveluscher, line 3562: You're doing this to cast from bigint to `Number`, right?

I planned for this!

What you have to do is to create an `IRPCTransport` that does _not_ have the bigint conversion feature.

Check this out: https://github.com/solana-labs/solana-web3.js/blob/9968a730655a79ec1bb25b665eb40a9ce4300b6a/packages/library/src/rpc-transport.ts#L21

In the _legacy_ library, just create a transport _without_ the request coalescer. At that point, a _lot_ more of these types should just be usable as-is.

steveluscher (changes requested):

packages/library-legacy/src/connection.ts
```diff
@@ -58,6 +58,19 @@ import type {Blockhash} from './blockhash';
...
+import {Blockhash as EXPERIMENTAL_Blockhash} from '../../rpc-core/src/blockhash';
+import {TransactionSignature as EXPERIMENTAL_TransactionSignature} from '../../rpc-core/src/transaction-signature';
+import {
+  Commitment as EXPERIMENTAL_Commitment,
+  Slot,
+  U64UnsafeBeyond2Pow53Minus1,
+} from '../../rpc-core/src/rpc-methods/common';
+import {createDefaultRpcTransport, createSolanaRpc} from '../../library/src';
+import type {IRpcTransport} from '../../rpc-transport/dist/types/transports/transport-types';
+import type {Rpc} from '../../rpc-transport/dist/types/json-rpc-types';
```
steveluscher, line 73: You can add all the packages to the package.json:

```
"@solana/rpc-core": "workspace:*",
```

And then `pnpm i` and then just import them like you will in prod:

```
import {Base58EncodedAddress} from '@solana/keys';
```

packages/library-legacy/src/connection.ts
```diff
@@ -58,6 +58,19 @@ import type {Blockhash} from './blockhash';
...
+import {Base58EncodedAddress} from '../../keys';
+import {SolanaRpcMethods} from '../../rpc-core/src';
+import {Blockhash as EXPERIMENTAL_Blockhash} from '../../rpc-core/src/blockhash';
+import {TransactionSignature as EXPERIMENTAL_TransactionSignature} from '../../rpc-core/src/transaction-signature';
+import {
+  Commitment as EXPERIMENTAL_Commitment,
+  Slot,
+  U64UnsafeBeyond2Pow53Minus1,
+} from '../../rpc-core/src/rpc-methods/common';
+import {createDefaultRpcTransport, createSolanaRpc} from '../../library/src';
```
steveluscher, line 71: Try to do this entire PR without ever importing `library/`. The point is for `library/` to have all of the modern configuration (like the `bigint` coalescer) and for `library-legacy/` to assemble all of the primitives in a ‘legacy compatibility’ style.

steveluscher (changes requested):

packages/library-legacy/src/connection.ts
buffalojoec: Gotcha, cool! This is what I ended up doing

packages/library-legacy/src/connection.ts
buffalojoec: that makes a lot of sense for the coalescing no doubt

packages/library-legacy/src/connection.ts
buffalojoec: smhhhhhhh thank you lol. I had to read a lot of how you built `library` before I realized exactly how it works

steveluscher: Sorry; I linked you to the wrong line earlier. Updated the comment above!

PR #1436, opened 2023-07-25
refactor(experimental): a function to turn a transaction into a durable nonce transaction

steveluscher:
refactor(experimental): a function to turn a transaction into a durable nonce transaction

## Summary

A durable nonce transaction is one whose:

1. Lifetime constraint is that of a durable nonce
2. First instruction is an `AdvanceNonceAccount` instruction against the system program

Use this function to turn a transaction into a durable nonce transaction by adding those two things.

## Test Plan

```
cd packages/transactions/
pnpm test:unit:browser
pnpm test:unit:node
```

### DIFF
--- a/packages/transactions/src/durable-nonce.ts
+++ b/packages/transactions/src/durable-nonce.ts
@@ -3,6 +3,7 @@ import { ReadonlyAccount, ReadonlySignerAccount, WritableAccount } from '@solana
 import { Base58EncodedAddress } from '@solana/keys';
 
 import { BaseTransaction } from './types';
+import { getUnsignedTransaction } from './unsigned-transaction';
 
 type AdvanceNonceAccountInstruction<
     TNonceAccountAddress extends string = string,
@@ -19,6 +20,15 @@ type AdvanceNonceAccountInstruction<
 type AdvanceNonceAccountInstructionData = Uint8Array & {
     readonly __advanceNonceAccountInstructionData: unique symbol;
 };
+type DurableNonceConfig<
+    TNonceAccountAddress extends string = string,
+    TNonceAuthorityAddress extends string = string,
+    TNonceValue extends string = string
+> = Readonly<{
+    readonly nonce: Nonce<TNonceValue>;
+    readonly nonceAccountAddress: Base58EncodedAddress<TNonceAccountAddress>;
+    readonly nonceAuthorityAddress: Base58EncodedAddress<TNonceAuthorityAddress>;
+}>;
 export type Nonce<TNonceValue extends string = string> = TNonceValue & { readonly __nonce: unique symbol };
 type NonceLifetimeConstraint<TNonceValue extends string = string> = Readonly<{
     nonce: Nonce<TNonceValue>;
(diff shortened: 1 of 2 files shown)


packages/transactions/src/unsigned-transaction.ts
```diff
@@ -0,0 +1,17 @@
...
+export function getUnsignedTransaction<TTransaction extends BaseTransaction>(
+    transaction: TTransaction | (TTransaction & ITransactionWithSignatures)
+): TTransaction | Omit<TTransaction & ITransactionWithSignatures, keyof ITransactionWithSignatures> {
+    if ('signatures' in transaction) {
+        // The implication of the lifetime constraint changing is that any existing signatures are invalid.
+        const {
+            signatures: _, // eslint-disable-line @typescript-eslint/no-unused-vars
+            ...unsignedTransaction
+        } = transaction;
+        return unsignedTransaction;
```
buffalojoec, line 13: This one is for my own understanding: Through the act of making this a durable nonce transaction, we void any existing signatures? Bc they are only valid for a recent blockhash, right

packages/transactions/src/instructions.ts
```diff
@@ -39,8 +18,10 @@ export function prependTransactionInstruction<TTransaction extends BaseTransacti
...
 ): TTransaction | Omit<TTransaction, keyof ITransactionWithSignatures> {
-    const nextInstructions = [instruction, ...transaction.instructions] as const;
-    const out = replaceInstructions(transaction, nextInstructions);
+    const out = {
+        ...getUnsignedTransaction(transaction),
+        instructions: [instruction, ...transaction.instructions],
+    };
     Object.freeze(out);
     return out;
 }
```
buffalojoec, line 27: Assuming you offer `prepend` and `append` to add the nonce ixs before or after a set of arbitrary instructions?

packages/transactions/src/durable-nonce.ts
```diff
@@ -16,20 +20,137 @@ type AdvanceNonceAccountInstruction<
...
+}
+
+function isDurableNonceTransaction(
+    transaction: BaseTransaction | (BaseTransaction & IDurableNonceTransaction)
+): transaction is BaseTransaction & IDurableNonceTransaction {
+    return (
+        'lifetimeConstraint' in transaction &&
+        typeof transaction.lifetimeConstraint.nonce === 'string' &&
+        transaction.instructions[0] != null &&
+        isAdvanceNonceAccountInstruction(transaction.instructions[0])
```
buffalojoec, line 117: I think this answers my one question.
This first instruction is always `AdvanceNonce` in a DNT ?

buffalojoec (approved):
Lgtm, I haven't worked with durable nonces so I asked a couple of questions for my own understanding

packages/transactions/src/unsigned-transaction.ts
steveluscher: Basically the part of any transaction that needs to be signed _includes_ the blockhash/nonce as _content_. A change in the blockhash/nonce changes the transaction's content. Consequently, any existing signatures are no longer valid.

packages/transactions/src/instructions.ts
steveluscher: In order for a transaction to make use of a nonce:

1. The nonce has to be set in place of the ‘recent blockhash’
2. The first instruction in the transaction _has_ to be the `AdvanceNonceAccount` instruction

So prepend only. No appending allowed.

https://docs.solana.com/implemented-proposals/durable-tx-nonces#runtime-support

steveluscher: In any case, these lines weren't part of this PR.

packages/transactions/src/durable-nonce.ts
steveluscher: Always!

PR #1448, opened 2023-07-27
refactor(experimental): a function to test if a public key is on the Ed25519 curve

steveluscher:
refactor(experimental): a function to test if a public key is on the Ed25519 curve

# Summary

The goal here was to craft the smallest possible utility to test whether a public key lies on the Ed25519 curve or not. This will be applicable to a forthcoming utility to generate off-curve addresses; addresses with no corresponding private key (eg. program derived addresses)

# Test Plan

```
cd packages/keys
pnpm test:unit:browser
pnpm test:unit:node
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/solana-labs/solana-web3.js/pull/1448).
* #1453
* #1452
* #1451
* __->__ #1448


packages/addresses/src/__tests__/curve-test.ts
```diff
@@ -0,0 +1,36 @@
...
+    ]),
+    new Uint8Array([
+        52, 94, 161, 109, 55, 62, 164, 12, 183, 165, 56, 112, 86, 103, 19, 109, 196, 33, 93, 42, 143, 6, 221, 172, 173,
+        21, 130, 96, 170, 101, 82, 200,
+    ]),
+];
+
+describe('compressedPointBytesAreOnCurve', () => {
+    it.each(OFF_CURVE_KEY_BYTES)(
+        'returns false when a public key does not lie on the Ed25519 curve [%#]',
```
mcintyre94, line 26: What does `[%#]` mean/do?

packages/addresses/src/curve.ts
```diff
@@ -0,0 +1,24 @@
...
+    return BigInt(integerLiteralString);
+}
+
+export async function compressedPointBytesAreOnCurve(bytes: Uint8Array): Promise<boolean> {
+    if (bytes.byteLength !== 32) {
+        return false;
+    }
+    const y = decompressPointBytes(bytes);
+    return pointIsOnCurve(y, bytes[31]);
+}
```
mcintyre94, line 24: Is there any explanation for the code in this file we could add a link to as a comment? Are these standard/well known ECC algorithms for example?

packages/addresses/src/vendor/noble/ed25519.ts
```diff
@@ -0,0 +1,94 @@
...
+    const x = uvRatio(u, v); // (uv³)(uv⁷)^(p-5)/8; square root
+    if (x === null) {
+        return false;
+    }
+    const isLastByteOdd = (lastByte & 0x80) !== 0; // x_0, last bit
+    if (x === 0n && isLastByteOdd) {
+        return false;
+    }
+    return true;
+}
```
mcintyre94, line 94: What's the reason for vendor-ing this? Is it because of the tree-shaking issue with the `noble-ed25519` lib that you mentioned before?

For maintaining this, where does it actually come from? Some of it looks like https://github.com/paulmillr/noble-ed25519/blob/be4b5cd15c3d1fcc82bcb173394c0afb17674f20/index.ts but eg `uvRatio` has a different signature and `pointIsOnCurve` isn't in that file at all 

Also how are you thinking about security if we vendor a modified(?) version of that lib? The original is audited/fuzzed and I'd be a bit worried about subtle bugs in our version

packages/addresses/src/__tests__/curve-test.ts
steveluscher: It spits out the index of the test case in `it.each()`.

https://jestjs.io/docs/api#1-testeachtablename-fn-timeout

packages/addresses/src/curve.ts
steveluscher: Yeah, I actually put the comment to the algorithm above the implementation of `pointIsOnCurve()` in `vendor/noble/ed25519.ts`

```ts
https://datatracker.ietf.org/doc/html/rfc8032#section-5.1.3
export function pointIsOnCurve(y: bigint, lastByte: number): boolean {
```

packages/addresses/src/vendor/noble/ed25519.ts
steveluscher: > Is it because of the tree-shaking issue with the noble-ed25519 lib that you mentioned before?

That's the one. More importantly though, it doesn't provide a standalone ‘does this compressed point decompress or not’ test without _also_ getting an instance of `Point` with all of the implementation of `Point` that we don't need.

> eg `uvRatio` has a different signature

That's right. The original returns the value even if it's invalid, and the modified one returns `null` if the value is invalid.

> Also how are you thinking about security if we vendor a modified(?) version of that lib?

On the face of it, all bets are off at that point, and the new code needs a new audit. I'm _honestly_ not that worried about it though, since this code isn't being used to generate or operate over private key material. All of the bits that flow through `pointIsOnCurve()` are public.

Paging @samkim-crypto to check if my thinking is wrong here.

steveluscher: I'm going to land this now, but we can absolutely keep discussing it and take a different direction if need be.

samkim-crypto: > > Also how are you thinking about security if we vendor a modified(?) version of that lib?
> 
> On the face of it, all bets are off at that point, and the new code needs a new audit. I'm _honestly_ not that worried about it though, since this code isn't being used to generate or operate over private key material. All of the bits that flow through `pointIsOnCurve()` are public.
> 
> Paging @samkim-crypto to check if my thinking is wrong here.
Yeah, as @steveluscher mentioned, the original library has a careful implementation since it is using it for parts that require secrets as well. If the function is only used for public operations like key derivation, then the implementation should be fine.

We can probably add a warning on the function doc that the function is not meant to be used for private key material. There are also components that enforce constant time evaluation, which we can probably remove if we really wanted to optimize it.

steveluscher: I mean hey, if you want to take a crack at writing our own routine, the offer is still open :)

PR #1453, opened 2023-07-28
refactor(experimental): a function to compute program derived addresses

steveluscher:
refactor(experimental): a function to compute program derived addresses

# Summary

Given a program address and a series of string or bytearray seeds, this function will compute the off-chain program derived address (PDA) corresponding to both.

# Test Plan

```
cd packages/addresses
pnpm test:unit:browser
pnpm test:unit:node
```

### DIFF
--- a/packages/addresses/README.md
+++ b/packages/addresses/README.md
@@ -60,3 +60,24 @@ import { getBase58EncodedAddressFromPublicKey } from '@solana/addresses';
 
 const address = await getBase58EncodedAddressFromPublicKey(publicKey);
 ```
+
+### `getProgramDerivedAddress()`
+
+Given a program's `Base58EncodedAddress` and up to 16 `Seeds`, this method will return the program derived address (PDA) associated with each.
+
+```ts
+import { getBase58EncodedAddressCodec, getProgramDerivedAddress } from '@solana/addresses';
+
+const { serialize } = getBase58EncodedAddressCodec();
+const { bumpSeed, pda } = await getProgramDerivedAddress({
+    programAddress: 'ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL' as Base58EncodedAddress,
+    seeds: [
+        // Owner
+        serialize('9fYLFVoVqwH37C3dyPi6cpeobfbQ2jtLpN5HgAYDDdkm' as Base58EncodedAddress),
+        // Token program
+        serialize('TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA' as Base58EncodedAddress),
+        // Mint
+        serialize('EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v' as Base58EncodedAddress),
+    ],
+});
+```
--- a/packages/addresses/src/__tests__/program-derived-address-test.ts
+++ b/packages/addresses/src/__tests__/program-derived-address-test.ts
@@ -1,0 +1,141 @@
+import { Base58EncodedAddress } from '../base58';
+import { getProgramDerivedAddress } from '../program-derived-address';
+
+describe('getProgramDerivedAddress()', () => {
+    it('fatals when supplied more than 16 seeds', async () => {
+        expect.assertions(1);
+        await expect(
+            getProgramDerivedAddress({
+                programAddress: 'FN2R9R724eb4WaxeDmDYrUtmJgoSzkBiQMEHELV3ocyg' as Base58EncodedAddress,
+                seeds: Array(17).fill(''),
+            })
+        ).rejects.toThrow(/A maximum of 16 seeds/);
+    });
+    it.each([
+        new Uint8Array(Array(33).fill(0)),
+        'a'.repeat(33),
+        '\uD83D\uDC68\u200D\uD83D\uDC68\u200D\uD83D\uDC67\u200D\uD83D\uDC66\uD83D\uDC69\u200D\uD83D\uDC69\u200D\uD83D\uDC67\u200D\uD83D\uDC66\uD83D\uDC68\u200D\uD83D\uDC69\u200D\uD83D\uDC67\u200D\uD83D\uDC66',
+    ])('fatals when supplied a seed that is 33 bytes long', async oversizedSeed => {
+        expect.assertions(1);
+        await expect(
+            getProgramDerivedAddress({
+                programAddress: '5eUi55m4FVaDqKubGH9r6ca1TxjmimmXEU9v1WUZJ47Z' as Base58EncodedAddress,
+                seeds: [oversizedSeed],
+            })
+        ).rejects.toThrow(/exceeds the maximum length of 32 bytes/);
+    });
+    it('returns a program derived address given a program address and no seeds', async () => {
+        expect.assertions(1);
+        await expect(
+            getProgramDerivedAddress({
+                programAddress: 'CZ3TbkgUYpDAJVEWpujQhDSgzNTeqbokrJmYa1j4HAZc' as Base58EncodedAddress,
+                seeds: [],
+            })
+        ).resolves.toStrictEqual({
+            bumpSeed: 255,
+            pda: '9tVtkyCGAHSDDBPwz7895aC3p2gJRjpu2v26o35FTUco',
+        });
+    });
+    it('returns a program derived address after having tried multiple bump seeds given a program address and no seeds', async () => {
+        expect.assertions(1);
+        await expect(
+            getProgramDerivedAddress({
+                programAddress: 'EfTbwNBrSqSuCNBhWUHsBoBdSMWgRU1S47daqRNgW7aK' as Base58EncodedAddress,
+                seeds: [],
+            })
+        ).resolves.toStrictEqual({
+            bumpSeed: 251,
+            pda: 'CKWT8KZ5GMzKpVRiAULWKPg1LiHt9U3NdAtbuTErHCTq',
+        });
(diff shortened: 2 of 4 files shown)


packages/addresses/LICENSE
```diff
@@ -0,0 +1,20 @@
+Copyright (c) 2018 Solana Labs, Inc
```
2501babe, line 1: :thinking:

packages/addresses/README.md
```diff
@@ -0,0 +1,83 @@
...
+        const balanceInLamports = await rpc.getBalance(address).send();
+    } catch (e) {
+        // `address` turned out not to be a base58-encoded address
+    }
+}
+```
+
+### `getBase58EncodedAddressFromPublicKey()`
+
+Given a public `CryptoKey`, this method will return its associated `Base58EncodedAddress`.
```
2501babe, line 56: not sure how this will feel in practice but in theory i like the distinction between pubkeys and addresses

2501babe: i wonder if the method name could be shorter tho, since it will be used a lot. `getAddressFromPublicKey`? or maybe have a `toAddress()` on the `CryptoKey`?

packages/addresses/src/__tests__/curve-test.ts
2501babe, line 1: not necessary for this pr but when in "finishing touches" mode this is the kind of thing i personally would have a set of tests like "generate a few thousand random base addresses and seed strings and assert the old and new libraries produce the same result"

packages/addresses/src/__tests__/program-derived-address-test.ts
```diff
@@ -0,0 +1,141 @@
+import { Base58EncodedAddress } from '../base58';
+import { getProgramDerivedAddress } from '../program-derived-address';
+
+describe('getProgramDerivedAddress()', () => {
+    it('fatals when supplied more than 16 seeds', async () => {
```
2501babe, line 5: huh interesting was this always a property of pda generation? i never had cause to check, and more than 16 is certainly pointless, but i always assumed it was happy hashing more inputs in forever

packages/addresses/src/__tests__/program-derived-address-test.ts
```diff
@@ -0,0 +1,141 @@
...
+                seeds: ['butterfly'],
+            }),
+            getProgramDerivedAddress({
+                programAddress: '9PyoV2rqNtoboSvg2JD7GWhM5RQvHGwgdDvK7MCfpgX1' as Base58EncodedAddress,
+                seeds: ['butter', 'fly'],
+            }),
+        ]);
+        expect(pdaButterfly).toStrictEqual(pdaButterFly);
+    });
+    // https://solana.stackexchange.com/questions/7253/what-combination-of-program-address-and-seeds-would-cause-findprogramaddress-t
```
2501babe, line 137: hahaha i love the completionism

packages/addresses/src/vendor/noble/ed25519.ts
2501babe, line 1: noting for posterity that obvi i cant review this but clearly its third party code anyway

2501babe:
thank you for your service, ill try using this soon

2501babe:
incidentally are we going back to not having a synchronous version of this... i know its probably something like "blah blah web workers 0.0001s speedup" but its just convenient not having pda generation infect the entire codebase with async/await

packages/addresses/README.md
```diff
@@ -60,3 +60,24 @@ import { getBase58EncodedAddressFromPublicKey } from '@solana/addresses';
...
+
+```ts
+import { getBase58EncodedAddressCodec, getProgramDerivedAddress } from '@solana/addresses';
+
+const { serialize } = getBase58EncodedAddressCodec();
+const { bumpSeed, pda } = await getProgramDerivedAddress({
+    programAddress: 'ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL' as Base58EncodedAddress,
+    seeds: [
+        // Owner
+        serialize('9fYLFVoVqwH37C3dyPi6cpeobfbQ2jtLpN5HgAYDDdkm' as Base58EncodedAddress),
```
mcintyre94, line 76: Does this return the same bytes as `new PublicKey('9fYLFVoVqwH37C3dyPi6cpeobfbQ2jtLpN5HgAYDDdkm').toBuffer()` in the old lib?

packages/addresses/src/__tests__/program-derived-address-test.ts
```diff
@@ -0,0 +1,141 @@
...
+        await expect(
+            getProgramDerivedAddress({
+                programAddress: 'H8gBP21L5ietkHgXcGbgQBCVVEdPUQyuP9Q5MPRLLSJu' as Base58EncodedAddress,
+                seeds: ['\uD83D\uDE80'],
+            })
+        ).resolves.toStrictEqual({
+            bumpSeed: 251,
+            pda: '46v3JvPtEPeQmH3euXydEbxYD6yfxeZjWSzkkYvvM5Pp',
+        });
+    });
```
mcintyre94, line 122: Just out of interest did you use some trick to generate these examples? Mainly asking because they all have bumpSeed 251

packages/addresses/src/program-derived-address.ts
```diff
@@ -0,0 +1,78 @@
+import { assertDigestCapabilityIsAvailable } from '@solana/assertions';
+
+import { Base58EncodedAddress, getBase58EncodedAddressCodec } from './base58';
+import { compressedPointBytesAreOnCurve } from './curve';
+
+type PDAInput = Readonly<{
+    programAddress: Base58EncodedAddress;
+    seeds: Seed[];
+}>;
+type Seed = string | Uint8Array;
```
mcintyre94, line 10: Not suggesting we should keep the existing interface, but do you know how you'd convert from a `Buffer` in the old code to this?

I remember that when I used `Buffer.from([posX, posY])` (posX and posY are 0-255, u8 in rust) as a seed, I couldn't switch it to a `Uint8Array` and get the same result in the old code

I'm not sure if the different implementation here means that `Uint8Array([posX, posY])` would be equivalent?

mcintyre94 (approved):
Nice!

packages/addresses/LICENSE
steveluscher: IANAL.

packages/addresses/README.md
steveluscher: I'm into it. If we can agree that the commonly used thing (the base58 string) is what we mean when we say ‘address’ then we can rename all the methods for the common case, and reserve the more wordy function names for the more uncommon cases (`getAddressBytesFromFoo()` etc).

I'll still keep the TypeScript types the way they are though.

packages/addresses/src/vendor/noble/ed25519.ts
steveluscher: oic you're not familiar with my stacked PR workflow and you probably reviewed every commit in here. Each PR in this stack is only ‘about’ the top commit. The other commits before the last one are just therem because GitHub.

steveluscher: In other words, wow thanks for reviewing the whole stack! :)

steveluscher:
> are we going back to not having a synchronous version of this...

1. I'm pretty sure that if I did enough digging through GitHub issues I'd find a bunch of folks who were really mad about the move to sync versions of our key manipulation functions because their apps were key-heavy and sync versions kneecapped their performance severely.
2. We absolutely have no choice. Native Ed25519 primitives are coming to every JavaScript engine, and every single bit of those APIs is async.

packages/addresses/README.md
steveluscher: It does!

```ts
> new PublicKey('9fYLFVoVqwH37C3dyPi6cpeobfbQ2jtLpN5HgAYDDdkm').toBuffer()
<Buffer 80 bc fe 45 2a 9a 00 99 12 7e 5e de 0f 4f 69 98 01 18 0f a0 75 af c9 f5 df 28 ec 9b 28 c0 5d ba>
```

```ts
> Buffer.from(getBase58EncodedAddressCodec().serialize('9fYLFVoVqwH37C3dyPi6cpeobfbQ2jtLpN5HgAYDDdkm'))
<Buffer 80 bc fe 45 2a 9a 00 99 12 7e 5e de 0f 4f 69 98 01 18 0f a0 75 af c9 f5 df 28 ec 9b 28 c0 5d ba>
```

packages/addresses/src/__tests__/program-derived-address-test.ts
steveluscher: Generating a few addresses for the tests until exactly that is the case.

packages/addresses/src/program-derived-address.ts
steveluscher: It definitely is. `Buffer` has always been a pain in the ass because it's a Node-only primitive that has required a polyfill in browser, for almost no gain.

https://github.com/solana-labs/solana-web3.js/issues/1100

```ts
> Buffer.from([64, 216])
<Buffer 40 d8>
> new Uint8Array([64, 216])
Uint8Array(2) [ 64, 216 ]
> 0x40 === 64
true
> 0xd8 === 216
true
```

steveluscher: > I remember that when I used Buffer.from([posX, posY]) (posX and posY are 0-255, u8 in rust) as a seed, I couldn't switch it to a Uint8Array and get the same result in the old code

Very sus! We should try to repro that and figure out why.

PR #1457, opened 2023-07-28
Experimental RPC API

buffalojoec:
# JSON RPC API Methods Implemented in Experimental Web3 JS

Keeping this PR as a draft for now, until the rest of the JSON RPC methods are added to the experimental Web3 JS library.

## HTTP

- [x] `getAccountInfo`
  - [x] base
  - [x] parsed
- [x] `getBalance`
- [ ] `getBlock` (WIP)
- [ ] `getBlockCommitment`
  - [x] Method
  - [ ] Needs tests
- [x] `getBlockHeight`
- [x] `getBlockProduction`
- [x] `getBlocks`
- [ ] `getBlocksWithLimit`
  - [x] Method
  - [ ] Needs tests
- [x] `getBlockTime`
- [x] `getClusterNodes`
- [x] `getEpochInfo`
- [x] `getEpochSchedule`
- [x] `getFeeforMessage`
- [x] `getFirstAvailableBlock`
- [x] `getGenesisHash`
- [ ] `getHealth`
  - [x] Method 
  - [ ] Needs tests
- [ ] `getHighestSnapshotSlot`
  - [x] Method 
  - [ ] Needs tests
- [ ] `getIdentity`
  - [x] Method 
  - [ ] Needs tests
- [x] `getInflationGovernor`
- [x] `getInflationRate`
- [x] `getInflationReward`
- [x] `getLargestAccounts`
- [x] `getLatestBlockhash`
- [x] `getLeaderSchedule`
- [ ] `getMaxRetransmitSlot`
  - [x] Method 
  - [ ] Needs tests
- [ ] `getMaxShredInsertSlot` 
  - [x] Method
  - [ ] Needs tests
- [x] `getMinimumBalanceForRentExemption`
- [x] `getMultipleAccounts`
  - [x] base
  - [x] parsed
- [ ] `getProgramAccounts`
- [x] `getRecentPerformanceSamples`
- [x] `getRecentPrioritizationFees`
- [x] `getSignaturesForAddress`
- [x] `getSignatureStatuses`
- [x] `getSlot`
- [x] `getSlotLeader`
- [x] `getSlotLeaders`
- [x] `getStakeActivation`
- [x] `getStakeMinimumDelegation`
- [x] `getSupply`
- [x] `getTokenAccountBalance`
- [ ] `getTokenAccountsByDelegate`
  - [x] Method
  - [ ] Needs tests
- [x] `getTokenAccountsByOwner`
- [x] `getTokenLargestAccounts`
- [x] `getTokenSupply`
- [ ] `getTransaction`
  - [x] base
  - [ ] parsed
- [x] `getTransactionCount`
- [x] `getVersion`
- [x] `getVoteAccounts`
- [x] `isBlockhashValid`
- [x] `minimumLedgerSlot`
- [x] `requestAirdrop`
- [x] `sendTransaction`
- [ ] `simulateTransaction` (WIP)


buffalojoec:
This is ready for review, but I'm debugging that CI build & test failure. Everything works with `pnpm` locally for me!

The culprit is my import here:
https://github.com/buffalojoec/solana-web3.js/blob/ed91d6ccd4691bcca4a0c26a3b195e3e6df6ee90/packages/library-legacy/src/rpc.ts#L3-L7

I feel like I'm maybe missing some kind of config within `tsconfig.json` or similar, since I noticed you're doing a similar import here:
https://github.com/buffalojoec/solana-web3.js/blob/ed91d6ccd4691bcca4a0c26a3b195e3e6df6ee90/packages/library/src/rpc-transport.ts#L2

packages/library-legacy/src/connection.ts
```diff
@@ -3212,23 +3312,25 @@ export class Connection {
...
+      .send()
+      .then(unsafeRes =>
+        create(unsafeRes, jsonRpcResultAndContextExperimental(number())),
+      )
+      .catch(e => {
+        throw new SolanaJSONRPCError(
+          e,
+          `failed to get balance for ${publicKey.toBase58()}`,
+        );
+      });
```
steveluscher, line 3333: You can write this more like this:

```ts
try {
    const unsafeRes = await this._solanaRpc.getBalance(...params).send();
    return create(unsafeRes, jsonRpcResultAndContextExperimental(number()));
} catch (e) {
    throw new SolanaJSONRPCError(e, `failed to get balance for ${publicKey.toBase58()}`);
}
```

buffalojoec: This was kinda how they were written beforehand. Now that I'm looking at it, I'm assuming this would be better because it catches errors on the RPC request and response coercion, rather than just request?

steveluscher: Uhhhh… hmm. Good question. I guess you probably _just_ want to wrap the rpc request in the try/catch so as to minimize the amount of change here (response coercer would throw straight up in both cases)

```ts
let unsafeRes;
try {
  unsafeRes = await this._solanaRpc.getBalance(...params).send();
} catch (e) {
  throw new SolanaJSONRPCError(e, `failed to get balance for ${publicKey.toBase58()}`); 
}
return create(unsafeRes, jsonRpcResultAndContextExperimental(number()));
```

steveluscher: The _style_ is up to you (I prefer try/catch style, but the other is fine) but the way you had it written above, `SolanaJSONRPCError` will get thrown on _both_ RPC errors _and_ failures to do runtime typechecking in `create`, which diverges from the current behavior and was maybe hard to spot.

buffalojoec: Makes sense, thanks! I sort of realized after I asked what you were driving at

buffalojoec:
@steveluscher is this stitch work still part of the release plan?

steveluscher:
I think it's worth doing, but it's lower on my list of priorities as compared to ‘getting a major ecosystem project completely migrated.’

PR #1458, opened 2023-07-28
Export Commitment type to consumers

mcintyre94:
- We use `Commitment` as part of the input config for many calls, so it's useful to export it so that consumers can pass it around/create it in their code. Otherwise you need to hardcode a compliant string value in code for each RPC call
- It's not feasible to use `Parameters ` or similar to obtain this type because it's usually optional and nested in a config object

I've also checked the other inputs for RPC methods and I can't see anything else likely to cause issues.

Tested by compiling `@packages/rpc-core` and then seeing `import type { Commitment } from '@solana/rpc-core` in a dependent package no longer errors.

### DIFF
--- a/packages/rpc-core/src/rpc-methods/index.ts
+++ b/packages/rpc-core/src/rpc-methods/index.ts
@@ -58,6 +58,8 @@ export type SolanaRpcMethods = GetAccountInfoApi &
     IsBlockhashValidApi &
     MinimumLedgerSlotApi;
 
+export type { Commitment } from './common';
+
 export function createSolanaRpcApi(config?: Config): IRpcApi<SolanaRpcMethods> {
     return new Proxy({} as IRpcApi<SolanaRpcMethods>, {
         defineProperty() {


steveluscher (changes requested):
I _just_ want to point out that if you're a complete lunatic, it _is_ possible to use `Parameters` for this purpose :)

```ts
import { createDefaultRpcTransport, createSolanaRpc } from '@solana/web3.js';

const rpc = createSolanaRpc({
    transport: createDefaultRpcTransport({
        url: 'https://api.devnet.solana.com',
    }),
});

async function foo(
  commitment: NonNullable<Parameters<typeof rpc.getLatestBlockhash>[0]>['commitment'],
) {
    return await rpc.getLatestBlockhash({ commitment }).send();
}
```

In any case, let's do one or more of these things.

1. Now that there are more sensible places for these types to live, move them there ([see this](https://github.com/solana-labs/solana-web3.js/blob/083193b71741e38b294bc6f62dc55672db6c344b/packages/rpc-core/src/rpc-methods/common.ts#L1-L2))
2. In the target(s), create a `types.ts` file, move all the exportable stuff into there, and then export the whole thing from `index.ts` (`export * from './types'`)

steveluscher (approved):
Uh. I don't even know if I agree with what I wrote here:

https://github.com/solana-labs/solana-web3.js/blob/083193b71741e38b294bc6f62dc55672db6c344b/packages/rpc-core/src/rpc-methods/common.ts#L1-L2

That doesn't really make any sense. Commitment is an RPC thing.

Carry on. Stamp!

mcintyre94:
Oops yep I meant to comment on that saying I think it makes more sense in RPC too!

Issue #1449, opened 2023-07-27
JSON RPC HTTP Methods Missing from Experimental Lib


buffalojoec:
- [x] [`getBlock`](https://docs.solana.com/api/http#getblock)
- [x] [`getBlockCommitment`](https://docs.solana.com/api/http#getblockcommitment)
- [x] [`getBlocksWithLimit`](https://docs.solana.com/api/http#getblockswithlimit)
- [x] [`getClusterNodes`](https://docs.solana.com/api/http#getclusternodes)
- [x] [`getFeeforMessage`](https://docs.solana.com/api/http#getfeeformessage)
- [x] [`getGenesisHash`](https://docs.solana.com/api/http#getgenesishash)
- [x] [`getHealth`](https://docs.solana.com/api/http#gethealth)
- [x] [`getHighestSnapshotSlot`](https://docs.solana.com/api/http#gethighestsnapshotslot)
- [x] [`getIdentity`](https://docs.solana.com/api/http#getidentity)
- [x] [`getInflationGovernor`](https://docs.solana.com/api/http#getinflationgovernor)
- [x] [`getInflationRate`](https://docs.solana.com/api/http#getinflationrate)
- [x] [`getLargestAccounts`](https://docs.solana.com/api/http#getlargestaccounts)
- [x] [`getLeaderSchedule`](https://docs.solana.com/api/http#getleaderschedule)
- [x] [`getMinimumBalanceForRentExemption`](https://docs.solana.com/api/http#getminimumbalanceforrentexemption)
- [x] [`getMultipleAccounts`](https://docs.solana.com/api/http#getmultipleaccounts)
- [x] [`getProgramAccounts`](https://docs.solana.com/api/http#getprogramaccounts)
- [x] [`getRecentPrioritizationFees`](https://docs.solana.com/api/http#getrecentprioritizationfees)
- [x] [`getSignatureStatuses`](https://docs.solana.com/api/http#getsignaturestatuses)
- [x] [`getSlotLeader`](https://docs.solana.com/api/http#getslotleader)
- [x] [`getStakeActivation`](https://docs.solana.com/api/http#getstakeactivation)
- [x] [`getTokenAccountBalance`](https://docs.solana.com/api/http#gettokenaccountbalance)
- [x] [`getTokenAccountsByDelegate`](https://docs.solana.com/api/http#gettokenaccountsbydelegate)
[...]

PR #1461, opened 2023-07-28
refactor(experimental): Add the `getBlockCommitment` RPC method

buffalojoec:
This PR adds the `getBlockCommitment` RPC method to the new experimental Web3 JS's arsenal.

Ref #1449

### DIFF
--- a/packages/rpc-core/src/rpc-methods/__tests__/get-block-commitment-test.ts
+++ b/packages/rpc-core/src/rpc-methods/__tests__/get-block-commitment-test.ts
@@ -1,0 +1,33 @@
+import { createHttpTransport, createJsonRpc } from '@solana/rpc-transport';
+import type { Rpc } from '@solana/rpc-transport/dist/types/json-rpc-types';
+import fetchMock from 'jest-fetch-mock-fork';
+
+import { createSolanaRpcApi, SolanaRpcMethods } from '../index';
+
+describe('getBlockCommitment', () => {
+    let rpc: Rpc<SolanaRpcMethods>;
+    beforeEach(() => {
+        fetchMock.resetMocks();
+        fetchMock.dontMock();
+        rpc = createJsonRpc<SolanaRpcMethods>({
+            api: createSolanaRpcApi(),
+            transport: createHttpTransport({ url: 'http://127.0.0.1:8899' }),
+        });
+    });
+
+    it('returns the block commitment for an older block, which has null commitment', async () => {
+        expect.assertions(1);
+        const getBlockCommitmentPromise = rpc.getBlockCommitment(0n).send();
+        await expect(getBlockCommitmentPromise).resolves.toMatchObject({
+            commitment: null,
+            totalStake: expect.any(BigInt),
+        });
+    });
+
+    // TODO: We need a good way to feed `getBlockCommitment` a recent block.
+    // This would actually return a value for commitment.
+    // This is tricky to do without `getSlot`, and we'll need some kind
+    // of manipulation capability over test-validator to pull it off without
+    // another RPC call.
+    it.todo('returns the block commitment for a recent block');
+});
--- a/packages/rpc-core/src/rpc-methods/getBlockCommitment.ts
+++ b/packages/rpc-core/src/rpc-methods/getBlockCommitment.ts
@@ -1,0 +1,14 @@
+import { LamportsUnsafeBeyond2Pow53Minus1, Slot } from './common';
+
+type GetBlockCommitmentApiResponse = Readonly<{
+    commitment: LamportsUnsafeBeyond2Pow53Minus1[] | null;
+    totalStake: LamportsUnsafeBeyond2Pow53Minus1;
+}>;
+
+export interface GetBlockCommitmentApi {
+    /**
+     * Returns the amount of cluster stake in lamports that has voted on
+     * a particular block, as well as the stake attributed to each vote account
+     */
+    getBlockCommitment(slot: Slot): GetBlockCommitmentApiResponse;
+}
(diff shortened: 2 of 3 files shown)


packages/rpc-core/src/rpc-methods/__tests__/get-block-commitment-test.ts
```diff
@@ -0,0 +1,36 @@
...
+        fetchMock.dontMock();
+        rpc = createJsonRpc<SolanaRpcMethods>({
+            api: createSolanaRpcApi(),
+            transport: createHttpTransport({ url: 'http://127.0.0.1:8899' }),
+        });
+    });
+
+    it('returns the block commitment for an older block, which has null commitment', async () => {
+        expect.assertions(1);
+        const getBlockCommitmentPromise = rpc.getBlockCommitment(BigInt(0)).send();
```
steveluscher, line 20:
```suggestion
        const getBlockCommitmentPromise = rpc.getBlockCommitment(0n).send();
```

packages/rpc-core/src/rpc-methods/getBlockCommitment.ts
```diff
@@ -0,0 +1,13 @@
+import { Slot, U64UnsafeBeyond2Pow53Minus1 } from './common';
+
+type GetBlockCommitmentApiResponse = Readonly<{
+    commitment: U64UnsafeBeyond2Pow53Minus1[] | null;
```
steveluscher, line 4: The awful thing about this is that it will never, ever be right with the current RPC. These numbers are way larger than JS numbers.

```shell
 % curl https://api.mainnet-beta.solana.com -X POST -H "Content-Type: application/json" -d '                                                                                                   
  {
    "jsonrpc": "2.0", "id": 1,
    "method": "getBlockCommitment",
    "params":[208151977]
  }
'
{"jsonrpc":"2.0","result":{"commitment":[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,382219010785520358],"totalStake":383784289656803160},"id":1}
```

```
> console.log(382219010785520358)
382219010785520400
> console.log(383784289656803160)
383784289656803140
```

We need to deprecate this API as soon as possible and do https://github.com/solana-labs/solana/issues/30741.

steveluscher: Related: https://github.com/solana-labs/solana/issues/32662.

steveluscher, line 4:
```suggestion
    commitment: LamportsUnsafeBeyond2Pow53Minus1[] | null;
```

packages/rpc-core/src/rpc-methods/getBlockCommitment.ts
```diff
@@ -0,0 +1,13 @@
...
+    totalStake: U64UnsafeBeyond2Pow53Minus1;
```
steveluscher, line 5:
```suggestion
    totalStake: LamportsUnsafeBeyond2Pow53Minus1;
```

packages/rpc-core/src/rpc-methods/getBlockCommitment.ts
```diff
@@ -0,0 +1,13 @@
...
+}>;
+
+export interface GetBlockCommitmentApi {
+    /**
+     * Returns commitment for particular block
```
steveluscher, line 10: :/

packages/rpc-core/src/rpc-methods/__tests__/get-block-commitment-test.ts
```diff
@@ -0,0 +1,36 @@
...
+        await expect(getBlockCommitmentPromise).resolves.toMatchObject({
+            commitment: null,
+            totalStake: expect.any(BigInt),
+        });
+    });
+
+    it('returns the block commitment for a recent block', async () => {
+        expect.assertions(1);
+        const slot = await rpc.getSlot({ minContextSlot: 0n }).send();
```
steveluscher, line 29: Tests like this worry me because:

1. Now we're testing _two_ RPC methods in one test
2. If `getSlot` ever returned `0n`, then this test would fail

I would rather have _no_ test than a test that could fail transiently.

Until we have [this level of control](https://github.com/solana-labs/solana/blob/3cc97c75e67b710124cf5749476bef92f7004de8/core/src/replay_stage.rs#L4727-L4748) over the test validator, I think it might be better to just note the cases that we wish we could tee up consistently, but can't with `it.todo`.

```ts
it.todo('...')
```

steveluscher (approved):
thank-you

packages/rpc-core/src/rpc-methods/getBlockCommitment.ts
buffalojoec: Lol this was in the RPC docs. Want more info?

packages/rpc-core/src/rpc-methods/__tests__/get-block-commitment-test.ts
buffalojoec: Ooooh yeah that's nifty. Do we have some kind of roadmap to this level of control with JS?

steveluscher: I think we either need to break out our 🦀 and make some changes to the test validator itself, or we need to find a champion in core eng.

It's just never as important as other things.

packages/rpc-core/src/rpc-methods/getBlockCommitment.ts
steveluscher: ¯\_(°ペ)_/¯

PR #1465, opened 2023-07-31
fix: getRecentPrioritizationFees works with commitment specified

### LINKED ISSUE #1464: getRecentPrioritizationFees fails if commitment is specified
## Steps to reproduce

```
const connectionWithCommitment = new Connection('https://api.mainnet-beta.solana.com', 'confirmed')
const recentPrioritizationFees = await connectionWithCommitment.getRecentPrioritizationFees();
```

## Description of bug

When a commitment level is specified with this RPC call, the call fails. This is due to the `_buildArgs` method adding commitment as a parameter to the RPC request: 
https://github.com/solana-labs/solana-web3.js/blob/91615fce739a0f7a2e926581dfb55a0193ce34ae/packages/library-legacy/src/connection.ts#L4550.

This is problematic because the RPC only expects (at most) an array of pubkeys: https://docs.solana.com/api/http#getrecentprioritizationfees

Lou-Kamades:
closes #1464

### DIFF
--- a/packages/library-legacy/src/connection.ts
+++ b/packages/library-legacy/src/connection.ts
@@ -4547,7 +4547,7 @@ export class Connection {
     config?: GetRecentPrioritizationFeesConfig,
   ): Promise<RecentPrioritizationFees[]> {
     const accounts = config?.lockedWritableAccounts?.map(key => key.toBase58());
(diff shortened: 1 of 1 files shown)


steveluscher (approved):
giphy

PR #1476, opened 2023-08-02
refactor(experimental): Add the `getBlocksWithLimit` RPC method

buffalojoec:
This PR adds the `getBlocksWithLimit` RPC method to the new experimental Web3 JS's arsenal.

Ref #1449 

---

I will say adding this method pointed out some strange behavior with the RPC that I'm curious about:
- If you provide a valid slot number and a limit of `0`, you get an empty array
- If you provide a slot number greater than the current slot, you get an empty array

The behavior of getting an empty array rather than an error seems strange, no?

### DIFF
--- a/packages/rpc-core/src/rpc-methods/__tests__/get-blocks-with-limit-test.ts
+++ b/packages/rpc-core/src/rpc-methods/__tests__/get-blocks-with-limit-test.ts
@@ -1,101 +1,32 @@
-import { createHttpTransport, createJsonRpc } from '@solana/rpc-transport';
-import { SolanaJsonRpcErrorCode } from '@solana/rpc-transport/dist/types/json-rpc-errors';
-import type { Rpc } from '@solana/rpc-transport/dist/types/json-rpc-types';
-import fs from 'fs';
-import fetchMock from 'jest-fetch-mock-fork';
-import path from 'path';
-
-import { Commitment, createSolanaRpcApi, SolanaRpcMethods } from '../index';
-
-const logFilePath = path.resolve(__dirname, '../../../../../test-ledger/validator.log');
-const confirmedPattern = /My next leader slot is (\d+)/g;
-const finalizedPattern = /new root (\d+)/g;
-
-async function waitForBlock(slot: number, commitment: 'confirmed' | 'finalized') {
-    const pattern = commitment === 'confirmed' ? confirmedPattern : finalizedPattern;
-    let slotFound = false;
-    while (!slotFound) {
-        await new Promise(resolve => setTimeout(resolve, 1000));
-        const logs = fs.readFileSync(logFilePath, 'utf8');
-        let match;
-        let latestSlot: number;
-        while ((match = pattern.exec(logs)) !== null) {
-            latestSlot = Number(match[1]);
-            console.log(`Found slot ${latestSlot}`);
-            if (latestSlot >= slot) {
-                slotFound = true;
-                return;
-            }
-        }
-        await new Promise(resolve => setTimeout(resolve, (slot - latestSlot) * 500));
-    }
-}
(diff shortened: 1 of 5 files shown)


packages/rpc-core/src/rpc-methods/__tests__/get-blocks-with-limit-test.ts
```diff
@@ -0,0 +1,63 @@
...
+            api: createSolanaRpcApi(),
+            transport: createHttpTransport({ url: 'http://127.0.0.1:8899' }),
+        });
+    });
+
+    describe('when called with a valid `startSlot` and a valid `limit`', () => {
+        it('returns 8 blocks for a limit of 8', async () => {
+            expect.assertions(1);
+            const res = await rpc.getBlocksWithLimit(5n, 8).send();
+            expect(res).toHaveLength(8);
```
mcintyre94, line 22: Does this only work if there are >13 blocks at the point this test runs? Just wondering if it's potentially going to be flaky in CI if this test ends up running very soon after the test validator starts

packages/rpc-core/src/rpc-methods/__tests__/get-blocks-with-limit-test.ts
```diff
@@ -0,0 +1,63 @@
...
+            expect(res).toHaveLength(0);
+        });
+    });
+
+    describe('when called with a `limit` higher than 500,000', () => {
+        it('throws an error', async () => {
+            expect.assertions(1);
+            const sendPromise = rpc.getBlocksWithLimit(0n, 500_001).send();
+            await expect(sendPromise).rejects.toMatchObject({
+                code: -32602,
```
mcintyre94, line 57:
```suggestion
                code: -32602 satisfies (typeof SolanaJsonRpcErrorCode)['JSON_RPC_INVALID_PARAMS'],
```

Nitpicky but this is the pattern we've used elsewhere for these JSON RPC errors

mcintyre94 (approved):
Nice!

packages/rpc-core/src/rpc-methods/__tests__/get-blocks-with-limit-test.ts
buffalojoec: Ahh yeah good point. Could just add a sleep timer then?

buffalojoec:
@mcintyre94 Let's see if the commit I just pushed works with CI, lol.

It's not pretty, but it will accomplish what you mentioned

buffalojoec:
young-frankenstein-meme

packages/rpc-core/src/rpc-methods/__tests__/get-blocks-with-limit-test.ts
```diff
@@ -0,0 +1,87 @@
...
+        await new Promise(resolve => setTimeout(resolve, 1000));
+        const logs = fs.readFileSync(logFilePath, 'utf8');
+        let match;
+        while ((match = pattern.exec(logs)) !== null) {
+            if (Number(match[1]) >= slot) {
+                slotFound = true;
+                return;
+            }
+        }
+        await new Promise(resolve => setTimeout(resolve, 10000));
```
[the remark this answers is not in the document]
buffalojoec: nice touches. I read 800 ms in the docs, where did you find 400 ms?

I'll make those changes rn

buffalojoec:
Alright, @mcintyre94 this is probably the best we can do for both commitment types.

If we don't want the tests to wait 17 seconds for a finalized block, then we can make these `todo` for now.

buffalojoec:
> Hmm for 17s I think we should just TODO it for now. Definitely starting to find a lot of RPC methods that we need more control over the validator for.

Agreed. Done!

PR #1478, opened 2023-08-03
refactor(experimental): add `getGenesisHash` API method

buffalojoec:
This PR adds the `getGenesisHash` RPC method to the new experimental Web3 JS's arsenal.

Ref #1449

### DIFF
--- a/packages/rpc-core/src/rpc-methods/__tests__/get-genesis-hash-test.ts
+++ b/packages/rpc-core/src/rpc-methods/__tests__/get-genesis-hash-test.ts
@@ -1,0 +1,41 @@
+import { Base58EncodedAddress } from '@solana/addresses';
+import { createHttpTransport, createJsonRpc } from '@solana/rpc-transport';
+import type { Rpc } from '@solana/rpc-transport/dist/types/json-rpc-types';
+import fetchMock from 'jest-fetch-mock-fork';
+
+import { createSolanaRpcApi, SolanaRpcMethods } from '../index';
+
+// TODO: We may want to replace the higher clusters with just the local
+// validator, since this is within the live test suite anyhow.
+// Also, with further manipulation capabilities over the local validator,
+// we can explore error testing as well
+const genesisHashMap: [string, Base58EncodedAddress][] = [
+    ['https://api.devnet.solana.com', 'EtWTRABZaYq6iMfeYKouRu166VU2xqa1wcaWoxPkrZBG' as Base58EncodedAddress],
+    ['https://api.testnet.solana.com', '4uhcVJyU9pJkvQyS88uRDiswHXSCkY3zQawwpjk2NsNY' as Base58EncodedAddress],
+    ['https://api.mainnet-beta.solana.com', '5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d' as Base58EncodedAddress],
+];
+
+describe('getGenesisHash', () => {
+    let rpc: Rpc<SolanaRpcMethods>;
+    const createRpcForTest = (url: string) => {
+        rpc = createJsonRpc<SolanaRpcMethods>({
+            api: createSolanaRpcApi(),
+            transport: createHttpTransport({ url }),
+        });
+    };
+    beforeEach(() => {
+        fetchMock.resetMocks();
+        fetchMock.dontMock();
+    });
+
+    genesisHashMap.forEach(([url, expectedGenesisHash]) => {
+        describe(`when sent to ${url}`, () => {
+            it(`returns ${expectedGenesisHash}`, async () => {
+                expect.assertions(1);
+                createRpcForTest(url);
+                const genesisHashPromise = rpc.getGenesisHash().send();
+                await expect(genesisHashPromise).resolves.toBe(expectedGenesisHash);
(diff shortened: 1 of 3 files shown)


packages/rpc-core/src/rpc-methods/__tests__/get-genesis-hash-test.ts
```diff
@@ -0,0 +1,41 @@
...
+
+// TODO: We may want to replace the higher clusters with just the local
+// validator, since this is within the live test suite anyhow.
+// Also, with further manipulation capabilities over the local validator,
+// we can explore error testing as well
+const genesisHashMap: [string, Base58EncodedAddress][] = [
+    ['https://api.devnet.solana.com', 'EtWTRABZaYq6iMfeYKouRu166VU2xqa1wcaWoxPkrZBG' as Base58EncodedAddress],
+    ['https://api.testnet.solana.com', '4uhcVJyU9pJkvQyS88uRDiswHXSCkY3zQawwpjk2NsNY' as Base58EncodedAddress],
+    ['https://api.mainnet-beta.solana.com', '5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d' as Base58EncodedAddress],
+];
```
mcintyre94, line 16: I'm a bit hesitant about hitting live RPC nodes in CI tests, but at the same time I don't think we can test this with the current local validator. I think I'd lean toward just having a `todo` test for now.

mcintyre94 (approved):

packages/rpc-core/src/rpc-methods/__tests__/get-genesis-hash-test.ts
buffalojoec: Let me see if this one idea works, if not I'll make it a `todo`

buffalojoec: That should do it!

packages/rpc-core/src/rpc-methods/__tests__/get-genesis-hash-test.ts
```diff
@@ -0,0 +1,32 @@
...
+        rpc = createJsonRpc<SolanaRpcMethods>({
+            api: createSolanaRpcApi(),
+            transport: createHttpTransport({ url: 'http://127.0.0.1:8899' }),
+        });
+    });
+
+    describe('when sent to a local validator', () => {
+        it('returns the genesis hash', async () => {
+            expect.assertions(1);
+            const logFile = fs.readFileSync(logFilePath, 'utf-8');
```
steveluscher, line 26: Ooh. Interesting.

One downside of this is that it will read the whole log into memory. The later this test runs, the larger that logfile is.

Alternatively, you could read lines and read _until_ you find the genesis hash.

```ts
import { open } from 'node:fs/promises';

const file = await open('./some/file/to/read');

let expectedGenesisHash;
for await (const line of file.readLines()) {
  const match = line.match(genesisHashPattern);
  if (match) {
    expectedGenesisHash = match[1];
    break;
  }
}
if (expectedGenesisHash === undefined) {
  throw new Error('...');
}
```

packages/rpc-core/src/rpc-methods/getGenesisHash.ts
```diff
@@ -0,0 +1,13 @@
+import { Base58EncodedAddress } from '@solana/addresses';
+
+type GetGenesisHashApiResponse = Base58EncodedAddress;
```
steveluscher, line 3: We should probably have a different type for this. It's not actually an address, so we don't want values of this kind to typecheck in places where an address is expected.

packages/rpc-core/src/rpc-methods/__tests__/get-genesis-hash-test.ts
buffalojoec: Hey! @mcintyre94 actually re-wrote this to use chunked reads for this reason! #1521

packages/rpc-core/src/rpc-methods/getGenesisHash.ts
buffalojoec: Would this be a blockhash instead? Genesis has is a blockhash right?

packages/rpc-core/src/rpc-methods/__tests__/get-genesis-hash-test.ts
buffalojoec: Hmm I guess we could trim it down to lines though if we want to! Nice touch

mcintyre94: Yep we hit this! https://github.com/solana-labs/solana-web3.js/pull/1521/files
I used a stream and it’s a bit more code than your example there, but similar idea

steveluscher: #1535

packages/rpc-core/src/rpc-methods/getGenesisHash.ts
steveluscher: Oh, maybe! I guess that's it.

PR #1479, opened 2023-08-03
refactor(experimental): add `getHighestSnapshotSlot` API method

buffalojoec:
This PR adds the `getHighestSnapshotSlot` RPC method to the new experimental Web3 JS's arsenal.

Ref #1449

### DIFF
--- a/packages/rpc-core/src/rpc-methods/__tests__/get-highest-snapshot-slot-test.ts
+++ b/packages/rpc-core/src/rpc-methods/__tests__/get-highest-snapshot-slot-test.ts
@@ -1,0 +1,14 @@
+// TODO: Again, until we can manipulate the local validator more,
+// this test can't be implemented properly.
+describe('getHighestSnapshotSlot', () => {
+    describe('when there is at least one valid snapshot', () => {
+        it.todo('returns the highest snapshot slot information that the node has snapshots for');
+    });
+
+    describe('when there are no snapshots yet', () => {
+        // TODO:
+        //  Error code -32008
+        //  Message: "No snapshot"
+        it.todo('throws an error');
+    });
+});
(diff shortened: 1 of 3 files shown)


packages/rpc-core/src/rpc-methods/getHighestSnapshotSlot.ts
```diff
@@ -0,0 +1,20 @@
+import { Slot } from './common';
+
+type GetHighestSnapshotSlotApiResponse = Readonly<{
+    full: Slot;
+    incremental: Slot | null;
+}>;
```
mcintyre94, line 6: Looks like this is correct, but it might be worth PRing a fix to the docs which have it as `incremental: <u64|undefined>`

packages/rpc-core/src/rpc-methods/__tests__/get-highest-snapshot-slot-test.ts
```diff
@@ -0,0 +1,7 @@
+describe('getHighestSnapshotSlot', () => {
+    // TODO: Again, until we can manipulate the local validator more,
+    // this test can't be implemented properly.
+    // Here we'll likely want to know the last epoch and a handful of recent
+    // slots and maybe compare against that?
+    it.todo('returns the highest snapshot slot information that the node has snapshots for');
```
mcintyre94, line 6: Would also be good to test the no snapshot case when we can add tests for this

mcintyre94 (approved):

packages/rpc-core/src/rpc-methods/getHighestSnapshotSlot.ts
buffalojoec: I actually meant to do that shortly!

PR #1483, opened 2023-08-03
refactor(experimental): add getMinimumBalanceForRentExemption

2501babe:
the test is kind of perfunctory im not sure what else wed want to look at tho

### DIFF
--- a/packages/rpc-core/src/rpc-methods/__tests__/get-minimum-balance-for-rent-exemption-test.ts
+++ b/packages/rpc-core/src/rpc-methods/__tests__/get-minimum-balance-for-rent-exemption-test.ts
@@ -20,7 +20,7 @@ describe('getMinimumBalanceForRentExemption', () => {
         describe(`when called with \`${commitment}\` commitment`, () => {
             it('returns an expected rent amount', async () => {
                 expect.assertions(1);
-                const result = await rpc.getMinimumBalanceForRentExemption(BigInt(0)).send();
+                const result = await rpc.getMinimumBalanceForRentExemption(BigInt(0), { commitment }).send();
                 expect(result).toEqual(BigInt(890880));
             });
         });
--- a/packages/rpc-core/src/rpc-methods/__tests__/get-minimum-balance-for-rent-exemption-test.ts
+++ b/packages/rpc-core/src/rpc-methods/__tests__/get-minimum-balance-for-rent-exemption-test.ts
@@ -1,0 +1,28 @@
+import { createHttpTransport, createJsonRpc } from '@solana/rpc-transport';
+import type { Rpc } from '@solana/rpc-transport/dist/types/json-rpc-types';
+import fetchMock from 'jest-fetch-mock-fork';
+
+import { Commitment } from '../common';
+import { createSolanaRpcApi, SolanaRpcMethods } from '../index';
+
+describe('getMinimumBalanceForRentExemption', () => {
+    let rpc: Rpc<SolanaRpcMethods>;
+    beforeEach(() => {
+        fetchMock.resetMocks();
+        fetchMock.dontMock();
+        rpc = createJsonRpc<SolanaRpcMethods>({
+            api: createSolanaRpcApi(),
+            transport: createHttpTransport({ url: 'http://127.0.0.1:8899' }),
+        });
+    });
+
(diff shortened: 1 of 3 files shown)


packages/rpc-core/src/rpc-methods/getMinimumBalanceForRentExemption.ts
```diff
@@ -0,0 +1,15 @@
+import { Commitment, LamportsUnsafeBeyond2Pow53Minus1, RpcResponse, U64UnsafeBeyond2Pow53Minus1 } from './common';
+
+type GetMinimumBalanceForRentExemptionApiResponse = RpcResponse<LamportsUnsafeBeyond2Pow53Minus1>;
+
+export interface GetMinimumBalanceForRentExemptionApi {
+    /**
+     * Returns the minimum balance to exempt an account of a certain size from rent
+     */
+    getMinimumBalanceForRentExemption(
+        size: U64UnsafeBeyond2Pow53Minus1, // XXX technically usize, idk how fussy we want to be
```
2501babe, line 10: pls note

packages/rpc-core/src/rpc-methods/getMinimumBalanceForRentExemption.ts
```diff
@@ -0,0 +1,15 @@
+import { Commitment, LamportsUnsafeBeyond2Pow53Minus1, RpcResponse, U64UnsafeBeyond2Pow53Minus1 } from './common';
+
+type GetMinimumBalanceForRentExemptionApiResponse = RpcResponse<LamportsUnsafeBeyond2Pow53Minus1>;
+
+export interface GetMinimumBalanceForRentExemptionApi {
+    /**
+     * Returns the minimum balance to exempt an account of a certain size from rent
+     */
+    getMinimumBalanceForRentExemption(
+        size: U64UnsafeBeyond2Pow53Minus1,
```
2501babe, line 10: note this type is technically uszie but i dunno if we care about that distinction in js, in practice it should always be equivalent to u64. also note the rpc docs indicate its optional but that seems silly

2501babe:
@steveluscher weirdly for some reason i dont see the gear that allows me to request a reviewer?

packages/rpc-core/src/rpc-methods/getMinimumBalanceForRentExemption.ts
buffalojoec: Actually it will throw you an error if you provide no args, so we should prob throw a flare up on the JSON RPC docs. 

This is one of 3 mistakes we've found in the docs (that I know about). I think I'll just collect this info and make a PR

buffalojoec (approved):

buffalojoec:
lgtm!!

packages/rpc-core/src/rpc-methods/getMinimumBalanceForRentExemption.ts
mcintyre94: We've been using the U64/bigint types for usize so this is right. And yea agreed looks like an error in the RPC docs

packages/rpc-core/src/rpc-methods/__tests__/get-minimum-balance-for-rent-exemption-test.ts
```diff
@@ -0,0 +1,28 @@
...
+            api: createSolanaRpcApi(),
+            transport: createHttpTransport({ url: 'http://127.0.0.1:8899' }),
+        });
+    });
+
+    (['confirmed', 'finalized', 'processed'] as Commitment[]).forEach(commitment => {
+        describe(`when called with \`${commitment}\` commitment`, () => {
+            it('returns an expected rent amount', async () => {
+                expect.assertions(1);
+                const result = await rpc.getMinimumBalanceForRentExemption(BigInt(0)).send();
```
mcintyre94, line 23:
```suggestion
                const result = await rpc.getMinimumBalanceForRentExemption(BigInt(0), { commitment }).send();
```

buffalojoec: @2501babe If you want to commit this change we can merge this in!

2501babe: ah my mistake, thank you!

PR #1484, opened 2023-08-03
Revert "refactor(experimental): Add the `getBlockCommitment` RPC method"

buffalojoec:
Reverts solana-labs/solana-web3.js#1461


mcintyre94:
Can we just change the test to a TODO and keep the RPC method?

buffalojoec:
> Can we just change the test to a TODO and keep the RPC method? 

Yeah that was my plan. Is it easier to just do that in this one commit you think?

mcintyre94:
I think so, should just be a single file changed then. Whatever's easiest though!

buffalojoec:
Closing in favor of #1493

PR #1485, opened 2023-08-03
refactor(experimental): add `getIdentity` API method

buffalojoec:
This PR adds the `getIdentity` RPC method to the new experimental Web3 JS's arsenal.

Ref #1449

### DIFF
--- a/packages/rpc-core/tsconfig.json
+++ b/packages/rpc-core/tsconfig.json
@@ -1,7 +1,8 @@
 {
     "$schema": "https://json.schemastore.org/tsconfig",
     "compilerOptions": {
-        "lib": ["DOM", "ES2017", "ES2020.BigInt", "ES2022.Error"]
+        "lib": ["DOM", "ES2017", "ES2020.BigInt", "ES2022.Error"],
+        "resolveJsonModule": true
     },
     "display": "@solana/rpc-core",
     "extends": "tsconfig/base.json",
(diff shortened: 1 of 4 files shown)


mcintyre94 (approved):
Looks good! Nice that we can test this one too

packages/rpc-core/tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
     "$schema": "https://json.schemastore.org/tsconfig",
     "compilerOptions": {
-        "lib": ["DOM", "ES2017", "ES2020.BigInt", "ES2022.Error"]
+        "lib": ["DOM", "ES2017", "ES2020.BigInt", "ES2022.Error"],
+        "resolveJsonModule": true
```
buffalojoec, line 5: @mcintyre94 is adding this to `rpc-core`'s `tsconfig.json` an issue?

mcintyre94: I assume it's required for importing that JSON keypair? Honestly I don't know but I can't see any harm, or any impact on consumers, looks like it just affects our TS compilation.

buffalojoec: Me too but I didn't want it to somehow affect bundle size or anything related. I tried adding it to the `tsconfig.json` within the test-configs dir but that didn't seem to solve it

buffalojoec: I don't see it anywhere in the bundle and from my initial research it appears it only adds JSON modules to the bundle if they're imported in TS code that is also added to the bundle.

So, types and tests really shouldn't affect anything if pulled from JSON

mcintyre94: That makes sense to me, I wouldn't expect it to make any difference in code that isn't importing JSON

PR #1486, opened 2023-08-04
refactor(experimental): add `getInflationGovernor` API method

buffalojoec:
This PR adds the `getInflationGovernor` RPC method to the new experimental Web3 JS's arsenal.

Ref #1449


packages/rpc-core/src/rpc-methods/getInflationGovernor.ts
```diff
@@ -0,0 +1,30 @@
+import { Commitment, U64UnsafeBeyond2Pow53Minus1 } from './common';
+
+type GetInflationGovernorApiResponse = Readonly<{
+    /** The initial inflation percentage from time 0 */
+    initial: number; // Until we land on best type for `f64`
+    /** Percentage of total inflation allocated to the foundation */
+    foundation: number; // Until we land on best type for `f64`
+    /** Duration of foundation pool inflation in years */
+    foundationTerm: U64UnsafeBeyond2Pow53Minus1; // Until we land on best type for `f64`
```
mcintyre94, line 9: I'd just use `number` here too TBH. If it's f64 on the backend then we should support an f64 (as best as we can) even if the public networks always use an int.

packages/rpc-core/src/rpc-methods/__tests__/get-inflation-governor-test.ts
```diff
@@ -0,0 +1,34 @@
...
+    let rpc: Rpc<SolanaRpcMethods>;
+    beforeEach(() => {
+        fetchMock.resetMocks();
+        fetchMock.dontMock();
+        rpc = createJsonRpc<SolanaRpcMethods>({
+            api: createSolanaRpcApi(),
+            transport: createHttpTransport({ url: 'http://127.0.0.1:8899' }),
+        });
+    });
+    // TODO: I honestly have no clue how to test this
```
mcintyre94, line 18: I think this is fine TBH! If it returns the right structure in CI then that's good enough

mcintyre94 (changes requested):
We should add these Number fields to the not-bigint list: https://github.com/solana-labs/solana-web3.js/blob/master/packages/rpc-core/src/response-patcher-allowed-numeric-values.ts 

The reason you're getting `Number` currently is because they're not integers, and [there's a check](https://github.com/solana-labs/solana-web3.js/blob/f8b14a795a73c9c2b83a3a2132db2a6baeb234fa/packages/rpc-core/src/response-patcher.ts#L35) that protects against trying to parse a non-integer as BigInt, which would error. But if one of these fields was an integer you'd be getting `BigInt`. To fix that you can add them to this list and then they'll be kept as Number no matter what.

packages/rpc-core/src/rpc-methods/getInflationGovernor.ts
buffalojoec: I forgot to remove that comment but that one is `u64`!

buffalojoec: nvm I must have misread it, it's `f64`. Thought it was `u64`!

buffalojoec:
> We should add these Number fields to the not-bigint list: https://github.com/solana-labs/solana-web3.js/blob/master/packages/rpc-core/src/response-patcher-allowed-numeric-values.ts

OK I've added them to the list, but it's worth noting these are `f64` which `Number` is also not going to accurately represent. Considering the comment above the list of exceptions:

```
These are values that are legitimately defined as `u8` or `usize` on the backend.
```

I'm assuming this is only temporary until we find a library that satisfies #1488 (which I meant to link out to here)

mcintyre94 (approved):

packages/rpc-core/src/rpc-methods/getInflationGovernor.ts
```diff
@@ -0,0 +1,30 @@
...
+    /** Duration of foundation pool inflation in years */
+    foundationTerm: number; // Until we land on best type for `f64`
+    /**
+     * Rate per year at which inflation is lowered.
+     * (Rate reduction is derived using the target
+     * slot time in genesis config)
+     */
+    taper: number; // Until we land on best type for `f64`
+    /** Terminal inflation percentage */
+    terminal: number; // Until we land on best type for `f64`
```
steveluscher, line 17: Please do make a type for these like `F64Unsafe`, move the comment there, link to the GitHub issue, and maybe add a link to this: https://stackoverflow.com/a/57225494/802047

PR #1487, opened 2023-08-04
refactor(experimental): add `getInflationRate` API method

buffalojoec:
This PR adds the `getInflationRate` RPC method to the new experimental Web3 JS's arsenal.

Ref #1449 

---

Hyper-relevant to this PR is #1488. You can see I'm using number in place of what would be an ideal floating-point number handler for `f64`

### DIFF
--- a/packages/rpc-core/src/response-patcher-allowed-numeric-values.ts
+++ b/packages/rpc-core/src/response-patcher-allowed-numeric-values.ts
@@ -60,6 +60,7 @@ export const ALLOWED_NUMERIC_KEYPATHS: Partial<
     ],
     getBlockTime: [[]],
     getInflationGovernor: [['initial'], ['foundation'], ['foundationTerm'], ['taper'], ['terminal']],
+    getInflationRate: [['foundation'], ['total'], ['validator']],
     getInflationReward: [[KEYPATH_WILDCARD, 'commission']],
     getRecentPerformanceSamples: [[KEYPATH_WILDCARD, 'samplePeriodSecs']],
     getTokenLargestAccounts: [
--- a/packages/rpc-core/src/rpc-methods/__tests__/get-inflation-rate-test.ts
+++ b/packages/rpc-core/src/rpc-methods/__tests__/get-inflation-rate-test.ts
@@ -1,0 +1,30 @@
+import { createHttpTransport, createJsonRpc } from '@solana/rpc-transport';
+import type { Rpc } from '@solana/rpc-transport/dist/types/json-rpc-types';
+import fetchMock from 'jest-fetch-mock-fork';
+
+import { createSolanaRpcApi, SolanaRpcMethods } from '../index';
+
+describe('getInflationRate', () => {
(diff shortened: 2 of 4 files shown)


packages/rpc-core/src/rpc-methods/__tests__/get-inflation-rate-test.ts
```diff
@@ -0,0 +1,30 @@
...
+    let rpc: Rpc<SolanaRpcMethods>;
+    beforeEach(() => {
+        fetchMock.resetMocks();
+        fetchMock.dontMock();
+        rpc = createJsonRpc<SolanaRpcMethods>({
+            api: createSolanaRpcApi(),
+            transport: createHttpTransport({ url: 'http://127.0.0.1:8899' }),
+        });
+    });
+    // TODO: I honestly have no clue how to test this
```
mcintyre94, line 17: Same comment as your other PR - I think this is all we need for the test

mcintyre94 (changes requested):
As with #1486 we should add these to the number parsing exceptions to make sure they always convert to Number and not BigInt: https://github.com/solana-labs/solana-web3.js/blob/master/packages/rpc-core/src/response-patcher-allowed-numeric-values.ts

buffalojoec:
> As with #1486 we should add these to the number parsing exceptions to make sure they always convert to Number and not BigInt: https://github.com/solana-labs/solana-web3.js/blob/master/packages/rpc-core/src/response-patcher-allowed-numeric-values.ts

Done!

packages/rpc-core/src/response-patcher-allowed-numeric-values.ts
```diff
@@ -59,6 +59,7 @@ export const ALLOWED_NUMERIC_KEYPATHS: Partial<
         ['value', 'data', 'parsed', 'info', 'votes', KEYPATH_WILDCARD, 'confirmationCount'],
     ],
     getBlockTime: [[]],
+    getInflationRate: [['foundation', 'total', 'validator']],
```
mcintyre94, line 62:
```suggestion
    getInflationRate: [['foundation'], ['total'], ['validator']],
```

I think you want this since each field is top-level

buffalojoec: Good catch. Done

mcintyre94 (approved):

packages/rpc-core/src/rpc-methods/getInflationRate.ts
```diff
@@ -0,0 +1,22 @@
+import { U64UnsafeBeyond2Pow53Minus1 } from './common';
+
+type GetInflationRateApiResponse = Readonly<{
+    /** Epoch for which these values are valid */
+    epoch: U64UnsafeBeyond2Pow53Minus1;
+    /** Inflation allocated to the foundation */
+    foundation: number; // Until we land on best type for `f64`
```
steveluscher, line 7: Related suggestion: https://github.com/solana-labs/solana-web3.js/pull/1486#discussion_r1300533763

PR #1492, opened 2023-08-04
refactor(experimental): add `getMultipleAccounts` API method 

mcintyre94:
This PR adds `getMultipleAccounts` to experimental web3js. The API code, numeric patcher exceptions and tests are based on `getAccountInfo`

Note that the change to `get-account-info-tests.ts` is just formatting: moving more tests into the when called with jsonParsed encoding > for an account with parse-able JSON data block 

Ref #1449

### DIFF
--- a/packages/rpc-core/src/response-patcher-allowed-numeric-values.ts
+++ b/packages/rpc-core/src/response-patcher-allowed-numeric-values.ts
@@ -66,6 +66,67 @@ export const ALLOWED_NUMERIC_KEYPATHS: Partial<
     getInflationGovernor: [['initial'], ['foundation'], ['foundationTerm'], ['taper'], ['terminal']],
     getInflationRate: [['foundation'], ['total'], ['validator']],
     getInflationReward: [[KEYPATH_WILDCARD, 'commission']],
+    getMultipleAccounts: [
+        // parsed AddressTableLookup account
+        ['value', KEYPATH_WILDCARD, 'data', 'parsed', 'info', 'lastExtendedSlotStartIndex'],
+        // parsed Config account
+        ['value', KEYPATH_WILDCARD, 'data', 'parsed', 'info', 'slashPenalty'],
+        ['value', KEYPATH_WILDCARD, 'data', 'parsed', 'info', 'warmupCooldownRate'],
+        // parsed Token/Token22 token account
+        ['value', KEYPATH_WILDCARD, 'data', 'parsed', 'info', 'tokenAmount', 'decimals'],
+        ['value', KEYPATH_WILDCARD, 'data', 'parsed', 'info', 'tokenAmount', 'uiAmount'],
+        ['value', KEYPATH_WILDCARD, 'data', 'parsed', 'info', 'rentExemptReserve', 'decimals'],
+        ['value', KEYPATH_WILDCARD, 'data', 'parsed', 'info', 'delegatedAmount', 'decimals'],
+        [
+            'value',
+            KEYPATH_WILDCARD,
+            'data',
+            'parsed',
+            'info',
+            'extensions',
+            KEYPATH_WILDCARD,
+            'state',
+            'olderTransferFee',
+            'transferFeeBasisPoints',
+        ],
+        [
+            'value',
+            KEYPATH_WILDCARD,
+            'data',
+            'parsed',
+            'info',
+            'extensions',
+            KEYPATH_WILDCARD,
+            'state',
+            'newerTransferFee',
+            'transferFeeBasisPoints',
+        ],
+        [
+            'value',
+            KEYPATH_WILDCARD,
+            'data',
+            'parsed',
+            'info',
+            'extensions',
+            KEYPATH_WILDCARD,
+            'state',
+            'preUpdateAverageRate',
+        ],
+        ['value', KEYPATH_WILDCARD, 'data', 'parsed', 'info', 'extensions', KEYPATH_WILDCARD, 'state', 'currentRate'],
+        // parsed Token/Token22 mint account
+        ['value', KEYPATH_WILDCARD, 'data', 'parsed', 'info', 'decimals'],
+        // parsed Token/Token22 multisig account
+        ['value', KEYPATH_WILDCARD, 'data', 'parsed', 'info', 'numRequiredSigners'],
+        ['value', KEYPATH_WILDCARD, 'data', 'parsed', 'info', 'numValidSigners'],
+        // parsed Stake account
+        ['value', KEYPATH_WILDCARD, 'data', 'parsed', 'info', 'stake', 'delegation', 'warmupCooldownRate'],
+        // parsed Sysvar rent account
+        ['value', KEYPATH_WILDCARD, 'data', 'parsed', 'info', 'exemptionThreshold'],
+        ['value', KEYPATH_WILDCARD, 'data', 'parsed', 'info', 'burnPercent'],
+        // parsed Vote account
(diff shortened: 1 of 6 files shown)


packages/rpc-core/src/rpc-methods/getMultipleAccounts.ts
```diff
@@ -0,0 +1,113 @@
...
+    lamports: LamportsUnsafeBeyond2Pow53Minus1;
+    /** pubkey of the program this account has been assigned to */
+    owner: Base58EncodedAddress;
+    /** the epoch at which this account will next owe rent */
+    rentEpoch: U64UnsafeBeyond2Pow53Minus1;
+    /** the data size of the account */
+    space: U64UnsafeBeyond2Pow53Minus1;
+} | null;
+
+type GetMultipleAccountsApiResponseWithBase58EncodedData_DEPRECATED = Readonly<{
```
buffalojoec, line 28: Do we need a `WithDefaultData` payload as well, like `getAccountInfo`?

```typescript
type GetAccountInfoApiResponseWithDefaultData = Readonly<{
    value: Readonly<{
        data: Base58EncodedBytes;
    }> | null;
}>;
```

packages/rpc-core/src/rpc-methods/__tests__/get-multiple-accounts-test.ts
```diff
@@ -0,0 +1,591 @@
...
+                expect(multipleAccounts.value).toHaveLength(2);
+                multipleAccounts.value.forEach(accountInfo => {
+                    expect(accountInfo).toMatchObject({
+                        data: expect.any(Array),
+                        executable: expect.any(Boolean),
+                        lamports: expect.any(BigInt),
+                        owner: expect.any(String),
+                        rentEpoch: expect.any(BigInt),
+                        space: expect.any(BigInt),
+                    });
```
buffalojoec, line 49: Do we want to verify the account's state considering we have it stored as a fixture? ie. validating the proper lamports, owner, data, etc.

buffalojoec (commented):
lgtm, just a few comments/questions

packages/rpc-core/src/rpc-methods/getMultipleAccounts.ts
```diff
@@ -0,0 +1,113 @@
...
+        // account the `data` field falls back to `base64`.
+        | Base64EncodedDataResponse;
+}>;
+
+type GetMultipleAccountsApiCommonConfig = Readonly<{
+    /** Defaults to `finalized` */
+    commitment?: Commitment;
+    /** The minimum slot that the request can be evaluated at */
+    minContextSlot?: Slot;
+}>;
```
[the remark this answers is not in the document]
mcintyre94: Maybe! I deliberately avoided trying to be clever with this PR because I noticed that `getProgramAccounts` introduces the exciting quirk that it can either return things nested in an RPC Response (like `getAccountInfo` and `getMultipleAccounts` do).. or not, depending on the inputs!

So once that is in I think it'd make sense to revisit this and see if there are actually common data structures that make sense to extract, or if the combined weirdness of the 3 RPC calls makes it easier to keep them separate

packages/rpc-core/src/rpc-methods/__tests__/get-multiple-accounts-test.ts
mcintyre94: We could do but have generally been pretty lazy with this in the other tests. I think if we did this we'd just be testing the data flow from fixtures into RPC responses really, I'm not sure we'd really improve coverage of our own code. I haven't thought about it too much though so not too sure.

packages/rpc-core/src/rpc-methods/getMultipleAccounts.ts
buffalojoec: omg ok wow

packages/rpc-core/src/rpc-methods/__tests__/get-multiple-accounts-test.ts
buffalojoec: I'll leave it up to you for this one! Since we have to comb through all of the tests at some point, prob not a big deal

buffalojoec (approved):

packages/rpc-core/src/response-patcher-allowed-numeric-values.ts
```diff
@@ -66,6 +66,67 @@ export const ALLOWED_NUMERIC_KEYPATHS: Partial<
     getInflationGovernor: [['initial'], ['foundation'], ['foundationTerm'], ['taper'], ['terminal']],
     getInflationRate: [['foundation'], ['total'], ['validator']],
     getInflationReward: [[KEYPATH_WILDCARD, 'commission']],
+    getMultipleAccounts: [
```
steveluscher, line 69: 😬

One of these two things is true:

1. gzip will crush this and the next developer who comes along will notice that there are two places to update/add account parsers
2. we should consider making this a memoized `getAllowedNumericKeypaths()` method so we can run some logic like:
    ```ts
    const accountParserConfigs = [...];
    return {
        getAccountInfo: accountParserConfigs.map(c => (['value', ...c])),
        getMultipleAccounts: accountParserConfigs.map(c => (['value', KEYPATH_WILDCARD, ...c])),
    }
    ```

packages/rpc-core/src/rpc-methods/getMultipleAccounts.ts
```diff
@@ -0,0 +1,113 @@
...
+     */
+    getMultipleAccounts(
+        /** An array of up to 100 Pubkeys to query */
+        addresses: Base58EncodedAddress[],
+        config: GetMultipleAccountsApiCommonConfig &
+            GetMultipleAccountsApiSliceableCommonConfig &
+            Readonly<{
+                encoding: 'base64';
+            }>
+    ): RpcResponse<(GetMultipleAccountsApiResponseBase & GetMultipleAccountsApiResponseWithBase64EncodedData)[]>;
```
steveluscher, line 77: I don't think this allows for holes (nulls) in the return array (eg. if an account doesn't exist).

Check out the approach in `getAccountInfo` where I layer over the possibility of an account turning up `null` using the `NestInRpcResponseOrNull` utility type.

steveluscher:
Oof. I wish TypeScript type mapping was good enough that we could have just mapped over the overloads of `getAccountInfo` and turn them all into arrays.

```
type GetMultipleAccountsApi = {
  [TMethod in keyof GetAccountInfoApi as 'getMultipleAccounts']: GetMultipleAccountsApi[TMethod] extends (...args: any[]) => any
    ? (
        addresses: Parameters<GetMultipleAccountsApi[TMethod]>[0],
        config: Parameters<GetMultipleAccountsApi[TMethod]>[1],
      ) => Readonly<{
        ...ReturnType<GetMultipleAccountsApi[TMethod]>,
        value: readonly (ReturnType<GetMultipleAccountsApi[TMethod]>['value'])[],
      }>
  : never,
};
```

Something like that doesn't work, right?

packages/rpc-core/src/response-patcher-allowed-numeric-values.ts
mcintyre94: I think that 2nd option is probably worth doing, it becomes a lot to maintain when there are so many methods returning this parsed data

steveluscher:
Ugh.

Would probably have to do something like the ‘expand all the overloads’ trick.

Oh well.

PR #1494, opened 2023-08-04
refactor(experimental): add `getLeaderSchedule` API method

buffalojoec:
This PR adds the `getLeaderSchedule` RPC method to the new experimental Web3 JS's arsenal.

Ref #1449


packages/rpc-core/src/rpc-methods/getLeaderSchedule.ts
```diff
@@ -0,0 +1,43 @@
...
+ *     39, 40, 41, 42, 43, 44, 45, 46, 47, 48, 49, 50, 51, 52, 53, 54, 55, 56,
+ *     57, 58, 59, 60, 61, 62, 63
+ *   ]
+ * }
+ * ```
+ */
+type GetLeaderScheduleApiResponse =
+    | Readonly<{
+          [key: Base58EncodedAddress]: Slot[];
+      }>[]
```
mcintyre94, line 24:
```suggestion
      }>
```

I think you just want a single object here. [Typescript Playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAMhCGATCAnAygYwBYUQVwBsIAlCAZzAHsA7M6AXilKRoJAB4BvAKCj6gDaAawggAXFDLAUAS2oBzALoTqeALYAjVAMUBubgF8AfDu7cA9OagBJKMCxyhdh2SgzXEAB6QMwXFBRyKloIbgwaKQCJOCRUTBx8IlIKCIYoHn4oACJ4DQwsiQEARgAaACYSgGZFEt5+LOQAMwLBABYSgFYSgDZFQzNLKABBakRndzdXDDwUQOpgSWk8XxnQ8NoFz2iEZHRsXEISINSoRgEM+tz8wtKK6trMhohmwvau3sMS9Lq+LPkHFoCADsJQAHCUAJx9Ax9bigSCwHZxfaJI4pEIAMRknn8jGYiFYHAuwlEEiksgUyigqk02j0hiMZnWkRQZW2sT2CUOyWCdCxOLGjAuvyugNuVRqP2yTUBbx60LMQA)

packages/rpc-core/src/rpc-methods/getLeaderSchedule.ts
```diff
@@ -0,0 +1,43 @@
...
+         * Fetch the leader schedule for the epoch that corresponds to the provided slot.
+         * If unspecified, the leader schedule for the current epoch is fetched
+         */
+        slot?: Slot,
+        config?: Readonly<{
+            commitment?: Commitment;
+            /** Only return results for this validator identity (base58 encoded address) */
+            identity?: Base58EncodedAddress;
+        }>
+    ): GetLeaderScheduleApiResponse;
```
mcintyre94, line 42: I think if `slot` is defined, then the response can't be null, since there's always a current epoch (I think!). Not sure if an overload would be useful to eliminate the null check in that case?

We could also return a narrower type when `identity` is defined, but would probably have the same issue as https://github.com/solana-labs/solana-web3.js/issues/1280 and not gain any benefit in practice.

mcintyre94 (changes requested):

packages/rpc-core/src/rpc-methods/getLeaderSchedule.ts
buffalojoec: Dictionary confused me, ty!

packages/rpc-core/src/rpc-methods/getLeaderSchedule.ts
buffalojoec: Yeah, looks like:
1. It will calculate the epoch for a provided slot ([link](https://github.com/solana-labs/solana/blob/master/rpc/src/rpc.rs#L2705))
2. Which is just a calculation ([link](https://github.com/solana-labs/solana/blob/master/sdk/program/src/epoch_schedule.rs#L131-L165))
3. Then if that epoch has no associated value in the cached leader schedule on the ledger, it returns `None` for `Option<Arc<LeaderSchedule>>` ([link](https://github.com/solana-labs/solana/blob/master/ledger/src/leader_schedule_cache.rs#L198C12-L198C37))

So, in other words, we can safely assume - as you said - the response can only be null if a slot parameter is provided, and that slot returns an epoch that doesn't exist

mcintyre94: Sorry yep, opposite way! Do you think it's a useful overload to remove the `null` return when `slot` is undefined? I don't have a strong view TBH

buffalojoec: I think since we did the digging we know it can't hurt! And the added benefit is not having to do the null check if you don't provide a slot, which is nice.

I wrote a new implementation with this feature, but because of TypeScript's tendency to accept `undefined` as a valid arg, I decided instead to use overloading. 

Let me know what you think!

mcintyre94: Looks good! Agreed that overloads is definitely the way to go there

PR #1496, opened 2023-08-04
test: fix genesis hash regex

buffalojoec:
This PR fixes the regex for grabbing the genesis hash from the validator logs. The length specifier previously used (44) is likely to fail whenever the raw bytes produce a shorter base58 encoding


mcintyre94 (approved):
LGTM!

PR #1499, opened 2023-08-05
refactor(experimental): add `getSlotLeader` method

buffalojoec:
This PR adds`getSlotLeader` RPC method to the new experimental Web3 JS's arsenal.

Ref #1449

### DIFF
--- a/packages/rpc-core/src/rpc-methods/__tests__/get-slot-leader-test.ts
+++ b/packages/rpc-core/src/rpc-methods/__tests__/get-slot-leader-test.ts
@@ -18,18 +18,18 @@ describe('getSlotLeader', () => {
 
     (['confirmed', 'finalized', 'processed'] as Commitment[]).forEach(commitment => {
         describe(`when called with \`${commitment}\` commitment`, () => {
-            describe('when called with no `mintContextSlot`', () => {
+            describe('when called with no `minContextSlot`', () => {
                 it('returns the node public key', async () => {
                     expect.assertions(1);
-                    const res = await rpc.getSlotLeader().send();
+                    const res = await rpc.getSlotLeader({ commitment }).send();
                     expect(res).toEqual(expect.any(String));
                 });
             });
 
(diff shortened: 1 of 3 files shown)


packages/rpc-core/src/rpc-methods/__tests__/get-slot-leader-test.ts
```diff
@@ -0,0 +1,54 @@
...
+            transport: createHttpTransport({ url: 'http://127.0.0.1:8899' }),
+        });
+    });
+
+    (['confirmed', 'finalized', 'processed'] as Commitment[]).forEach(commitment => {
+        describe(`when called with \`${commitment}\` commitment`, () => {
+            describe('when called with no `mintContextSlot`', () => {
+                it('returns the node public key', async () => {
+                    expect.assertions(1);
+                    const res = await rpc.getSlotLeader().send();
```
mcintyre94, line 24:
```suggestion
                    const res = await rpc.getSlotLeader({ commitment }).send();
```

packages/rpc-core/src/rpc-methods/__tests__/get-slot-leader-test.ts
```diff
@@ -0,0 +1,54 @@
...
+                    expect(res).toEqual(expect.any(String));
+                });
+            });
+
+            describe('when called with a valid `mintContextSlot`', () => {
+                it('returns the node public key', async () => {
+                    expect.assertions(1);
+                    const res = await rpc.getSlotLeader({ minContextSlot: 0n }).send();
```
mcintyre94, line 32:
```suggestion
                    const res = await rpc.getSlotLeader({ commitment, minContextSlot: 0n }).send();
```

mcintyre94 (changes requested):
Looks good - just missing the `commitment` param in the tests

buffalojoec:
Added `commitment` !

mcintyre94 (approved):

PR #1502, opened 2023-08-05
refactor(experimental): add `getTokenSupply` API method

buffalojoec:
This PR adds the `getTokenSupply` RPC method to the new experimental Web3 JS's arsenal.

Ref #1449

### DIFF
--- a/packages/rpc-core/src/rpc-methods/__tests__/get-token-supply-test.ts
+++ b/packages/rpc-core/src/rpc-methods/__tests__/get-token-supply-test.ts
@@ -33,7 +33,8 @@ describe('getTokenSupply', () => {
                     value: {
                         amount: expect.any(String),
                         decimals: expect.any(Number),
-                        uiAmount: expect.anything(), // Number or null
+                        // This can be Number or null, but we're using a fixture so it should be Number
+                        uiAmount: expect.any(Number),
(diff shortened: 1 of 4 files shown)


packages/rpc-core/src/rpc-methods/__tests__/get-token-supply-test.ts
```diff
@@ -0,0 +1,60 @@
...
+                expect.assertions(1);
+                const tokenAccountBalancePromise = rpc.getTokenSupply(tokenMintAddress, { commitment }).send();
+                await expect(tokenAccountBalancePromise).resolves.toMatchObject({
+                    context: {
+                        slot: expect.any(BigInt),
+                    },
+                    value: {
+                        amount: expect.any(String),
+                        decimals: expect.any(Number),
+                        uiAmount: expect.anything(), // Number or null
```
mcintyre94, line 36: Same suggestion to use `any(Number)` here

mcintyre94 (approved):
LGTM! Nice to see these fixtures being useful!

packages/rpc-core/src/rpc-methods/__tests__/get-token-supply-test.ts
buffalojoec: Done !

PR #1503, opened 2023-08-05
refactor(experimental): add `getVersion` API method

buffalojoec:
This PR adds the`getVersion` RPC method to the new experimental Web3 JS's arsenal.

Ref #1449


buffalojoec:
> Looks good! I took a look at the backend code and it looks like `feature-set` is a `u32`: https://github.com/solana-labs/solana/blob/849525735f784f8e2e0fca19ba5699aea6b1724e/version/src/lib.rs#L33
> 
> I don't think it really matters but we could opt it out of the bigint parsing

I think since it's a `u32` it makes sense to make it a `Number` and opt it out. Will add

packages/rpc-core/src/rpc-methods/__tests__/get-version-test.ts
```diff
@@ -0,0 +1,28 @@
...
+        fetchMock.resetMocks();
+        fetchMock.dontMock();
+        rpc = createJsonRpc<SolanaRpcMethods>({
+            api: createSolanaRpcApi(),
+            transport: createHttpTransport({ url: 'http://127.0.0.1:8899' }),
+        });
+    });
+
+    describe('when called on a valid node', () => {
+        it('returns the version', async () => {
```
steveluscher, line 19: You could read the version string out of `validator.log` too, using that `readLines()` trick, if you like.

PR #1504, opened 2023-08-05
refactor(experimental): add `getFeeForMessage` API method

buffalojoec:
This PR adds the `getFeeForMessage` RPC method to the new experimental Web3 JS's arsenal.

Ref #1449

### DIFF
--- a/packages/rpc-core/src/rpc-methods/__tests__/get-fee-for-message-test.ts
+++ b/packages/rpc-core/src/rpc-methods/__tests__/get-fee-for-message-test.ts
@@ -1,0 +1,144 @@
+import { base58, base64, fixSerializer } from '@metaplex-foundation/umi-serializers';
+import { createHttpTransport, createJsonRpc } from '@solana/rpc-transport';
+import type { SolanaJsonRpcErrorCode } from '@solana/rpc-transport/dist/types/json-rpc-errors';
+import type { Rpc } from '@solana/rpc-transport/dist/types/json-rpc-types';
+import { Blockhash } from '@solana/transactions';
+import fetchMock from 'jest-fetch-mock-fork';
+
+import { Base64EncodedBytes, Commitment } from '../common';
+import { createSolanaRpcApi, SolanaRpcMethods } from '../index';
+
+// See scripts/fixtures/send-transaction-fee-payer.json
+const MOCK_PUBLIC_KEY_BYTES = // DRtXHDgC312wpNdNCSb8vCoXDcofCJcPHdAw4VkJ8L9i
+    // prettier-ignore
+    new Uint8Array([
+        0xb8, 0xac, 0x70, 0x4f, 0xaf, 0xc7, 0xa5, 0xfc, 0x8c, 0x5d, 0x1f, 0x0a, 0xc8, 0xcf, 0xaa, 0xe0,
+        0x42, 0xfa, 0x3b, 0xb8, 0x25, 0xf0, 0xec, 0xfc, 0xe2, 0x27, 0x4d, 0x7d, 0xad, 0xad, 0x51, 0x2d,
+    ]);
+
+function getMockTransactionMessage(blockhash: Blockhash) {
+    const memoString = 'Hello from the web3.js tests!';
+    const blockhashBytes = fixSerializer(base58, 32).serialize(blockhash);
+    // prettier-ignore
+    const message = new Uint8Array([
+        /** VERSION HEADER */
+        0x80, // 0 + version mask
+
+        /** MESSAGE HEADER */
+        0x01, // numSignerAccounts
+        0x00, // numReadonlySignerAccount
+        0x01, // numReadonlyNonSignerAccounts
+
+        /** STATIC ADDRESSES */
+        0x02, // Number of static accounts
+            ...MOCK_PUBLIC_KEY_BYTES,
+            0x05, 0x4a, 0x53, 0x5a, 0x99, 0x29, 0x21, 0x06, 0x4d, 0x24, 0xe8, 0x71, 0x60, 0xda, 0x38, 0x7c, 0x7c, 0x35, 0xb5, 0xdd, 0xbc, 0x92, 0xbb, 0x81, 0xe4, 0x1f, 0xa8, 0x40, 0x41, 0x05, 0x44, 0x8d, // MemoSq4gqABAXKb96qnH8TysNcWxMyWCqXgDLGmfcHr
+
+        /** TRANSACTION LIFETIME TOKEN (ie. the blockhash) */
+        ...blockhashBytes,
+
+        /* INSTRUCTIONS */
+        0x01, // Number of instructions
+
+            // First instruction
+            0x01, // Program address index
+            0x00, // Number of address indices
+            memoString.length, // Length of instruction data
+                ...new TextEncoder().encode(memoString),
+
+        /** ADDRESS TABLE LOOKUPS */
+        0x00, // Number of address table lookups
+    ]);
+    const messageBase64 = base64.deserialize(message)[0];
+    return messageBase64 as Base64EncodedBytes;
+}
+
+describe('getFeeForMessage', () => {
+    let rpc: Rpc<SolanaRpcMethods>;
+    beforeEach(() => {
+        fetchMock.resetMocks();
+        fetchMock.dontMock();
+        rpc = createJsonRpc<SolanaRpcMethods>({
+            api: createSolanaRpcApi(),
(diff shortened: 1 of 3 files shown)


packages/rpc-core/src/rpc-methods/getFeeForMessage.ts
```diff
@@ -0,0 +1,18 @@
...
+/** Fee corresponding to the message at the specified blockhash */
+type GetFeeForMessageApiResponse = RpcResponse<U64UnsafeBeyond2Pow53Minus1 | null>;
+
+export interface GetFeeForMessageApi {
+    /**
+     * Returns the fee the network will charge for a particular Message
+     */
+    getFeeForMessage(
+        /** Base-64 encoded message */
+        message: string, // TODO: May not be string
```
mcintyre94, line 12: `Base64EncodedBytes`?

mcintyre94 (approved):
Looks good!

packages/rpc-core/src/rpc-methods/__tests__/get-fee-for-message-test.ts
```diff
@@ -0,0 +1,144 @@
...
+    });
+
+    (['confirmed', 'finalized', 'processed'] as Commitment[]).forEach(commitment => {
+        describe(`when called with \`${commitment}\` commitment`, () => {
+            if (commitment === 'finalized') {
+                it.todo(
+                    'returns the result as a bigint (test broken; see https://discord.com/channels/428295358100013066/560496939779620864/1132048104728825926)'
+                );
+                return;
+            }
```
mcintyre94, line 74: Hmm is this a timing issue that could affect `confirmed` too if the test runs early enough?

packages/rpc-core/src/rpc-methods/getFeeForMessage.ts
buffalojoec: I think that makes sense, is that de facto for tx message?

packages/rpc-core/src/rpc-methods/__tests__/get-fee-for-message-test.ts
buffalojoec: I'm not sure on confirmed. I plucked this from https://github.com/solana-labs/solana-web3.js/blob/master/packages/rpc-core/src/rpc-methods/__tests__/send-transaction-test.ts#L115-L120

It's definitely related to the fact that it takes ~17s for the test validator to finalize slot 0, but `confirmed` seems to be almost immediate so maybe that's not a concern? I don't see `confirmed` omitted anywhere else in our test suites

packages/rpc-core/src/rpc-methods/getFeeForMessage.ts
mcintyre94: I want to say yes but there's probably somewhere we use some other encoding for some reason 😅 In this case though the RPC docs are clear that it's base64 and that's exactly what this type is for!

buffalojoec: Got it, going with that then!

packages/rpc-core/src/rpc-methods/__tests__/get-fee-for-message-test.ts
```diff
@@ -0,0 +1,144 @@
...
+            describe('when called with a recent blockhash', () => {
+                it('returns the result as a bigint', async () => {
+                    expect.assertions(1);
+                    const latestBlockhash = await rpc.getLatestBlockhash().send();
+                    const message = getMockTransactionMessage(latestBlockhash.value.blockhash);
+                    console.log('Printing mock message', message);
```
steveluscher, line 80: 👀

packages/rpc-core/src/rpc-methods/getFeeForMessage.ts
```diff
@@ -0,0 +1,18 @@
...
+/** Fee corresponding to the message at the specified blockhash */
+type GetFeeForMessageApiResponse = RpcResponse<U64UnsafeBeyond2Pow53Minus1 | null>;
+
+export interface GetFeeForMessageApi {
+    /**
+     * Returns the fee the network will charge for a particular Message
+     */
+    getFeeForMessage(
+        /** Base-64 encoded message */
+        message: Base64EncodedBytes,
```
steveluscher, line 12: We should really create an opaque type here to represent ‘a bunch of bytes that are definitely a serialized message.’

1. Create such an opaque type
2. Update `packages/transaction/src/serializers/message.ts` to return that type from `serialize`
3. Make this function only accept that type

packages/rpc-core/src/rpc-methods/getFeeForMessage.ts
```diff
@@ -0,0 +1,18 @@
+import { Base64EncodedBytes, Commitment, RpcResponse, Slot, U64UnsafeBeyond2Pow53Minus1 } from './common';
+
+/** Fee corresponding to the message at the specified blockhash */
+type GetFeeForMessageApiResponse = RpcResponse<U64UnsafeBeyond2Pow53Minus1 | null>;
```
steveluscher, line 4: The RPC does not wrap this in a `RpcResponse` wrapper.

```diff
- type GetFeeForMessageApiResponse = RpcResponse<U64UnsafeBeyond2Pow53Minus1 | null>;
+ type GetFeeForMessageApiResponse = U64UnsafeBeyond2Pow53Minus1 | null;
```

buffalojoec: I believe it does mate

https://docs.solana.com/api/http#getfeeformessage
https://github.com/solana-labs/solana/blob/67d6d688ccf10d09a012559a43f524aa35e87919/rpc/src/rpc.rs#L3996
https://github.com/solana-labs/solana/blob/67d6d688ccf10d09a012559a43f524aa35e87919/rpc/src/rpc.rs#L120

packages/rpc-core/src/rpc-methods/__tests__/get-fee-for-message-test.ts
```diff
@@ -0,0 +1,144 @@
...
+
+// See scripts/fixtures/send-transaction-fee-payer.json
+const MOCK_PUBLIC_KEY_BYTES = // DRtXHDgC312wpNdNCSb8vCoXDcofCJcPHdAw4VkJ8L9i
+    // prettier-ignore
+    new Uint8Array([
+        0xb8, 0xac, 0x70, 0x4f, 0xaf, 0xc7, 0xa5, 0xfc, 0x8c, 0x5d, 0x1f, 0x0a, 0xc8, 0xcf, 0xaa, 0xe0,
+        0x42, 0xfa, 0x3b, 0xb8, 0x25, 0xf0, 0xec, 0xfc, 0xe2, 0x27, 0x4d, 0x7d, 0xad, 0xad, 0x51, 0x2d,
+    ]);
+
+function getMockTransactionMessage(blockhash: Blockhash) {
```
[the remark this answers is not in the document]
buffalojoec: I think the hard part is hard-coding a valid blockhash for the recent blockhash.
Even in the test for `sendTransaction` that's what we're doing. 

https://github.com/solana-labs/solana-web3.js/blob/dc614aff8d85929e95452e4af0e4f5d3310ee82e/packages/rpc-core/src/rpc-methods/__tests__/send-transaction-test.ts#L123-L131

A solution for blockhash might be to somehow grab it from the validator logs, then we get around the _retrieval_ of a valid blockhash, but if that blockhash isn't deterministic within our test suite then I don't know how we can hard-code the serialized value of it.

Blockhashes should be _kind of_ deterministic with a local validator right? I mean, if you know precisely the slots and transactions you can maybe reproduce it, but I'm not sure on the timing of all the test suites firing at once.

packages/rpc-core/src/rpc-methods/getFeeForMessage.ts
steveluscher: Oh, oops. I was reading just the docs, which are wrong. They should mention that an RpcResponse object is returned.

packages/rpc-core/src/rpc-methods/__tests__/get-fee-for-message-test.ts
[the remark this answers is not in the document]
steveluscher: Oh crap. I wasn’t thinking of the fact that the fee schedule is tied to a moment in time, ie. a blockhash.

Carry on!

packages/rpc-core/src/rpc-methods/getFeeForMessage.ts
buffalojoec: Yeah looks like it's in their payload just not the description like the others

PR #1510, opened 2023-08-07
refactor(experimental): Fix getClusterNodes test

buffalojoec:
Turns out these ports change!

Also added the `number` response fields to the bigint exemptions

### DIFF
--- a/packages/rpc-core/src/rpc-methods/__tests__/get-cluster-nodes-test.ts
+++ b/packages/rpc-core/src/rpc-methods/__tests__/get-cluster-nodes-test.ts
@@ -15,18 +15,19 @@ describe('getClusterNodes', () => {
         });
     });
 
-    it('returns the cluster nodes', async () => {
-        expect.assertions(1);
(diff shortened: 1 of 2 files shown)


packages/rpc-core/src/rpc-methods/__tests__/get-cluster-nodes-test.ts
```diff
@@ -15,18 +15,19 @@ describe('getClusterNodes', () => {
...
+    describe('when run against the test validator', () => {
+        it('returns RPC and validator info', async () => {
+            expect.assertions(1);
+            const res = await rpc.getClusterNodes().send();
+            expect(res).toEqual(
+                expect.arrayContaining([
+                    expect.objectContaining({
+                        gossip: expect.any(String),
+                        rpc: expect.any(String),
+                        tpu: expect.any(String),
```
steveluscher, line 27: Take another pass at this with `expect.stringMatching()`? That should help you make this resilient against changing port numbers.

PR #1521, opened 2023-08-10
Fix: remove fs.readFileSync from genesis hash test

### LINKED ISSUE #1517: Error running get-genesis-hash-test locally on long-running validator
## Overview

- You can run the shared validator locally: `./scripts/start-shared-test-validator.sh`
- If you do this then the unit tests for eg rpc-core will use this validator
- This is useful for being able to make RPC calls manually while creating tests
- Over time the `validator.log` will fill up
- And eventually a test that tries to read the whole file will start failing

## Steps to reproduce

- Run the shared validator locally: `./scripts/start-shared-test-validator.sh`
- A few hours later, in `packages/rpc-core`, run `pnpm test:unit:node`

## Description of bug

The genesis hash test fails:

```
Summary of all failing tests
 FAIL  src/rpc-methods/__tests__/get-genesis-hash-test.ts
  ● getGenesisHash › when sent to a local validator › returns the genesis hash

    RangeError: File size (5599667301) is greater than 2 GiB

      24 |         it('returns the genesis hash', async () => {
      25 |             expect.assertions(1);
    > 26 |             const logFile = fs.readFileSync(logFilePath, 'utf-8');
         |                                ^
      27 |             const expectedGenesisHash = logFile.match(genesisHashPattern)?.[1];
      28 |             const genesisHashPromise = rpc.getGenesisHash().send();
      29 |             await expect(genesisHashPromise).resolves.toBe(expectedGenesisHash);

      at Object.readFileSync (src/rpc-methods/__tests__/get-genesis-hash-test.ts:26:32)

  ● getGenesisHash › when sent to a local validator › returns the genesis hash

    expect.assertions(1)

    Expected one assertion to be called but received zero assertion calls.

      23 |     describe('when sent to a local validator', () => {
      24 |         it('returns the genesis hash', async () => {
    > 25 |             expect.assertions(1);
         |                    ^
      26 |             const logFile = fs.readFileSync(logFilePath, 'utf-8');
      27 |             const expectedGenesisHash = logFile.match(genesisHashPattern)?.[1];
      28 |             const genesisHashPromise = rpc.getGenesisHash().send();

      at Object.assertions (src/rpc-methods/__tests__/get-genesis-hash-test.ts:25:20)


Test Suites: 1 failed, 50 passed, 51 total
Tests:       1 failed, 66 todo, 260 passed, 327 total
Snapshots:   0 total
Time:        2.419 s
```

We'll need to find a different way to find the genesis hash pattern in validator.log without loading the whole file into memory.

This will probably always be fine in CI because the validator has just started, but locally it's useful to be able to keep a long-running test validator up and run the test suite against it.

mcintyre94:
We instead use a read stream to process the file in chunks

This avoids a bug where when validator.log gets too big locally the `fs.readFileSync` would fail

Fixes #1517

### DIFF
--- a/packages/rpc-core/src/rpc-methods/__tests__/get-genesis-hash-test.ts
+++ b/packages/rpc-core/src/rpc-methods/__tests__/get-genesis-hash-test.ts
@@ -9,6 +9,24 @@ import { createSolanaRpcApi, SolanaRpcMethods } from '../index';
 const logFilePath = path.resolve(__dirname, '../../../../../test-ledger/validator.log');
 const genesisHashPattern = /genesis hash: ([\d\w]{32,})/;
 
+async function getGenesisHashFromLogFile() {
+    return new Promise<string | undefined>((resolve, reject) => {
+        const logFileStream = fs.createReadStream(logFilePath, { end: 64 * 1024 });
+        logFileStream.on('error', e => reject(e))
+
+        logFileStream.on('data', data => {
+            const chunk = data.toString('utf-8');
+            const expectedGenesisHash = chunk.match(genesisHashPattern)?.[1];
+            if (expectedGenesisHash) {
+                logFileStream.destroy();
+                resolve(expectedGenesisHash)
+            }
+        })
+
+        logFileStream.on('end', () => resolve(undefined));
+    })
+}
+
 describe('getGenesisHash', () => {
     let rpc: Rpc<SolanaRpcMethods>;
     beforeEach(() => {
@@ -23,8 +41,7 @@ describe('getGenesisHash', () => {
     describe('when sent to a local validator', () => {
         it('returns the genesis hash', async () => {
             expect.assertions(1);
-            const logFile = fs.readFileSync(logFilePath, 'utf-8');
-            const expectedGenesisHash = logFile.match(genesisHashPattern)?.[1];
+            const expectedGenesisHash = await getGenesisHashFromLogFile();
             const genesisHashPromise = rpc.getGenesisHash().send();
             await expect(genesisHashPromise).resolves.toBe(expectedGenesisHash);
         });


packages/rpc-core/src/rpc-methods/__tests__/get-genesis-hash-test.ts
```diff
@@ -9,6 +9,24 @@ import { createSolanaRpcApi, SolanaRpcMethods } from '../index';
...
+        logFileStream.on('data', data => {
+            const chunk = data.toString('utf-8');
+            const expectedGenesisHash = chunk.match(genesisHashPattern)?.[1];
+            if (expectedGenesisHash) {
+                logFileStream.destroy();
+                resolve(expectedGenesisHash)
+            }
+        })
+
+        logFileStream.on('end', () => resolve(undefined));
```
buffalojoec, line 26: This should be good, just curious if you think we should have some kind of retry loop, like `numRetries = 2`

buffalojoec (approved):
Solid fix mate, nicely done

packages/rpc-core/src/rpc-methods/__tests__/get-genesis-hash-test.ts
mcintyre94: Do you mean we should just fail if it's not in the first N chunks? I think that if we want to do that the best way might be to set the `end` parameter when creating the stream, so it only reads X bytes (https://nodejs.org/api/fs.html#filehandlecreatereadstreamoptions). We don't seem to have much control over how big chunks are (I think `highWaterMark` sets an upper bound, that seems to be it) so I'm a bit hesitant to rely on a chunk count. But maybe we could just set the `end` at about 16Kb or something?

buffalojoec: No actually I was thinking what if it doesn't find the genesis hash in it's first attempt to read the chunks (maybe the file is only a few lines and the validator is still spinning up)?

The more I think about it, the more I feel like it's not an issue since if the validator is not in a ready state, everything's going to fail anyway.

If this causes any issues with any tests across the board, we're better off making an issue and adjusting [this bad boy](https://github.com/solana-labs/solana-web3.js/blob/master/scripts/start-shared-test-validator.sh)

mcintyre94: Yep agreed, the existing code seems to work fine in CI too just reading it right away. I've added a 64kb `end` just in case someone somehow ends up with a very broken validator.log. It's about 4.5kb into the file for my current one so pretty sure it'll always end up in the first 64kb!

buffalojoec: Ship it!

steveluscher: Even easier, tho :)

https://github.com/solana-labs/solana-web3.js/pull/1478#discussion_r1300523142

Issue #1280, opened 2023-04-26
Figure out how to improve type safety for `getBlockProduction` API


mcintyre94:
See: https://github.com/solana-labs/solana-web3.js/pull/1263#issuecomment-1521613024

Ideally we'd like this to be a type error:

```ts
const blockProduction = await rpc.getBlockProduction({identity: 'abc' as Base58EncodedAddress})
blockProduction.value.byIdentity['def' as Base58EncodedAddress]
```

[...]

PR #1522, opened 2023-08-10
Add get program accounts RPC method

mcintyre94:
This PR adds the `getProgramAccounts` RPC method for experimental web3js

This is very similar to `getMultipleAccounts`, with the fun extra quirk that it can return either an RPC response or not, depending on the value of `withContext`. Annoyingly my attempts to use a conditional type ([Typescript playground](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcBKYDGbgGdIB2uwAPACoBqAhgDYCuwAfHALzrBUAmEBNiJAbwBQcUXEw8YoGAC52XHn0Eixq3DQiy4BOgFsARsCgBuFaIC+jU6oButBnMr3gpy6aFIUcAOIAFAIIAwhC6ujzBBABmAJYA5uQA6tEwABYRUiAwzGw4Crz8wqoSocm6wAQwAPxyuDBQ0QSx1mIA7slpktLVcGRJqenSrlZCHsioGNh4hMQAkpF9HRXSie0DmQA0PTMEBEbZPQtr8NLlnLgIUAxwlehYOPg8xOTbu1DMji9G7pF0BJgw0R4PgCK36nUycBOBDOcH0EAgNA4BFYcEitGIjAAFGZVOIeDFYt0-EEQmECBECaDFhkskIAJRyCb3abAOaHcEwKlHTY6Ax7OCFMSpKAQFraYBigCiUBFUExdNcIwkRHgBAgFLicl5higKOJ8tMytq2ggvioUCoui1eh1eoCmIExV0pXKWgA5FQ3eYFUIjfA2mClplGXcpo9SNr+Wx9QIA9TpHI6gxvYbHqr1RzrXzddH7bHVpnUejgCmhEA)) didn't work, it just always returned `T | RpcResponse<T>`. I suspect this is the same underlying issue as #1280 

So instead this method has 10 overloads, (4 encodings + no encoding) * 2 options for `withContext`. The previous limit was 5, so that's why there's a change to [json-rpc-types.ts](https://github.com/solana-labs/solana-web3.js/compare/master...mcintyre94:get-program-accounts?expand=1#diff-6a020e524e197bb6c72e960a513d3a6305a6c9b71e5d3d991873c32ebffd6b30) 

I also decided against refactoring to try to share data structures between `getAccountInfo`, `getMultipleAccounts` and `getProgramAccounts`. They have subtly different behaviours/structures and I think it's just clearer to keep them as-is for now. 

Ref #1449


packages/rpc-core/src/rpc-methods/getProgramAccounts.ts
```diff
@@ -0,0 +1,185 @@
...
+            }>
+    ): RpcResponse<AccountWithPubkey<AccountInfoBase & AccountInfoWithBase64EncodedData>[]>;
+
+    getProgramAccounts(
+        program: Base58EncodedAddress,
+        config: GetProgramAccountsApiCommonConfig &
+            GetProgramAccountsApiSliceableCommonConfig &
+            Readonly<{
+                encoding: 'base64';
+                withContext?: boolean;
```
buffalojoec, line 106: This looks super weird, but it works apparently? If provided as `true` it's going to the first overload instead, so this covers `undefined` or `false` yeah?

buffalojoec: I also only recently discovered you can do [default parameters in TypeScript](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAcwKZQAoCc7KwQwFsBBCCOcKAZwAoAoRRxfAExa1SqoC5EqosMMMgA0iBk3JhgMZLwDeEpk1SQ4LIcgD8vfoOEilyxAHcYUABYBhBFFQAPKDsQAjOHAA2qfGEPKAvogAvIiKxipqGsK8AOQu+FSoAKwAHDF+4abm1rYOULzA+B6JGYz+dACUoUZSVJ6oAHQeuDSs7JxUFQDcNQh1Xk0tUjLI3XT+QA), like Python:

packages/rpc-core/src/rpc-methods/__tests__/get-program-accounts-test.ts
```diff
@@ -0,0 +1,1213 @@
...
+                    'DXngmJfjurhnAwbMPgpUGPH6qNvetCKRJ6PiD4ag4PTj' as Base58EncodedAddress<'DXngmJfjurhnAwbMPgpUGPH6qNvetCKRJ6PiD4ag4PTj'>;
+
+                const accountInfo = await rpc
+                    .getProgramAccounts(program, {
+                        encoding: 'base58',
+                        withContext: true,
+                    })
+                    .send();
+
+                expect(accountInfo.value[0].account.data).toStrictEqual(['2Uw1bpnsXxu3e', 'base58']);
```
buffalojoec, line 150: Seems a little unnecessary, but should we check the whole response to ensure it matches `RpcResponse`? ie. add the context bit to the `expect(..)`

buffalojoec (commented):
Lgtm!! What a beast

packages/rpc-transport/src/json-rpc-types.ts
```diff
@@ -54,38 +54,93 @@ type PendingRpcRequestBuilder<TMethodImplementations> = UnionToIntersection<
 type Callable = (...args: any[]) => any;
 type Flatten<T> = T extends (infer Item)[] ? Item : never;
 type Overloads<T> =
-    // Have an RPC method with more than 5 overloads? Add another section and update this comment
+    // Have an RPC method with more than 10 overloads? Add another section and update this comment
```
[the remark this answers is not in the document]
mcintyre94: image

packages/rpc-core/src/rpc-methods/getProgramAccounts.ts
mcintyre94: Yep exactly, it's basically that the param is optional and the default is `false`. So you define `withContext: true` with its `RpcResponse<...>` and then `withContext?: boolean` will capture both explicit false and undefined -> default false.
Annoyingly it doesn't work with `encoding` though because (for this RPC method...) no defined encoding has its own annoying special case
I don't think default parameters will solve it here either because you want a different type on the response depending on the value of that parameter - it's not just input. So I think that requires an overload, or the conditional typing that I couldn't get to work

packages/rpc-core/src/rpc-methods/__tests__/get-program-accounts-test.ts
mcintyre94: It's probably worth checking this somewhere, and since this is the first `withContext: true` test it's probably the place to do it

One thing I will point out is that these tests that traverse the data like `accountInfo.value[0].account.data` are actually really valuable for the type checking. Eg if you mess up the typing like when I was trying to do the conditional types, the tests that use `toMatchObject` etc. won't fail because the actual return data is correct. But `accountInfo.value` will error because it doesn't know what the type is. I suspect we have some RPC methods that we've only tested using object matchers that we're not really testing the typing of.

packages/rpc-core/src/rpc-methods/getProgramAccounts.ts
buffalojoec: Sounds good to me. You're right that is a mega-pain lol

packages/rpc-core/src/rpc-methods/__tests__/get-program-accounts-test.ts
buffalojoec: What about if you are expecting:

```javascript
{
  context: number,
  value: [
    { account: { data: string }}
  ]
}
```

and you get:

```javascript
{
  foo: string,
  value: [
    { account: { data: string }}
  ]
}
```

Wouldn't `accountInfo.value[0].account.data` still work?

mcintyre94: Yep you have to test `context` for that (or use `toMatchObject` and check both fields)

I've added `expect(accountInfo.context.slot).toEqual(expect.any(BigInt));` to this test which covers it for this case

The rest of the `account` structure is IMO adequately tested elsewhere, but the RPC bit is only tested by these tests that have `withContext: true`

buffalojoec (approved):

packages/rpc-core/src/response-patcher-allowed-numeric-values.ts
```diff
@@ -127,6 +127,142 @@ export const ALLOWED_NUMERIC_KEYPATHS: Partial<
         ['value', KEYPATH_WILDCARD, 'data', 'parsed', 'info', 'commission'],
         ['value', KEYPATH_WILDCARD, 'data', 'parsed', 'info', 'votes', KEYPATH_WILDCARD, 'confirmationCount'],
     ],
+    getProgramAccounts: [
```
steveluscher, line 130: Oof. Related comment: https://github.com/solana-labs/solana-web3.js/pull/1492#discussion_r1300732679

packages/rpc-core/src/rpc-methods/__tests__/get-program-accounts-test.ts
```diff
@@ -0,0 +1,1214 @@
...
+            });
+            expect(accountInfo[1]).toMatchObject({
+                account: expect.objectContaining({
+                    data: expect.any(String),
+                    executable: expect.any(Boolean),
+                    lamports: expect.any(BigInt),
+                    owner: expect.any(String),
+                    rentEpoch: expect.any(BigInt),
+                    space: expect.any(BigInt),
+                }),
```
steveluscher, line 109: Should we actually assert on the values of the accounts, since they come from fixtures?

packages/rpc-core/src/rpc-methods/getProgramAccounts.ts
```diff
@@ -0,0 +1,185 @@
...
+     */
+    getProgramAccounts(
+        program: Base58EncodedAddress,
+        config: GetProgramAccountsApiCommonConfig &
+            GetProgramAccountsApiSliceableCommonConfig &
+            Readonly<{
+                encoding: 'base64';
+                withContext: true;
+            }>
+    ): RpcResponse<AccountWithPubkey<AccountInfoBase & AccountInfoWithBase64EncodedData>[]>;
```
steveluscher, line 98: What if…

https://www.typescriptlang.org/play?ssl=20&ssc=1&pln=20&pc=9#code/JYOwLgpgTgZghgYwgAgGIHt3IN4FgBQyyA5hGACoAWoxAPOQMLrgQAeYTIMwxybkIACYBnHAHdgYSp0jsA-AC5kAI0wAbCHBABfAHwAKBM27FFyRs1kdjPZADIcBIkSMBbV5NcRwZ4WCg0ADROyNoAlEoWLOycJgDaAOQSUjL8CQC6fOzeIsj+AK4QIchyyCD5rsrQcenFSuWV0ATaBK34RiB+yHDIALzI+uVqat2i+SAA1iDoYiCjaJhhAHSkFNQgxPrYydKW-EoFEOEA3AQA9GfOyAB6cm0dXcp9A0MjcGOT07PzGOjLq1QaFsdql2Ep4GphEcwqd8Bcrrd7swughnoN8sN5uMpjM5u8Fn8VmRARstm4PGAvOAlAkAA5QdBIYRQwQJE7nS7ORH4AgPMDIQRo15Yz64n6LIlrIEwjkIuRAA

Does that work?

mcintyre94: I suspect it won't - I did something similar here: https://solanalabs.slack.com/archives/C050QNCV6BU/p1691418848861819

[This playground works correctly](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcBKYDGbgGdIB2uwAPACoBqAhgDYCuwAfHALzrBUAmEBNiJAbwBQcUXEw8YoGAC52XHn0Eixq3DQiy4BOgFsARsCgBuFaIC+jU6oButBnMr3gpy6aFIUcAOIAFAIIAwhC6ujzBBABmAJYA5uQA6tEwABYRUiAwzGw4Crz8wqoSocm6wAQwAPxyuDBQ0QSx1mIA7slpktLVcGRJqenSrlZCHsioGNh4hMQAkpF9HRXSie0DmQA0PTMEBEbZPQtr8NLlnLgIUAxwlehYOPg8xOTbu1DMji9G7pF0BJgw0R4PgCK36nUycBOBDOcH0EAgNA4BFYcEitGIjAAFGZVOIeDFYt0-EEQmECBECaDFhkskIAJRyCb3abAOaHcEwKlHTY6Ax7OCFMSpKAQFraYBigCiUBFUExdNcIwkRHgBAgFLicl5higKOJ8tMytq2ggvioUCoui1eh1eoCmIExV0pXKWgA5FQ3eYFUIjfA2mClplGXcpo9SNr+Wx9QIA9TpHI6gxvYbHqr1RzrXzddH7bHVpnUejgCmhEA)

But when I used the same pattern in this codebase it wouldn't work, and just treated the output as like `number | RpcResponse<number>` for any input

I think it might be the same underlying issue as https://github.com/solana-labs/solana-web3.js/issues/1280 but I'm not sure

packages/rpc-core/src/rpc-methods/__tests__/get-program-accounts-test.ts
mcintyre94: We definitely could, we just have a lot of tests already that don't

packages/rpc-core/src/rpc-methods/getProgramAccounts.ts
steveluscher: I fucked around and found out (that there's a TypeScript bug). https://github.com/solana-labs/solana-web3.js/pull/1531

packages/rpc-core/src/rpc-methods/__tests__/get-program-accounts-test.ts
buffalojoec: I sort of think we should. It's an easy lay-up for completing some of these tests and asserting proper functionality. Happy to start this work

PR #1524, opened 2023-08-11
refactor(experimental): add getTokenAccountsByDelegate method

mcintyre94:
This PR adds the `getTokenAccountsByDelegate` RPC function

This is kinda similar to `getProgramAccounts` and co, but only returns token accounts. This means that the tests for `jsonParsed` are sane, which is a nice change.

The tests rely on a few new fixtures:
- spl-token-mint-account-with-delegated.json is a token mint account
- spl-token-token-account-delegated.json is a token account of the above mint account, that has been delegated
- spl-token-mint-no-token-accounts.json is a token mint account with no token accounts (and therefore no delegated ones)

Also note that the solana account decoder currently limits base58 accounts to 128 bytes. As token accounts are >128 bytes this means that the `base58` and no encoding (which falls back to plain base58 bytes) options aren't usable in practice. I decided not to restrict this in web3js because it's just a function of current account sizes/hardcoded limitation, not the RPC call itself, and these could change. But it does mean that some tests can't be written with the current validator.

Ref #1449


packages/rpc-core/src/rpc-methods/common.ts
```diff
@@ -40,3 +45,25 @@ export type Base64EncodedDataResponse = [Base64EncodedBytes, 'base64'];
 export type Base64EncodedZStdCompressedDataResponse = [Base64EncodedZStdCompressedBytes, 'base64+zstd'];
 
 export type Base58EncodedTransactionSignature = string & { readonly __base58EncodedSignature: unique symbol };
+
+type TokenAmount = Readonly<{
+    amount: StringifiedBigInt;
+    decimals: number;
+    uiAmount?: number;
```
buffalojoec, line 52: This is actually `number | null`, which could cause problems if someone expects it to be a valid `number` whenever it's defined, rather than defined but `null`

https://docs.solana.com/api/http#token-balances-structure

packages/rpc-core/src/rpc-methods/getTokenAccountsByDelegate.ts
```diff
@@ -0,0 +1,136 @@
...
+    executable: boolean;
+    /** number of lamports assigned to this account */
+    lamports: LamportsUnsafeBeyond2Pow53Minus1;
+    /** pubkey of the program this account has been assigned to */
+    owner: Base58EncodedAddress;
+    /** the epoch at which this account will next owe rent */
+    rentEpoch: U64UnsafeBeyond2Pow53Minus1;
+    /** the data size of the account */
+    space: U64UnsafeBeyond2Pow53Minus1;
+}>;
```
buffalojoec, line 28: I nudged on this before, but a lot of these methods are starting to redefine the same types over and over again, this one being [`GetAccountInfoApiResponseBase`](https://github.com/solana-labs/solana-web3.js/blob/master/packages/rpc-core/src/rpc-methods/getAccountInfo.ts#L16-L22) or more practically `AccountInfo`

packages/rpc-core/src/rpc-methods/__tests__/get-token-accounts-by-delegate-test.ts
```diff
@@ -0,0 +1,387 @@
...
+            const accountInfoPromise = rpc.getTokenAccountsByDelegate(delegate, { mint }).send();
+
+            await expect(accountInfoPromise).resolves.toMatchObject({
+                value: [],
+            });
+        });
+    });
+
+    describe('when called with a program that is not a Token program', () => {
+        it('throws an error for unrecognized program', async () => {
```
buffalojoec, line 189: This one is kind of funny. So it will only recognize the two token programs for this call?

buffalojoec:
Lgtm! Just check my comment about `number | null`

packages/rpc-core/src/rpc-methods/getTokenAccountsByDelegate.ts
mcintyre94: Yep agreed! I wanted to review `getTokenAccountsByOwner` before poking any further, but it looks like that one actually mirrors this basically exactly. I did put the new data structure here in `common`, and as part of that next PR I'll pull the common data structures there. I think whatever makes sense to extract from what will then be 5 methods counts as common now!

packages/rpc-core/src/rpc-methods/common.ts
mcintyre94: Thankyou, fixed!

packages/rpc-core/src/rpc-methods/getTokenAccountsByDelegate.ts
buffalojoec: Sounds like a good plan!

buffalojoec (approved):

packages/rpc-core/src/rpc-methods/getTokenAccountsByDelegate.ts
```diff
@@ -0,0 +1,136 @@
...
+    >;
+
+    getTokenAccountsByDelegate(
+        program: Base58EncodedAddress,
+        filter: AccountsFilter,
+        config: GetTokenAccountsByDelegateApiCommonConfig &
+            Readonly<{
+                encoding: 'jsonParsed';
+            }>
+    ): RpcResponse<TokenAccountWithPubkey<TokenAccountInfoBase & TokenAccountInfoWithJsonData>[]>;
```
steveluscher, line 119: This needs a base64 fallback, right?

> If jsonParsed is requested but a parser cannot be found, the field falls back to base64 encoding, detectable when the data field is type string.

See:

https://github.com/solana-labs/solana-web3.js/blob/958cc6b5ceaa8eddc79d46ed133e2ceac4593476/packages/rpc-core/src/rpc-methods/common.ts#L80-L91

mcintyre94: I didn't include that because the accounts returned are known to be token accounts owned by a known token program, so AFAIK they will always be parsed. That is dependent on current behaviour of the solana codebase though, so I'm not sure whether relying on that at the type level here is the right thing to do? Happy to add the base64 fallback if you'd prefer or I'm missing something though!

steveluscher: Mmmmm. Smart. Got it!

PR #1525, opened 2023-08-15
refactor(experimental): add getTokenAccountsByOwner method 

mcintyre94:
This PR adds the `getTokenAccountsByOwner` RPC function

The structure and behaviour  is exactly the same as `getTokenAccountsByDelegate`, it just queries for token accounts owned by instead of delegated to a particular address

In this PR I've also refactored out the common aspects of the output data in `getAccountInfo`, `getMultipleAccounts`, `getProgramAccounts`, `getTokenAccountsByDelegate` and `getTokenAccountsByOwner`, which removes a lot of repetition! 

Ref #1449


packages/rpc-core/src/rpc-methods/__tests__/get-token-accounts-by-owner-test.ts
```diff
@@ -0,0 +1,387 @@
...
+
+import { Commitment } from '../common';
+import { createSolanaRpcApi, SolanaRpcMethods } from '../index';
+
+// Change all these to getTokenAccountsByOwner
+// Mint fixture: fixtures/spl-token-mint-account-with-owner.json
+// Fixture: fixtures/spl-token-token-account-owner.json
+// Owner: G4QJANEpvEN8vLaaMZoWwZtqHfWxuWpd5RrVVYXPCgeL
+// Mint: 2nBoNW5B9SdpJYEg9neii7ecCJFwh6UrbXS6HFxkK7Gf
+// Program: TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA
```
buffalojoec, line 15: I don't think I understand what this comment is instructing

buffalojoec (approved):

packages/rpc-core/src/rpc-methods/__tests__/get-token-accounts-by-owner-test.ts
mcintyre94: That would be because it was a note-to-self that was never supposed to be checked in! Sorry, removed!

buffalojoec: Lol no prob 😂

packages/rpc-core/src/rpc-methods/getTokenAccountsByOwner.ts
```diff
@@ -0,0 +1,101 @@
...
+    ): RpcResponse<AccountInfoWithPubkey<AccountInfoBase & AccountInfoWithBase64EncodedZStdCompressedData>[]>;
+
+    getTokenAccountsByOwner(
+        program: Base58EncodedAddress,
+        filter: AccountsFilter,
+        config: GetTokenAccountsByOwnerApiCommonConfig &
+            Readonly<{
+                encoding: 'jsonParsed';
+            }>
+    ): RpcResponse<AccountInfoWithPubkey<AccountInfoBase & TokenAccountInfoWithJsonData>[]>;
```
steveluscher, line 84: Related comment: https://github.com/solana-labs/solana-web3.js/pull/1524/files#r1300774805

PR #1526, opened 2023-08-21
refactor(experimental): add simulateTransaction RPC call

mcintyre94:
This PR adds the `simulateTransaction` RPC call to the experimental web3js

- `commitment` and `minContextSlot` are optional, as in most of these configs. Note that here `commitment` defaults to `finalized`. Like with `sendTransaction` this means that tests specify the commitment because `finalized` may fail shortly after the validator starts up.  
- There are two mutually exclusive optional params: `sigVerify` and `replaceRecentBlockhash`, which both default to false. Their mutual exclusivity is enforced with types
- An optional `encoding` parameter defaults to base58, but should be set to base64 since this is deprecated. As with `sendTransaction` this means that excluding the config object, or not setting `encoding` on it, is deprecated. Most tests use the `base64` version and I haven't duplicated them all with the deprecated `base58` one. 
- There's an optional `accounts` parameter which can be used to request account data for a subset of accounts involved in a transaction. This contains a list of addresses, and an encoding (unrelated to the transaction level one) for those accounts returned. This is responsible for most of the overloads, because this `encoding` affects the structure of the returned `accounts`. Note that `base58` is not supported: https://github.com/solana-labs/solana/blob/7902ac106d1c537149b44143d080247a6ec85891/rpc/src/rpc.rs#L3774 

The tests in this PR are modified from the `sendTransaction` ones, including the specific transaction structure used. 

Ref #1449


packages/rpc-core/src/rpc-methods/simulateTransaction.ts
```diff
@@ -0,0 +1,160 @@
...
+    Base64EncodedDataResponse,
+    Commitment,
+    Slot,
+    U64UnsafeBeyond2Pow53Minus1,
+} from './common';
+
+type SimulateTransactionConfigBase = Readonly<{
+    /**
+     * Commitment level to simulate the transaction at
+     * @default finalized
```
buffalojoec, line 20: Ahh I wonder if we should start tagging all of these configs haha.

btw, looks like it's actually [`defaultValue`](https://typedoc.org/tags/defaultValue/)

packages/rpc-core/src/rpc-methods/simulateTransaction.ts
```diff
@@ -0,0 +1,160 @@
...
+}>;
+
+// Both are optional booleans, but conflict - so cannot both be true
+type SigVerifyAndReplaceRecentBlockhashConfig =
+    | Readonly<{
+          /** if `true` the transaction signatures will be verified (conflicts with `replaceRecentBlockhash`) */
+          sigVerify: true;
+          /** if `true` the transaction recent blockhash will be replaced with the most recent blockhash. (conflicts with `sigVerify`) */
+          replaceRecentBlockhash?: false;
+      }>
```
buffalojoec, line 34: I think since these two fields directly conflict with each other, it's better to provide one or the other, but not both, at a type level.

It doesn't look like these configs affect the return type, so you don't need any generic `<T>` handling, and instead you could use "or" `|` to make sure the configs are mutually exclusive at the typecheck level. [Here's an example](https://www.typescriptlang.org/play?#code/C4TwDgpgBAyglgWwK4BsCGwIBUBOaB2AzmgMbBwD2+AwlQGZwDmAQmodALxQBKEaAJlRQgAPAG8AUFGlQA9ACp5UmVHlRaCBHGAII+YFBQQAbhBRRgFKIUSoM0YAAsHeIqXJUoGZTLUABfgg6NFQDBnw0FDgALwh+H2k1eVkEkgpNbV19AH4ALmtgHDh8RgBuOVkoAHktYEx+KDoKHCgAIxwTbRAEhTUsZygtfFsEaxQKAycMCwGOgEckCEIDEgI26BNIpHsG6eSEodp9CAAPYBhx4DyofCQEVogcctlKmu16xua2juMuiQBfAB8pQkElAkFgTAAao84HQQEcGIwoFxeAIhKJJCpelA4VAAAaFRb4mYuAjEMiUfDWJgRYBIDqEKAAdzgKHMDygpiKDDiUAAFGl8HQomQmaynASOmB0CQILw5fpmOMSABrRxsRz4gCUqhSKhsjBhPJA+SJEBBQNB4OgvBlpHlEEVwGVFDVGsIjkRTBRPD4gnwwnEPUUuLoBPNJKcZLclM8HWdbRV6s1LLZHOg0tlfIljlJgwoyygCb0BlayY9jgAdAKhSK4GK05L8YbjXCQDq9Qksw6FaXXe7Nddgih2JbAaCJHQkPg49SbMh0JhcOT3FT+QlWmwIAA2AAsAFFZxRAvwAOpwDor2MefD5ZZFEoAGgq1VqHyaLXanVAqXoTHyeBF3sa8KVvb0WG3KAADIoDEKA9DSfhikYfIAHIt3Yfc0Kgf4YIFeAjVheEIKgAAfP17TlPslQrTUIO1CRdTEPDXzeOo+U-b4fxAIA). What do you think?

buffalojoec: I think it's confusing as a user that they _can_ provide both configs, but won't really know that one overrides the other until they observe behavior or read the docs

packages/rpc-core/src/rpc-methods/__tests__/simulate-transaction-test.ts
```diff
@@ -0,0 +1,717 @@
...
+                accounts: null,
+                err: 'BlockhashNotFound',
+                logs: expect.any(Array),
+                returnData: null,
+                unitsConsumed: expect.any(BigInt),
+            }),
+        });
+    });
+
+    it('replaces the invalid blockhash when `replaceRecentBlockhash` is true', async () => {
```
buffalojoec, line 303: Just continuing on my comment of the mutually exclusive configs, if we're going to keep the ability to provide both, we should have a test that observes the API's behavior if both are provided. 
Although I think we shouldn't 😝

packages/rpc-core/src/rpc-methods/__tests__/simulate-transaction-test.ts
```diff
@@ -0,0 +1,717 @@
...
+                accounts: null,
+                err: 'AccountNotFound',
+                logs: expect.any(Array),
+                returnData: null,
+                unitsConsumed: expect.any(BigInt),
+            }),
+        });
+    });
+
+    it('returns account data for a transaction account with base64 encoding', async () => {
```
buffalojoec, line 440: Just curious - bc the docs aren't super clear - this returns the account data post-simulation? In other words, it returns the simulated change in state for any provided keys? Pretty cool if true, didn't know that

packages/rpc-core/src/rpc-methods/simulateTransaction.ts
mcintyre94: TY! Probably not worth adding it to all of them, just improve docstrings as we're working on stuff over time. But I should probably fix the 4 other places I've used `@default`!

packages/rpc-core/src/rpc-methods/__tests__/simulate-transaction-test.ts
mcintyre94: That's a type error:

Screenshot 2023-08-21 at 18 14 02
Screenshot 2023-08-21 at 18 14 14

FWIW the RPC message is:
```
    "error": {
        "code": -32602,
        "message": "sigVerify may not be used with replaceRecentBlockhash"
    },
```

But I think a type error is definitely better!

packages/rpc-core/src/rpc-methods/__tests__/simulate-transaction-test.ts
mcintyre94: I haven't used this API much but IIRC yep it's the account after the transaction. I'm pretty sure I used this in a fake-wallet to show the change in SOL that would result from approving a transaction, by comparing lamports with a `getAccountInfo` or similar call against the same account.

mcintyre94:
Moved this to draft - this API can return `jsonParsed` accounts so need to add to the allowed numeric accounts

Should probably do after #1528 to make it less annoying!

packages/rpc-core/src/rpc-methods/__tests__/simulate-transaction-test.ts
buffalojoec: Interesting!

packages/rpc-core/src/rpc-methods/simulateTransaction.ts
buffalojoec: I'm moreso suggesting if you provide both as `true`, do you get a type error? Or do you have to wait for an RPC error? I think if it's the latter, we should make it the former.

However, seeing your other comment I think this is what I was getting after. As long as you get a type error for providing both I'm good with it!

buffalojoec: Also I just realized I fat-fingered my typescript playground and it doesn't even throw a type error lol

packages/rpc-core/src/rpc-methods/__tests__/simulate-transaction-test.ts
buffalojoec: Yeah, that's a rough error. But nonetheless it is a type error!

I'm starting to think _a lot_ of our type errors for parameters are going to be balls of yarn to decode for the unassuming developer. Perhaps later we can comb through and try to reduce suffering across many parameter types.

Regardless, making sure we got a type error instead of RPC error was my main concern, thanks for clarifying with this example!

buffalojoec:
lgtm but summoning @steveluscher

steveluscher:
> Their mutual exclusivity is enforced with types

<3

packages/rpc-core/src/rpc-methods/simulateTransaction.ts
```diff
@@ -0,0 +1,160 @@
...
+        replaceRecentBlockhash: true;
+        /** if `true` the transaction signatures will be verified (conflicts with `replaceRecentBlockhash`) */
+        sigVerify?: false;
+    }>
+    | Readonly<{
+        /** if `true` the transaction signatures will be verified (conflicts with `replaceRecentBlockhash`) */
+        sigVerify?: false;
+        /** if `true` the transaction recent blockhash will be replaced with the most recent blockhash. (conflicts with `sigVerify`) */
+        replaceRecentBlockhash?: false;
+    }>;
```
steveluscher, line 46: damn-so-good-perfection

packages/rpc-core/src/rpc-methods/__tests__/simulate-transaction-test.ts
```diff
@@ -0,0 +1,717 @@
...
+import { createHttpTransport, createJsonRpc } from '@solana/rpc-transport';
+import { SolanaJsonRpcErrorCode } from '@solana/rpc-transport/dist/types/json-rpc-errors';
+import type { Rpc } from '@solana/rpc-transport/dist/types/json-rpc-types';
+import { Base64EncodedWireTransaction } from '@solana/transactions';
+import fetchMock from 'jest-fetch-mock-fork';
+
+import { Base58EncodedBytes, Commitment } from '../common';
+import { createSolanaRpcApi, SolanaRpcMethods } from '../index';
+
+function getMockTransactionMessage({
```
steveluscher, line 12: Related comment: https://github.com/solana-labs/solana-web3.js/pull/1504#discussion_r1300697601

packages/rpc-core/src/rpc-methods/__tests__/simulate-transaction-test.ts
```diff
@@ -0,0 +1,717 @@
...
+                        ...message,
+                    ])
+                ).toString('base64') as Base64EncodedWireTransaction,
+                { commitment: 'processed', encoding: 'base64' }
+            )
+            .send();
+
+        await expect(resultPromise).rejects.toMatchObject({
+            code: -32602 satisfies (typeof SolanaJsonRpcErrorCode)['JSON_RPC_INVALID_PARAMS'],
+            message: expect.stringContaining('expected a valid transaction message version'),
```
steveluscher, line 368: This might need to change now. See https://github.com/solana-labs/solana-web3.js/pull/1534/files#diff-a9e1b8ed269c088358fa54dabec9128cf162fd8c1a1f91d09fd4d8a0ea5edff0

packages/rpc-core/src/rpc-methods/__tests__/simulate-transaction-test.ts
```diff
@@ -0,0 +1,717 @@
...
+                        ...message,
+                    ])
+                ).toString('base64') as Base64EncodedWireTransaction,
+                { commitment: 'processed', encoding: 'base64' }
+            )
+            .send();
+
+        await expect(resultPromise).rejects.toMatchObject({
+            code: -32602 satisfies (typeof SolanaJsonRpcErrorCode)['JSON_RPC_INVALID_PARAMS'],
+            message: expect.stringContaining('failed to fill whole buffer'),
```
steveluscher, line 393: 🙄

packages/rpc-core/src/rpc-methods/__tests__/simulate-transaction-test.ts
```diff
@@ -0,0 +1,717 @@
...
+                ],
+                err: null,
+                logs: expect.any(Array),
+                returnData: null,
+                unitsConsumed: expect.any(BigInt),
+            }),
+        });
+    });
+
+    it('returns account data for a transaction account with jsonParsed encoding', async () => {
```
steveluscher, line 534: Add a test for ‘asked for jsonParsed but fell back to base64’?

steveluscher (approved):
giphy-2

packages/rpc-core/src/rpc-methods/__tests__/simulate-transaction-test.ts
mcintyre94: Probably the same conclusion here - we need to test with different blockhashes

packages/rpc-core/src/rpc-methods/__tests__/simulate-transaction-test.ts
mcintyre94: That's actually the test I wrote here, there's no JSON parser for the account I'm testing. The missing one is one that actually does parse as JSON, I'll see how easily I can add that.

mcintyre94: Got both now!

packages/rpc-core/src/rpc-methods/__tests__/simulate-transaction-test.ts
```diff
@@ -0,0 +1,830 @@
...
+                    encoding: 'base64',
+                }
+            )
+            .send();
+
+        await expect(resultPromise).resolves.toMatchObject({
+            value: expect.objectContaining({
+                accounts: [
+                    expect.objectContaining({
+                        // falls back to base64
```
steveluscher, line 619: This comment is out of place, right?

mcintyre94: Oops yep, removed!

PR #1527, opened 2023-08-21
refactor(experimental): fix micro lamport type symbol

buffalojoec:


mcintyre94 (approved):
Looks good!

@steveluscher out of interest why do we need to use different brands for these? [This playground](https://www.typescriptlang.org/play?target=7#code/C4TwDgpgBAMghgWzAewE7AM5QLxQEYCWA5gQHbBQBkUA3lKhHACbKkA2IUA+l24iugwAuKAFdSBAI6joGEAjzI2UAL4BuALAAoUJCgBZAgGNUyeEjSYc+YmQrU6DZqw7de-S8LETps+YuV1bW0AM3EjYAJWMQwIcwFMAAo2EXjPAEpaFWCtI1YMCgQRQxMzD0FrAEZSKDgsEtM0wU0tUVimpIR0tSA) re-uses the symbol and has a type error when you use the wrong one

steveluscher:
Oh yeah! You're right, because of that `unique symbol` bit. Maybe we should change all of the property names to just be `__brand` then, everywhere.

buffalojoec:
> Oh yeah! You're right, because of that `unique symbol` bit. Maybe we should change all of the property names to just be `__brand` then, everywhere.

From my understanding, the main purpose of `& { readonly __brand: unique symbol };` is to distinguish this type from another that may similar or otherwise the same? ie. `Lamports` and `MicroLamports` which are based on `bigint`

Lol "brand" sounds very... ant-computers 🫢
`__identifier`? `__id`?  `__typeId`? I think `__symbol` would also work!

steveluscher:
It's weird, but they're called ‘Branded Types’ in TypeScript. https://github.com/kourge/ts-brand

PR #1528, opened 2023-08-22
Refactor (experimental): Tweaks to jsonParsed account encoding

mcintyre94:
This PR addresses some of the comments for the various methods that fetch account info

- Refactor the allowed numeric types list, to use a shared list of `jsonParsed` account fields. The various functions that return this account data now map over this list and add the relevant prefixes for their return structure
- Update `getMultipleAccounts` types to include that `null` is returned for accounts that aren't found


steveluscher:
Wrapped this puppy in a memoized function so that the library still remains tree-shakable. You could go even further and change the signature to like `getAllowedNumericKeypathsForMethod(methodName: ...)` and then have a `switch` statement in there that prepares _just_ the right config, then memoizes that.

This is probably fine though.

PR #1530, opened 2023-08-22
refactor(experimental): better build command

2501babe:
currently `pnpm build` runs the full test suite, which is very not conducive to working with the library as an external dependency. it makes it hard to tell when the compile is actually done and the rest can be aborted, and also it kills any `solana-test-validator` running and leaves behind its own after SIGINT

this changes the build command to just build

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     ],
     "scripts": {
         "build": "turbo run build",
-        "compile": "turbo run compile:js && turbo run compile:typedefs",
+        "compile": "turbo run compile:js compile:typedefs",
         "lint": "turbo run test:lint",
         "publish-packages": "turbo run publish-packages",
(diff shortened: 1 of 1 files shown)


2501babe:
oops yea, ive pushed a change just adding compile as a new command instead

2501babe:
i cant merge because of repo permissions so feel free to, ty!

package.json
```diff
@@ -5,6 +5,7 @@
     ],
     "scripts": {
         "build": "turbo run build",
+        "compile": "turbo run compile:js && turbo run compile:typedefs",
```
steveluscher, line 8: This doesn't parallelize the work, but this does!

```diff
-         "compile": "turbo run compile:js && turbo run compile:typedefs",
+         "compile": "turbo run compile:js compile:typedefs",
```

PR #1531, opened 2023-08-22
refactor(experimental): reduce `getProgramAccounts` types using generics

steveluscher:
refactor(experimental): reduce `getProgramAccounts` types using generics


mcintyre94:
This is awesome! Do you have any idea how tricky the fix would be on the Typescript side?

steveluscher:
I'm not sure. I feel like it's a straightforward bug with `infer`, but nothing ever is.

It infers _just fine_ if you:

1. Supply `<any>` for the type parameter
2. Don't have the return type depend on the type parameter

I feel like it should absolutely infer the full range of possible types when the type parameter is not supplied.

PR #1534, opened 2023-08-23
test: rewrite tests to target `latest` validator, as of today

steveluscher:
test: rewrite tests to target `latest` validator, as of today

# Summary

The only reason these tests succeed is:

1. You're running 1.17.x locally
2. That test result gets cached in Turborepo

Rather, we should all be running the same validator as is running in CI.

In this PR, we eliminate all of the tested values that are not yet present in 1.14.

# Test Plan

```
pnpm test:live-with-test-validator:setup
pnpm turbo test:unit:node --force
```


steveluscher:
Good idea? Bad idea?

buffalojoec (approved):
Definitely a good idea. We want to stay consistent with mainnet I'm assuming?

I had no idea 1.14 had no `space` field, that's strange

mcintyre94:
Should we remove `space` from the `getAccountInfo` etc. response objects too?

https://github.com/solana-labs/solana-web3.js/blob/b9efb9c46cbbe69e8a48c11ffb426b24466a51ca/packages/rpc-core/src/rpc-methods/common.ts#L59

steveluscher:
bfb1537a149ea525ba28da476b11155fcda23c92

PR #1535, opened 2023-08-23
test: get genesis hash from logfile with `readLines` API

steveluscher:
test: get genesis hash from logfile with `readLines` API


buffalojoec (approved):
Awesome, thanks for throwing this together!

mcintyre94 (approved):
Nice, much simpler!

PR #1536, opened 2023-08-24
test: get version info from logfile with `readLines` API

buffalojoec:
test: get genesis hash from logfile with `readLines` API


steveluscher (approved):
giphy-2

PR #1538, opened 2023-08-25
refactor(experimental): added unsafe f64 type

buffalojoec:
refactor(experimental): added unsafe f64 type

### DIFF
--- a/packages/rpc-core/src/rpc-methods/common.ts
+++ b/packages/rpc-core/src/rpc-methods/common.ts
@@ -29,6 +29,15 @@ export type Slot = U64UnsafeBeyond2Pow53Minus1;
 // the JSON-RPC transport.
 export type U64UnsafeBeyond2Pow53Minus1 = bigint;
 
+// FIXME(solana-labs/solana/issues/30341)
+// <https://stackoverflow.com/questions/45929493/node-js-maximum-safe-floating-point-number/57225494#57225494>
+// Beware that floating-point value precision can vary widely:
+// - For precision of 1 decimal place, anything above 562949953421311
+// - For precision of 2 decimal places, anything above 70368744177663
+// can be truncated or rounded because of a downcast to JavaScript `number` between your calling
+// code and the JSON-RPC transport.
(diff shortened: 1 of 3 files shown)


packages/rpc-core/src/rpc-methods/common.ts
```diff
@@ -29,6 +29,15 @@ export type Slot = U64UnsafeBeyond2Pow53Minus1;
...
 export type U64UnsafeBeyond2Pow53Minus1 = bigint;
 
+// FIXME(solana-labs/solana/issues/30341)
+// <https://stackoverflow.com/questions/45929493/node-js-maximum-safe-floating-point-number/57225494#57225494>
+// Beware that floating-point value precision can vary widely:
+// - For precision of 1 decimal place, anything above 562949953421311
+// - For precision of 2 decimal places, anything above 70368744177663
+// can be truncated or rounded because of a downcast to JavaScript `number` between your calling
+// code and the JSON-RPC transport.
+export type F64UnsafeSingleDecimalPointBeyond2Pow53Over16Minus1 = number;
```
steveluscher, line 39: I mean, I love it, but probably better to just be like `F64UnsafeSeeCodeCommentForWhy` or something :)

steveluscher (approved):
giphy-1

packages/rpc-core/src/rpc-methods/common.ts
buffalojoec: I figured the longer the name the more obvious of a warning it is. Your suggestion is also quite long

mcintyre94: `__F64_UNSAFE___DO_NOT_USE_OR_YOU_WILL_BE_FIRED`

mcintyre94 (approved):

packages/rpc-core/src/rpc-methods/common.ts
buffalojoec: I went with `F64UnsafeSeeDocumentation`

PR #1543, opened 2023-08-25
refactor(experimental): add opaque type for serialized message

buffalojoec:
This PR adds an opaque type for serialized transaction messages.

You can see there's one type for serialized bytes in binary format (`Uint8Array`) and there's another for `base64`.


steveluscher (approved):
giphy

buffalojoec:
Just updated symbol names to be different

steveluscher:
> Just updated symbol names to be different

I'm also still not opposed to changing them to `__brand` everywhere, so that we don't even have to think about it anymore.

buffalojoec:
> > Just updated symbol names to be different
> 
> I'm also still not opposed to changing them to `__brand` everywhere, so that we don't even have to think about it anymore.

I just made a note for this. Will submit in a new PR

PR #1544, opened 2023-08-26
test: use fixtures for account data

buffalojoec:
This PR adds the fixed data from our account fixtures into various account-oriented tests to validate the response data.

It appears to me that the test validator is using a random keypair for the stake account, and it doesn't emit this account to a JSON file anywhere within the `test-ledger` folder. So we can't read this address from JSON. Exploring alternatives.

Related PR: https://github.com/solana-labs/solana/pull/33014


packages/rpc-core/src/rpc-methods/__tests__/get-account-info-test.ts
```diff
@@ -34,11 +34,11 @@ describe('getAccountInfo', () => {
...
-                        data: expect.any(String),
-                        executable: expect.any(Boolean),
-                        lamports: expect.any(BigInt),
-                        owner: expect.any(String),
-                        rentEpoch: expect.any(BigInt),
+                        data: '2Uw1bpnsXxu3e',
+                        executable: false,
+                        lamports: 5000000n,
+                        owner: '11111111111111111111111111111111',
+                        rentEpoch: 18446744073709551616n, // TODO: This number loses precision
```
steveluscher, line 41: In #1534, I changed CI to run against `stable` (ie. 1.14.24 as of now) which I think emits `0` for this value.

1. If you're not running `stable` locally when you run these tests, you should nuke your validator and re-run `pnpm test:live-with-test-validator:setup`
2. I should figure out a way to install the `stable` validator _locally_ (ie. in the repo, rather than globally) so that you don't have to nuke your global copy

packages/rpc-core/src/rpc-methods/__tests__/get-token-accounts-by-owner-test.ts
```diff
@@ -314,33 +344,40 @@ describe('getTokenAccountsByOwner', () => {
...
-                                        state: expect.any(String),
-                                        tokenAmount: {
-                                            amount: expect.any(String),
-                                            decimals: expect.any(Number),
-                                            uiAmount: expect.any(Number),
-                                            uiAmountString: expect.any(String),
-                                        },
-                                    }),
-                                    type: 'account',
+            expect(accountInfos).toMatchObject({
```
steveluscher, line 347: Oh, why no more `expect(...).resolves.toMatchObject`? When you write it that way, Jest catches and formats exceptions in the case that the promise rejects. When written the way you have it here, the exception just blows up all over the test body itself.

steveluscher (approved):
giphy-1

packages/rpc-core/src/rpc-methods/__tests__/get-account-info-test.ts
buffalojoec: The second bullet shouldn't be too bad. I can mess around with it.
I'm thinking something like building it from source and referencing the compiled Rust binary directly ie:

```shell
target/debug/solana-test-validator
```

steveluscher: I _think_ the install script has a `--data-dir` option, but I haven't tried it yet. Give it a shot? We should just install into like `./.solana/` and then add `.solana/` to the `.gitignore`.

https://github.com/solana-labs/solana/blob/4d452fc5e9fd465c50b2404354bbc5d84a30fbcb/install/solana-install-init.sh#L38

steveluscher: In any case we probably don't want to build from source because that would be slow and make it difficult for people to contribute to web3.js if they don't have the whole compilation toolchain installed.

buffalojoec: So I just finally circled back to this and I realized this _doesn't_ emit a 0, and maybe it's because our fixture has a value for rentEpoch?

buffalojoec:
@steveluscher curious if you could maybe tell me what I'm missing for the test matcher I added?

It seems to work in the tests just fine, but the typecheck fails and says the check isn't part of `Expect`.

Tried to mimic your `toBeFrozenObject()` check, but on an object field

packages/rpc-core/src/rpc-methods/__tests__/get-account-info-test.ts
buffalojoec: > I _think_ the install script has a `--data-dir` option, but I haven't tried it yet. Give it a shot? We should just install into like `./.solana/` and then add `.solana/` to the `.gitignore`.

This worked like a charm

steveluscher:
1. Maybe this approach instead? https://stackoverflow.com/a/51136198/802047
2. Did you `pnpm turbo compile:typedefs` after adding your matcher?

mcintyre94:
The `rentEpoch` value in these fixtures is just copied from some devnet or mainnet account. We could just change the fixtures to 0 if you want to, or we can leave it as-is and it'll be fixed when we fix bigint responses from the RPC.

buffalojoec:
I ended up just dropping the custom matcher and omitting `rootSlot`, since most of the forums and stuff I was finding kept saying "ackshually you don't want to have unpredictable null" and jest has intentionally skirted around this as a matcher for expectedly good reason

buffalojoec:
OK I had to take out _some_ of the hard-coded fixtures/values for the `get_largest_accounts` tests. But now this thing should be ready to rock n' roll. 

Our CI seems to be behaving pretty good lately so wanted to make sure this thing didn't nuke it all.

If anyone wants to give it another look @mcintyre94 @steveluscher give it a rip

mcintyre94 (commented):
This generally looks good, but there are a lot of tests where I think changing to `arrayContaining` checks is making them quite a bit less strict, since it just checks for a subset and doesn't care how long the actual array you're asserting against is. 

Where we have an array of length 1 and we fully know the data, could we just do an array equality check?

For the ones with a longer length, I'm not sure if the return order is consistent? If it is then again can we do an array equality? And if it isn't then I think it'd be better to assert on the length too.

buffalojoec:
> This generally looks good, but there are a lot of tests where I think changing to `arrayContaining` checks is making them quite a bit less strict, since it just checks for a subset and doesn't care how long the actual array you're asserting against is.
> 
> Where we have an array of length 1 and we fully know the data, could we just do an array equality check?

Yes, we can definitely do this. I observed some variable-length behavior in a lot of the tests _outside of_ `getMultipleAccounts`, so perhaps we leave those as `arrayContaining`, but when we're querying a specific number of addresses, I think it's fair to go with the fixed equality check. Maybe I changed too many of them to `arrayContaining` when my tests kept failing (?).
 
> For the ones with a longer length, I'm not sure if the return order is consistent? If it is then again can we do an array equality? And if it isn't then I think it'd be better to assert on the length too.

The only ones where we can assert the length are `getMultipleAccounts`. Anything else it seems to vary wildly. Take `getProgramAccounts` for example, there's other tests firing that may create accounts at different moments in time, and this test _may_ or _may not_ pick them up.

buffalojoec:
> The problem I was facing with this was the ordering was seldom preserved. I went with `arrayContaining` so it wouldn't care about the order. Maybe I just also add a length check?

Actually, this ordering should be preserved in `getMultipleAccounts`, so I've added a commit for that.

mcintyre94:
> I observed some variable-length behavior in a lot of the tests outside of getMultipleAccounts, so perhaps we leave those as arrayContaining, but when we're querying a specific number of addresses, I think it's fair to go with the fixed equality check.

Ah yep you're right, I remember that we can't control the program ID of some of the tests because the RPC only renders accounts owned by those programs with `jsonParsed`. So makes sense that we're limited on those, and agree that just using `arrayContaining` there and ignoring the rest makes sense.

> The problem I was facing with this was the ordering was seldom preserved. I went with arrayContaining so it wouldn't care about the order. Maybe I just also add a length check?

I think that makes sense for `getMultipleAccounts`, annoying that the order on that one isn't preserved. Annoyingly I guess we'll have to switch it back to awaiting the promise to be able to assert twice? Unless there's some trick I don't know for that!

buffalojoec:
I think this is gtg, let's see what CI says on a force p

mcintyre94 (approved):

PR #1549, opened 2023-08-29
refactor(experimental): The WebSocket implementation now waits until the buffer is empty to send new messages

steveluscher:
refactor(experimental): The WebSocket implementation now waits until the buffer is empty to send new messages

# Summary

I have this sneaking suspicion that _this_ is the source of a lot of grief, particularly for long-lived server-side applications. WebSocket connections can buffer data provided to `send()` if the network is such that it can't handle it. Crucially, if this buffer fills up, the client will abruptly kill the websocket.

Instead of that, let's make `send()` suspend execution in cases where the buffer is above some high watermark. Let's debate the watermark, but start the default at zero.

# Test Plan

```
cd packages/rpc-transport
pnpm test:unit:browser
pnpm test:unit:node
```


packages/rpc-transport/src/transports/websocket/websocket-transport.ts
```diff
@@ -8,6 +8,10 @@ type Config = Readonly<{
 export function createWebSocketTransport({ url }: Config): IRpcWebSocketTransport {
     return async function ({ signal }: Parameters<IRpcWebSocketTransport>[0]) {
         signal.throwIfAborted();
-        return (await createWebSocketConnection({ signal, url })) as RpcWebSocketConnection;
+        return (await createWebSocketConnection({
+            sendBufferHighWatermark: 0,
```
steveluscher, line 12: I really don't know what the right number is for this. I can't find _anything_ that says how large Node or the browser's buffer is (ie. how much `send()` data it will buffer before killing the `WebSocket`).

packages/rpc-transport/src/transports/websocket/__tests__/websocket-connection-test.ts
```diff
@@ -1,7 +1,10 @@
 import WS from 'jest-websocket-mock';
+import { Client } from 'mock-socket';
 
 import { createWebSocketConnection, RpcWebSocketConnection } from '../websocket-connection';
 
+const MOCK_SEND_BUFFER_HIGH_WATERMARK = 42069;
```
buffalojoec, line 6: Nice 😉

packages/rpc-transport/src/transports/websocket/websocket-connection.ts
```diff
@@ -54,11 +60,48 @@ export async function createWebSocketConnection({ signal, url }: Config): Promis
...
+                                    // TODO: Coded error
+                                    new Error('WebSocket was closed before payload could be sent')
+                                );
+                            };
+                        });
+                        bufferDrainWatcher = {
+                            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+                            // @ts-ignore
+                            onCancel,
+                            promise,
```
buffalojoec, line 99: TypeScript yells at you for this?

packages/rpc-transport/src/transports/websocket/websocket-transport.ts
buffalojoec: I don't see much either, except for some huge numbers:

https://itecnote.com/tecnote/node-js-way-to-limit-the-maximum-size-of-a-websocket-message/
https://stackoverflow.com/questions/24153996/is-there-a-limit-on-the-size-of-a-string-in-json-with-node-js/24154610#24154610

buffalojoec (approved):
Lgtm, but we should prob land on the watermark number

packages/rpc-transport/src/transports/websocket/websocket-connection.ts
steveluscher: The problem here is that:

1. Promise callbacks run synchonously
2. I know that
3. TypeScript pretends that it doesn't know that

If you search around their GitHub you'll find lots of complaints about that.

steveluscher:
I really feel like I need to put a number in for the highwatermark other than zero, otherwise basic things like ‘starting a shitton of transaction confirmations’ will create this massive queue.

OTOH, maybe people starting tons of transaction confirmations is _the thing_ that takes down their websockets. ¯\_(°ペ)_/¯

steveluscher:
OK, look at me, punting on deciding a value.

buffalojoec:
> OK, look at me, punting on deciding a value.

This value sounds like it's more of a guardrail than anything performance-inhibitive, no? In other words, "here, we'll cap you at this watermark so you don't blow your connections" but we could also say "good luck (Taken guy voice)"

PR #1563, opened 2023-09-02
refactor(experimental): implement getBase58EncodedAddressForSeed

2501babe:
doing this now because its a blocker for me. straightforward reimpl of `createWithSeed`:

```
  static async createWithSeed(
    fromPublicKey: PublicKey,
    seed: string,
    programId: PublicKey,
  ): Promise<PublicKey> {
    const buffer = Buffer.concat([
      fromPublicKey.toBuffer(),
      Buffer.from(seed),
      programId.toBuffer(),
    ]);
    const publicKeyBytes = sha256(buffer);
    return new PublicKey(publicKeyBytes);
  }
```

not ready to merge as-is, pls let me know:
* how to run `lint:fix`, theres no command for it so i have no idea
* if you want this in a new file
* any style complaints, i tried to copy existing code as much as possible

### DIFF
--- a/packages/addresses/src/__tests__/computed-address-test.ts
+++ b/packages/addresses/src/__tests__/computed-address-test.ts
@@ -167,6 +167,7 @@ describe('createAddressWithSeed', () => {
     it('fails with a malicious programAddress meant to produce an address that would collide with a PDA', async () => {
         expect.assertions(1);
         const baseAddress = 'Bh1uUDP3ApWLeccVNHwyQKpnfGQbuE2UECbGA6M4jiZJ' as Base58EncodedAddress;
+        // The ending bytes of this address decode to the ASCII string 'ProgramDerivedAddress'
         const programAddress = '4vJ9JU1bJJE96FbKdjWme2JfVK1knU936FHTDZV7AC2' as Base58EncodedAddress;
 
         await expect(createAddressWithSeed({ baseAddress, programAddress, seed: 'seed' })).rejects.toThrow(
--- a/packages/addresses/src/__tests__/computed-address-test.ts
+++ b/packages/addresses/src/__tests__/computed-address-test.ts
@@ -1,5 +1,5 @@
 import { Base58EncodedAddress } from '../base58';
-import { getProgramDerivedAddress } from '../program-derived-address';
+import { createAddressWithSeed, getProgramDerivedAddress } from '../computed-address';
 
 describe('getProgramDerivedAddress()', () => {
     it('fatals when supplied more than 16 seeds', async () => {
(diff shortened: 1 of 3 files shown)


buffalojoec:
> * how to run `lint:fix`, theres no command for it so i have no idea

This works from root:

```shell
pnpm eslint --fix packages/addresses/src/__tests__/base58-test.ts
```

Curious what you think of `createBase58AddressWithSeed`? To match original

2501babe:
>This works from root
this doesnt run the other linter... it turns out the command was `pnpm prettier -w src/base58.ts`. if theres a way to run both linters on all files in write mode, that would be a worthwhile command to add to toplevel `package.json`

>Curious what you think of createBase58AddressWithSeed? To match original
sounds good!

2501babe:
oh my god theres _three_ linters ```/home/hana/work/solana/web3/packages/addresses/src/__tests__/base58-test.ts
  1:1  error  Run autofix to sort these imports!  simple-import-sort/imports
```

**Review by buffalojoec — COMMENTED** [review]
This looks good and definitely matches the legacy library's implementation, however I wonder why the legacy lib doesn't look more like the Rust side of things.

https://github.com/solana-labs/solana/blob/adee97fe38b175ae185d34a7aa71b2e7bfd570e8/sdk/program/src/pubkey.rs#L206-L224

Why doesn't it have the checks that `solana_program` does (`MAX_SEED_LENGTH`, `PDA_MARKER`), any idea?

File: packages/addresses/src/base58.ts [hunk 1]
```diff
@@ -52,3 +52,18 @@ export function getBase58EncodedAddressComparator(): (x: string, y: string) => n
...
     }).compare;
 }
+
+export async function createBase58AddressWithSeed(
+    baseAddress: Base58EncodedAddress,
+    seed: string,
+    programAddress: Base58EncodedAddress
+): Promise<Base58EncodedAddress> {
+    const { serialize, deserialize } = getBase58EncodedAddressCodec();
+    const seedBytes = new TextEncoder().encode(seed);
```
steveluscher, line 62: I was going to point out that `TextEncoder` isn't available in React Native, but:

* We also use it in `program-derived-address.ts`
* We should presume that, in React Native, we need to call out to an async API to do this encoding, but you've already made this method `async` so… carry on.

packages/addresses/src/__tests__/base58-test.ts
```diff
@@ -132,4 +133,16 @@ describe('base58', () => {
             ]);
         });
     });
+    describe('createBase58AddressWithSeed', () => {
+        it('works', async () => {
```
steveluscher, line 137: For the benefit of the next reader who doesn't know what this is supposed to do in the first place.

```suggestion
        it('returns an address that is the SHA-256 hash of the concatenated base address, seed, and program address', async () => {
```

packages/addresses/src/base58.ts
```diff
@@ -52,3 +52,18 @@ export function getBase58EncodedAddressComparator(): (x: string, y: string) => n
         usage: 'sort',
     }).compare;
 }
+
+export async function createBase58AddressWithSeed(
```
steveluscher, line 56: For pedantry:

```suggestion
export async function createBase58EncodedAddressWithSeed(
```

steveluscher (changes requested):
> …I wonder why the legacy lib doesn't look more like the Rust side of things.

Yeah, 100%. I remember seeing that code and being like ‘hmm, we should roll that in.’ Add to this implementation:

* A check for max seed length like we did for [program-derived-address](https://github.com/solana-labs/solana-web3.js/blob/6a124b1e580e104c00f9b41c0585466d7737c948/packages/addresses/src/program-derived-address.ts#L31-L34)
* The Rust-style check for a forged PDA marker. What you're trying to protect against is this super weird edge case: `create(baseAddress, seed, '4vJ9JU1bJJE96FbKdjWme2JfVK1knU936FHTDZV7AC2')`. The last 21 bytes of `4vJ9JU1bJJE96FbKdjWme2JfVK1knU936FHTDZV7AC2` just _happen_ to decode to the ASCII string `ProgramDerivedAddress`.

Just one other thought: I think this is one of those APIs where a struct arg would be better. See `createProgramDerivedAddress(pdaInput: PdaInput)`.

packages/addresses/src/base58.ts
steveluscher: Akshually; I just renamed everything to `*Address*` didn't I? So `createAddressWithSeed` is the thing that would match everything else now.

2501babe:
ok i
* renamed the function to `createAddressWithSeed`
* renamed `program-derived-address*.ts` to `computed-address*.ts` and moved the functionality there since they both require the same constants and have a certain harmony
* added guardrails as in rust and tests for them

2501babe:
note ci failure is due to a bug in node https://github.com/nodejs/node/issues/49497

steveluscher (approved):
giphy-7

PR #1568, opened 2023-09-04
refactor(experimental): add lint:fix

2501babe:
this creates a toplevel command to run the linters, feel free to chastize me if ive invoked to tools in a scurrilous manner

### DIFF
--- a/packages/addresses/package.json
+++ b/packages/addresses/package.json
@@ -38,6 +38,7 @@
         "dev": "jest -c node_modules/test-config/jest-dev.config.ts --rootDir . --watch",
         "prepublishOnly": "version-from-git --no-git-tag-version --template experimental.short",
         "publish-packages": "pnpm publish --tag experimental --access public --no-git-checks",
+        "style:fix": "pnpm eslint --fix src/* && pnpm prettier -w src/*",
         "test:lint": "jest -c node_modules/test-config/jest-lint.config.ts --rootDir . --silent",
         "test:prettier": "jest -c node_modules/test-config/jest-prettier.config.ts --rootDir . --silent",
         "test:treeshakability:browser": "agadoo dist/index.browser.js",
--- a/turbo.json
+++ b/turbo.json
@@ -41,6 +41,10 @@
             "dependsOn": ["build", "^publish-packages"],
             "outputs": []
         },
+        "style:fix": {
+            "inputs": ["src/**"],
+            "outputs": []
+        },
         "test:lint": {
             "inputs": ["src/**", "test/**"],
             "outputs": []
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
         "compile": "turbo run compile:js compile:typedefs",
         "lint": "turbo run test:lint",
         "publish-packages": "turbo run publish-packages",
+        "style:fix": "turbo run style:fix",
         "test": "turbo run test:unit:browser test:unit:node",
         "test:live-with-test-validator": "turbo run test:live-with-test-validator",
         "test:live-with-test-validator:setup": "./scripts/setup-test-validator.sh"
--- a/packages/assertions/package.json
+++ b/packages/assertions/package.json
@@ -38,6 +38,7 @@
         "dev": "jest -c node_modules/test-config/jest-dev.config.ts --rootDir . --watch",
         "prepublishOnly": "version-from-git --no-git-tag-version --template experimental.short",
         "publish-packages": "pnpm publish --tag experimental --access public --no-git-checks",
+        "style:fix": "pnpm eslint --fix src/* && pnpm prettier -w src/*",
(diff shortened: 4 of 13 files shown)


turbo.json
```diff
@@ -36,6 +36,10 @@
             "inputs": ["rollup.config.types.js", "tsconfig.*", "src/**"],
             "outputs": ["declarations/**", "dist/**/*.d.ts", "lib/**/*.d.ts"]
         },
+        "lint:fix": {
+            "inputs": ["src/**", "test/**"],
```
mcintyre94, line 40: Should we be passing `test/*` to the `eslint --fix` and `prettier -w` calls too?

packages/addresses/package.json
```diff
@@ -36,6 +36,7 @@
         "compile:js": "tsup --config build-scripts/tsup.config.library.ts",
         "compile:typedefs": "tsc -p ./tsconfig.declarations.json",
         "dev": "jest -c node_modules/test-config/jest-dev.config.ts --rootDir . --watch",
+        "lint:fix": "pnpm eslint --fix src/* && pnpm prettier -w src/*",
```
steveluscher, line 39: Running these serially kneecaps Turborepo from giving us MAX PARALLELISM. Make `test:lint:fix` and `test:prettier:fix` and then make a root task `"style:fix": "turbo run test:lint:fix test:prettier:fix"`

steveluscher (approved):
Small nit, otherwise thanks!

packages/addresses/package.json
mcintyre94: I was going to suggest this too but what does turborepo do if you run two jobs in parallel that modify the same set of files? Is it smart enough to make that work without them ever colliding on the same file?

steveluscher: I was about to write something something race conditions something something file locks but now that I think about it for two seconds, Prettier _has_ to run last in case Eslint has fixes.

Yeah OK ignore me parallelism doesn't make sense here.

Still prefer the term `"style:fix"` though because Prettier isn't _really_ lint.

steveluscher: The ideal situation would be to get Prettier to run _only_ on changed files, but at this point I think we've probably collectively burned enough time on this.

turbo.json
2501babe: there arent any `test/` dirs tho, other than `library-legacy` (which i didnt add this command for because it has its own flows and is a separate library). everything is in `src/`

2501babe: i removed `test/**` from this bc id just naively copy-pasted it

2501babe:
i believe this is now failing because ts doesnt like something that babel is doing...?

```
  cache miss, executing b909c70e542c957a
  
  > @solana/web3.js@0.0.0-development compile:js /home/runner/work/solana-web3.js/solana-web3.js/packages/library-legacy
  > cross-env NODE_ENV=production rollup -c
  
  [!] TypeError: Cannot redefine property: File
      at Function.defineProperty (<anonymous>)
      at Object.<anonymous> (/home/runner/work/solana-web3.js/solana-web3.js/node_modules/.pnpm/@babel+core@7.22.10/node_modules/@babel/core/lib/index.js:7:8)
      at Module._compile (node:internal/modules/cjs/loader:1241:14)
      at Object.Module._extensions..js (node:internal/modules/cjs/loader:1295:10)
      at Module.load (node:internal/modules/cjs/loader:1091:32)
      at Function.Module._load (node:internal/modules/cjs/loader:938:12)
      at Module.require (node:internal/modules/cjs/loader:1115:19)
      at require (node:internal/modules/helpers:130:18)
      at Object.<anonymous> (/home/runner/work/solana-web3.js/solana-web3.js/node_modules/.pnpm/@babel+core@7.22.10/node_modules/@babel/core/src/config/helpers/config-api.ts:4:1)
      at Module._compile (node:internal/modules/cjs/loader:1241:14)
```

its strange that this is happening now because i havent updated anything, i guess ci is fetching different dependency versions than a couple days ago?

buffalojoec:
> its strange that this is happening now because i havent updated anything, i guess ci is fetching different dependency versions than a couple days ago?

I just ran into this too on another PR. Is that what the problem is?

2501babe:
yea its definitely either "babel updated doing something tsc doesnt like" or "tsc updated and decided it doesnt like babel anymore" because i get the exact same build failure in #1563 too now. either dependabot updated on red or the lockfiles arent locking everything...?

2501babe:
ok its a bad dependabot commit, im pring a revert

2501babe:
apparently this is a bug in node? lol. i will leave it for you to do what you will https://github.com/angular/angular-cli/issues/25782

steveluscher:
> apparently this is a bug in node? lol.

Yeah, just do what I do and ignore all of your problems until they go away.

steveluscher (approved):
giphy-6

PR #1575, opened 2023-09-07
ci: pin node to under 20.5.1

buffalojoec:
I spotted this in one of the PRs linked in https://github.com/nodejs/node/issues/49497 .

It should solve our CI problem until the issue is resolved (?).


steveluscher:
I'm going to presume we can preemptively allowlist `>=20.6.1`. Let's see if CI is OK with that.

buffalojoec:
> I'm going to presume we can preemptively allowlist `>=20.6.1`. Let's see if CI is OK with that.

Didn't work on mine either, weird.

steveluscher:
> Didn't work on mine either, weird.

Well, it depends on what you mean by ‘work.’ The engines thing doesn't affect this, I don't think:

https://github.com/solana-labs/solana-web3.js/blob/94b4d494590b2ffbf60751bb74ed267b7228f012/.github/workflows/pull-requests.yml#L60-L63

That's the only thing that will fix CI. The `engines` thing is for people who install _us_, I think.

steveluscher:
I wouldn't worry about it too much. Let's just let them ship https://github.com/nodejs/node/pull/49528.

2501babe:
perhaps temporarily remove the rule where a failure on node current aborts the run on node lts? assuming this is trivial anyway

buffalojoec:
> That's the only thing that will fix CI. The `engines` thing is for people who install _us_, I think.

This _does_ sound more logical ha

PR #1576, opened 2023-09-09
refactor(experimental): Add `@solana/functional`

### LINKED ISSUE #1573: [experimental] API for transaction builder
# Problem

Awesome things about the purely functional transaction builder interface in the new web3.js:

1. You only bundle the code that you _actually_ use. Don't use nonces in your transactions? Don't get any of that logic.
2. You can bake in certain values via currying, pass those functions around as values, and use them over and over.
    ```ts
    const makeStevePayTheFees = setTransactionFeePayer.bind(null, 'S1eV33JH21H32G312gHASDjksda4s');
    const paidTransactions = unpaidTransactions.map(makeStevePayTheFees);
    ```

Bad things:

1. TypeScript is MAX PAIN if you try to reassign like this:
    ```ts
    let tx = createTransaction({version: 0});
    tx = setTransactionFeePayer('S1eV33JH21H32G312gHASDjksda4s', tx); // `tx` is _not_ `ITransactionWithFeePayer` now.
    ```
2. The alternative is even worse:
    ```ts
    const tx = createTransaction({version: 0});
    const txWithFeesPaid = setTransactionFeePayer('S1eV33JH21H32G312gHASDjksda4s', tx);
    const txWithFeesPaidAndBlockhashLifetimeSet = setTransactionLifetimeUsingBlockhash({/* ... */}, tx);
    ```

# Solutions

## Stop time until the JS pipeline operator is ready

One day, you will interact with this API [like this](https://github.com/tc39/proposal-pipeline-operator):

```ts
const tx = createTransaction({version: 0})
  |> setTransactionFeePayer('S1eV33JH21H32G312gHASDjksda4s', %)
  |> setTransactionLifetimeUsingBlockhash({/* ... */}, %);
```

## Create a builder

Doing this would undo _all_ of the benefits of tree shaking. There would be no way to shake away the code that you don't use (eg. nonce builders) because _every_ function would be bound to the builder instance.

This API _kills_ the ability to customize the steps. It also makes it impossible for an async step sneak in the middle; it either has to be _all_ async or _not at all_. See the cool `blockhashMeTriton` example below, which is made impossible with this ‘fluent’ builder API.

```ts
const tx = createTransaction({version: 0})
  .setTransactionFeePayer('S1eV33JH21H32G312gHASDjksda4s')
  .setTransactionLifetimeUsingBlockhash({/* ... */});
```

## Create a builder into which you have to inject dependencies as ‘capabilities’

This would give you the API of a builder, but demand that you declare the dependencies explicitly to imbue it with ‘capabilities’. This will be an interesting TypeScript exercise, but we've already achieved the impossible with the RPC API, right?

This API might be more customizable than the one above, but there may be some rough edges when it comes to async code.

```ts
const createTransaction = makeTransactionBuilderWithCapabilities({
  setTransactionFeePayer,
  setTransactionLifetimeUsingBlockhash,
});
const tx = createTransaction({version: 0})
  .setTransactionFeePayer('S1eV33JH21H32G312gHASDjksda4s')
  .setTransactionLifetimeUsingBlockhash({/* ... */});
```

## Tell people to compose all of their transactions in pointfree style

LOL no.

```ts
const tx =
  setTransactionLifetimeUsingBlockhash({/* ... */},
    setTransactionFeePayer('S1eV33JH21H32G312gHASDjksda4s',
      createTransaction({version: 0}),
    ),
  );
```

## Ship a compose helper

This has all of the bad characteristics of composing functions in pointfree style, while also being able to confuse people who don't already know that compose operates from right-to-left.

```ts
const tx = compose(
  setTransactionLifetimeUsingBlockhash.bind(null, {/* ... */}),
  setTransactionFeePayer.bind(null, 'S1eV33JH21H32G312gHASDjksda4s'),
  createTransaction,
)({version: 0});
```

## Ship a pipe helper

Pipe is all the bad things about compose, except with the nice feature of at least reading left-to-right and taking in a value as the first argument.

```ts
const tx = pipe(
  createTransaction({version: 0}),
  setTransactionFeePayer.bind(null, 'S1eV33JH21H32G312gHASDjksda4s'),
  setTransactionLifetimeUsingBlockhash.bind(null, {/* ... */}),
);
```

Now at least you can do nice things and create ‘readable’ (for some definition) builders like this:

```ts
/**
 * All the setup code
 */
function makeStevePay(tx: Transaction) {
  return setTransactionFeePayer(
    'S1eV33JH21H32G312gHASDjksda4s' as Base58EncodedAddress,
    tx,
  );
}
async function blockhashMe(rpc: Rpc<GetLatestBlockhashApi>, tx: Transaction) {
  return setTransactionLifetimeUsingBlockhash(
    await rpc.getLatestBlockhash(),
    tx,
  );
}
const tritonDevnetTransport = createDefaultRpcTransport({ url: 'https://api.devnet.solana.com' });
const rpc = createSolanaRpc({transport: tritonDevnetTransport});
const blockhashMeTriton = blockhashMe.bind(null, rpc);

/**
 * Your actual app code
 */
const tx = await pipe(
  createTransaction({version: 0}),
  makeStevePay,
  blockhashMeTriton,
);
```

## Pipe, with autobinding

We could develop a custom syntax that expresses ‘take this value, and apply these functions with these arguments.’

```ts
const tx = pipe(
  createTransaction({version: 0}),
  [setTransactionFeePayer, 'S1eV33JH21H32G312gHASDjksda4s' as Base58EncodedAddress],
  [setTransactionLifetimeUsingBlockhash, await rpc.getLatestBlockhash()],
);
```

The implementation would look something like this, but with a whole ton of TypeScript magic:

```ts
function pipe(value, ...fnsToApply) {
  return fnsToApply.reduce(([fn, ...args], acc) => fn(...args, acc), value);
}
```

One drawback of this API is that you no longer get TypeScript _autocomplete_ for your function arguments (ie. `setTransactionFeePayer`)

## Edit this issue to suggest more!

buffalojoec:
This PR introduces the `@solana/functional` library with a pipe helper, as discussed in #1573.

This pipe helper can be used for a variety of functional programming use cases, but the tests in this PR demonstrate using it for building a transaction, which could arguably be its primary use case in the context of the new Web3 JS API.

Closes #1573 

### One Area of Concern
Combining types works nicely for the resulting fields and values, and you can see the tests I wrote where I deliberately overwrite certain values to verify only the last one set it used. However, the type that comes up in IntelliSense/TypeCheck is a giant conglomerate of all the return types.

So, maybe you set the fee payer twice. The resulting value for `feePayer` is going to be the last one you set. BUT, if you hover over your `tx` variable, you'll see both, ie: `TransactionV0 & { feePayer: feePayer1 } & { feePayer: feePayer2 }`.

This could be problematic for larger pipes, and typechecking could get blown out, but it works I guess?


buffalojoec:
Requesting preemptive reviews to get some feedback before moving forward with more development/testing.

I want to make sure that `Chain` type combining all of the return types makes sense, and also have some discussion on the points I raised in the PR description.

buffalojoec:
I decided to address my first point raised in the PR description in a separate PR #1577

packages/functional/src/pipe.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Pipe for creating transactions.
+ * This function drives the general `pipe` function, but provides type safety for
+ * working with transactions.
+ * @param baseTransaction   The initial transaction, which can be any type of transaction
+ * @param mutations         Any number of mutating functions to pipe the transaction through
+ * @returns                 The final transaction with all mutations applied
+ */
+export function pipeTransaction<T extends BaseTransaction, R extends BaseTransaction, F extends Func<T, R>[]>(
```
steveluscher, line 34: We shouldn't need this.

```ts
const tx = pipe(
   createTransaction({ version: 0 }),
   () => { return { joe: 'string' } },
   setFeePayer,
   setBlockhash,
)
```

That generic invocation alone should be able to know, statically, that `{joe: 'string'}` does not conform to any of the input types _demanded_ by `setTransactionFeePayer`.

But you know what _can_ know that, statically?

https://www.typescriptlang.org/play?ssl=10&ssc=1&pln=10&pc=2#code/KYDwDg9gTgLgBAMwK4DsDGMCWEVzJsYAHgBUBJFTLAQwBsA+ACk0pgC45zXM6BKDrlR60AUKEixEqDNlz5CpCkLoAaOACUAjExZUBSmrTW6YAfSiaOzVvu584AXnobN-F2PDR4ydFhx4CYkFDNS1QgCYdG04DYWNWc0s4az0Yu1peR2cwuAtzcKsLDi1Mpw1wt3VwjwlvaT85QMV0qNTguLgAOm6EFABnDgBhOlpqACNaYAAxetkAbQBdTIBvOABfERF5YEYROH24Oc01cLUAZgWVPYPGUudGVYgxgCtgDA4AchQIAHcP9d4VwOyWoUCgHBQSAAtmNgFBFnc4KCoJ0odQwIxcGVcABqOCuES8IA

Your old friend… :)

steveluscher (changes requested):

packages/functional/src/pipe.ts
buffalojoec: sob

That's hilarious. I guess I'll have to add those and we'll have to have a max cap eh?

buffalojoec: Does this work to combine objects? I see it does change the type but one problem I faced was when the first couple of functions remove signatures, then the others add them, and there's also multiple signatures being appended on each other.

Rhetorical, I can mess around with it.

packages/functional/src/__tests__/pipe-test.ts
```diff
@@ -0,0 +1,325 @@
...
+const secondFeePayer =
+    '6JkwLherbVYPVF5sXGHm7qd9Lpd6gzinU4P792FkgdfS' as Base58EncodedAddress<'6JkwLherbVYPVF5sXGHm7qd9Lpd6gzinU4P792FkgdfS'>;
+
+const blockhashLifetime = { blockhash: '111' as Blockhash, lastValidBlockHeight: 5n };
+const secondBlockhashLifetime = { blockhash: '222' as Blockhash, lastValidBlockHeight: 8n };
+
+// The generic below is needed to shut TypeScript up and infer the proper return type.
+function setFeePayer<T extends BaseTransaction>(tx: T) {
+    return setTransactionFeePayer(feePayer, tx);
+}
```
mcintyre94, line 23: Will consumers have the same issue of having to wrap our functions like `setTransactionFeePayer` in their own function with this generic (or use arrow functions) to use them with `pipe`?

packages/functional/src/__tests__/pipe-transaction-test.ts
```diff
@@ -0,0 +1,325 @@
...
+                expect(tx1.version).toBe('legacy');
+                expect(tx1.lifetimeConstraint).toBe(secondBlockhashLifetime);
+                expect(tx1.feePayer).toBe(feePayer);
+                expect(tx2.version).toBe(0);
+                expect(tx2.lifetimeConstraint).toBe(secondBlockhashLifetime);
+                expect(tx2.feePayer).toBe(feePayer);
+            });
+        });
+    });
+});
```
mcintyre94, line 325: Are these tests the same as those in `pipe-test.ts`?

It might make sense to do non-transaction-specific tests for `pipe-test` instead IMO, which would also make it easier to follow for someone using `pipe` for something else.

packages/functional/src/__tests__/pipe-test.ts
buffalojoec: Sort of, yeah. It depends on how dynamic they have to make their helper. For example, if they know they'll always be adding a fee payer to a `TransactionV0`, then no. But if it may change and/or may have other configs like a lifetime, then yeah kinda.

There's always `any`!

packages/functional/src/__tests__/pipe-transaction-test.ts
buffalojoec: Yes, and yes you're correct I was just lazy lol. This isn't ready for merge yet

buffalojoec:
Revised it back to the generic pipe only. Ready for another look!

packages/functional/README.md
```diff
@@ -0,0 +1,28 @@
...
+is also exported as part of the Solana JavaScript SDK
+[`@solana/web3.js@experimental`](https://github.com/solana-labs/solana-web3.js/tree/master/packages/library).
+
+## Types
+
+TODO
+
+## Functions
+
+TODO
```
steveluscher, line 28: This reminds me; I have some work to do to update some READMEs that I forgot about.

packages/functional/src/pipe.ts
```diff
@@ -0,0 +1,26 @@
...
+    r1_r2: (r1: R1) => R2,
+    r2_r3: (r2: R2) => R3
+): R3;
+export function pipe<TInitial, R1, R2, R3, R4>(
+    init: TInitial,
+    init_r1: (init: TInitial) => R1,
+    r1_r2: (r1: R1) => R2,
+    r2_r3: (r2: R2) => R3,
+    r3_r4: (r3: R3) => R4
+): R4;
```
steveluscher, line 23: People are generally going to need way more overloads than this, for any kind of serious transaction. I don't know how to think about where to stop.

packages/functional/src/__tests__/pipe-transaction-test.ts
```diff
@@ -0,0 +1,215 @@
+import { Base58EncodedAddress } from '@solana/addresses';
+import {
+    BaseTransaction,
+    Blockhash,
+    createTransaction,
+    setTransactionFeePayer,
+    setTransactionLifetimeUsingBlockhash,
+} from '@solana/transactions';
```
steveluscher, line 8: I probably should have said something about this earlier 😓 but some thoughts now on this test.

1. The unit test doesn't have to be – and probably shouldn't – be specific to transactions. You probably only need to test:
    * What's the return value when you _only_ supply an initial value
    * What's the return value when you supply at least one function that operates over the previous value
    * What happens when one of the transformation functions throws
2. The _more_ interesting test, probably, is a type test.
    * If you only supply an initial value, does the final value match the initial value's type?
    * If you transform a previous value using a function with generics, does the previous type properly flow through the generic?

Like, this is a type test that you'd want to make sure passes.

```ts
function asyncify<TValue>(value: TValue): Promise<TValue> {
  return value;
}
pipe(123, asyncify) satisfies Promise<number>;
pipe('123', asyncify) satisfies Promise<string>;
pipe('123', () => {}, asyncify) satisfies Promise<void>;
```

In any case, these test functions don't have to – and probably shouldn't – come from `@solana/transactions`. Imagine: a breaking change to `@solana/transactions` will break these tests – for no good reason.

steveluscher (changes requested):

packages/functional/src/pipe.ts
buffalojoec: Ahh yeah I was going to ask. I stopped at 4 for the tests, and figured we add more if we need. But I guess client-facing we don't know that upper bound.

This is a tough position to be in, but worst case if someone needs more, they can always nest a pipe in the initial value.

How about arbitrarily starting with 10? 😅

packages/functional/src/__tests__/pipe-transaction-test.ts
buffalojoec: All fair points, I'll cut `@solana/transactions` out of the lib & tests

packages/functional/src/pipe.ts
steveluscher: Maybe just figure out what approach other functional programming utility libraries took and follow suit?

packages/functional/LICENSE
```diff
@@ -0,0 +1,20 @@
+Copyright (c) 2018 Solana Labs, Inc
```
mcintyre94, line 1: I figure you copied this from another license in this repo, but should this be 2023?

mcintyre94 (approved):
Looks good, and nice tests!

packages/functional/LICENSE
buffalojoec: Yeah, I copied it from I think `addresses`

#1602

steveluscher: \~IANAL\~
ANALO\~
ANALOS
SNALOS
SNALOA
SOALOA
SOALNA
SOLLNA
SOLANA

packages/functional/src/__tests__/pipe-test.ts
```diff
@@ -0,0 +1,382 @@
+import { pipe } from '../pipe';
+
+describe('pipe', () => {
+    it('can pipe a single value', () => {
+        expect.assertions(6);
```
steveluscher, line 5: You only need `expect.assertions()` if the test is `async`. It's designed to let Jest know that one of the awaitables didn't trigger, so the test should fail even though the others might have passed.

packages/functional/src/__tests__/pipe-test.ts
```diff
@@ -0,0 +1,382 @@
...
+        ).toBe('3!');
+    });
+    describe('combining objects', () => {
+        function combine<T extends object, U extends object>(a: T, b: U): T & U {
+            return { ...a, ...b };
+        }
+        it('can combine two objects', () => {
+            expect.assertions(1);
+            expect(pipe({ a: 1 }, value => combine(value, { b: 2 }))).toEqual({ a: 1, b: 2 });
+        });
```
steveluscher, line 80: The only _possible_ addition that I could make here would be a test that asserts that there's no copy-funny-business happening inside `pipe` itself. Like I'd want to assert that the object I put in is referentially equal to the one that comes out.

```ts
const startObj = {};
const endObj = pipe(
  startObj,
  (obj) => {
    obj.hello = 'there'; // A mutation of the passed-in-object
    return obj;
  },
);
expect(startObj).toBe(endObj);
```

steveluscher (approved):

packages/functional/src/__tests__/pipe-test.ts
buffalojoec: ohhh did not know that, will remove

steveluscher:
I needed this today, and not for transactions :) Can't wait.

buffalojoec:
> I needed this today, and not for transactions :) Can't wait.

I’m gonna start using it too!

I just need to add the last test you requested.

packages/functional/src/__tests__/pipe-test.ts
buffalojoec: I actually can't even set this test up without TypeScript throwing

```
Property 'hello' does not exist on type '{}'
```

But I did add 2 tests to address your concerns!

steveluscher:
I think Github is just passive-aggressively trying to get us to pay for Actions.

buffalojoec:
> I think Github is just passive-aggressively trying to get us to pay for Actions.

Lol, I noticed a recent incident with Actions/PRs, but I am going to debug some more. Just curious if you guys ran into this previously.

https://www.githubstatus.com/history

steveluscher:
Yeah, I would kick up your feet and come back later. If it passes locally just shipit.

PR #1577, opened 2023-09-10
refactor(experimental): revise `setFeePayer` overloads

buffalojoec:
This PR introduces some changes to `setTransactionFeePayer` because the original type configuration was always returning a union for a provided base transaction.

With separated overloads, TypeScript can successfully resolve the return type based on the transaction provided.


mcintyre94 (approved):
This looks good, would it be worth adding a type test to avoid a regression here in future?

buffalojoec:
> This looks good, would it be worth adding a type test to avoid a regression here in future?

I don't see why not! What if I instead create an issue for a new PR to write type tests for all transaction functions?

PR #1581, opened 2023-09-12
refactor(experimental): bring the HTTP transport and the WebSocket closer together, spiritually

steveluscher:
refactor(experimental): bring the HTTP transport and the WebSocket closer together, spiritually

# Summary

The HTTP transport returns a method that you can call to perform requests and to return responses. My initial version of the WebSocket transport did not fit this interface; instead it returned an object that encapsulated a send method and an iterator of the connection's messages.

Let's bring these spiritually closer together. After this PR you will instead get a method that behaves like a ‘one shot’ request function. When you call it, it creates a WebSocket connectio, sends your payload down it, then returns the connection iterator _and_ the send method.

Now you can more easily write wrapper transports around it that do stuff like:

* Connection persistence
* Auto-pinging
* Payload/response transformations

# Test Plan

```
cd packages/rpc-transport
pnpm test:unit:browser
pnpm test:unit:node
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/solana-labs/solana-web3.js/pull/1581).
* #1598
* #1586
* #1585
* #1584
* #1583
* #1582
* __->__ #1581
* #1580

### DIFF
--- a/packages/rpc-transport/src/transports/transport-types.ts
+++ b/packages/rpc-transport/src/transports/transport-types.ts
@@ -10,9 +10,16 @@ export interface IRpcTransport {
 }
 
 type RpcWebSocketTransportConfig = Readonly<{
+    payload: unknown;
     signal: AbortSignal;
 }>;
 
 export interface IRpcWebSocketTransport {
-    (config: RpcWebSocketTransportConfig): Promise<RpcWebSocketConnection>;
+    (config: RpcWebSocketTransportConfig): Promise<
+        Readonly<
+            Omit<RpcWebSocketConnection, 'send'> & {
+                send_DO_NOT_USE_OR_YOU_WILL_BE_FIRED: RpcWebSocketConnection['send'];
+            }
+        >
+    >;
 }
--- a/packages/rpc-transport/src/transports/websocket/__tests__/websocket-connection-test.ts
+++ b/packages/rpc-transport/src/transports/websocket/__tests__/websocket-connection-test.ts
@@ -5,9 +5,51 @@ import { createWebSocketConnection, RpcWebSocketConnection } from '../websocket-
 
 const MOCK_SEND_BUFFER_HIGH_WATERMARK = 42069;
 
+describe('createWebSocketConnection', () => {
+    let ws: WS;
+    function getLatestClient() {
+        const clients = ws.server.clients();
+        return clients[clients.length - 1];
+    }
+    beforeEach(async () => {
+        ws = new WS('wss://fake', {
+            jsonProtocol: true,
+        });
+    });
+    afterEach(() => {
+        WS.clean();
+    });
+    it('does not resolve until the socket is open', async () => {
+        expect.assertions(2);
+        const connectionPromise = createWebSocketConnection({
+            sendBufferHighWatermark: 0,
+            signal: new AbortController().signal,
+            url: 'wss://fake',
+        });
+        const client = getLatestClient();
+        expect(client).toHaveProperty('readyState', WebSocket.CONNECTING);
+        await connectionPromise;
+        expect(client).toHaveProperty('readyState', WebSocket.OPEN);
+    });
+    // https://github.com/thoov/mock-socket/issues/384
+    it.failing('throws when the connection is aborted before the connection is established', async () => {
+        expect.assertions(1);
+        const abortController = new AbortController();
+        const connectionPromise = createWebSocketConnection({
+            sendBufferHighWatermark: 0,
+            signal: abortController.signal,
+            url: 'wss://fake',
+        });
+        const client = getLatestClient();
+        expect(client).toHaveProperty('readyState', WebSocket.CONNECTING);
+        abortController.abort();
+        await expect(connectionPromise).rejects.toThrow();
+    });
+});
+
 describe('RpcWebSocketConnection', () => {
     let abortController: AbortController;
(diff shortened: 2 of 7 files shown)


packages/rpc-transport/src/transports/websocket/__tests__/websocket-transport-test.ts
```diff
@@ -1,54 +1,112 @@
...
-        await transportPromise;
-        expect(ws.server.clients()[0]).toHaveProperty('readyState', WebSocket.OPEN);
+        let resolveConnection: CallableFunction;
+        jest.mocked(createWebSocketConnection).mockReturnValue(
+            new Promise(r => {
+                resolveConnection = r;
+            })
+        );
+        const transportPromise = sendWebSocketMessage({ payload: 'hello', signal: abortController.signal });
+        await expect(Promise.race([transportPromise, 'pending'])).resolves.toBe('pending');
```
steveluscher, line 69: This might be worth extracting into a `await expect(promise).toBePending()` Jest matcher.

packages/rpc-transport/src/transports/websocket/__tests__/websocket-connection-test.ts
```diff
@@ -5,6 +5,48 @@ import { createWebSocketConnection, RpcWebSocketConnection } from '../websocket-
...
+        const connectionPromise = createWebSocketConnection({
+            sendBufferHighWatermark: 0,
+            signal: abortController.signal,
+            url: 'wss://fake',
+        });
+        const client = getLatestClient();
+        expect(client).toHaveProperty('readyState', WebSocket.CONNECTING);
+        abortController.abort();
+        await expect(connectionPromise).rejects.toThrow();
+    });
```
buffalojoec, line 47: Nice, solid.

buffalojoec (approved):
lgtm, I definitely like the idea of making the two interfaces similar

mcintyre94 (approved):

steveluscher:
So, just for completeness, @jordansexton and I talked at length about this API and collectively wondered if this could ever manifest as a race condition:

```ts
const notifications = await rpcSubscriptions
  .thingNotifications()
  // The `subscribe` method:
  //   1. Acquires an open socket connection
  //   2. Sends a message to the server asking for the subscription to be set up
  //   3. Vends the `notifications` iterable above after the server answers with the subscription confirmation
  .subscribe();
// Is it possible for a notification, for that subscription, to come down the socket between
// the time you *obtain* the iterable and the time you start to *iterate* over it, here:
for await (const n of notifications) {
  // Did I miss a notification?
  console.log(n);
}
```

I don't _think_ this is possible, because this is how I think the js task queue will see this:

1. Subscription confirmation comes in and fires `message` on the socket.
2. `handleMessage()` will get called and will resolve the promise
3. The resolution of that promise will resume execution up until the `for await` starts pumping the async iterator
4. If there was a subscription notification in the queue right behind the confirmation, it would fire after all of this is done, and the iterator would receive it.

I think this is a consequence of the ‘[run to completion](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Event_loop#run-to-completion)’ property of messages in the message queue.

From [LogRocket](https://blog.logrocket.com/complete-guide-node-js-event-loop/#poll):
> When the event loop enters the poll phase, it queues pending I/O events and executes them until the queue is empty or a system-dependent limit is reached. In between the execution of the JavaScript callbacks, the “next tick” and microtask queues are drained, like in the other phases.

Promise thenables are part of the microtask queue, so… QED? 🤞🏻

buffalojoec:
I mean, that all makes perfect sense to me. Is there some way we can test it?

steveluscher:
> Is there some way we can test it?

…maybe? Not with the live validator, but I could try to set something up with `jest-websocket-mock`. Thing is, I've found so many inconsistencies between `mock-socket` and _actual_ `WebSockets` already that I'm not sure the test would be airtight.

packages/rpc-transport/src/transports/transport-types.ts
```diff
@@ -10,9 +8,10 @@ export interface IRpcTransport {
 }
 
 type RpcWebSocketTransportConfig = Readonly<{
+    payload: unknown;
     signal: AbortSignal;
 }>;
 
 export interface IRpcWebSocketTransport {
-    (config: RpcWebSocketTransportConfig): Promise<RpcWebSocketConnection>;
+    (config: RpcWebSocketTransportConfig): Promise<AsyncIterable<unknown>>;
```
steveluscher, line 16: Ugh. This isn't going to work. Wrappers like the auto-ping wrapper need to be able to send messages down the _exact same connection_ that they're holding on to an iterator for, so they need a `send()` method to call.

Most of this PR can stay the way it is, but I'll have to expose `send()` on this type. I'll consider whether I need to call it `send_DO_NOT_USE_OR_YOU_WILL_BE_FIRED` or whether misuse is unlikely.

jordansexton: Steve never met a `DO_NOT_USE_OR_YOU_WILL_BE_FIRED` he didn't love 😆

steveluscher:
Added a test for that race condition. Exposed `send` as `send_DO_NOT_USE_OR_YOU_WILL_BE_FIRED`.

PR #1582, opened 2023-09-12
test: use Jest fake timers properly in WebSocket transport test

steveluscher:
test: use Jest fake timers properly in WebSocket transport test

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/solana-labs/solana-web3.js/pull/1582).
* #1604
* #1603
* #1598
* #1586
* #1585
* #1584
* #1583
* __->__ #1582

PR #1583, opened 2023-09-12
refactor(experimental): RPC subscriptions

steveluscher:
refactor(experimental): RPC subscriptions

# Summary

This new subscriptions API uses async iterables to vend subscription notifications, rather than a callback API.

```ts
// Subscribe for account notifications
const accountNotifications = await rpc
  .accountNotifications(address, {encoding: 'jsonParsed'})
  .subscribe();

// The subscription is now set up. Consume notifications.
try {
  for await (const notification of accountNotifications) {
    // Loops forever.
    console.log(notification);
  }
} catch(e) {
  // The subscription failed.
  // Use this `catch` to decide how/whether to re-establish it.
} finally {
  // Whether the subscription failed or was aborted,
  // you can run additional cleanup logic here.
}
```

Subscriptions respond to your app's regular abort controllers, so you can unsubscribe them automatically as part of your normal unmount, navigation, and other events.

```ts
// Create a cancellation controller.

// Subscribe for slot notifications.
const slotNotifications = await rpc
  .slotNotification()
  .subscribe({abortSignal: abortController.signal});

// Now, from *anywhere* in your app, triggering `abort()` will unsubscribe.
// For instance, in a React effect's teardown hook.
useEffect(() => {
  // ...
  return () => {
    // Abort everything when the component unmounts.
    abortControllerRef.current.abort();
  }
}, []);
```

# Test Plan

```
cd packages/rpc-transport
pnpm test:unit:browser
pnpm test:unit:node
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/solana-labs/solana-web3.js/pull/1583).
* #1604
* #1603
* #1598
* #1586
* #1585
* #1584
* __->__ #1583
* #1582

### DIFF
--- a/packages/rpc-transport/src/json-rpc-subscription.ts
+++ b/packages/rpc-transport/src/json-rpc-subscription.ts
@@ -1,0 +1,142 @@
+import { JsonRpcResponse } from './json-rpc';
+import { SolanaJsonRpcError } from './json-rpc-errors';
+import { createJsonRpcMessage } from './json-rpc-message';
+import {
+    PendingRpcSubscription,
+    RpcSubscription,
+    RpcSubscriptionConfig,
+    RpcSubscriptions,
+    SubscribeOptions,
+} from './json-rpc-types';
+
+type JsonRpcNotification<TNotification> = Readonly<{
+    params: Readonly<{
+        result: TNotification;
+        subscription: number;
+    }>;
+}>;
+type SubscriptionId = number;
+
+function registerIterableFatal(iterable: AsyncIterable<unknown>, catchFn: CallableFunction) {
+    (async () => {
+        try {
+            // eslint-disable-next-line @typescript-eslint/no-unused-vars
+            for await (const _ of iterable);
+        } catch (e) {
+            // Run the catch function if the iterator throws.
+            catchFn(e);
+        }
+    })();
+}
+
+function createPendingRpcSubscription<TRpcSubscriptionMethods, TNotification>(
+    rpcConfig: RpcSubscriptionConfig<TRpcSubscriptionMethods>,
+    { params, subscribeMethodName, unsubscribeMethodName, responseProcessor }: RpcSubscription<TNotification>
+): PendingRpcSubscription<TNotification> {
+    return {
+        async subscribe(options?: SubscribeOptions): Promise<AsyncIterable<TNotification>> {
+            options?.abortSignal?.throwIfAborted();
+            let subscriptionId: number | undefined;
+            function handleCleanup() {
+                if (subscriptionId !== undefined) {
+                    const payload = createJsonRpcMessage(unsubscribeMethodName, [subscriptionId]);
+                    connection.send_DO_NOT_USE_OR_YOU_WILL_BE_FIRED(payload).finally(() => {
+                        connectionAbortController.abort();
+                    });
+                } else {
+                    connectionAbortController.abort();
+                }
+            }
+            options?.abortSignal?.addEventListener('abort', handleCleanup);
+            /**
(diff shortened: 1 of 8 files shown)


packages/rpc-transport/src/json-rpc-subscription.ts
```diff
@@ -0,0 +1,142 @@
...
+            let subscriptionId: number | undefined;
+            function handleCleanup() {
+                if (subscriptionId !== undefined) {
+                    const payload = createJsonRpcMessage(unsubscribeMethodName, [subscriptionId]);
+                    const unsubscribeAbortController = new AbortController();
+                    rpcConfig.transport({ payload, signal: unsubscribeAbortController.signal }).finally(() => {
+                        unsubscribeAbortController.abort();
+                    });
+                }
+                connectionAbortController.abort();
```
buffalojoec, line 48: Seems like this could just use one abort controller, no? why not just pass `connectionAbortController` into the closure?

packages/rpc-transport/src/json-rpc-subscription.ts
```diff
@@ -0,0 +1,142 @@
...
+                    : {
+                          params: rawParams,
+                          subscribeMethodName: methodName.replace(/Notifications$/, 'Subscribe'),
+                          unsubscribeMethodName: methodName.replace(/Notifications$/, 'Unsubscribe'),
+                      };
+                return createPendingRpcSubscription(rpcConfig, newRequest);
+            };
+        },
+    }) as RpcSubscriptions<TRpcSubscriptionMethods>;
+}
```
buffalojoec, line 136: Nice, this looks just like the HTTP proxy setup. Do we dare combine them? 😂 😂

buffalojoec (approved):
holy smokes

mcintyre94 (approved):

packages/rpc-transport/src/json-rpc-subscription.ts
steveluscher: You're right. I just need to change this to ‘kill the one now if you don't have a subscription’ or ‘kill both once you've sent the unsubscribe message.’

packages/rpc-transport/src/json-rpc-subscription.ts
steveluscher: Sure does. Just with the exception of having one `methodName` for one-shot requests and separate subscribe/unsubscribe method names for subscriptions. Also, the subscriptions one does that janky pattern matching on `*Notifications` thing.

I thought about combining them but thought it would be confusing in the end for little gain.

packages/rpc-transport/src/json-rpc-subscription.ts
```diff
@@ -0,0 +1,142 @@
...
+function registerIterableFatal(iterable: AsyncIterable<unknown>, catchFn: CallableFunction) {
+    (async () => {
+        try {
+            // eslint-disable-next-line @typescript-eslint/no-unused-vars
+            for await (const _ of iterable);
+        } catch (e) {
+            // Run the catch function if the iterator throws.
+            catchFn(e);
+        }
+    })();
```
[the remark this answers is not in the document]
steveluscher: I don't ever want someone to accidentally block on it by doing this:

```ts
await registerIterableFatal(connection, handleConnectionFatal);
```

That would pump the iterator indefinitely, which is definitely not what you intended.

packages/rpc-transport/src/json-rpc-subscription.ts
buffalojoec: I would have to agree

PR #1586, opened 2023-09-12
refactor(experimental): The slot notifications API

steveluscher:
refactor(experimental): The slot notifications API

# Summary

Use this method to subscribe for a notification every time the slot advances.

# Test Plan

```
cd packages/rpc-core
pnpm test:unit:browser
pnpm test:unit:node
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/solana-labs/solana-web3.js/pull/1586).
* #1604
* #1603
* #1598
* __->__ #1586

### DIFF
--- a/packages/rpc-core/src/rpc-subscriptions/__typetests__/slot-notifications-type-test.ts
+++ b/packages/rpc-core/src/rpc-subscriptions/__typetests__/slot-notifications-type-test.ts
@@ -1,0 +1,22 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
+
+import { RpcSubscriptions } from '@solana/rpc-transport/dist/types/json-rpc-types';
+
+import { U64UnsafeBeyond2Pow53Minus1 } from '../../rpc-methods/common';
+import { SolanaRpcSubscriptions } from '../index';
+
+async () => {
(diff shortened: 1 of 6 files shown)


buffalojoec (approved):
Nice! Looking quite familiar...

buffalojoec:
> Do you think we'll want typescript tests here at some point? If we will then it'd be worth adding them now IMO so we don't get the catch up work that we have on the RPC methods

I would say fo sho

packages/rpc-core/src/rpc-subscriptions/__typetests__/slot-notifications-type-test.ts
```diff
@@ -0,0 +1,22 @@
...
+        Readonly<{
+            parent: U64UnsafeBeyond2Pow53Minus1;
+            root: U64UnsafeBeyond2Pow53Minus1;
+            slot: U64UnsafeBeyond2Pow53Minus1;
+        }>
+    >;
+
+    // @ts-expect-error Takes no params.
+    rpcSubcriptions.slotNotifications({ commitment: 'finalized' });
+};
```
steveluscher, line 22: Added type tests cc/ @buffalojoec

PR #1596, opened 2023-09-14
Export LamportsUnsafe type and add an assertion function

mcintyre94:
This is required because the `requestAirdrop` RPC call requires it as an input.

So as with our other branded inputs we need to be able to write code like:

```ts
const address = 'abc' as Base58EncodedAddress
const lamports = 1_000_000n
assertIsLamports(lamports)
await rpc.requestAirdrop(address, lamports).send()

// or just cast it
await rpc.requestAirdrop(address, 1_000_000n as LamportsUnsafeBeyond2Pow53Minus1).send()
```

Any BigInt that is greater than zero and is less than u64::MAX is valid as a number of lamports and will succeed in `assertIsLamports`.

### DIFF
--- a/packages/rpc-core/src/lamports.ts
+++ b/packages/rpc-core/src/lamports.ts
@@ -1,0 +1,21 @@
+// FIXME(solana-labs/solana/issues/30341) Beware that any value above 9007199254740991 may be
+// truncated or rounded because of a downcast to JavaScript `number` between your calling code and
+// the JSON-RPC transport.
+export type LamportsUnsafeBeyond2Pow53Minus1 = bigint & { readonly __brand: unique symbol };
+
+// Largest possible value to be represented by a u64
+const maxU64Value = 18446744073709551615n; // 2n ** 64n - 1n
+
+export function assertIsLamports(
+    putativeLamports: bigint
+): asserts putativeLamports is LamportsUnsafeBeyond2Pow53Minus1 {
+    if (putativeLamports < 0) {
+        // TODO: Coded error.
+        throw new Error('Input for 64-bit unsigned integer cannot be negative');
+    }
+
+    if (putativeLamports > maxU64Value) {
+        // TODO: Coded error.
+        throw new Error('Input number is too large to be represented as a 64-bit unsigned integer');
+    }
+}
--- a/packages/rpc-core/src/__tests__/lamports-test.ts
+++ b/packages/rpc-core/src/__tests__/lamports-test.ts
@@ -1,0 +1,32 @@
+import { assertIsLamports } from '../lamports';
+
+describe('assertIsLamports()', () => {
+    it('throws when supplied a negative number', () => {
+        expect(() => {
+            assertIsLamports(-1n);
+        }).toThrow();
+        expect(() => {
+            assertIsLamports(-1000n);
+        }).toThrow();
+    });
+    it('throws when supplied a too large number', () => {
+        expect(() => {
+            assertIsLamports(2n ** 64n);
+        }).toThrow();
+    });
+    it('does not throw when supplied zero lamports', () => {
+        expect(() => {
+            assertIsLamports(0n);
+        }).not.toThrow();
+    });
+    it('does not throw when supplied a valid non-zero number of lamports', () => {
+        expect(() => {
+            assertIsLamports(1_000_000_000n);
+        }).not.toThrow();
+    });
+    it('does not throw when supplied the max valid number of lamports', () => {
+        expect(() => {
+            assertIsLamports(2n ** 64n - 1n);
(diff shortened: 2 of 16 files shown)


packages/rpc-core/src/lamports.ts
```diff
@@ -0,0 +1,19 @@
...
+export type LamportsUnsafeBeyond2Pow53Minus1 = bigint & { readonly __brand: unique symbol };
+
+// Largest possible value to be represented by a u64
+const maxU64Value = 2n ** 64n - 1n;
+
+export function assertIsLamports(
+    putativeLamports: bigint
+): asserts putativeLamports is LamportsUnsafeBeyond2Pow53Minus1 {
+    if (putativeLamports < 0) {
+        throw new Error('Input number is negative');
```
buffalojoec, line 13: nit: "Input for 64-bit unsigned integer cannot be negative"

buffalojoec (approved):
This code looks good, but I have some qualms about the idea of seeing code snippets & examples born out of this new API all using this cast.

```
rpc.getAirdrop(lamports as LamportsUnsafeBeyond2Pow53Minus1).send();
```

packages/rpc-core/src/lamports.ts
```diff
@@ -0,0 +1,19 @@
+// FIXME(solana-labs/solana/issues/30341) Beware that any value above 9007199254740991 may be
+// truncated or rounded because of a downcast to JavaScript `number` between your calling code and
+// the JSON-RPC transport.
+export type LamportsUnsafeBeyond2Pow53Minus1 = bigint & { readonly __brand: unique symbol };
+
+// Largest possible value to be represented by a u64
+const maxU64Value = 2n ** 64n - 1n;
```
steveluscher, line 7: Prefer to bake this rather than to do math in the module constructor.
```suggestion
const maxU64Value = 18446744073709551615n; // 2n ** 64n - 1n
```

packages/rpc-core/src/lamports.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export function assertIsLamports(
+    putativeLamports: bigint
+): asserts putativeLamports is LamportsUnsafeBeyond2Pow53Minus1 {
+    if (putativeLamports < 0) {
+        throw new Error('Input number is negative');
+    }
+
+    if (putativeLamports > maxU64Value) {
+        throw new Error('Input number is too large to be represented as a 64-bit unsigned integer');
```
steveluscher, line 17:
```suggestion
        // TODO: Coded error.
        throw new Error('Input number is too large to be represented as a 64-bit unsigned integer');
```

packages/rpc-core/src/lamports.ts
steveluscher, line 13:
```suggestion
        // TODO: Coded error.
        throw new Error('Input number is negative');
```

steveluscher (approved):
I imagine that, at some point soon – rather than to simplify the type names – we should offer a set of assertion-coercion utilities that turn this:

```ts
await rpc.requestAirdrop(address, 1_000_000n as LamportsUnsafeBeyond2Pow53Minus1).send()
```

…into this:

```ts
await rpc.requestAirdrop(address, lamports(1_000_000n)).send()
```

`lamports()` being essentially:

```ts
function lamports(putativeLamports: bigint): LamportsUnsafeBeyond2Pow53Minus1 {
  assertIsLamports(putativeLamports);
  return putativeLamports;
}
```

buffalojoec:
> I imagine that, at some point soon – rather than to simplify the type names – we should offer a set of assertion-coercion utilities that turn this:
> 
> ```ts
> await rpc.requestAirdrop(address, 1_000_000n as LamportsUnsafeBeyond2Pow53Minus1).send()
> ```
> 
> …into this:
> 
> ```ts
> await rpc.requestAirdrop(address, lamports(1_000_000n)).send()
> ```
> 
> `lamports()` being essentially:
> 
> ```ts
> function lamports(putativeLamports: bigint): LamportsUnsafeBeyond2Pow53Minus1 {
>   assertIsLamports(putativeLamports);
>   return putativeLamports;
> }
> ```

I like this, but where shall it live? It's not directly related to the RPC, so `rpc-core` seems weird. _Another_ package?

steveluscher:
> Another package?

Probably. I'd say ‘closest to the type itself’ but yeah maybe `@solana/rpc-core` is a bit weird for that.

I really expect that before we go to release candidate status we're going to end up moving a _lot_ of this code around.

mcintyre94:
Agree with all this, I'd rather have nicely named functions that do the assert for you than less descriptive types

buffalojoec:
#1600

PR #1604, opened 2023-09-15
refactor(experimental): teach the auto-pinger *not* to ping the connection when you're offline

steveluscher:
refactor(experimental): teach the auto-pinger *not* to ping the connection when you're offline

# Summary

There is zero point to slamming ping messages down a socket if the server can't hear them. At most we should send a ping the moment we come back online, then resume as normal.

# Test Plan

```
cd packages/library
pnpm test:unit:browser
pnpm test:unit:node
```

1. Paste [this](https://gist.github.com/steveluscher/4254aba293222daded59bdceafbdaa4c) code into a Chrome console.
2. Toggle the network between ‘no throttling’ and ‘offline’
3. Observe the ping behavior, as described in the unit tests

### DIFF
--- a/packages/library/src/rpc-websocket-autopinger.ts
+++ b/packages/library/src/rpc-websocket-autopinger.ts
@@ -18,11 +18,12 @@ export function getWebSocketTransportWithAutoping({ intervalMs, transport }: Con
     return async (...args) => {
         const connection = await transport(...args);
         let intervalId: string | number | NodeJS.Timeout | undefined;
+        function sendPing() {
+            connection.send_DO_NOT_USE_OR_YOU_WILL_BE_FIRED(PING_PAYLOAD);
+        }
         function restartPingTimer() {
             clearInterval(intervalId);
-            intervalId = setInterval(() => {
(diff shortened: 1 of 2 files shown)


packages/library/src/rpc-websocket-autopinger.ts
```diff
@@ -45,9 +46,28 @@ export function getWebSocketTransportWithAutoping({ intervalMs, transport }: Con
                 } finally {
                     pingableConnections.delete(connection);
                     clearInterval(intervalId);
+                    if (handleOffline && handleOnline) {
+                        globalThis.window.removeEventListener('offline', handleOffline);
+                        globalThis.window.removeEventListener('online', handleOnline);
+                    }
```
mcintyre94, line 52: Is it worth splitting this into 2 ifs, one for each listener? The current code always defines them together, but it'd probably be easy to refactor it so that's not guaranteed and we might not notice we need to update this

steveluscher: Good call. This is one of those ‘no really TypeScript trust me these will always be defined’ things, but none the less it's worth splitting them up for future protection against refactoring as you've mentioned!

PR #1608, opened 2023-09-18
feat(experimental): add core codec types and helpers

lorisleiva:
This PR brings [Umi's core serializers](https://github.com/metaplex-foundation/umi/tree/main/packages/umi-serializers-core) library to Web3.js and renames `Serializer` to `Codec` in order to make room for more granular `Encoder` and `Decoder` types.


packages/codecs-core/src/bytes.ts
```diff
@@ -0,0 +1,31 @@
...
+    return result;
+};
+
+/**
+ * Pads a `Uint8Array` with zeroes to the specified length.
+ * If the array is longer than the specified length, it is returned as-is.
+ */
+export const padBytes = (bytes: Uint8Array, length: number): Uint8Array => {
+    if (bytes.length >= length) return bytes;
+    const paddedBytes = new Uint8Array(length).fill(0);
```
steveluscher, line 21: Is this as performant as doing two `set()` operations with only the bytes needed?

```ts
const out = new Uint8Array(length);
out.set(bytes);
out.set(new Uint8Array(length - bytes.length).fill(0), bytes.length);
return out;
```

packages/codecs-core/src/bytes.ts
```diff
@@ -0,0 +1,31 @@
...
+    paddedBytes.set(bytes);
+    return paddedBytes;
+};
+
+/**
+ * Fixes a `Uint8Array` to the specified length.
+ * If the array is longer than the specified length, it is truncated.
+ * If the array is shorter than the specified length, it is padded with zeroes.
+ */
+export const fixBytes = (bytes: Uint8Array, length: number): Uint8Array => padBytes(bytes.slice(0, length), length);
```
steveluscher, line 31: Is it worth preventing one copy in the case the thing is shorter?

```ts
export const fixBytes = (bytes: Uint8Array, length: number): Uint8Array => padBytes(
  bytes.length <= length ? bytes : bytes.slice(0, length),
  length,
);
```

packages/codecs-core/src/errors.ts
```diff
@@ -0,0 +1,24 @@
...
+    constructor(codec: string, expected: bigint | number, actual: bigint | number) {
+        super(`Codec [${codec}] expected ${expected} bytes, got ${actual}.`);
+    }
+}
+
+export class ExpectedFixedSizeCodecError extends Error {
+    readonly name: string = 'ExpectedFixedSizeCodecError';
+
+    constructor(message?: string) {
+        message ??= 'Expected a fixed-size coded, got a variable-size one.';
```
steveluscher: ‘codec?’

packages/codecs-core/src/errors.ts
```diff
@@ -0,0 +1,24 @@
...
+        super(message);
+    }
+}
```
steveluscher: In general, error subclasses bum me out, because:

* they take up space
* they encourage `instanceof` checks which _always_ get us into trouble when someone has multiple copies of a library

The approach I want to try in web3.js is to have _coded errors_. That is to say that there's _one_ `Error` subclass for all of the entire library that has a `code` instance variable. That code is a Typescript const enum. Now you can do stuff like this:

```ts
} catch (e) {
    if (e instanceof Object && 'code' in e && e.code === SolanaCodedError.EXPECTED_FIXED_SIZE_CODEC) {
        /* ... */
    }
}
```

And that compiles down to basically:

```ts
catch (e) {
    if (e instanceof Object && 'code' in e && e.code === 0 /* SolanaCodedError.EXPECTED_FIXED_SIZE_CODEC */) {
        /* ... */
    }
}
```

Here's an example: https://github.com/solana-labs/solana-web3.js/blob/55e373b2e48becf999a67ce4cb236f8265987a60/packages/rpc-transport/src/json-rpc-errors.ts#L30-L42

I'm pretty sure there's a way to get it so that depending on what `code` ends up being, TypeScript refines `data` to be a particular type other than `unknown`. I can't find an example right now.

packages/codecs-core/src/fixCodec.ts
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable sort-keys-fix/sort-keys-fix */
```
steveluscher: giphy-4

packages/codecs-core/src/fixCodec.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @param fixedBytes - The fixed number of bytes to read.
+ * @param description - A custom description for the codec.
+ */
+export function fixCodec<T, U extends T = T>(
+    codec: Codec<T, U>,
+    fixedBytes: number,
+    description?: string
+): Codec<T, U> {
+    return {
+        description: description ?? `fixed(${fixedBytes}, ${codec.description})`,
```
steveluscher: How about we throw these out except in dev mode?

```diff
-         description: description ?? `fixed(${fixedBytes}, ${codec.description})`,
+         description: __DEV__
+             ? (description ?? `fixed(${fixedBytes}, ${codec.description})`)
+             : undefined,
```

packages/codecs-core/src/fixCodec.ts
```diff
@@ -0,0 +1,39 @@
...
+        fixedSize: fixedBytes,
+        maxSize: fixedBytes,
+        encode: (value: T) => fixBytes(codec.encode(value), fixedBytes),
+        decode: (buffer: Uint8Array, offset = 0) => {
+            // Slice the buffer to the fixed size.
+            buffer = buffer.slice(offset, offset + fixedBytes);
```
steveluscher: Prevent a copy in the case that `offset === 0 && fixedBytes >= buffer.length`?

packages/codecs-core/src/fixCodec.ts
```diff
@@ -0,0 +1,39 @@
...
+    codec: Codec<T, U>,
+    fixedBytes: number,
+    description?: string
+): Codec<T, U> {
+    return {
+        description: description ?? `fixed(${fixedBytes}, ${codec.description})`,
+        fixedSize: fixedBytes,
+        maxSize: fixedBytes,
+        encode: (value: T) => fixBytes(codec.encode(value), fixedBytes),
+        decode: (buffer: Uint8Array, offset = 0) => {
```
steveluscher: Is it worth scrubbing the word ‘buffer’ from this codebase, since we're taking a stand against `Buffer` in favour of `Uint8Array`?

packages/codecs-core/src/fixCodec.ts
```diff
@@ -0,0 +1,39 @@
...
+            // Slice the buffer to the fixed size.
+            buffer = buffer.slice(offset, offset + fixedBytes);
+            // Ensure we have enough bytes.
+            if (buffer.length < fixedBytes) {
+                throw new NotEnoughBytesError('fixCodec', fixedBytes, buffer.length);
```
steveluscher: An example of my coded error comment above:

```ts
throw new WhateverWeCallThisError(
  'message',
  WhateverWeCallThisErrorCode.NOT_ENOUGH_BYTES,
  {codec: 'fixCodec', expected: fixedBytes, actual: bytes.length},
);

File: packages/codecs-core/src/reverseCodec.ts [hunk 10]
```diff
@@ -0,0 +1,26 @@
...
+        throw new ExpectedFixedSizeCodecError('Cannot reverse a codec of variable size.');
+    }
+    return {
+        ...codec,
+        encode: (value: T) => codec.encode(value).reverse(),
+        decode: (bytes: Uint8Array, offset = 0) => {
+            const newBytes = mergeBytes([
+                bytes.slice(0, offset),
+                bytes.slice(offset, offset + fixedSize).reverse(),
+                bytes.slice(offset + fixedSize),
```
steveluscher: Don't know if you feel like eliminating the unnecessary copies that happen when:

* `offset` is `0`, or
* `offset + bytes.length === fixedSize`

steveluscher (approved):
giphy-5


Sidenote: I _think_ that using the `feat(experimental)` commit message pattern might cause a release of the legacy library, which is not what we want. That's why I've been using `refactor(experimental)`. I also hate conventional-commit and want to get rid of it soon.

steveluscher:
OMG I only looked at the first commit in this PR.

packages/codecs-core/src/__tests__/fix-codec-test.ts
```diff
@@ -0,0 +1,97 @@
...
+        const b = (s: string) => base16.encode(s);
+
+        expect(fixEncoder(a1z26, 42).description).toBe('fixed(42, a1z26)');
+        expect(fixEncoder(a1z26, 42, 'my fixed').description).toBe('my fixed');
+        expect(fixEncoder(a1z26, 12).fixedSize).toBe(12);
+        expect(fixEncoder(a1z26, 12).maxSize).toBe(12);
+
+        expect(fixEncoder(a1z26, 10).encode('helloworld')).toStrictEqual(b('08050c0c0f170f120c04'));
+        expect(fixEncoder(a1z26, 5).encode('helloworld')).toStrictEqual(b('08050c0c0f'));
+        expect(fixEncoder(a1z26, 10).encode('hello')).toStrictEqual(b('08050c0c0f0000000000'));
```
steveluscher, line 77: OK, I see what `__setup__` is for now.

I'd play these tests a different way.

I'd set up a complete mock encoder.

```ts
const mockEncoder = {
  /* ... */
  decode: jest.fn(),
  encode: jest.fn(),
};
```

Then test not what the encoder does, but what the fixer does to the encode call.

```ts
fixEncoder(mockEncoder, 10).encode('helloworld');
expect(mockEncoder.encode).toHaveBeenCalledWith('helloworld');

fixEncoder(mockEncoder, 5).encode('helloworld');
expect(mockEncoder.encode).toHaveBeenCalledWith('hello');
```

You shouldn't need a concrete encoder in any case. The fixer should just play the mock like a puppet in all the ways you need it to.

steveluscher (approved):

packages/codecs-core/package.json
```diff
@@ -0,0 +1,97 @@
...
+        "publish-packages": "pnpm publish --tag experimental --access public --no-git-checks",
+        "style:fix": "pnpm eslint --fix src/* && pnpm prettier -w src/*",
+        "test:lint": "jest -c node_modules/test-config/jest-lint.config.ts --rootDir . --silent",
+        "test:prettier": "jest -c node_modules/test-config/jest-prettier.config.ts --rootDir . --silent",
+        "test:treeshakability:browser": "agadoo dist/index.browser.js",
+        "test:treeshakability:native": "agadoo dist/index.node.js",
+        "test:treeshakability:node": "agadoo dist/index.native.js",
+        "test:typecheck": "tsc --noEmit",
+        "test:unit:browser": "jest -c node_modules/test-config/jest-unit.config.browser.ts --rootDir . --silent",
+        "test:unit:node": "jest -c node_modules/test-config/jest-unit.config.node.ts --rootDir . --silent"
```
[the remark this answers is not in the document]
lorisleiva: Resolving as this package has tests.

packages/codecs-core/src/bytes.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * Concatenates an array of `Uint8Array`s into a single `Uint8Array`.
+ */
+export const mergeBytes = (bytesArr: Uint8Array[]): Uint8Array => {
```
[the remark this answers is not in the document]
lorisleiva: I like that!

packages/codecs-core/src/fixCodec.ts
lorisleiva: I don't understand the point of that change? All of these descriptions can be overridden by the end-user and provide a helpful way to debug which codec failed. What would be the benefit of stripping them? Would that even be tree-shaken?

packages/codecs-core/src/fixCodec.ts
lorisleiva: Agreed! I've been renaming them to `bytes` instead but must have missed a few. I'll do a proper search and replace.

packages/codecs-core/src/fixCodec.ts
steveluscher: ¯\\\_(°ペ)_\/¯

lorisleiva:
Closing this PR in favour of #1630. I will update the new PR with the comments in this PR before marking it as ready for review.

packages/codecs-core/package.json
```diff
@@ -0,0 +1,97 @@
+{
+    "name": "@solana/codecs-core",
+    "version": "2.0.0-development",
+    "description": "Core types and helpers for encoding and decoding on Solana",
```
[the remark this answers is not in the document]
lorisleiva: Done on https://github.com/solana-labs/solana-web3.js/pull/1630

packages/codecs-core/src/bytes.ts
lorisleiva: I don't know but my guess would be no because it's instantiating an extra `Uint8Array`.

packages/codecs-core/src/bytes.ts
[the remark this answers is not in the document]
lorisleiva: Done on https://github.com/solana-labs/solana-web3.js/pull/1630

packages/codecs-core/src/bytes.ts
lorisleiva: Done on https://github.com/solana-labs/solana-web3.js/pull/1630

packages/codecs-core/src/errors.ts
lorisleiva: In general, I like that solution. What I don't like about it is that each package will have to define its error code in some other package because they all need to be in the same enum. I'd much prefer each package to be responsible for their own errors much like the current `SolanaJsonRpcError` does.

packages/codecs-core/src/errors.ts
lorisleiva: This was actually fixed in one of the next commit.

packages/codecs-core/src/fixCodec.ts
lorisleiva: Gotcha! Yeah then I don't see why not. I'll make the changes.

lorisleiva: Although, the problem with that is it makes it impossible to use the description of a codec inside an error message which could be convenient.

packages/codecs-core/src/fixCodec.ts
lorisleiva: Updated in #1630. 👌

packages/codecs-core/src/errors.ts
lorisleiva: I think this warrants a deeper conversation anyway. I'll use simple `Error`s for now and add `// TODO: Coded error.` to stay consistent with the rest of the libraries.

packages/codecs-core/src/fixCodec.ts
lorisleiva: Done in #1630.

packages/codecs-core/src/fixCodec.ts
lorisleiva: Reverted to plain `Error`s for now with the `// TODO: Coded error.` prefix.

packages/codecs-core/src/fix-codec.ts
```diff
@@ -0,0 +1,68 @@
...
+ * @param decoder - The decoder to wrap into a fixed-size decoder.
+ * @param fixedBytes - The fixed number of bytes to read.
+ * @param description - A custom description for the decoder.
+ */
+export function fixDecoder<T>(decoder: Decoder<T>, fixedBytes: number, description?: string): Decoder<T> {
+    return {
+        ...fixCodecHelper(decoder, fixedBytes, description),
+        decode: (buffer: Uint8Array, offset = 0) => {
+            // Slice the buffer to the fixed size.
+            buffer = buffer.slice(offset, offset + fixedBytes);
```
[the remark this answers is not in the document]
lorisleiva: Done in #1630.

packages/codecs-core/src/__tests__/fix-codec-test.ts
steveluscher: Sure you can! You could do:

```ts
mockEncoder.encode.mockReturnValue('08050c0c0f')
expect(fixEncoder(mockEncoder, 10).encode('hello')).toStrictEqual(b('08050c0c0f0000000000'));
```

lorisleiva: Done in #1630.

packages/codecs-core/src/reverseCodec.ts
lorisleiva: Done in #1630.

packages/codecs-core/src/fixCodec.ts
lorisleiva: Done in #1630.

PR #1609, opened 2023-09-19
feat(experimental): add codecs-numbers package

lorisleiva:
Stacked PR based on #1608.

---

This PR adds number codecs to web3.js. It adds the following functions `getXEncoder`, `getXDecoder` and `getXCodec` for each of the following `X`s:

- `F32`
- `F64`
- `I128`
- `I16`
- `I32`
- `I64`
- `I8`
- `ShortU16`
- `U128`
- `U16`
- `U32`
- `U64`
- `U8`

PR #1610, opened 2023-09-19
feat(experimental): add codecs-strings package

lorisleiva:
Stacked PR based on #1609 .

---

This PR adds string codecs to web3.js. It adds the following functions `getXEncoder`, `getXDecoder` and `getXCodec` for each of the following `X`s:

- `BaseX`
- `BaseXReslice`
- `Base10` (uses `BaseX`)
- `Base16`
- `Base58` (uses `BaseX`)
- `Base64` (uses `atob` and `btoa` on the browser, otherwise uses `BaseXReslice`)
- `Utf8` (uses `TextEncoder` and `TextDecoder`)
- `String` (to support multiple size and encoding strategies)


packages/codecs-strings/src/base64.ts
```diff
@@ -0,0 +1,53 @@
...
+import { combineCodec, Decoder, Encoder, mapDecoder, mapEncoder } from '@solana/codecs-core';
+
+import { getBaseXResliceDecoder, getBaseXResliceEncoder } from './baseX-reslice';
+import { InvalidBaseStringError } from './errors';
+
+const alphabet = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/';
+
+/** Encodes strings in base64. */
+export const getBase64Encoder = (): Encoder<string> => {
+    if (__BROWSER__) {
```
steveluscher, line 11: Is it the case that we can do…

```ts
if (__BROWSER__) {
    // What you have here.
} else if (__NODEJS__) {
    return Buffer.from(value, 'base64');
} else {
    // Your userspace implementation
}
```

packages/codecs-strings/src/utf8.ts
```diff
@@ -0,0 +1,32 @@
...
+import { removeNullCharacters } from './null-characters';
+
+/** Encodes UTF-8 strings using the native `TextEncoder` API. */
+export const getUtf8Encoder = (): Encoder<string> => {
+    let textEncoder: TextEncoder;
+    return {
+        description: 'utf8',
+        fixedSize: null,
+        maxSize: null,
+        encode: (value: string) => new Uint8Array((textEncoder ||= new TextEncoder()).encode(value)),
```
steveluscher, line 13: Maybe now is the right time to make a `text-encoding-impl` package just like `fetch-impl` and `ws-impl` so that you can build separate browser/node and react-native bundles. The React Native one would basically take on some polyfill as a dependency.

Be sure to search for `fetch-impl` because there are some build scripts that have to be updated.

packages/test-config/setup-text-encoder.ts
```diff
@@ -1,5 +1,11 @@
-import { TextEncoder } from 'util';
+import { TextDecoder, TextEncoder } from 'util';
 
 if (typeof globalThis.TextEncoder === 'undefined') {
```
steveluscher, line 3: 👍🏻

steveluscher (approved):
spiderman-spidey

lorisleiva:
Closing this PR in favour of #1632. I will update the new PR with the comments in this PR before marking it as ready for review.

packages/codecs-strings/src/base64.ts
lorisleiva: We've only got tests for the browser and node.js environments though so it won't test the baseXReslice implementation. 🤔

PR #1611, opened 2023-09-19
refactor(experimental): make sure that subscriptions clean themselves up no matter how the connection closes

steveluscher:
refactor(experimental): make sure that subscriptions clean themselves up no matter how the connection closes

# Summary

The bug here was that, if the connection closes for a non-fatal reason (eg. it was explicitly aborted) we want to unhook the subscription abort signal unconditionally. This prevents a situation where aborting a subscription after the connection has dropped runs the unsubscribe logic.

# Test Plan

```
cd packages/rpc-transport
pnpm test:unit:browser
pnpm test:unit:node
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/solana-labs/solana-web3.js/pull/1611).
* #1613
* #1612
* __->__ #1611

### DIFF
--- a/packages/rpc-transport/src/__tests__/json-rpc-subscription-test.ts
+++ b/packages/rpc-transport/src/__tests__/json-rpc-subscription-test.ts
@@ -13,16 +13,22 @@ interface TestRpcSubscriptionNotifications {
 }
 
 describe('JSON-RPC 2.0 Subscriptions', () => {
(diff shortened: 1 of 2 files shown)


mcintyre94 (approved):

buffalojoec (approved):
lgtm, good testing here

PR #1617, opened 2023-09-21
refactor(experimental): make signTransaction take multiple keypairs

### LINKED ISSUE #1615: [experimental] `signTransaction` should probably allow you to pass multiple signers
As a convenience:

```ts
const signedTx = await signTransaction([pk1, pk2], tx);
```

Rather than what you have to do now:

```ts
const partiallySignedTx = await signTransaction(pk1, tx);
const signedTx = await signTransaction(pk2, tx);
```

buffalojoec:
Modifies the `signTransaction` function to take multiple signers instead of one.

Closes #1615

### DIFF
--- a/packages/transactions/src/signatures.ts
+++ b/packages/transactions/src/signatures.ts
@@ -24,11 +24,15 @@ export async function signTransaction<TTransaction extends Parameters<typeof com
     const compiledMessage = compileMessage(transaction);
     const nextSignatures: Record<Base58EncodedAddress, Ed25519Signature> =
         'signatures' in transaction ? { ...transaction.signatures } : {};
-    for (const keyPair of keyPairs) {
-        const [signerPublicKey, signature] = await Promise.all([
-            getAddressFromPublicKey(keyPair.publicKey),
-            getCompiledMessageSignature(compiledMessage, keyPair.privateKey),
-        ]);
+    const publicKeySignaturePairs = await Promise.all(
+        keyPairs.map(keyPair =>
+            Promise.all([
+                getAddressFromPublicKey(keyPair.publicKey),
+                getCompiledMessageSignature(compiledMessage, keyPair.privateKey),
+            ])
+        )
+    );
+    for (const [signerPublicKey, signature] of publicKeySignaturePairs) {
         nextSignatures[signerPublicKey] = signature;
     }
     const out = {
--- a/packages/transactions/src/signatures.ts
+++ b/packages/transactions/src/signatures.ts
@@ -18,18 +18,19 @@ async function getCompiledMessageSignature(message: CompiledMessage, secretKey: 
 }
 
 export async function signTransaction<TTransaction extends Parameters<typeof compileMessage>[0]>(
-    keyPair: CryptoKeyPair,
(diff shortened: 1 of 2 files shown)


packages/transactions/src/signatures.ts
```diff
@@ -18,18 +18,19 @@ async function getCompiledMessageSignature(message: CompiledMessage, secretKey:
...
-    };
+    const nextSignatures: Record<Base58EncodedAddress, Ed25519Signature> =
+        'signatures' in transaction ? { ...transaction.signatures } : {};
+    for (const keyPair of keyPairs) {
+        const [signerPublicKey, signature] = await Promise.all([
+            getAddressFromPublicKey(keyPair.publicKey),
+            getCompiledMessageSignature(compiledMessage, keyPair.privateKey),
+        ]);
+        nextSignatures[signerPublicKey] = signature;
+    }
```
steveluscher, line 33: Totally, but don't await these in a loop; parallelize!

```suggestion
    const publicKeySignaturePairs = await Promise.all(
        keyPairs.map(keyPair => Promise.all([
            getAddressFromPublicKey(keyPair.publicKey),
            getCompiledMessageSignature(compiledMessage, keyPair.privateKey),
        ]),
    );
    for (const [signerPublicKey, signature] of publicKeySignaturePairs) {
        nextSignatures[signerPublicKey] = signature;
    }
```

packages/transactions/src/signatures.ts
```diff
@@ -18,18 +18,19 @@ async function getCompiledMessageSignature(message: CompiledMessage, secretKey:
 }
 
 export async function signTransaction<TTransaction extends Parameters<typeof compileMessage>[0]>(
-    keyPair: CryptoKeyPair,
+    keyPairs: CryptoKeyPair[],
```
steveluscher, line 21: Do you think this is the way to go, or should we offer two function overloads, one with a single and one with an array, and then check which we got?

steveluscher: ie. So that you can do both of these.

```ts
signTransaction(signerKeyPair, tx);
signTransaction([signerKeyPairA, signerKeyPairB], tx);
```

packages/transactions/src/signatures.ts
buffalojoec: Ah, nice call

packages/transactions/src/signatures.ts
buffalojoec: I could go either way honestly, but I went with the simpler approach.
It's not too much to ask people to bracket a single keypair, but if you think we should do the overloads I am fine with that as well, for some extra UX sugar

packages/transactions/src/signatures.ts
steveluscher: Man. We have to install that ‘no-await-in-loop’ linter.

packages/transactions/src/signatures.ts
steveluscher: The way you have it is probably right!

buffalojoec:
I'm not sure why CI continues to hang for me, but this is ready for another look.

PR #1618, opened 2023-09-21
refactor(experimental): transaction type tests

### LINKED ISSUE #1597: Write type tests for the `@solana/transactions` library
As discussed in #1577, it's probably a good idea to write some type tests for the functions inside of `@solana/transaction`.

An example of these type tests can be found here:

https://github.com/solana-labs/solana-web3.js/blob/bba4a25bdaf07425f91ebb256da5a71e7b8641fe/packages/rpc-core/src/rpc-methods/__typetests__/get-block-type-test.ts#L45-L57

buffalojoec:
Adds typetests for the `@solana/transactions` library.

Closes #1597

### DIFF
--- a/packages/transactions/src/__typetests__/transactions-typetests.ts
+++ b/packages/transactions/src/__typetests__/transactions-typetests.ts
@@ -1,0 +1,463 @@
+import { Base58EncodedAddress } from '@solana/addresses';
+
+import {
+    appendTransactionInstruction,
+    Blockhash,
+    IDurableNonceTransaction,
+    ITransactionWithBlockhashLifetime,
+    ITransactionWithSignatures,
+    Nonce,
+    prependTransactionInstruction,
+    setTransactionLifetimeUsingBlockhash,
+    setTransactionLifetimeUsingDurableNonce,
+    signTransaction,
+} from '..';
+import { createTransaction } from '../create-transaction';
+import { ITransactionWithFeePayer, setTransactionFeePayer } from '../fee-payer';
+import { CompiledMessage, compileMessage } from '../message';
+import { Transaction } from '../types';
+import { getUnsignedTransaction } from '../unsigned-transaction';
+
+const mockFeePayer = null as unknown as Base58EncodedAddress<'feePayer'>;
+const mockBlockhash = null as unknown as Blockhash;
+const mockBlockhashLifetime = {
+    blockhash: mockBlockhash,
+    lastValidBlockHeight: 0n,
+};
+const mockSigner = {} as CryptoKeyPair;
+const mockNonceConfig = {
+    nonce: null as unknown as Nonce<'nonce'>,
+    nonceAccountAddress: null as unknown as Base58EncodedAddress<'nonce'>,
+    nonceAuthorityAddress: null as unknown as Base58EncodedAddress<'nonceAuthority'>,
+};
+const mockInstruction = {
+    accounts: [],
+    data: Uint8Array.of(0),
+    programAddress: null as unknown as Base58EncodedAddress<'program'>,
+} as Transaction['instructions'][number];
+
+async () => {
+    // createTransaction
+    createTransaction({ version: 'legacy' }) satisfies Extract<Transaction, { version: 'legacy' }>;
+    // @ts-expect-error version should match
+    createTransaction({ version: 0 }) satisfies Extract<Transaction, { version: 'legacy' }>;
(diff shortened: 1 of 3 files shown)


buffalojoec:
Same CI problem here

packages/transactions/src/__typetests__/transactions-typetests.ts
```diff
@@ -0,0 +1,408 @@
...
+        const mockV0Transaction = createTransaction({ version: 0 });
+        mockV0Transaction satisfies Transaction & { version: 0 };
+    }
+    // setTransactionFeePayer
+    {
+        const mockLegacyTransaction = createTransaction({ version: 'legacy' });
+        const legacyTxWithFeePayer = setTransactionFeePayer(mockFeePayer, mockLegacyTransaction);
+        legacyTxWithFeePayer satisfies Transaction & { version: 'legacy' } & {
+            feePayer: Base58EncodedAddress<'feePayer'>;
+        };
```
steveluscher, line 56: Isn't the thing we want to assert that:


```suggestion
        legacyTxWithFeePayer satisfies ITransactionWithFeePayer<'feePayer'>;
```

That's the critical part. You _maybe_ want to assert that it _also_ satisfies the inbound type.

```suggestion
        legacyTxWithFeePayer satisfies (typeof mockLegacyTransaction) & ITransactionWithFeePayer<'feePayer'>;
```

packages/transactions/src/__typetests__/transactions-typetests.ts
```diff
@@ -0,0 +1,408 @@
...
+function assertNotAProperty<T extends object, TPropName extends string>(
+    _: { [Prop in keyof T]: Prop extends TPropName ? never : T[Prop] },
+    _propName: TPropName
+): void {}
+
+async () => {
+    // createTransaction
+    {
+        const mockLegacyTransaction = createTransaction({ version: 'legacy' });
+        mockLegacyTransaction satisfies Transaction & { version: 'legacy' };
```
steveluscher, line 46: I think this is more like ‘satisfies the member of the union whose version is …’


```suggestion
        mockLegacyTransaction satisfies Extract<Transaction, { version: 'legacy' }>;
```

steveluscher (changes requested):
In general, I think you can make this way simpler by keeping the tests tighter. By this I mean not to create the actual objects, just pass types in and make sure you get the right types out.

```ts
const LEGACY_TX = null as unknown as Extract<Transaction, {version: 'legacy'}>;

// ...then later
const feePayer = null as unknown as Base58EncodedAddress<'abc'>;
setTransactionFeePayer(LEGACY_TX, feePayer) satisfies ITransactionWithFeePayer<'abc'>;
```

Next, what we're _really_ missing from these tests is the unhappy path. For instance we want to make sure that:

* if you pass a transaction _without_ a fee payer or lifetime to `signTransaction()` that you get a type error (that we have to suppress with `// @ts-expect-error`)
* if you pass a _signed_ transaction into pretty much any other function it comes out _without_ the `ITransactionWithSignatures` type (ie. `satisfies ITransactionWithSignatures` raises a type error)

packages/transactions/src/__typetests__/transactions-typetests.ts
buffalojoec: Done!

buffalojoec (commented):
OK so I've gone ahead and re-did these type tests and covered all of the cases described.

These bled out a couple small bugs.

### Fee Payer Mismatch

I think it's OK for us to support overwriting the fee payer of a transaction, and our types should support this. Basically, with the current implementation, if you provide a transaction with one fee payer and try to assign it another fee payer, you get a type error:

```typescript
setTransactionFeePayer(
        mockFeePayer,
        null as unknown as Extract<Transaction, { version: 'legacy' }> & ITransactionWithFeePayer<'NOTfeePayer'>
    ) satisfies Extract<Transaction, { version: 'legacy' }> & ITransactionWithFeePayer<'feePayer'>;

// Type '"NOTfeePayer" & { readonly __brand: unique symbol; }' is not assignable to type '"feePayer"'.
```

I changed this type signature.

### Durable Nonce Does Not Drop Signatures

Unlike it's blockhash lifetime counterpart, `setTransactionLifetimeUsingDurableNonce` does not remove signatures from the transaction. It should, since it's modifying the instruction set.

I've added this to the function.

PR #1619, opened 2023-09-21
refactor(experimental): add coercion utility functions

### LINKED ISSUE #1600: Introduce a set of assertion-coercion utilities for experimental Web3 JS types
Introduce a set of assertion-coercion utilities that turn this:

```typescript
await rpc.requestAirdrop(address, 1_000_000n as LamportsUnsafeBeyond2Pow53Minus1).send()
```

…into this:

```typescript
await rpc.requestAirdrop(address, lamports(1_000_000n)).send()
```

`lamports()` being essentially:

```typescript
function lamports(putativeLamports: bigint): LamportsUnsafeBeyond2Pow53Minus1 {
  assertIsLamports(putativeLamports);
  return putativeLamports;
}
```

Aside from `lamports()`, cover other types such as `address()`: `Base58EncodedAddress`.

buffalojoec:
This PR adds type cercion utility functions such as `lamports()` and `address()` to our API.

Note: I originally felt like these didn't belong in `@solana/rpc-core`, but I followed the pattern of placing these coercion functions in the same file as their assertions. It seems to make sense. Lmk any thoughts on where to place these.

Closes #1600

### DIFF
--- a/packages/addresses/src/base58.ts
+++ b/packages/addresses/src/base58.ts
@@ -4,6 +4,27 @@ export type Base58EncodedAddress<TAddress extends string = string> = TAddress & 
     readonly __brand: unique symbol;
 };
 
+export function isBase58EncodedAddress(
+    putativeBase58EncodedAddress: string
+): putativeBase58EncodedAddress is Base58EncodedAddress<typeof putativeBase58EncodedAddress> {
+    // Fast-path; see if the input string is of an acceptable length.
+    if (
+        // Lowest address (32 bytes of zeroes)
+        putativeBase58EncodedAddress.length < 32 ||
+        // Highest address (32 bytes of 255)
+        putativeBase58EncodedAddress.length > 44
+    ) {
+        return false;
+    }
+    // Slow-path; actually attempt to decode the input string.
+    const bytes = base58.serialize(putativeBase58EncodedAddress);
+    const numBytes = bytes.byteLength;
+    if (numBytes !== 32) {
+        return false;
+    }
+    return true;
+}
+
 export function assertIsBase58EncodedAddress(
     putativeBase58EncodedAddress: string
 ): asserts putativeBase58EncodedAddress is Base58EncodedAddress<typeof putativeBase58EncodedAddress> {
@@ -28,6 +49,13 @@ export function assertIsBase58EncodedAddress(
             cause: e,
         });
     }
+}
+
+export function address<TAddress extends string = string>(
+    putativeBase58EncodedAddress: TAddress
+): Base58EncodedAddress<TAddress> {
+    assertIsBase58EncodedAddress(putativeBase58EncodedAddress);
+    return putativeBase58EncodedAddress as Base58EncodedAddress<TAddress>;
 }
 
 export function getBase58EncodedAddressCodec(
--- a/packages/addresses/src/__tests__/coercions-test.ts
+++ b/packages/addresses/src/__tests__/coercions-test.ts
@@ -1,0 +1,17 @@
+import { address, Base58EncodedAddress } from '../base58';
+
+describe('coercions', () => {
+    describe('address', () => {
+        it('can coerce to `Base58EncodedAddress`', () => {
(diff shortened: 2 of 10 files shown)


mcintyre94 (approved):
Nice! These are way nicer :)
Do you have any thoughts on whether we should still export the `assertIs...` functions?

buffalojoec:
> Do you have any thoughts on whether we should still export the `assertIs...` functions?

Good call. I would say no, we shouldn't actually. Want me to add to this PR?

mcintyre94:
> Good call. I would say no, we shouldn't actually. Want me to add to this PR?

I think we might as well remove them here yep. Just have to make sure we export a similar conversion function for any new ones going forward but that's not a bad thing!

lorisleiva:
Personally, I'd vote for the following consistent API for all types:

```ts
type SomeType = ...

export function isSomeType (value): value is SomeType {...}

export function assertIsSomeType (value): asserts value is SomeType {...}

export function someType (value): SomeType {...}
```

Perhaps with the exception of `address` since it's gonna be used everywhere.

buffalojoec:
OK yeah I can dig that. I'll update this PR. Do we think just exporting all of each is OK? Each module is wildcard-exported rn

buffalojoec:
I guess each one would look like this, since we want the `assert` to throw but `is` to return a boolean:

```typescript
export type UnixTimestamp = number & { readonly __brand: unique symbol };

export function isUnixTimestamp(putativeTimestamp: number): putativeTimestamp is UnixTimestamp {
    // see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#the_epoch_timestamps_and_invalid_date
    if (putativeTimestamp > 8.64e15 || putativeTimestamp < -8.64e15) {
        return false;
    }
    return true;
}

export function assertIsUnixTimestamp(putativeTimestamp: number): asserts putativeTimestamp is UnixTimestamp {
    // see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#the_epoch_timestamps_and_invalid_date
    try {
        if (putativeTimestamp > 8.64e15 || putativeTimestamp < -8.64e15) {
            throw new Error('Expected input number to be in the range [-8.64e15, 8.64e15]');
        }
    } catch (e) {
        throw new Error(`\`${putativeTimestamp}\` is not a timestamp`, {
            cause: e,
        });
    }
}

export function unixTimestamp(putativeTimestamp: number): UnixTimestamp {
    assertIsUnixTimestamp(putativeTimestamp);
    return putativeTimestamp;
}
```

packages/addresses/src/base58.ts
```diff
@@ -30,6 +30,13 @@ export function assertIsBase58EncodedAddress(
     }
 }
 
+export function address<TAddress extends string = string>(
+    putativeBase58EncodedAddress: TAddress
+): Base58EncodedAddress<TAddress> {
+    assertIsBase58EncodedAddress(putativeBase58EncodedAddress);
```
steveluscher, line 36: This is obviously the right thing to do, semantically, I just wish that doing this didn't have a CPU cost.

```ts
// I get this for free.
const address = '555555555555555555555555` as Base58EncodedAddress<'555555555555555555555555'>;

// This calls `base58.serialize` to perform the assertion, ffs
const address = address('555555555555555555555555');
```

steveluscher (approved):
giphy-4

buffalojoec:
Added @lorisleiva's API suggestions. Good to go!

packages/addresses/src/base58.ts
buffalojoec: Doesn't seem like we can avoid it, right? If someone uses the former, they get the CPU bonus but no assertion, and if they instead use the latter, it's the opposite.

Ideally the real slick devs could just _make sure_ their string is a valid base58 address and go with the former

steveluscher: Yeah. I'm just whining out loud.

steveluscher:
giphy-7

PR #1627, opened 2023-09-26
feat(experimental): add curry to functional lib

buffalojoec:
This PR adds the `curry` function to the `@solana/functional` library.

This function is analogous to using JavaScript's `.bind(..)` on a function, however, it provides a more intuitive and simpler approach to performing function currying.

```javascript
const add = (a: number, b: number) => a + b;
// Using bind
const addOneWithBind = add.bind(null, 1);
addOneWithBind(2); // 3
// Using curry
const addOneWithCurry = curry(add, 1);
addOneWithCurry(2); // 3
```

Motivation: most people may not really know how to use `bind`, and seeing `curry` in an example could be more intuitive to most.

### DIFF
--- a/packages/functional/src/curry.ts
+++ b/packages/functional/src/curry.ts
@@ -1,0 +1,382 @@
+/**
+ * General curry function.
+ *
+ * The largest arity present in all `@solana` packages belongs to `pipe`,
+ * which supports arity of 10. Thus `curry` also supports arity of 10.
+ * @see https://github.com/solana-labs/solana-web3.js/blob/master/packages/functional/src/pipe.ts
+ *
+ * Using this function you can curry any function, binding any number of arguments.
+ * ```typescript
+ * // Using some defined function
+ * const add = (a: number, b: number) => a + b;
+ * // You can curry in one line
+ * const sum = curry(add)(1, 2);    // 3
+ * // Or you can curry with a variable
+ * const addStuff = curry(add);
+ * const sum = addStuff(1, 2);      // 3
+ * // You can also bind some arguments
+ * const addOne = curry(add, 1);
+ * const sum = addOne(2);           // 3
+ * // Or bind all arguments
+ * const addOneAndTwo = curry(add, 1, 2);
+ * const sum = addOneAndTwo();      // 3
+ * ```
+ * @param fn    The function to be curried
+ * @param args  Any default arguments to bind to the function
+ * @returns     The curried function
+ */
+export function curry<R>(fn: () => R): () => R;
+// 1 arg
+export function curry<A, R>(fn: (a: A) => R): (a: A) => R;
+export function curry<A, R>(fn: (a: A) => R, a: A): () => R;
+// 2 args
+export function curry<A, B, R>(fn: (a: A, b: B) => R): (a: A, b: B) => R;
+export function curry<A, B, R>(fn: (a: A, b: B) => R, a: A): (b: B) => R;
+export function curry<A, B, R>(fn: (a: A, b: B) => R, a: A, b: B): () => R;
+// 3 args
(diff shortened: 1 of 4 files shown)


mcintyre94 (commented):
I'm a bit hesitant about adding this. Unlike pipe I don't think this adds a lot over the `bind` primitive

And if people do want to use something like this then I think it might be better to use the functional library of their choice. You could make the same argument for our `pipe` (that it's in every functional lib), but we have code using that in the library ourselves, and it's pretty essential to our transaction API.

Do you have any thoughts on where we'd want to use this/encourage its use?

buffalojoec:
> I'm a bit hesitant about adding this. Unlike pipe I don't think this adds a lot over the `bind` primitive
> 
> And if people do want to use something like this then I think it might be better to use the functional library of their choice. You could make the same argument for our `pipe` (that it's in every functional lib), but we have code using that in the library ourselves, and it's pretty essential to our transaction API.
> 
> Do you have any thoughts on where we'd want to use this/encourage its use?

Definitely a fair take.

My thoughts were basically:
- It adds more useful tooling to `@solana/functional`
- It's slightly more brief and intuitive to use than `bind`
   - Seeing some modules written in functional style with a nice blend of `curry` and `pipe` would be pretty cool

As far as specific use cases, pretty much anywhere someone might use `bind`. Steve has thrown a lot of `bind` examples around for transaction building, etc. and you could instead curry it a few times (for different functionality or defaults) and provide these curried functions to `pipe`.

I definitely understand your point(s), but I think if we end up exporting `@solana/functional` from the main `@solana/web3.js` or encouraging folks to use it in their clients, we should have some of the common functional methods built with type-safety in mind. Also even more pertinent if any of `@solana/functional` lands in our docs.

mcintyre94:
I think we definitely will be using `pipe` pretty heavily in our docs because we really don't have a good way to write transaction code without that. So I think you're right that if there is a main `web3.js` it'll be exported from there and used in the docs a lot

I think the other good argument in favour is that nothing in our code will use `this`, so using `bind` with our stuff will always have a pointless first argument, which your `curry` removes. And these functions are written with the transaction last so currying/binding will work well.

packages/functional/src/curry.ts
```diff
@@ -0,0 +1,382 @@
...
+    e: E,
+    f: F,
+    g: G,
+    h: H,
+    i: I,
+    j: J
+): () => R;
+// Implementation
+export function curry(fn: (...args: unknown[]) => unknown, ...args: unknown[]) {
+    return fn.bind(null, ...args);
```
steveluscher, line 381: Couple notes about this.

* Currying hipsters are used to being able to chain calls like this: `sum3(1)(2)(3)`
* Currying hipsters are also used to the functions auto-invoking once you've met or exceeded the arity of the original function: `sum3(1)(2, 3); // 6`
* It would be nice in any case to have the property that `curry()` either invokes immediately, or returns the original function when the thing has no effect, so that we're not needlessly creating copies
    ```ts
    const f = () => {};
    const fCurried = curry(f);
    f === fCurried; // true
    ````

steveluscher (changes requested):
Whatever we decide, I don't think this is what most functional programming hipsters think of when they think of a currying utility. What we've built here, they know by the name ‘partial application.’

packages/functional/src/curry.ts
buffalojoec: OK, I can add this behavior, as long as we're in favor of adding it to the lib. Sounds like @mcintyre94 was on the fence, and I am in favor. Steve is... ?

steveluscher: I think I'm also sort of on the fence. I haven't needed to do this yet, at least not in quite the way that Real Currying™ allows.

buffalojoec:
I'm gonna let this stew a little, considering the feedback from the gang, and probably close eventually.

RIP

PR #1643, opened 2023-09-29
refactor(experimental): change u64 types to slot in slot notifications

buffalojoec:

PR #1645, opened 2023-09-29
feat(experimental): add `slotsUpdatesNotifications` websocket method

buffalojoec:


buffalojoec:
* master
  * **PR #1643** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1643" target="_blank">Graphite</a> 
    * **PR #1644** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1644" target="_blank">Graphite</a> 
      * **PR #1645** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1645" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1645?utm_source=stack-comment).

packages/rpc-core/src/rpc-subscriptions/slots-updates-notifications.ts
```diff
@@ -0,0 +1,20 @@
...
+    slot: Slot;
+    timestamp: UnixTimestamp;
+    type: 'completed' | 'createdBank' | 'dead' | 'firstShredReceived' | 'frozen' | 'optimisticConfirmation' | 'root';
+}>;
+
+export interface SlotsUpdatesNotificationsApi {
+    /**
+     * Subscribe to receive a notification from the validator on a variety of updates on every slot
+     */
+    slotsUpdatesNotifications(
```
steveluscher, line 16: Oof. This really underscores how grammatically incorrect this method name is. It should have been `slotUpdateNotification`/`slotUpdateSubscribe`.

packages/rpc-core/src/response-patcher-allowed-numeric-values.ts
```diff
@@ -1,22 +1,28 @@
...
  */
 export function getAllowedNumericKeypathsForNotification(): AllowedNumericKeypaths<
-    ReturnType<typeof createSolanaRpcSubscriptionsApi>
+    IRpcSubscriptionsApi<SolanaRpcSubscriptions & SolanaRpcSubscriptionsUnstable>
 > {
     if (!memoizedNotificationKeypaths) {
-        memoizedNotificationKeypaths = {};
+        memoizedNotificationKeypaths = {
+            slotsUpdatesNotifications: [['timestamp']],
+        };
```
steveluscher, line 25: Knew I created this empty object for a reason.

packages/rpc-core/src/rpc-subscriptions/__typetests__/slots-updates-notifications-type-test.ts
```diff
@@ -0,0 +1,33 @@
...
+
+import { RpcSubscriptions } from '@solana/rpc-transport/dist/types/json-rpc-types';
+
+import { Slot } from '../../rpc-methods/common';
+import { UnixTimestamp } from '../../unix-timestamp';
+import { SolanaRpcSubscriptions, SolanaRpcSubscriptionsUnstable } from '../index';
+
+async () => {
+    const rpcSubscriptions = null as unknown as RpcSubscriptions<
+        SolanaRpcSubscriptions & SolanaRpcSubscriptionsUnstable
```
steveluscher, line 11: You should _only_ include the relevant one.
```suggestion
        SlotsUpdatesNotificationsApi
```

packages/rpc-core/src/rpc-subscriptions/__typetests__/slots-updates-notifications-type-test.ts
```diff
@@ -0,0 +1,33 @@
...
+            type:
+                | 'completed'
+                | 'createdBank'
+                | 'dead'
+                | 'firstShredReceived'
+                | 'frozen'
+                | 'optimisticConfirmation'
+                | 'root';
+        }>
+    >;
```
steveluscher, line 29: I think we'd be better off taking the approach that at every juncture the thing is what we think it is.
```suggestion
    type TNotification = Readonly<{
        parent?: Slot;
        slot: Slot;
        timestamp: UnixTimestamp;
        type:
            | 'completed'
            | 'createdBank'
            | 'dead'
            | 'firstShredReceived'
            | 'frozen'
            | 'optimisticConfirmation'
            | 'root';
    }>;
    rpcSubscriptions
        .slotsUpdatesNotifications() satisfies Promise<PendingRpcSubscription<TNotification>>;
    rpcSubscriptions    
        .slotsUpdatesNotifications()
        .subscribe() satisfies Promise<AsyncIterable<TNotification>>;
```

packages/rpc-core/src/rpc-subscriptions/index.ts
```diff
@@ -48,3 +54,9 @@ export function createSolanaRpcSubscriptionsApi(config?: Config): IRpcSubscripti
         },
     });
 }
+
+export function addSolanaUnstableSubscriptionsApi<TRpcMethods>(
+    rpc: RpcSubscriptions<TRpcMethods>
+): RpcSubscriptions<TRpcMethods & SolanaRpcSubscriptionsUnstable> {
+    return rpc as RpcSubscriptions<TRpcMethods & SolanaRpcSubscriptionsUnstable>;
+}
```
steveluscher, line 62: I wonder if you shouldn't actually do this in `library/src/rpc.ts`. Like, offer two different methods `createSolanaRpcSubscriptions()` and `createSolanaRpcSubscriptions_UNSTABLE()`.

That way we could reserve the right, later on, to modify the `api` we instantiate in each, without people needing to update their code in any way. Meaning: we could inject _actual_ program logic into that API to transform the subscription requests / notification responses.

steveluscher (approved):
[looksgood](https://app.graphite.dev/macros/solana-labs)

packages/rpc-core/src/rpc-subscriptions/index.ts
buffalojoec: Investigating.......

buffalojoec: I added what you suggested!

packages/rpc-core/src/rpc-subscriptions/index.ts
```diff
@@ -3,33 +3,48 @@ import { IRpcSubscriptionsApi, RpcSubscription } from '@solana/rpc-transport/dis
...
-                        ? SolanaRpcSubscriptions[TNotificationName]
+                    (SolanaRpcSubscriptions &
+                        SolanaRpcSubscriptionsUnstable)[TNotificationName] extends CallableFunction
+                        ? (SolanaRpcSubscriptions & SolanaRpcSubscriptionsUnstable)[TNotificationName]
                         : never
                 >
-            ): RpcSubscription<ReturnType<SolanaRpcSubscriptions[TNotificationName]>> {
+            ): RpcSubscription<
+                ReturnType<(SolanaRpcSubscriptions & SolanaRpcSubscriptionsUnstable)[TNotificationName]>
+            > {
```
steveluscher, line 47: Wait, I don't think we need this right? Instead, in your `_UNSTABLE` method you just _force_ the type of this thing to be `IRpcSubscriptionsApi<SolanaRpcSubscriptions & SolanaRpcSubscriptionsUnstable>` because we know it's compatible.

I general, I want someone who uses this directly to get none of the unstable methods. The only way to get the unstable stuff is to override the types (or use the helper you've created that overrides them for you).

steveluscher (approved):
Other than some extra nits, ship it!

[abandonship](https://app.graphite.dev/macros/solana-labs)

packages/rpc-core/src/rpc-subscriptions/index.ts
buffalojoec: OK I'll check that out but I thought I was getting a Type Error when trying to provide both. But maybe just a straight up `as` cast is what you meant? I'll check it out.

buffalojoec: Cool, I think I found a good solution that works better than what I had. I just introduced a forced type from `createSolanaSubscriptionsApi_UNSTABLE()` so you can pass that as an `api` arg in the `createJsonSubscriptionsRpc()` function.

Should be good to go.

packages/rpc-core/src/response-patcher-allowed-numeric-values.ts
```diff
@@ -1,22 +1,28 @@
...
  * to a `bigint`. These are values that are legitimately defined as `u8` or `usize` on the backend.
  */
 export function getAllowedNumericKeypathsForNotification(): AllowedNumericKeypaths<
-    ReturnType<typeof createSolanaRpcSubscriptionsApi>
+    IRpcSubscriptionsApi<SolanaRpcSubscriptions & SolanaRpcSubscriptionsUnstable>
 > {
     if (!memoizedNotificationKeypaths) {
-        memoizedNotificationKeypaths = {};
+        memoizedNotificationKeypaths = {
+            slotsUpdatesNotifications: [['timestamp']],
```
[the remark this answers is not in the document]
buffalojoec: If it's a `u64` in the code then yeah you're right shouldn't be

PR #1654, opened 2023-09-29
Add initial version of `fromVersionedTransactionWithBlockhash`

mcintyre94:
This PR adds a new library `@solana/compat` to contain compatibility helpers for converting between old web3js classes and new data structures

For now it exposes one function, `fromVersionedTransactionWithBlockhash`, which converts from an old web3js transaction with a blockhash lifetime to a new transaction object

It currently has some limitations:

- As the name suggests, only blockhash lifetimes are supported. Durable nonces will be added with a different function, since it will have a slightly different function signature (we differentiate lifetime constraints at type level now)

- Address table lookups aren't yet supported. These will need to decompile instruction account keys to `IAccountLookupMeta`. This will probably be done using a different function that takes an `AddressLookupTableAccount[]` parameter.


packages/compat/README.md
```diff
@@ -0,0 +1,25 @@
+[![npm][npm-image]][npm-url]
+[![npm-downloads][npm-downloads-image]][npm-url]
+[![semantic-release][semantic-release-image]][semantic-release-url]
+<br />
+[![code-style-prettier][code-style-prettier-image]][code-style-prettier-url]
+
+[code-style-prettier-image]: https://img.shields.io/badge/code_style-prettier-ff69b4.svg?style=flat-square
+[code-style-prettier-url]: https://github.com/prettier/prettier
+[npm-downloads-image]: https://img.shields.io/npm/dm/@solana/addresses/experimental.svg?style=flat
+[npm-image]: https://img.shields.io/npm/v/@solana/addresses/experimental.svg?style=flat
```
buffalojoec, line 10: Ya gotta change these bad boys!

packages/compat/src/transaction.ts
```diff
@@ -0,0 +1,158 @@
...
+        role,
+    };
+}
+
+// While the type `MessageCompiledInstruction` says data is a `Uint8Array`,
+// it is actually a Buffer in tests. Presumably this is a bug in the old web3js compile
+// message code. We will just assume it can be either here
+type MessageCompiledInstructionButDataMightBeABuffer = Omit<MessageCompiledInstruction, 'data'> & {
+    data?: Uint8Array | Buffer;
+};
```
buffalojoec, line 56: Buffer is a sub-type of Uint8Array!

packages/compat/src/transaction.ts
```diff
@@ -0,0 +1,158 @@
...
+        if (allZeros) continue;
+
+        const signature = transaction.signatures[i] as Ed25519Signature;
+        signatures[address] = signature;
+    }
+
+    return signatures;
+}
+
+export function fromOldVersionedTransactionWithBlockhash(
```
buffalojoec, line 115: I know you're probably using `Old` here because the transaction versions also use the term "legacy", but I gotta say I am much more in favor of "legacy" than "old" lol.

How can we rename this and still use some flavor of the terminology "Legacy Web3 JS"?

packages/compat/src/transaction.ts
```diff
@@ -0,0 +1,158 @@
...
+        // we don't do that for the new tx model. So just skip if it's all 0s
+        const allZeros = transaction.signatures[i].every(byte => byte === 0);
+        if (allZeros) continue;
+
+        const signature = transaction.signatures[i] as Ed25519Signature;
+        signatures[address] = signature;
+    }
+
+    return signatures;
+}
```
buffalojoec, line 113: If this is just for type coercion, can you just do:

```typescript
function convertSignatures(
    transaction: VersionedTransaction,
    staticAccountKeys: PublicKey[]
): ITransactionWithSignatures['signatures'] {
    return transaction.signatures.filter(s => !s.every(byte => byte === 0)) as Ed25519Signature[];
}
```

buffalojoec: Or something similar. That might not totally work

packages/compat/README.md
mcintyre94: dammit I knew my copying would leak somewhere! TY!

packages/compat/src/transaction.ts
mcintyre94: Hmm, that'd explain this annoyance!
The old one takes a `Buffer`, then compiles it to `Uint8Array` but actually just returns the buffer (which I guess isn't a bug given this sub-type relationship)
But in the tests the `Buffer` ends up as a different and I think worse representation:

```diff
-     "data": Uint8Array [
+     "data": Object {
+       "data": Array [
          0,
          1,
          2,
          3,
          4,
        ],
+       "type": "Buffer",
+     },
```

Maybe it doesn't really matter since it's a `Uint8Array` anyway and we should just leave it as a Buffer?

packages/compat/src/transaction.ts
mcintyre94: Yea same, I prefer legacy to old too. But I think legacy in the context of transactions is just going to be way confusing. Especially because when address lookup tables get added here, this probably becomes 2 different functions (either explicitly or under the hood) depending if it is a legacy transaction version or v0.

packages/compat/src/transaction.ts
mcintyre94: It's not just type coercion, it also re-shapes it to a dictionary, keyed by public key. The old `VersionedTransaction` just has an array of signatures, and they map implicitly to the first N accounts of the transaction. So (in functional terms) we basically zip the first N static accounts with the N signatures.

So you need to maintain the original index, to get to the right account key for each one, which means you can't filter the all 0s out first.

I think there is a purely functional way to do it using reduce, because you can get the index in the callback. That should work here.

packages/compat/src/transaction.ts
mcintyre94: Alright I've just dropped this, it's typed as a `Uint8Array` so we shouldn't care that under the hood it's a `Buffer`. No point adding complexity just to make the tests the exact shape I expected, consumers shouldn't care.

packages/compat/src/transaction.ts
mcintyre94: Refactored in 78764a44b33449822b743b887676882687547f20 - I think that's probably better?

buffalojoec: Slick!!

packages/compat/src/transaction.ts
steveluscher: The reason that people lean on `Buffer` is because it has this convenience that `Uint8Array` does _not_:

```ts
buf.toString('base64'); // Actually works.
```

So long as you can presume that peculiar feature is not needed on the way in or out of here, you can type it `Uint8Array` and it'll be fine.

mcintyre94:
I've added a typetest, which is currently very simple but later on when we have different functions with different signatures they'll be more important

One point it highlights is that currently at the type level you can't tell if the returned transaction is signed or not (`ITransactionWithSignatures`). I don't think it's possible to have a type-level overload purely based on whether `transaction.signatures` is empty because of typescript's runtime limitation.

So a few options:

- Leave it as is. Distinguish between the branches on the output using `if('signatures' in convertedTransaction)`. Or if you know it's signed you can just do `as ITransactionWithSignatures`. There's limited help from intellisense here though and it seems to type `convertedTransaction.signatures` as unknown if you just do the if check, which is probably the same typescript bug we've seen before.

- Create two separate functions, one for signed and one for unsigned, with different return types. The signed one checks if the transaction has any signatures and throws if it doesn't. The caller must check first, and then will get the correct type back depending which function they call. Under the hood they're the same implementation after that check. 

- Create a new type `SignedVersionedTransaction` which specifies that `signatures` is non-empty, and add a type assertion function that checks. Then we can just use TS overloads to say that when that's passed in you get `ITransactionWithSignatures`. The caller must again check first, and either cast or assert to convert their transaction to `SignedVersionedTransaction`. This would just use TS overloads so there's a single implementation with no check in our code.

I think I prefer the last option, because it's most similar to what we do elsewhere in our new libraries. But the downside of both the last two options is that it forces a caller to care about whether the transaction is signed. Currently if you don't care (like you just want the instructions in the new format or something) you don't have to do anything and the type you're getting is fine. But if you do care about signatures then the current implementation is a bit rubbish.

packages/compat/src/transaction.ts
```diff
@@ -0,0 +1,141 @@
...
+): Transaction & ITransactionWithFeePayer & ITransactionWithBlockhashLifetime & ITransactionWithSignatures;
+
+export function fromOldVersionedTransactionWithBlockhash(
+    transaction: VersionedTransaction,
+    lastValidBlockHeight: bigint
+): Transaction & ITransactionWithFeePayer & ITransactionWithBlockhashLifetime;
+
+export function fromOldVersionedTransactionWithBlockhash(
+    transaction: VersionedTransaction,
+    lastValidBlockHeight: bigint
```
buffalojoec, line 102: Ahhh this is annoying lol. I almost wonder what would be better: Requiring you to hang onto `lastValidBlockHeight` whenever you added the blockhash to your old tx, or simply passing both here - `blockhash` and `lastValidBlockHeight`

packages/compat/src/transaction.ts
buffalojoec: I wonder what it might look like if we just removed the word `Old` completely?

Or maybe we used `Deprecated` (idk if it really is yet). Or something worse: `LegacyWeb3JsVersionedTransaction1 lol.

buffalojoec (commented):
This lgtm I just have a few small things I asked about

buffalojoec:
> The caller must again check first, and either cast or assert to convert their transaction to `SignedVersionedTransaction`

I also like option 3 but this bit does sound a little annoying, but it's not the end of the world. I think it's more important to have the proper types.

I thought your implementation of it looked good though, so option 3 is the way for me! Callers would probably just cast the same way you are.

packages/compat/src/transaction.ts
mcintyre94: Yep fair point! Might be better to just have you pass both. I guess otherwise you have to mutate the old transaction to have the matching blockhash.

packages/compat/src/transaction.ts
mcintyre94: I'll drop Old, we can just have `fromVersionedTransactionWithBlockhash` etc. for now. Will have to figure something else out if we end up writing compatibility for the legacy-legacy `Transaction` class, but can figure that out if it happens!

packages/compat/src/__tests__/signed-versioned-transaction-test.ts
```diff
@@ -0,0 +1,142 @@
...
+import {
+    assertIsSignedVersionedTransaction,
+    isSignedVersionedTransaction,
+    signedVersionedTransaction,
+} from '../signed-versioned-transaction';
+
+describe('signed-versioned-transactions', () => {
+    const payerKey = Keypair.generate().publicKey;
+    const blockhashString = '4JgzET199353yzPPZ7QUAFa3FggaLmkeMczKwqW4VqTT';
+    const signature = new Uint8Array(Array(64).fill(1));
```
steveluscher, line 12: Small nit: since this puppy is mutable, you might consider re-initializing it on every pass.
```suggestion
    let signature: Uint8Array;
    beforeEach(() => {
        signature = new Uint8Array(Array(64).fill(1));
    });
```

steveluscher: Same with `payerKey`.

packages/compat/src/signed-versioned-transaction.ts
```diff
@@ -0,0 +1,37 @@
...
+
+function signatureIsAllZeros(signature: Uint8Array) {
+    return signature.every(byte => byte === 0);
+}
+
+export function isSignedVersionedTransaction(
+    versionedTransaction: VersionedTransaction
+): versionedTransaction is SignedVersionedTransaction {
+    return versionedTransaction.signatures.findIndex(sig => !signatureIsAllZeros(sig)) > -1;
+}
```
steveluscher, line 17: This isn't really compat-y. This is an operation over exclusively the legacy data type. It probably doesn't belong here.

Instead, I'd do one of:

1. Put this in the legacy library, if it's useful.
2. _Let_ them convert it to modern _first_, and then use the modern assertion `assertIsFullySignedTransaction` (which we have yet to write in `@solana/transactions`).

steveluscher (changes requested):
These are going to be great. In general I think this layer _has_ to be as dumb as possible, and some of the more useful stuff you've added here should be pushed into the modern library.

steveluscher (approved):
I read your longer comment about signatures in general @mcintyre94. Rather than to create a method that's like `fromSignedVersionedTransaction()` into which you pass a ‘`VersionedTransaction` that you definitely know to be signed’ that you do one of two things:

## Use the assertions of the modern library to refine the types

```ts
const tx = fromVersionedTransaction(versionedTransactionIKnowToBeSigned);
assertIsFullySignedTransaction(tx);
getBase64EncodedWireTransaction(tx);
```

## Just tell Typescript what to think

```ts
const tx = fromVersionedTransaction(versionedTransactionIKnowToBeSigned);
getBase64EncodedWireTransaction(tx as (typeof tx & ITransactionWithSignatures));
```

steveluscher (changes requested):
.

packages/compat/src/transaction.ts
```diff
@@ -0,0 +1,144 @@
...
+
+        const address = staticAccountKeys[index].toBase58() as Base58EncodedAddress;
+        return { ...acc, [address]: sig as Ed25519Signature };
+    }, {});
+}
+
+export function fromVersionedTransactionWithBlockhash(
+    transaction: SignedVersionedTransaction,
+    blockhash: Blockhash,
+    lastValidBlockHeight: bigint
```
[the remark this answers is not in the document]
mcintyre94: Previously I used the last blockhash from their transaction but required them to pass `lastValidBlockHeight`. Then I refactored to have them pass both, since they'd get both together. If just defaulting `lastValidBlockHeight` to `U64_MAX` works for us then I'm happy to change to doing that, for some reason I thought you had to use the value that matches what you got when you fetched the blockhash. Will update, thanks!

packages/compat/src/signed-versioned-transaction.ts
mcintyre94: I don't think `IFullySignedTransaction` is what we want here, you should be able to convert a transaction regardless of if it's signed (eg for simulation I want it to be unsigned/partially signed)

But you're right that we could do the assertion on the modern transaction, just to `ITransactionWithSignatures`. So we just do `fromVersionedTransactionWithBlockhash`, that gives you the transaction but not `ITransactionWithSignatures`, and you can then assert it has signatures  or cast to get `ITransactionWithSignatures` if you care about that.

The remaining question though is what to do with those all-zero signatures when I convert them? The new transaction model (AFAICT) just stores actual signatures in a dictionary. We could remove the filter so when you pass a transaction it ends up with the all-zero signature in that dictionary. Presumably that'd work fine, it's just not what you'd really expect with the new transaction model IMO.

packages/compat/src/transaction.ts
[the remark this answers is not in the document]
buffalojoec: yeah I didn't know you could do this tbh

mcintyre94:
Okay cool I've stripped this way back

- You no longer need to pass `lastValidBlockHeight`. The `blockhash` from the transaction is used, and `lastValidBlockHeight` is set to U64_MAX
- There's no longer a separate function for signed vs unsigned. The return type is always `Transaction & ITransactionWithFeePayer & ITransactionWithBlockhashLifetime`
- Signatures (if any) are copied over, and I'm still filtering out the all-zero ones

There's no assertion function for `ITransactionWithSignatures` yet or signature validity check, we can add those to the new transactions library independently of this.

packages/compat/src/transaction.ts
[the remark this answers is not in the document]
steveluscher: The function of `lastValidBlockHeight` is to signal to the confirmation function when it’s time to give up. As in: hey, if the slot is higher than X, stop waiting for the transaction to confirm.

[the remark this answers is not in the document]
steveluscher: Defaulting this to some very high number essentially disables this feature. That isn’t great either, and come to think of it is a behavioural change from what happens now if you don’t supply a block height (ie. we timeout after X seconds).

Sigh.

Options:

1. We create first class support in modern for not specifying a block height (please no)
2. We make you supply a block height when you `fromVersionedTransaction`; halfway between this and what you had before.
3. We make the confirmation function unconditionally time out after X, meaning that all a block height could possibly do would be to make it time out sooner (pretty much the intent of it)
4. Sort of like 3 crossed with 1, make a unique `Symbol` that’s acceptable in the type position for `lastValidBlockHeight` that represents `UNKNOWN_BLOCK_HEIGHT` and use that.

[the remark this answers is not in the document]
buffalojoec: > 1. We create first class support in modern for not specifying a block height (please no)

Agreed, definitely don't want to change experimental behavior as a result of how legacy is written

2 is probably the most functionally correct, but the dev ex is rough. Why don't we implement the behavior of 3, but add an overload for providing a block height instead of X? That way, if you convert over and don't pass it, you get default X, if you pass it _or_ if you're doing send/confirm in the new lib without `compat`, you get the proper control?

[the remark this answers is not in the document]
buffalojoec: I think I reiterated 3 more than I realized, but emphasis on the overloading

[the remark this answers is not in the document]
steveluscher: I think that's probably the best option? Make `lastValidBlockHeight` optional, default it to some extremely large number that effectively makes it useless.

The only question is whether to ship a version of `confirmTransaction` that:

* times out after some fixed number of seconds
* never times out. no time out. no more.

I'm sort of inclined to eliminate fixed timeouts completely. This means that if you convert a legacy transaction, and _don't_ supply a `lastValidBlockHeight`, your confirmations will hang indefinitely in the event that the transactions fail. Timing out is up to you at that point.

Actually, I kind of like that. Thoughts?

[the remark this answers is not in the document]
buffalojoec: Wouldn't this take up a thread on your app though? What if you send a bunch of transactions and they are all waiting for confirmation and just keep one or more subscriptions open? Might be bad for performance. I also am not sure how you'd go about force-cancelling that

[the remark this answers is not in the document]
steveluscher: Not really a thread, it would just hang on a continuation that will never continue. It would keep pumping `slotSubscribe()` though, for no reason. If your app gets tired of waiting it can cancel it with the `AbortSignal` you passed in.

I think it would look something like this:

```ts
// MyApp

// Missing sensible `lastValidBlockHeight` so this will try to confirm forever.
const tx = fromVersionedTransaction(legacyTransaction);
const confirmationPromise = confirmTransaction({abortSignal: abortController.signal}, tx);
const timeoutPromise = new Promise((_, reject) => {
  AbortSignal
    .timeout(60_000)
    .addEventListener('abort', reject, {signal: abortController.signal});
});
try {
  await Promise.race([
    confirmationPromise,
    timeoutPromise,
  ]);
  console.log('confirmed');
} catch(e) {
  console.error('o no', e);
} finally {
  abortController.abort();
}
```

packages/compat/src/transaction.ts
```diff
@@ -87,20 +85,10 @@ function convertSignatures(
...
-export function fromVersionedTransactionWithBlockhash(
-    transaction: VersionedTransaction,
-    lastValidBlockHeight: bigint
-): Transaction & ITransactionWithFeePayer & ITransactionWithBlockhashLifetime;
-
 export function fromVersionedTransactionWithBlockhash(
     transaction: VersionedTransaction,
     lastValidBlockHeight: bigint
-): Transaction {
+): Transaction & ITransactionWithFeePayer & ITransactionWithBlockhashLifetime {
```
buffalojoec, line 90: So with this change it will copy over my signatures but I won't get a `ITransactionWithSignatures` type response, right?
You might be able to wrangle it with generics, ie:

```typescript
type OldTransactionUnsigned = VersionedTransaction;
type OldTransactionSigned = VersionedTransaction & { signatures: [] };

type NewTransactionUnsigned = Transaction & ITransactionWithFeePayer & ITransactionWithBlockhashLifetime
type NewTransactionSigned = Transaction & ITransactionWithFeePayer & ITransactionWithBlockhashLifetime & ITransactionWithSignatures

// Psuedo-code
fn<TOldTransaction extends OldTransactionUnsigned | OldTransactionSigned>(old: TOldTransaction): TOldTransaction extends OldTransactionUnsigned ? NewTransactionUnsigned : NewTransactionSigned
```

packages/compat/src/transaction.ts
```diff
@@ -86,8 +86,7 @@ function convertSignatures(
 }
 
 export function fromVersionedTransactionWithBlockhash(
-    transaction: VersionedTransaction,
-    lastValidBlockHeight: bigint
+    transaction: VersionedTransaction
 ): Transaction & ITransactionWithFeePayer & ITransactionWithBlockhashLifetime {
```
buffalojoec, line 90: Also you decided not to overload here? You are just going to default to `u64::MAX` no matter what?

packages/compat/src/__tests__/signed-versioned-transaction-test.ts
mcintyre94: Resolving since these tests are removed

packages/compat/src/transaction.ts
mcintyre94: I tried stuff like this and the problem is that it can't inspect the old transaction and determine whether it has signatures. I guess because of the lack of a Typescript runtime. Which is why I ended up introducing `SignedVersionedTransaction`

packages/compat/src/transaction.ts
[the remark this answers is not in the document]
mcintyre94: Updated to have an optional `lastValidBlockHeight` and default to U64 max

packages/compat/src/transaction.ts
mcintyre94: Updated to add an optional `lastValidBlockHeight` per the latest discussion!

packages/compat/src/transaction.ts
buffalojoec: Ohhhhhh right I _did_ totally read this when you first wrote it a few commits ago, but now I get it

buffalojoec: This actually works though (or appears to)!
[playground link](https://www.typescriptlang.org/play?#code/PTAEBkFMHMEMGMCeAoALogDpUA1SAnAZwEsB7AO0gBMAVfWcwhVM80AXlAG9lQ-RijVPgCu8FhUIAuUIWGDoAbQC6AGl78AZpEgAFWIgIy5+Bev6gARgBtS8ANYALWIUfH55aOf4lo5WKgi+JCEAPwy5CIAtpYEKsgAvsjIIKAAogAeWKZRkOSosNZomNh0DEzirBzcGnyCJmISjO6mnvFJ6FigAJJljMysAOrEqI4AYjr6hvjVPBbaegZGsh5eicVdvfT9lRTDowBCtg7OruDE2iy5s7VWx04ubiutax0lPX0VTfuOAMrEfgCQRCNwsvn8gWC0meChU6iSG2wAHlrLRtl9WABVRgAyhUap4IisaifAYUADciNAKLR5TJ5H+fmoBIIJAoJPR9NAADJuLJcUCoRForF8CpQAlKVSAHKQADupN25Gx4OZnEVTR5H05Sp+E0W0y1WzpupGjiOdgeZwukCukBl8o1rEZeOqTooRvd5D1kyWM15xp23zNFpOj3Ol2I1wDXp+LsFIWSmhE5CVoE05AAPDQaV7QJAMqg8lRCNTUV6VbjmQAfMu0oPOqtUAB8AApSKiZDnyzqmgBKLu53tVAtF8glutel3M0KgWUK4cUaf4mTzis4pn4ub8YKQtiRazWUAuUAp+zkUhytgn1vd+sYj2j4ulocmprL0CzteLhlN0Crx0f0rTc+3WZB4EkVBQAvaciGqDNW23OohFEJVoUUAByYgMgwtRbgWKZlgwgArUhIAw7w+BsS1TieDDqIcCjElA8DINAOUzVg0tOAQpCBBQxpWHQrCcLw+ZfWmGQSLIpiLAYq06Pk2SfAFSEQhkRQAFZlGYoA)

mcintyre94: Nice! Seems to work with empty vs non-empty array too which is cool: [playground](https://www.typescriptlang.org/play?#code/PTAEBkFMHMEMGMCeAoALogDpUA1SAnAZwEsB7AO0gBMAVfWcwhVM80AXlAG9lQ-RijVPgCu8FhUIAuUIWGDoAbQC6AGl78AZpEgAFWIgIy5+Bev6gARgBtS8ANYALWIUfH55aOf4lo5WKgi+JDSoOQiALaWBCrIAL7IyCCgAKIAHlimEZDkqLDWaJjYdAxM4qwc3Bp8giZiEozupp6xCehYoACSJYzMrADqxKiOAGI6+ob4lTwW2noGRrIeXvGFHd30veUUg8MAQrYOzq7gxNos2dPVVodOLm5LzSttRV09ZQ27jgDKxH4BQRCVwsvn8gWCoRMChU6gSa2wADkKCkIhh0ABBfD0RAAHhoAD5KooaKpQAA6Ck0FTKRLtbAAeWstE2H1YAFVGH9KFRKngiKxqO8+hQANzw0CM5mlYXkX5+ai8ggkCiClky0AAMm4si5AIhMiR5BRaMQmOxOPCUQIhLiYvFCMgAHchdtyBzQQrOC6Gpq3mrXV8xvNJr6NtKA0NHAc7HcTmdIBdIPand7WHLuZVUxRQ1nyIHxgsplqw1tPpHo0d7qdzsRLsXc1903qQolNCJyK7QJpyHjJbnQJA0qgclRCBKmbn3VyFQAfcdS0tp6dUfEAClITJkND7-oaAEotzvwz7B8PyKP57n0wqAPygB3O3dL+U8g0pp8UKcvqoWYLgtjhNY1igC4oDtvY5CkI6bCgau24Th+bCniOY5HouFDXjyd4Pley6gG+j7HuynIvnuqzIPAkioGEpDXkQlTdquPDJBYAB6N7XLUwj1KwoSKAA5MQaT8Wo1xzBMiz8QAVqQkD8d4fA2DGxwPPxSkOPJ1ygs2fE0nEZEUVRoCOpGdFjpwjHMWAbEcRYXGiK6fGCcJomzAWkwyNJsmaRY6mxqpfk+T4urgiEMiKAArHpe5AA)

I don't think we'll be able to get to differentiating how it actually is on the old transactions though, you need to differentiate all signatures are all-zero arrays vs not. The old signature stuff seems designed to stop us doing this lol

buffalojoec: ohhh yeah I see what you mean. So if you provide "with signatures" and it's just one single all-zero signature, you'll get back "with signatures" but it will be empty

buffalojoec: Is it all-zero automatically? In other words, this whole thing isn't ever going to resolve to "unsigned"?

mcintyre94: I don't know for sure but in my testing all the ways you're meant to add a signature or a signer add an empty signature to the list of signatures. And AFAICT using that index is the only way to actually match an account to a signature, so it's probably a bug if there are any cases where they become disconnected. It's not a type-level guarantee though and you can probably mess it up pretty easily if you move outside of the class methods.

buffalojoec: Since signatures are fixed length, we could maybe use something like this in our conditional type-handling:

```typescript
type AllZeroSignature = [0, 0, 0, 0];
type SomeSignature = [number, number, number, number];

type ValidSignature = SomeSignature extends AllZeroSignature ? never : SomeSignature;
```

buffalojoec: but this might not work within another array :(

mcintyre94: Ah wait I think this is why it wasn't working for me: [playground](https://www.typescriptlang.org/play?#code/PTAEBkFMHMEMGMCeAoALogDpUA1SAnAZwEsB7AO0gBMAVfWcwhVM80AXlAG9lQ-RijVPgCu8FhUIAuUIWGDoAbQC6AGl78AZpEgAFWIgIy5+Bev6gARgBtS8ANYALWIUfH55aOf4lo5WKgi+JCEAPwy5CIAtpYEKsgAvsjIIKAAogAeWKZRkOSosNZomNh0DEzirBzcGnyCJmISjO6mnvFJ6FigAJJljMysAOrEqI4AYjr6hvjVPBbaegZGsh5eicVdvfT9lRTDowBCtg7OruDE2iy5s7VWx04ubiutax0lPX0VTfuOAMrEfgCQRCNwsvn8gWC0meChU6iSG2wAHlrLRtl9WABVRgAyhUap4IisaifAYUADciNAKLR5TJ5H+fmoBIIJAoJPR9NAADJuLJcUCoRForF8CpQAlKVSAHKQADupN25Gx4OZnEVTR5H05Sp+E0W0y1WzpupGjiOdgeZwukCukBl8o1rEZeOqTooRvd5D1kyWM15xp23zNFpOj3Ol2I1wDXp+LsFIWSmhE5CVoE05AAPDQaV7QJAMqg8lRCNTUV6VbjmQAfMu0oPOqtUAB8AApSKiZDnyzqmgBKLu53tVAtF8glutel3M0KgWUK4cUaf4mTzis4pn4ub8YKQtiRazWUAuUAp+zkUhytgn1vd+sYj2j4ulocmprL0CzteLhlN0Crx0f0rTc+3WZB4EkVBQAvaciBkQk2TxPNOG3OohFEJVoUUAByYgMmwtRbgWKZlmwgArUhIGw7w+BsS1TiebC6IcaiwIgoRoNIWDCAAYQoAA3Agi3xTgM1bGCqyIPtKVSfgAD1QmSdi5FAOUzW4+DWWJe8uRQ256mERpWCw3D8MI+ZfWmGRyMo1iLGYq1GIcuyfAFSEQhkRQAFZlDYyDVPUyTeIEoS1XTchWzU0ZuOklIwHk0IgA)

AFAICT as soon as you add the indirection of 'hey typescript this is a VersionedTransaction from somewhere else' and not 'hey typescript here is a VersionedTransaction being explicitly created right here so you can see it' it loses the ability to do anything that involves inspecting the transaction. I don't know if I'm missing something that unlocks this though because even if it doesn't work here because signatures are weird that would be awesome!

buffalojoec: Ahhh damn, it looks like just casting it to a `VersionedTransaction` from anywhere breaks it. Smh

mcintyre94:
* master
  * **PR #1654** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1654" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1654?utm_source=stack-comment).

buffalojoec:
Lots of convo on this PR lol, what's the status @mcintyre94 ? Are you waiting on another review?

mcintyre94:
@buffalojoec Yep should be good to go!

This comment still covers most of the status since last review: https://github.com/solana-labs/solana-web3.js/pull/1654#issuecomment-1746900075

Plus added the optional lastValidBlockHeight param w/ default to u64 max

buffalojoec (approved):
Lgtm. The type checking thing is a bummer, but it sounds like you & steve have plans for making this still not suck for converting and checking signatures

mcintyre94:
CI keeps failing on `@solana/web3.js#test:live-with-test-validator` on this branch. It passes locally on this branch, and nothing in this PR should affect the legacy library tests. I'm going to merge this, will keep an eye on CI after though!

PR #1660, opened 2023-10-02
feat(experimental): create an RPC GraphQL

buffalojoec:
This PR lays down the initial skeleton initializer for an RPC GraphQL. This will be the primary way Solana developers will instantiate an instance of the Solana RPC GraphQL, including the Solana GraphQL schema.

PR #1659, opened 2023-10-02
feat(experimental): init rpc-graphql package

buffalojoec:
This PR introduces the RPC GraphQL library to the monorepo. No code yet, just docs and configs.

PR #1661, opened 2023-10-02
feat(experimental): a cache for RPC GraphQL

buffalojoec:
This PR introduces a simple cache, usable by both Node JS and browser applications, for caching RPC requests within GraphQL queries.

Note: the cache is flushed after each query, since we don't want to hold cached state that could otherwise change between queries. However, within the exacution of a single query, no matter how large, cache will be used in favor of re-requesting the same resource from the RPC.


buffalojoec:
* master
  * **PR #1659** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1659" target="_blank">Graphite</a> 
    * **PR #1660** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1660" target="_blank">Graphite</a> 
      * **PR #1661** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1661" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1661?utm_source=stack-comment).

packages/rpc-graphql/src/cache.ts
```diff
@@ -0,0 +1,63 @@
...
+
+// Basic in-memory cache for Node.js
+const inMemoryCache: { [key: string]: string } = {};
+
+const stringifyValue = (value: unknown) =>
+    JSON.stringify(value, (_, value) => {
+        if (typeof value === 'bigint') {
+            return value.toString() + 'n';
+        }
+        return value;
```
mcintyre94, line 15: Nitpick: shadowing `value` here (if I'm understanding it correctly) is quite confusing! Same on `parseValue`

mcintyre94 (approved):

packages/rpc-graphql/src/cache.ts
buffalojoec: Yep totally fair

buffalojoec: Although technically that _is_ what `JSON.stringify(..)` does, too! I can change it though.

PR #1663, opened 2023-10-02
refactor(experimental): picks for creating GraphQL JS types

buffalojoec:
These picks are designed to reduce some of the verbosity involved with initializing GraphQL JS objects and defining their fields and types. The result is GraphQL JS object definitions that look more like TypeScript type definitions.

PR #1664, opened 2023-10-02
refactor(experimental): an account resolver

buffalojoec:
A resolver to hit the Solana RPC for an `AccountInfo` from a GraphQL query.

### DIFF
--- a/packages/rpc-graphql/src/context.ts
+++ b/packages/rpc-graphql/src/context.ts
@@ -2,13 +2,62 @@ import { SolanaRpcMethods } from '@solana/rpc-core';
 import { Rpc } from '@solana/rpc-transport/dist/types/json-rpc-types';
 
 import { createGraphQLCache, GraphQLCache } from './cache';
+import { AccountQueryArgs } from './schema/account/query';
 
 export interface RpcGraphQLContext {
     cache: GraphQLCache;
+    resolveAccount(args: AccountQueryArgs): ReturnType<typeof resolveAccount>;
     rpc: Rpc<SolanaRpcMethods>;
+}
+
+// Default to jsonParsed encoding if none is provided
+async function resolveAccount(
+    { address, encoding = 'jsonParsed', ...config }: AccountQueryArgs,
+    cache: GraphQLCache,
+    rpc: Rpc<SolanaRpcMethods>
+) {
+    const requestConfig = { encoding, ...config };
+
+    const cached = cache.get(address, requestConfig);
+    if (cached !== null) {
+        return cached;
+    }
+
+    const account = await rpc
+        .getAccountInfo(address, requestConfig as Parameters<SolanaRpcMethods['getAccountInfo']>[1])
+        .send()
+        .then(res => res.value)
+        .catch(e => {
+            throw e;
+        });
+
+    if (account === null) {
+        return null;
+    }
+
+    const [data, responseEncoding] = Array.isArray(account.data)
+        ? encoding === 'jsonParsed'
+            ? [account.data[0], 'base64']
+            : [account.data[0], encoding]
+        : [account.data, 'jsonParsed'];
+    const queryResponse = {
+        ...account,
+        data,
+        encoding: responseEncoding,
+    };
+
+    cache.insert(address, requestConfig, queryResponse);
+
+    return queryResponse;
 }
 
 export function createSolanaGraphQLContext(rpc: Rpc<SolanaRpcMethods>): RpcGraphQLContext {
     const cache = createGraphQLCache();
-    return { cache, rpc };
+    return {
+        cache,
+        resolveAccount(args) {
+            return resolveAccount(args, this.cache, this.rpc);
+        },
+        rpc,
+    };
 }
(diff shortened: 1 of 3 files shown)


buffalojoec:
* master
  * **PR #1659** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1659" target="_blank">Graphite</a> 
    * **PR #1660** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1660" target="_blank">Graphite</a> 
      * **PR #1661** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1661" target="_blank">Graphite</a> 
        * **PR #1662** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1662" target="_blank">Graphite</a> 
          * **PR #1663** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1663" target="_blank">Graphite</a> 
            * **PR #1664** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1664" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1664?utm_source=stack-comment).

packages/rpc-graphql/src/context.ts
```diff
@@ -2,13 +2,59 @@ import { SolanaRpcMethods } from '@solana/rpc-core';
...
 import { createGraphQLCache, GraphQLCache } from './cache';
+import { AccountQueryArgs } from './schema/account/query';
 
 export interface RpcGraphQLContext {
     cache: GraphQLCache;
+    resolveAccount(args: AccountQueryArgs): ReturnType<typeof resolveAccount>;
     rpc: Rpc<SolanaRpcMethods>;
 }
 
+// Default to jsonParsed encoding if none is provided
```
mcintyre94, line 13: In general do you expect this layer to be more opinionated than `rpc-core`/the RPC itself, and to have defaults like this in quite a lot of places?

packages/rpc-graphql/src/context.ts
```diff
@@ -2,13 +2,59 @@ import { SolanaRpcMethods } from '@solana/rpc-core';
...
+
+    const cached = cache.get(address, requestConfig);
+    if (cached !== null) {
+        return cached;
+    }
+
+    const account = await rpc
+        .getAccountInfo(address, requestConfig as Parameters<SolanaRpcMethods['getAccountInfo']>[1])
+        .send()
+        .then(res => res.value);
```
mcintyre94, line 29: Does `graphql` do something sensible if this promise throws?

packages/rpc-graphql/src/schema/account/query.ts
```diff
@@ -0,0 +1,13 @@
...
+export type AccountQueryArgs = {
+    address: Base58EncodedAddress;
+    commitment?: Commitment;
+    dataSlice?: {
+        offset: number;
+        length: number;
+    };
+    encoding?: 'base58' | 'base64' | 'base64+zstd' | 'jsonParsed';
+    minContextSlot?: bigint;
+};
```
mcintyre94, line 13: It's annoying but not surprising that we can't just re-use `rpc-core` types for this stuff

How do you think we should deal with invalid combinations of args though? Eg encoding=base64 and dataSlice is defined. `rpc-core` shouldn't let you call `getAccountInfo` with that, but this would let you call the graphql method with it.

mcintyre94 (approved):

packages/rpc-graphql/src/context.ts
buffalojoec: In general, yes, but to what extent I'm not sure yet.

For starters, the API favors parsed data unless you specifically tell it not to parse (encoding). My thought here was basically: if you're using GraphQL, you are much more likely to be querying parsed data than encoded data, so why not make the parsed ones the default, and if you don't want to parse, you specifically say that.

I'd love your thoughts on this though.

packages/rpc-graphql/src/schema/account/query.ts
buffalojoec: As per your comment about the error handling from the RPC call, I'll add a change to bubble those throws up, so if `getAccountInfo` throws on an invalid combination, you'll get that from GraphQL. 

However, if we want the type-checking to enforce this, unfortunately we'll have to re-write that bit.

I'm exploring re-using more of the types from `rpc-core` now.

packages/rpc-graphql/src/context.ts
buffalojoec: I'm not really sure, but I can explicitly bubble the error up

packages/rpc-graphql/src/schema/account/query.ts
buffalojoec: I managed to import a few more types from `@solana/rpc-core`, but the annoying part is we have to re-create them as GraphQL objects 😡 😡

buffalojoec:
### Merge Activity

* **Oct 11, 11:19 AM**: @buffalojoec started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1664).

PR #1676, opened 2023-10-04
refactor(experimental): add options package

lorisleiva:
This PR adds a new package that allows us to manage and encode/decode Rust-like `Option` types.

```ts
type Option<T> = Some<T> | None;
type Some<T> = { __option: "Some"; value: T };
type None = { __option: "None" };
```

This is particularly useful when generating clients from data structures defined in Rust. Whilst one could define an `Option<T>` type as `T | null` (currently serializable using the `getNullableCodec` method), this would loose information when using nested options. For instance, `Option<Option<T>>` would resolve to `T | T | null` which would resolve to `T | null` which is equivalent to `Option<T>`. In other words, `Some(None)` and `None` would not be distinguishable.

This new package provides everything we need when dealing with Rust-like `Option` types in JavaScript.

This is imported from the Umi framework as it is required when generating clients using Kinobi. See [Umi's Option documentation](https://github.com/metaplex-foundation/umi/blob/main/docs/helpers.md#options) for a more details description.


lorisleiva:
* master
  * **PR #1676** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1676" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1676?utm_source=stack-comment).

packages/codecs-data-structures/src/nullable.ts
steveluscher, line 1: Were all these changes supposed to be in another PR?

packages/options/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+This package allows us to manage and serialize Rust-like Option types in JavaScript. It can be used standalone, but it is also exported as part of the Solana JavaScript SDK [`@solana/web3.js@experimental`](https://github.com/solana-labs/solana-web3.js/tree/master/packages/library).
+
+## Types
+
+TODO
+
+## Functions
+
+TODO
```
steveluscher, line 25: I'm guilty of this too. We need to catch all the READMEs up, soon.

packages/options/src/option.ts
```diff
@@ -0,0 +1,63 @@
...
+ * @see {@link Option}
+ */
+export const some = <T>(value: T): Option<T> => ({ __option: 'Some', value });
+
+/**
+ * Creates a new {@link Option} of type `T` that has no value.
+ *
+ * @see {@link Option}
+ */
+export const none = <T>(): Option<T> => ({ __option: 'None' });
```
steveluscher, line 42: How would you feel about literally returning the same (frozen) object?
```suggestion
const NONE = Object.freeze({ __option: 'None' });
export const none = <T>(): Option<T> => NONE;
```

…then adding tests that `expect(none()).toBe(none())`?

packages/options/src/option.ts
```diff
@@ -0,0 +1,63 @@
...
+ * @see {@link Option}
+ */
+export type Some<T> = { __option: 'Some'; value: T };
+
+/**
+ * Represents an option of type `T` that has no value.
+ *
+ * @see {@link Option}
+ */
+export type None = { __option: 'None' };
```
steveluscher, line 28: These _really, really_ need to be `Readonly`.
```suggestion
/**
 * Represents an option of type `T` that has a value.
 *
 * @see {@link Option}
 */
export type Some<T> = Readonly<{ __option: 'Some'; value: T }>;

/**
 * Represents an option of type `T` that has no value.
 *
 * @see {@link Option}
 */
export type None = Readonly<{ __option: 'None' }>;
```

packages/options/src/option.ts
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Whether the given {@link Option} is a {@link Some}.
+ */
+export const isSome = <T>(option: Option<T>): option is Some<T> => option.__option === 'Some';
+
+/**
+ * Whether the given {@link Option} is a {@link None}.
+ */
+export const isNone = <T>(option: Option<T>): option is None => option.__option === 'None';
```
steveluscher, line 63: Then…
```suggestion
export const isNone = <T>(option: Option<T>): option is None => option === NONE;
```

packages/options/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+    "$schema": "https://json.schemastore.org/tsconfig",
+    "display": "@solana/options",
+    "extends": "tsconfig/base.json",
+    "include": ["src"],
+    "compilerOptions": {
+        "lib": ["DOM", "ES2020", "ES2022.Error"]
```
steveluscher, line 7: Copypasta? Eliminate items from this array until typechecking fails, and then put back only that which is needed to make it pass.

steveluscher (approved):
giphy-7

packages/options/src/option.ts
lorisleiva: I'm happy to make that change but couldn't this create a dual-package hazard if an application ends up with multiple instances of this library (either because it's dependency chains ends up requiring both ESM and CJS or slightly different versions)? I've been trying to avoid adding top-level state for that purpose. Is this a valid concern here?

packages/options/README.md
lorisleiva: Yeah don't worry I'll make sure I tackle all these codec READMEs soon.

packages/options/src/option.ts
lorisleiva: I've applied this suggestion but definitely think it's worth talking about it. 🙂

Issue #1621, opened 2023-09-21
[experimental] Implement `importKey()` in `@solana/webcrypto-ed25519-polyfill`


steveluscher:
Implement a polyfill for `crypto.subtle.importKey()` that basically shoves the private key bytes into `storageKeyBySecretKey_INTERNAL_ONLY_DO_NOT_EXPORT` in the exact same way that you would if you _generated_ a key using `generateKey()`.

Pattern the PR off a PR like #1397.

PR #1681, opened 2023-10-05
Add Ed25519 importKey polyfill

ronanyeah:
#1621

### DIFF
--- a/packages/webcrypto-ed25519-polyfill/src/__tests__/secrets-test.ts
+++ b/packages/webcrypto-ed25519-polyfill/src/__tests__/secrets-test.ts
@@ -21,6 +21,39 @@ const MOCK_PUBLIC_KEY_BYTES = new Uint8Array([
     166, 132, 114, 186, 49, 163, 23, 12, 11, 14, 119, 219, 102, 96, 26, 226, 91, 97, 238, 217, 236, 84, 232, 204, 62,
     212, 179, 252, 20, 37, 179, 52,
 ]);
+const MOCK_PKCS8_HEADER =
+    // prettier-ignore
+    new Uint8Array([
+        /**
+         * PKCS#8 header
+         */
+        0x30, // ASN.1 sequence tag
+        0x2e, // Length of sequence (46 more bytes)
+
+            0x02, // ASN.1 integer tag
+            0x01, // Length of integer
+                0x00, // Version number
+
+            0x30, // ASN.1 sequence tag
+            0x05, // Length of sequence
+                0x06, // ASN.1 object identifier tag
+                0x03, // Length of object identifier
+                    // Edwards curve algorithms identifier https://oid-rep.orange-labs.fr/get/1.3.101.112
+                        0x2b, // iso(1) / identified-organization(3) (The first node is multiplied by the decimal 40 and the result is added to the value of the second node)
+                        0x65, // thawte(101)
+                    // Ed25519 identifier
+                        0x70, // id-Ed25519(112)
+
+        /**
+         * Private key payload
+         */
+        0x04, // ASN.1 octet string tag
+        0x22, // String length (34 more bytes)
+
+            // Private key bytes as octet string
+            0x04, // ASN.1 octet string tag
+            0x20  // String length (32 bytes)
+    ]);
 
 describe('exportKeyPolyfill', () => {
     it.each(['jwk', 'pkcs8', 'spki'] as const)('throws an unimplemented error when the format is %s', format => {
@@ -52,10 +85,7 @@ describe('exportKeyPolyfill', () => {
 });
 
 describe('importKeyPolyfill', () => {
-    const PKCS8_HEADER = new Uint8Array([
-        0x30, 0x2e, 0x02, 0x01, 0x00, 0x30, 0x05, 0x06, 0x03, 0x2b, 0x65, 0x70, 0x04, 0x22, 0x04, 0x20,
-    ]);
-    const pkcs8PrivateKey = new Uint8Array([...PKCS8_HEADER, ...MOCK_SECRET_KEY_BYTES]);
+    const pkcs8PrivateKey = new Uint8Array([...MOCK_PKCS8_HEADER, ...MOCK_SECRET_KEY_BYTES]);
     it('stores secret key bytes in an internal cache', () => {
         const weakMapSetSpy = jest.spyOn(WeakMap.prototype, 'set');
         importKeyPolyfill('pkcs8', pkcs8PrivateKey, false, ['sign']);
@@ -69,6 +99,27 @@ describe('importKeyPolyfill', () => {
         const publicKey = importKeyPolyfill('raw', MOCK_PUBLIC_KEY_BYTES, false, ['verify']);
 
         expect(verifyPolyfill(publicKey, signature, MOCK_DATA)).toBe(true);
+    });
+    it.each([true, false])('has correct extractability', extractability => {
+        const privateKey = importKeyPolyfill('pkcs8', pkcs8PrivateKey, extractability, ['sign']);
+        expect(privateKey.extractable).toBe(extractability);
+
+        const publicKey = importKeyPolyfill('raw', MOCK_PUBLIC_KEY_BYTES, extractability, ['verify']);
+        expect(publicKey.extractable).toBe(true);
+    });
+    describe('correct properties are present', () => {
+        const privateKey = importKeyPolyfill('pkcs8', pkcs8PrivateKey, false, ['sign']);
+        const publicKey = importKeyPolyfill('raw', MOCK_PUBLIC_KEY_BYTES, true, ['verify']);
+        it.each([publicKey, privateKey])(`has the algorithm "Ed25519"`, key => {
+            expect(key).toHaveProperty(['algorithm', 'name'], 'Ed25519');
+        });
+        it.each([publicKey, privateKey])('has the string tag "CryptoKey"', key => {
+            expect(key).toHaveProperty([Symbol.toStringTag], 'CryptoKey');
+        });
+        it(`has the correct type`, () => {
+            expect(publicKey).toHaveProperty(['type'], 'public');
+            expect(privateKey).toHaveProperty(['type'], 'private');
+        });
     });
 });
 
--- a/packages/webcrypto-ed25519-polyfill/src/__tests__/secrets-test.ts
+++ b/packages/webcrypto-ed25519-polyfill/src/__tests__/secrets-test.ts
@@ -1,4 +1,11 @@
-import { exportKeyPolyfill, generateKeyPolyfill, isPolyfilledKey, signPolyfill, verifyPolyfill } from '../secrets';
+import {
+    exportKeyPolyfill,
+    importKeyPolyfill,
+    generateKeyPolyfill,
+    isPolyfilledKey,
+    signPolyfill,
+    verifyPolyfill,
+} from '../secrets';
 
 const MOCK_DATA = new Uint8Array([1, 2, 3]);
 const MOCK_DATA_SIGNATURE = new Uint8Array([
@@ -41,6 +48,27 @@ describe('exportKeyPolyfill', () => {
         jest.spyOn(globalThis.crypto, 'getRandomValues').mockReturnValue(MOCK_SECRET_KEY_BYTES);
         const { publicKey } = generateKeyPolyfill(/* extractable */ false, ['sign', 'verify']);
         expect(exportKeyPolyfill('raw', publicKey)).toEqual(MOCK_PUBLIC_KEY_BYTES);
+    });
+});
+
+describe('importKeyPolyfill', () => {
+    const PKCS8_HEADER = new Uint8Array([
+        0x30, 0x2e, 0x02, 0x01, 0x00, 0x30, 0x05, 0x06, 0x03, 0x2b, 0x65, 0x70, 0x04, 0x22, 0x04, 0x20,
+    ]);
+    const pkcs8PrivateKey = new Uint8Array([...PKCS8_HEADER, ...MOCK_SECRET_KEY_BYTES]);
+    it('stores secret key bytes in an internal cache', () => {
+        const weakMapSetSpy = jest.spyOn(WeakMap.prototype, 'set');
+        importKeyPolyfill('pkcs8', pkcs8PrivateKey, false, ['sign']);
+        expect(weakMapSetSpy).toHaveBeenCalledWith(expect.anything(), MOCK_SECRET_KEY_BYTES);
+    });
+    it('signs and verifies bytes', () => {
+        const privateKey = importKeyPolyfill('pkcs8', pkcs8PrivateKey, false, ['sign']);
+
+        const signature = signPolyfill(privateKey, MOCK_DATA);
+
+        const publicKey = importKeyPolyfill('raw', MOCK_PUBLIC_KEY_BYTES, false, ['verify']);
+
+        expect(verifyPolyfill(publicKey, signature, MOCK_DATA)).toBe(true);
     });
 });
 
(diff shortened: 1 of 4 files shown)


packages/webcrypto-ed25519-polyfill/src/secrets.ts
```diff
@@ -138,3 +138,25 @@ export function verifyPolyfill(key: CryptoKey, signature: BufferSource, data: Bu
...
+
+export function importKeyPolyfill(
+  format: KeyFormat,
+  keyData: BufferSource,
+  extractable: boolean,
+  keyUsages: readonly KeyUsage[]
+): CryptoKeyPair {
+  const bytes = bufferSourceToUint8Array(keyData);
+
+  if (format !== "raw") {
```
steveluscher, line 150: If my memory serves me correctly, I _think_ that `raw` _is_ permitted for public keys (ie. verify usage) but not for private keys (ie. sign usage) which must come in PKCS#8 format.

packages/webcrypto-ed25519-polyfill/src/secrets.ts
```diff
@@ -138,3 +138,25 @@ export function verifyPolyfill(key: CryptoKey, signature: BufferSource, data: Bu
...
+    throw new DOMException(
+      `Importing Ed25519 keys in the "${format}" format is unimplemented`,
+      "NotSupportedError"
+    );
+  }
+
+  if (!keyUsages.includes("sign") || !keyUsages.includes("verify")) {
```
steveluscher, line 157: Isn’t it also the case that supplying _both_ sign and verify is a runtime error?

packages/webcrypto-ed25519-polyfill/src/secrets.ts
```diff
@@ -138,3 +138,25 @@ export function verifyPolyfill(key: CryptoKey, signature: BufferSource, data: Bu
         return false;
     }
 }
+
+export function importKeyPolyfill(
+  format: KeyFormat,
+  keyData: BufferSource,
+  extractable: boolean,
+  keyUsages: readonly KeyUsage[]
+): CryptoKeyPair {
```
steveluscher, line 147: I don’t think that `importKey` returns key pairs. Only `generateKey` does that.

steveluscher (changes requested):
Check out the PKCS#8 import technique in the tests. That will tell you a lot about what has to happen here!

Also, this is a super helpful resource to get all the input and output types right. https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/importKey

ronanyeah:
I removed checks for now, and I am able to submit a successful transaction from a browser using an imported pkcs8 key, which has been built from a Solana private key.
```ts
  const solanaKeypair = new Uint8Array([...]);
  const privateKey = solanaKeypair.slice(0, 32);
  const publicKey = solanaKeypair.slice(32);

  const pkc = await crypto.subtle.importKey(
    "pkcs8",
    privateKey,
    "Ed25519",
    false,
    ["sign"]
  );

  const bh = await rpc.getLatestBlockhash().send();
  const message = getTransactionMessage({
    blockhash: bh.value.blockhash,
    feePayerAddressBytes: publicKey,
    memoString: `Hello from the RONAN tests! [${performance.now()}]`,
  });

  const signature = new Uint8Array(
    await crypto.subtle.sign("Ed25519", pkc, message)
  );
```

packages/webcrypto-ed25519-polyfill/src/index.ts
```diff
@@ -116,4 +116,30 @@ if (!__BROWSER__ || globalThis.isSecureContext) {
...
+    originalSubtleCrypto.importKey = (async (
+      ...args: Parameters<SubtleCrypto["importKey"]>
+    ) => {
+      const [format, keyData, algorithm, extractable, keyUsages] = args;
+      if (algorithm === "Ed25519") {
+        return await importKeyPolyfill(
+          format,
+          keyData,
+          extractable,
+          keyUsages as any
```
buffalojoec, line 135: you might be able to sneak around this `any` check by following what some of the other methods are doing, ie:

https://github.com/solana-labs/solana-web3.js/blob/eb5fd160f78c17f360d679861f487b42c693bd69/packages/webcrypto-ed25519-polyfill/src/index.ts#L95

buffalojoec (commented):
This is really on the right track, thanks for submitting this!

Would you mind writing some tests for this? You could do something simple like using the imported private key to sign a transaction, then the imported public key to validate the signature.

In your code block in the description, it doesn't look like you're adding the `pkcs8` header to the `ed25519` private key before importing.

Here's an area where we've used `importKey()` in the codebase that might be of help. It shows the `pkcs8` bytes and then the 32 private key bytes.
https://github.com/solana-labs/solana-web3.js/blob/master/packages/keys/src/__tests__/signatures-test.ts

ronanyeah:
Added some tests and hardcoded handling of the pkcs8 header. Had to make some changes to the verify polyfill, which are probably need more work.

packages/webcrypto-ed25519-polyfill/src/secrets.ts
```diff
@@ -131,10 +131,54 @@ export function verifyPolyfill(key: CryptoKey, signature: BufferSource, data: Bu
...
+
+export function importKeyPolyfill(
+    format: KeyFormat,
+    keyData: BufferSource,
+    extractable: boolean,
+    keyUsages: readonly KeyUsage[]
+): CryptoKey {
+    const type =
+        // could be more rigorous
+        format === 'pkcs8' ? 'private' : 'public';
```
mcintyre94, line 156: I think a better way to structure this would be to branch the code on `format`, something like this:

```ts
if (format === 'raw') {
  // you might need some checks here that you're not getting from `createKeyPairFromBytes`
  // code to import + return the public key
} else if (format === 'pkcs8') {
  // code to import + return the private key
} else {
  // TODO: coded error
  throw new Error(`cannot import key with format ${format} format`)
}
```

packages/webcrypto-ed25519-polyfill/src/secrets.ts
```diff
@@ -131,10 +131,54 @@ export function verifyPolyfill(key: CryptoKey, signature: BufferSource, data: Bu
...
+
+    const bytes = bufferSourceToUint8Array(keyData);
+
+    const keyBytes =
+        format === 'pkcs8'
+            ? // Remove header
+              bytes.slice(16)
```
mcintyre94, line 163: I think a more robust way to do this without parsing the header would be to extract the last 32 bytes
We know that this is an ED25519 private key with a potentially variable length header, so we know the length of the key after the header
This assumes that the input is valid, while properly parsing the header would let us catch and report potential issues with it, but for now I think this would be sufficient

packages/webcrypto-ed25519-polyfill/src/__tests__/secrets-test.ts
```diff
@@ -44,6 +51,27 @@ describe('exportKeyPolyfill', () => {
     });
 });
 
+describe('importKeyPolyfill', () => {
+    const PKCS8_HEADER = new Uint8Array([
+        0x30, 0x2e, 0x02, 0x01, 0x00, 0x30, 0x05, 0x06, 0x03, 0x2b, 0x65, 0x70, 0x04, 0x22, 0x04, 0x20,
+    ]);
```
mcintyre94, line 57: Could you format and comment this the same as https://github.com/solana-labs/solana-web3.js/blob/eb5fd160f78c17f360d679861f487b42c693bd69/packages/keys/src/__tests__/signatures-test.ts#L11-L45 please? It'll make this test much easier to maintain. You can probably just copy-paste from there exactly if you're not already using the same header.

packages/webcrypto-ed25519-polyfill/src/__tests__/secrets-test.ts
```diff
@@ -44,6 +51,27 @@ describe('exportKeyPolyfill', () => {
...
+    const pkcs8PrivateKey = new Uint8Array([...PKCS8_HEADER, ...MOCK_SECRET_KEY_BYTES]);
+    it('stores secret key bytes in an internal cache', () => {
+        const weakMapSetSpy = jest.spyOn(WeakMap.prototype, 'set');
+        importKeyPolyfill('pkcs8', pkcs8PrivateKey, false, ['sign']);
+        expect(weakMapSetSpy).toHaveBeenCalledWith(expect.anything(), MOCK_SECRET_KEY_BYTES);
+    });
```
mcintyre94, line 63: This test is great!

I'd suggest adding a simple one that verifies the return too, for each of private and public. You can see the assertions we're using for the `generatePolyfill` tests: https://github.com/solana-labs/solana-web3.js/blob/eb5fd160f78c17f360d679861f487b42c693bd69/packages/webcrypto-ed25519-polyfill/src/__tests__/secrets-test.ts#L60-L69 - I think these all apply here too 

We should also check `extractable` which I think should be false for a private key and true for a public key: https://github.com/solana-labs/solana-web3.js/blob/eb5fd160f78c17f360d679861f487b42c693bd69/packages/webcrypto-ed25519-polyfill/src/__tests__/secrets-test.ts#L70-L76

I also think we should have these tests from `generatePolyfill`:

- empty keyUsages is an error
- prohibited keyUsages are an error

See https://github.com/solana-labs/solana-web3.js/blob/21e994fd3b1403f9843b600276dcb7a510a8b683/packages/webcrypto-ed25519-polyfill/src/secrets.ts#L48-L53

packages/webcrypto-ed25519-polyfill/src/secrets.ts
```diff
@@ -131,10 +131,54 @@ export function verifyPolyfill(key: CryptoKey, signature: BufferSource, data: Bu
...
+            : bytes;
+
+    if (type === 'public') {
+        const publicKey = {
+            [Symbol.toStringTag]: 'CryptoKey',
+            algorithm: Object.freeze({ name: 'Ed25519' }),
+            extractable,
```
mcintyre94, line 170: I think that like in `generatePolyfill` a public key should always be extractable. Probably just ignore the user's input here to match the behaviour there

packages/webcrypto-ed25519-polyfill/src/secrets.ts
```diff
@@ -131,10 +131,54 @@ export function verifyPolyfill(key: CryptoKey, signature: BufferSource, data: Bu
...
+            type: 'public' as KeyType,
+            usages: [...keyUsages],
```
mcintyre94, line 172: Can this just be `usages: keyUsages`?

mcintyre94 (commented):
This is looking great, thanks for your contribution! I think we do need some more checks/tests here to match the behaviour we have elsewhere.

mcintyre94 (commented):
This is looking great, thanks for your contribution! I think we do need some more checks/tests here to match the behaviour we have elsewhere.

packages/webcrypto-ed25519-polyfill/src/__tests__/secrets-test.ts
ronanyeah: The test `sets the private key's 'extractable' to 'false'` looks to be incorrect. From the test description extractable for private keys should always be false. I have made the necessary changes in the test and secrets file.

mcintyre94: You're right that there's a mismatch between the test description and the code behaviour there. I'm not sure which is correct, so let's not change that code yet. I'll open an issue to make sure this gets cleared up, but for now please keep and re-use the existing behaviour.

mcintyre94 (changes requested):

mcintyre94 (approved):
This is looking really good, thanks for all your work on this!

The only thing I think would be worth adding still is tests for the `index.ts` logic to make sure it's calling the polyfill/original implementation in the correct circumstances. We have tests like this for the `generateKey` with similar logic

I'll approve because I don't think those tests are a blocker. If you'd like to add them that'd be awesome, otherwise could you just add a TODO and we'll loop back to them later.

Thanks again! :)

ronanyeah:
I just added a placeholder for that. Thanks for your review!

packages/webcrypto-ed25519-polyfill/src/secrets.ts
```diff
@@ -131,10 +131,53 @@ export function verifyPolyfill(key: CryptoKey, signature: BufferSource, data: Bu
...
+                type: 'public' as KeyType,
+                usages: Object.freeze(keyUsages.filter(usage => usage === 'verify')) as KeyUsage[],
+            };
+
+            const cache = (storageKeyBySecretKey_INTERNAL_ONLY_DO_NOT_EXPORT ||= new WeakMap());
+            cache.set(publicKey, bytes);
+
+            return publicKey;
+        }
+        case 'pkcs8': {
```
steveluscher, line 172: You have to also validate that the usages includes _only_ `'sign'`. You can't have PKCS#8 public keys. Trying to import a pkcs-encoded key with the `'verify'` usage should throw `Unsupported key usage for a Ed25519 key`.

packages/webcrypto-ed25519-polyfill/src/secrets.ts
```diff
@@ -131,10 +131,53 @@ export function verifyPolyfill(key: CryptoKey, signature: BufferSource, data: Bu
...
+        throw new DOMException('Usages cannot be empty when creating a key.', 'SyntaxError');
+    }
+    if (keyUsages.some(usage => PROHIBITED_KEY_USAGES.has(usage))) {
+        throw new DOMException('Unsupported key usage for an Ed25519 key.', 'SyntaxError');
+    }
+
+    const bytes = bufferSourceToUint8Array(keyData);
+
+    switch (format) {
+        case 'raw': {
```
steveluscher, line 158: You have to validate that the usages includes _only_ `'verify'`. Anything else should throw `Unsupported key usage for a Ed25519 key`

steveluscher (changes requested):
[closetreadmillcat](https://app.graphite.dev/macros/solana-labs) 

I added a bunch of pointers regarding how the native implementation behaves. The polyfill needs to mimic those precisely, and there need to be tests for each behaviour.

In general, I find it useful to pop open a Node REPL to exercise the _actual_ `importKey` implementation. Permute the inputs and observe its behaviour. That will tell you how the polyfill should behave!

packages/webcrypto-ed25519-polyfill/src/secrets.ts
steveluscher: It's actually a runtime fatal if you try to import a public key with `sign` or a private key with `verify`.

packages/webcrypto-ed25519-polyfill/src/secrets.ts
steveluscher: Sorry for overruling you in a later comment. I think leaving the PKCS#8 payload unvalidated is wandering into dark territory. If we're going to sell this thing as a polyfill I want it to stick as close to the spec as possible.

packages/webcrypto-ed25519-polyfill/src/secrets.ts
```diff
@@ -131,10 +131,54 @@ export function verifyPolyfill(key: CryptoKey, signature: BufferSource, data: Bu
...
     }
-    const publicKeyBytes = ed25519.getPublicKey(getSecretKeyBytes_INTERNAL_ONLY_DO_NOT_EXPORT(key));
+    const storedBytes = getSecretKeyBytes_INTERNAL_ONLY_DO_NOT_EXPORT(key);
     try {
-        return ed25519.verify(bufferSourceToUint8Array(signature), bufferSourceToUint8Array(data), publicKeyBytes);
+        return (
+            ed25519.verify(
+                bufferSourceToUint8Array(signature),
+                bufferSourceToUint8Array(data),
+                ed25519.getPublicKey(storedBytes)
```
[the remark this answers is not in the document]
steveluscher: Sorry again for the overrule. The idea was that keygen should be super fast, and we should defer this (actually expensive) derivation until the last possible moment (ie. when it's actually needed).

mcintyre94:
I've updated this on a new branch (#1755) so I'll close this PR now. Thanks again @ronanyeah for the contribution! :)

PR #1690, opened 2023-10-06
refactor(experimental): add fromVersionedTransactionWithDurableNonce

mcintyre94:
This PR adds a new `@solana/compat` function to convert a versioned transaction that uses a durable nonce lifetime

This is similar to the blockhash version, with a few differences:

- It does not take `lastValidBlockHeight` since that is only used for blockhash lifetimes
- It errors if the input transaction does not have an advance durable nonce instruction as its first instruction
- It returns an `IDurableNonceTransaction`

Note that the instruction check is done on the modern instruction. I've just used the `isAdvanceNonceAccountInstruction` function that we already have

I opted for a separate function because I don't think there's an easy way to differentiate these types in the legacy library, since it puts the `nonce` value in the `blockhash` field

Callers will likely often already know which sort of transaction they're dealing with, so I'd rather they just tell us by using a different function and then we get clearer typing too

### DIFF
--- a/packages/compat/src/__tests__/transaction-test.ts
+++ b/packages/compat/src/__tests__/transaction-test.ts
@@ -1,12 +1,12 @@
 import { Base58EncodedAddress } from '@solana/addresses';
 import { AccountRole, IInstruction } from '@solana/instructions';
 import { Ed25519Signature } from '@solana/keys';
-import { ITransactionWithSignatures } from '@solana/transactions';
+import { ITransactionWithSignatures, Nonce } from '@solana/transactions';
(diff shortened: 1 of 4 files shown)


mcintyre94:
* cm-tx-compat
  * **PR #1690** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1690" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1690?utm_source=stack-comment).

buffalojoec (approved):
Lgtm, I can't believe we just stuck nonce accounts in the `recentBlockhash` for legacy 😅

PR #1702, opened 2023-10-08
fix: unclosed string literal in code example, async keyword for the function

IgorShadurin:
In the code example within `packages/library/README.md`:
1. Corrected an unclosed string literal that could cause confusion for readers or implementation issues.
2. Added the `async` keyword to the function to align with asynchronous operations.


buffalojoec (approved):
Thanks for doing this!!

PR #1706, opened 2023-10-09
refactor(experimental): add ProgramDerivedAddress type and helpers

lorisleiva:
- Define `ProgramDerivedAddress` type
- Define `ProgramDerivedAddressBump` type
- Add `isProgramDerivedAddress()` helper
- Add `assertIsProgramDerivedAddress()` helper
- Update return value of `getProgramDerivedAddress` function to `Promise<ProgramDerivedAddress>` (Breaking change)
- Rename `computed-address.ts` to `program-derived-address.ts`
- Rename `computed-address-test.ts` to `program-derived-address-test.ts`
- Update tests in `program-derived-address-test.ts`

PR #1705, opened 2023-10-09
refactor(experimental): rename Base58EncodedAddress helpers

lorisleiva:
- Rename `isBase58EncodedAddress()` to `isAddress()`
- Rename `assertIsBase58EncodedAddress()` to `assertIsAddress()`
- Rename `getBase58EncodedAddressCodec()` to `getAddressCodec()`
- Rename `getBase58EncodedAddressComparator()` to `getAddressComparator()`
- Rename `base58.ts` and `base58-test.ts` files to `address.ts` and `address-test.ts`

PR #1710, opened 2023-10-09
refactor(experimental): add signers package

lorisleiva:
This PR adds the `@solana/signers` package **with the first part of its content** that focuses on adding the five signer interfaces with their type guards as well as a concrete implementation of these signers via the `KeyPairSigner` type.

I made sure to document the README immediately so you can have the full picture of what's available in this PR and how to use it.

Below is a copy/paste of that README:

---

You can think of signers as an abstract way to sign messages and transactions. This could be using a Crypto KeyPair, a wallet adapter in the browser, a Noop signer for testing purposes, or anything you want. Here's an example using a `CryptoKeyPair` signer:

```ts
import { pipe } from '@solana/functional';
import { generateKeyPairSigner } from '@solana/signers';
import { createTransaction } from '@solana/transactions';

// Generate a key pair signer.
const mySigner = await generateKeyPairSigner();
mySigner.address; // Address;

// Sign one or multiple messages.
const myMessage = createSignableMessage('Hello world!');
const [messageSignatures] = await mySigner.signMessages([myMessage]);

// Sign one or multiple transactions.
const myTransaction = pipe(
    createTransaction({ version: 0 })
    // Add instructions, fee payer, lifetime, etc.
);
const [transactionSignatures] = await mySigner.signTransactions([myTransaction]);
```

As you can see, this provides a consistent API regardless of how things are being signed behind the scenes. If tomorrow we need to use a browser wallet instead, we'd simply need to swap the `generateKeyPairSigner` function with the signer factory of our choice.

This package offers a total of five different types of signers that may be used in combination when applicable. Three of them allow us to sign transactions whereas the other two are used for regular message signing.

They are separated into three categories:

-   **Partial signers**: Given a message or transaction, provide one or more signatures for it. These signers are not able to modify the given data which allows us to run many of them in parallel.
-   **Modifying signers**: Can choose to modify a message or transaction before signing it with zero or more private keys. Because modifying a message or transaction invalidates any pre-existing signatures over it, modifying signers must do their work before any other signer.
-   **Sending signers**: Given a transaction, signs it and sends it immediately to the blockchain. When applicable, the signer may also decide to modify the provided transaction before signing it. This interface accommodates wallets that simply cannot sign a transaction without sending it at the same time. This category of signers does not apply to regular messages.

Thus, we end up with the following interfaces.

|                     | Partial signers            | Modifying signers            | Sending signers            |
| ------------------- | -------------------------- | ---------------------------- | -------------------------- |
| `TransactionSigner` | `TransactionPartialSigner` | `TransactionModifyingSigner` | `TransactionSendingSigner` |
| `MessageSigner`     | `MessagePartialSigner`     | `MessageModifyingSigner`     | N/A                        |

We will go through each of these five signer interfaces and their respective characteristics in the documentation below.

This package also provides the following concrete signer implementations:

-   The `KeyPairSigner` which uses a `CryptoKeyPair` to sign messages and transactions.
-   The Noop signer which does not sign anything and is mostly useful for testing purposes or for indicating that an account will be signed in a different environment (e.g. sending a transaction to your server so it can sign it).

Additionally, this package allows transaction signers to be stored inside the account meta of an instruction. This allows us to create instructions by passing around signers instead of addresses when applicable which, in turn, allows us to sign an entire transaction automatically without having to scan through its instructions to find the required signers.

In the sections below, we'll go through all the provided signers in more detail before diving into storing signers inside instruction account metas and how to benefit from it.

#### `SignableMessage`

Defines a message with any of the signatures that might have already been provided by other signers. This interface allows modifying signers to decide on whether or not they should modify the provided message depending on whether or not signatures already exist for such message. It also helps create a more consistent API by providing a structure analogous to transactions which also keep track of their signature dictionary.

```ts
type SignableMessage = {
    content: Uint8Array;
    signatures: SignatureDictionary; // Record<Address, SignatureBytes>
};
```

#### `MessagePartialSigner<TAddress>`

An interface that signs an array of `SignableMessages` without modifying their content. It defines a `signMessages` function that returns a `SignatureDictionary` for each provided message. Such signature dictionaries are expected to be merged with the existing ones if any.

```ts
const myMessagePartialSigner: MessagePartialSigner<'1234..5678'> = {
    address: address('1234..5678'),
    signMessages: async (messages: SignableMessage[]): Promise<SignatureDictionary[]> => {
        // My custom signing logic.
    },
};
```

**Characteristics**:

-   **Parallel**. When multiple signers sign the same message, we can perform this operation in parallel to obtain all their signatures.
-   **Flexible order**. The order in which we use these signers for a given message doesn’t matter.

#### `MessageModifyingSigner<TAddress>`

An interface that potentially modifies the content of the provided `SignableMessages` before signing them. E.g. this enables wallets to prefix or suffix nonces to the messages they sign. For each message, instead of returning a `SignatureDirectory`, its `modifyAndSignMessages` function returns its updated `SignableMessage` with a potentially modified content and signature dictionary.

```ts
const myMessageModifyingSigner: MessageModifyingSigner<'1234..5678'> = {
    address: address('1234..5678'),
    modifyAndSignMessages: async (messages: SignableMessage[]): Promise<SignableMessage[]> => {
        // My custom signing logic.
    },
};
```

**Characteristics**:

-   **Sequential**. Contrary to partial signers, these cannot be executed in parallel as each call can modify the content of the message.
-   **First signers**. For a given message, a modifying signer must always be used before a partial signer as the former will likely modify the message and thus impact the outcome of the latter.
-   **Potential conflicts**. If more than one modifying signer is provided, the second signer may invalidate the signature of the first one. However, modifying signers may decide not to modify a message based on the existence of signatures for that message.

#### `MessageSigner<TAddress>`

Union interface that uses any of the available message signers.

```ts
type MessageSigner<TAddress extends string = string> =
    | MessagePartialSigner<TAddress>
    | MessageModifyingSigner<TAddress>;
```

#### `createSignableMessage(content, signatures?)`

Creates a `SignableMessage` from a `Uint8Array` or a UTF-8 string. It optionally accepts a signature dictionary if the message already contains signatures.

```ts
const myMessage = createSignableMessage(new Uint8Array([1, 2, 3]));
const myMessageFromText = createSignableMessage('Hello world!');
const myMessageWithSignatures = createSignableMessage('Hello world!', {
    '1234..5678': new Uint8Array([1, 2, 3]),
});
```

#### Type guards

Each of the message interfaces described above comes with two type guards that allow us to check whether or not a given value is a message signer of the requested type. One that returns a boolean and one that asserts by throwing an error if the provided value is not of the expected interface.

```ts
const myAddress = address('1234..5678');

isMessagePartialSigner({ address: myAddress, signMessages: async () => {} }); // ✅ true
isMessagePartialSigner({ address: myAddress }); // ❌ false
assertIsMessagePartialSigner({ address: myAddress, signMessages: async () => {} }); // ✅ void
assertIsMessagePartialSigner({ address: myAddress }); // ❌ Throws an error.

isMessageModifyingSigner({ address: myAddress, modifyAndSignMessages: async () => {} }); // ✅ true
isMessageModifyingSigner({ address: myAddress }); // ❌ false
assertIsMessageModifyingSigner({ address: myAddress, modifyAndSignMessages: async () => {} }); // ✅ void
assertIsMessageModifyingSigner({ address: myAddress }); // ❌ Throws an error.

isMessageSigner({ address: myAddress, signMessages: async () => {} }); // ✅ true
isMessageSigner({ address: myAddress, modifyAndSignMessages: async () => {} }); // ✅ true
assertIsMessageSigner({ address: myAddress, signMessages: async () => {} }); // ✅ void
assertIsMessageSigner({ address: myAddress, modifyAndSignMessages: async () => {} }); // ✅ void
```

#### `TransactionPartialSigner<TAddress>`

An interface that signs an array of `CompilableTransactions` without modifying their content. It defines a `signTransactions` function that returns a `SignatureDictionary` for each provided transaction. Such signature dictionaries are expected to be merged with the existing ones if any.

```ts
const myTransactionPartialSigner: TransactionPartialSigner<'1234..5678'> = {
    address: address('1234..5678'),
    signTransactions: async (transactions: CompilableTransaction[]): Promise<SignatureDictionary[]> => {
        // My custom signing logic.
    },
};
```

**Characteristics**:

-   **Parallel**. It returns a signature directory for each provided transaction without modifying them, making it possible for multiple partial signers to sign the same transaction in parallel.
-   **Flexible order**. The order in which we use these signers for a given transaction doesn’t matter.

#### `TransactionModifyingSigner<TAddress>`

An interface that potentially modifies the provided `CompilableTransactions` before signing them. E.g. this enables wallets to inject additional instructions into the transaction before signing them. For each transaction, instead of returning a `SignatureDirectory`, its `modifyAndSignTransactions` function returns an updated `CompilableTransaction` with a potentially modified set of instructions and signature dictionary.

```ts
const myTransactionModifyingSigner: TransactionModifyingSigner<'1234..5678'> = {
    address: address('1234..5678'),
    modifyAndSignTransactions: async <T extends CompilableTransaction>(transactions: T[]): Promise<T[]> => {
        // My custom signing logic.
    },
};
```

**Characteristics**:

-   **Sequential**. Contrary to partial signers, these cannot be executed in parallel as each call can modify the provided transactions.
-   **First signers**. For a given transaction, a modifying signer must always be used before a partial signer as the former will likely modify the transaction and thus impact the outcome of the latter.
-   **Potential conflicts**. If more than one modifying signer is provided, the second signer may invalidate the signature of the first one. However, modifying signers may decide not to modify a transaction based on the existence of signatures for that transaction.

#### `TransactionSendingSigner<TAddress>`

An interface that signs one or multiple transactions before sending them immediately to the blockchain. It defines a `signAndSendTransactions` function that returns the transaction signature (i.e. its identifier) for each provided `CompilableTransaction`. This interface is required for PDA wallets and other types of wallets that don't provide an interface for signing transactions without sending them.

Note that it is also possible for such signers to modify the provided transactions before signing and sending them. This enables use cases where the modified transactions cannot be shared with the app and thus must be sent directly.

```ts
const myTransactionSendingSigner: TransactionSendingSigner<'1234..5678'> = {
    address: address('1234..5678'),
    signAndSendTransactions: async (transactions: CompilableTransaction[]): Promise<SignatureBytes[]> => {
        // My custom signing logic.
    },
};
```

**Characteristics**:

-   **Single signer**. Since this signer also sends the provided transactions, we can only use a single `TransactionSendingSigner` for a given set of transactions.
-   **Last signer**. Trivially, that signer must also be the last one used.
-   **Potential conflicts**. Since signers may decide to modify the given transactions before sending them, they may invalidate previous signatures. However, signers may decide not to modify a transaction based on the existence of signatures for that transaction.
-   **Potential confirmation**. Whilst this is not required by this interface, it is also worth noting that most wallets will also wait for the transaction to be confirmed (typically with a `confirmed` commitment) before notifying the app that they are done.

#### `TransactionSigner<TAddress>`

Union interface that uses any of the available transaction signers.

```ts
type TransactionSigner<TAddress extends string = string> =
    | TransactionPartialSigner<TAddress>
    | TransactionModifyingSigner<TAddress>
    | TransactionSendingSigner<TAddress>;
```

#### Type guards

Each of the transaction interfaces described above comes with two type guards that allow us to check whether or not a given value is a transaction signer of the requested type. One that returns a boolean and one that asserts by throwing an error if the provided value is not of the expected interface.

```ts
const myAddress = address('1234..5678');

isTransactionPartialSigner({ address: myAddress, signTransactions: async () => {} }); // ✅ true
isTransactionPartialSigner({ address: myAddress }); // ❌ false
assertIsTransactionPartialSigner({ address: myAddress, signTransactions: async () => {} }); // ✅ void
assertIsTransactionPartialSigner({ address: myAddress }); // ❌ Throws an error.

isTransactionModifyingSigner({ address: myAddress, modifyAndSignTransactions: async () => {} }); // ✅ true
isTransactionModifyingSigner({ address: myAddress }); // ❌ false
assertIsTransactionModifyingSigner({ address: myAddress, modifyAndSignTransactions: async () => {} }); // ✅ void
assertIsTransactionModifyingSigner({ address: myAddress }); // ❌ Throws an error.

isTransactionSendingSigner({ address: myAddress, signAndSignTransaction: async () => {} }); // ✅ true
isTransactionSendingSigner({ address: myAddress }); // ❌ false
assertIsTransactionSendingSigner({ address: myAddress, signAndSignTransaction: async () => {} }); // ✅ void
assertIsTransactionSendingSigner({ address: myAddress }); // ❌ Throws an error.

isTransactionSigner({ address: myAddress, signTransactions: async () => {} }); // ✅ true
isTransactionSigner({ address: myAddress, modifyAndSignTransactions: async () => {} }); // ✅ true
isTransactionSigner({ address: myAddress, signAndSignTransaction: async () => {} }); // ✅ true
assertIsTransactionSigner({ address: myAddress, signTransactions: async () => {} }); // ✅ void
assertIsTransactionSigner({ address: myAddress, modifyAndSignTransactions: async () => {} }); // ✅ void
assertIsTransactionSigner({ address: myAddress, signAndSignTransaction: async () => {} }); // ✅ void
```

#### `KeyPairSigner<TAddress>`

Defines a signer that uses a `CryptoKeyPair` to sign messages and transactions. It implements both the `MessagePartialSigner` and `TransactionPartialSigner` interfaces and keeps track of the `CryptoKeyPair` instance used to sign messages and transactions.

```ts
import { generateKeyPairSigner } from '@solana/signers';

const myKeyPairSigner = generateKeyPairSigner();
myKeyPairSigner.address; // Address;
myKeyPairSigner.keyPair; // CryptoKeyPair;
const [myMessageSignatures] = await myKeyPairSigner.signMessages([myMessage]);
const [myTransactionSignatures] = await myKeyPairSigner.signTransactions([myTransaction]);
```

#### `createSignerFromKeyPair()`

Creates a `KeyPairSigner` from a provided Crypto KeyPair. The `signMessages` and `signTransactions` functions of the returned signer will use the private key of the provided key pair to sign messages and transactions. Note that both the `signMessages` and `signTransactions` implementations are parallelized, meaning that they will sign all provided messages and transactions in parallel.

```ts
import { generateKeyPair } from '@solana/keys';
import { createSignerFromKeyPair, KeyPairSigner } from '@solana/signers';

const myKeyPair: CryptoKeyPair = await generateKeyPair();
const myKeyPairSigner: KeyPairSigner = await createSignerFromKeyPair(myKeyPair);
```

#### `generateKeyPairSigner()`

A convenience function that generates a new Crypto KeyPair and immediately creates a `KeyPairSigner` from it.

```ts
import { generateKeyPairSigner } from '@solana/signers';

const myKeyPairSigner = await generateKeyPairSigner();
```

#### `isKeyPairSigner()`

A type guard that returns `true` if the provided value is a `KeyPairSigner`.

```ts
const myKeyPairSigner = await generateKeyPairSigner();
isKeyPairSigner(myKeyPairSigner); // ✅ true
isKeyPairSigner({ address: address('1234..5678') }); // ❌ false
```

#### `assertIsKeyPairSigner()`

A type guard that throws an error if the provided value is not a `KeyPairSigner`.

```ts
const myKeyPairSigner = await generateKeyPairSigner();
assertIsKeyPairSigner(myKeyPairSigner); // ✅ void
assertIsKeyPairSigner({ address: address('1234..5678') }); // ❌ Throws an error.
```

#### `createNoopSigner()`

_Coming soon..._

Creates a Noop (No-Operation) signer from a given address. It will return an implementation of both the `MessagePartialSigner` and `TransactionPartialSigner` interfaces that do not sign anything. Namely, signing a transaction or a message will return an empty `SignatureDictionary`.

This signer may be useful:

-   For testing purposes.
-   For indicating that a given account is a signer and taking the responsibility to provide the signature for that account ourselves. For instance, if we need to send the transaction to a server that will sign it and send it for us.

```ts
import { createNoopSigner } from '@solana/signers';

const myAddress = address('1234..5678');
const myNoopSigner = createNoopSigner(myAddress);
// ^ MessagePartialSigner<'1234..5678'> & TransactionPartialSigner<'1234..5678'>
```

### Types

_Coming soon..._

### Functions

_Coming soon..._


lorisleiva:
* master
  * **PR #1705** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1705" target="_blank">Graphite</a> 
    * **PR #1706** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1706" target="_blank">Graphite</a> 
      * **PR #1710** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1710" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1710?utm_source=stack-comment).

packages/signers/src/keypair-signer.ts
```diff
@@ -0,0 +1,64 @@
...
+        isTransactionSigner(value)
+    );
+}
+
+/** Asserts that the provided value implements the {@link KeyPairSigner} interface. */
+export function assertIsKeyPairSigner<TAddress extends string>(value: {
+    address: Base58EncodedAddress<TAddress>;
+}): asserts value is KeyPairSigner<TAddress>;
+export function assertIsKeyPairSigner(value: unknown): asserts value is KeyPairSigner;
+export function assertIsKeyPairSigner(value: unknown): asserts value is KeyPairSigner {
```
mcintyre94, line 33: How come we have these two `unknown` ones? Won't everything valid meet the first overload with the `{address}` field? 
And nitpick: can we put the implementation on the first `value: unknown` one instead of writing it out twice?

packages/signers/src/keypair-signer.ts
```diff
@@ -0,0 +1,64 @@
...
+        // TODO: Coded error.
+        throw new Error('The provided value does not implement the KeyPairSigner interface');
+    }
+}
+
+/** Creates a KeyPairSigner from the provided Crypto KeyPair. */
+export async function createSignerFromKeyPair(keyPair: CryptoKeyPair): Promise<KeyPairSigner> {
+    return {
+        keyPair,
+        address: await getAddressFromPublicKey(keyPair.publicKey),
```
mcintyre94, line 44: This is quite nice to have available when creating transactions, since we sensibly put string addresses in instructions now!

packages/signers/README.md
```diff
@@ -0,0 +1,281 @@
...
+
+## Creating Noop signers
+
+### Functions
+
+#### `createNoopSigner()`
+
+_Coming soon..._
+
+Creates a Noop (No-Operation) signer from a given address. It will return an implementation of both the `MessageSigner` and `TransactionSigner` interfaces that do not sign anything. Namely, signing a transaction will return the transaction as-is and signing a message will provide empty byte arrays as signatures. This is mostly useful for testing purposes.
```
mcintyre94, line 264: Do we have a testing use case for this currently? Or has it been useful in Kinobi etc? I'd be tempted to just skip this and let people define it themselves in tests if they need it.

mcintyre94 (approved):
This all LGTM so far! Amazing documentation too!

packages/signers/README.md
```diff
@@ -0,0 +1,281 @@
...
+
+[code-style-prettier-image]: https://img.shields.io/badge/code_style-prettier-ff69b4.svg?style=flat-square
+[code-style-prettier-url]: https://github.com/prettier/prettier
+[npm-downloads-image]: https://img.shields.io/npm/dm/@solana/signers/experimental.svg?style=flat
+[npm-image]: https://img.shields.io/npm/v/@solana/signers/experimental.svg?style=flat
+[npm-url]: https://www.npmjs.com/package/@solana/signers/v/experimental
+[semantic-release-image]: https://img.shields.io/badge/%20%20%F0%9F%93%A6%F0%9F%9A%80-semantic--release-e10079.svg
+[semantic-release-url]: https://github.com/semantic-release/semantic-release
+
+# @solana/signers
```
steveluscher, line 15: <3

packages/signers/src/keypair-signer.ts
lorisleiva: Oh that's because I though it might be helpful to have guard that accept any old piece of data and tell you if it's a signer or not. So technically you could do `assertIsKeyPairSigner("some_address_as_a_string")` and it will throw but type-check. That way if you have an input that's like `Based58EncodedAddress | KeyPairSigner` then you can thrown that input in there too. Would you prefer if the input guard was more narrow?

packages/signers/README.md
lorisleiva: I don't mind not including this out-of-the-box. There are two main use-cases I've seen for noop signers (as people used them in Umi):
- For testing purposes
- In production, to create transactions on the client that require a server signature. Since the generated instruction function requires a signer, you pass a noop signer to make it happy and then you send it over to your server (serialised) to it can sign it for real before sending it.

packages/signers/src/keypair-signer.ts
mcintyre94: I think I'd prefer to restrict it to types that might succeed, and make people narrow the type before calling it. I think that's what we've done with the other guards, eg you can't call `address(String | Number)`

packages/signers/README.md
mcintyre94: Ah got you, this is `guestIdentityDriver` or something like that in the Metaplex SDK? Definitely found that useful for partial signing! Since we have tree shaking I can't see the harm in adding it.

packages/signers/src/keypair-signer.ts
lorisleiva: Gotcha, thanks! I'll update accordingly when I adjust the signer interfaces.

packages/signers/src/__tests__/keypair-signer-test.ts
```diff
@@ -0,0 +1,178 @@
...
+        expect(() => assertIsKeyPairSigner({ address: myAddress })).toThrow(expectedMessage);
+        expect(() => assertIsKeyPairSigner({ ...mySigner, signMessage: 42 })).toThrow(expectedMessage);
+        expect(() => assertIsKeyPairSigner({ ...mySigner, signTransaction: 42 })).toThrow(expectedMessage);
+        expect(() => assertIsKeyPairSigner({ ...mySigner, keyPair: 42 })).toThrow(expectedMessage);
+    });
+});
+
+describe('createSignerFromKeyPair', () => {
+    it('creates a KeyPairSigner from a given CryptoKeypair', async () => {
+        expect.hasAssertions();
```
steveluscher, line 69: Prefer `expect.assertions(number)` and enumerate them.

packages/signers/src/signable-message.ts
```diff
@@ -0,0 +1,17 @@
+import { TextEncoder } from 'text-encoding-impl';
+
+import { SignatureDictionary } from './types';
+
+/** Defines a message that needs signing and its current set of signatures if any. */
+export type SignableMessage = {
+    content: Uint8Array;
```
steveluscher, line 7: I didn't dig at all; is it possible to spec TypedArrays as readonly somehow? Modifying the content invalidates the signatures.

packages/signers/src/signable-message.ts
```diff
@@ -0,0 +1,17 @@
...
+    signatures: SignatureDictionary;
+};
```
steveluscher, line 9: Readonly, everywhere.

```suggestion
export type SignableMessage = Readonly<{
    content: Uint8Array;
    signatures: SignatureDictionary;
}>;
```

packages/signers/src/signable-message.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Creates a signable message from a provided content.
+ * If a string is provided, it will be UTF-8 encoded.
+ */
+export function createSignableMessage(content: string | Uint8Array, signatures: SignatureDictionary = {}) {
```
steveluscher, line 15: I'll just write this comment once, rather than all over the place (eg. same applies to `createSignerFromKeyPair`); I think all of this should be immutable and frozen; the wrapper and the dictionaries. We should force folks who modify this stuff to return new objects updated in the immutable style.

So I'd:

* Freeze the signature dictionary
* Freeze the wrapper
* Do that same thing everywhere
* Write tests that assert that they're frozen
* Wherever possible write tests that assert that if the signatures don't _actually_ change, the exact same dictionary is returned. Same with the message/transaction content.

Examples of all of this can be found in `@solana/transactions`.

steveluscher (changes requested):
This is all super cool.

Are we going to be able to eventually _enforce_ things like ‘don't put a modifying signer after a regular signer’ in the type system, or in the runtime? Like, if we made the various signer types participate in _discriminated_ unions (ie. by adding a ` __type` property that can be asserted on) would we be able to catch problems like that?

packages/signers/README.md
```diff
@@ -0,0 +1,366 @@
...
+```ts
+export type SignableMessage = {
+    content: Uint8Array;
+    signatures: SignatureDictionary; // Record<Address, SignatureBytes>
+};
+```
+
+#### `MessagePartialSigner<TAddress>`
+
+An interface that signs an array of `SignableMessage`s without modifying their content. It defines a `signMessage` function that returns a `SignatureDictionary` for each provided message. Such signature dictionaries are expected to be merged with the existing ones if any.
```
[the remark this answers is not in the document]
lorisleiva: I prefer it that way too but assumed you guys wouldn't haha. Making the change. ☺️

packages/signers/README.md
```diff
@@ -0,0 +1,366 @@
...
+```
+
+As you can see, this provides a consistent API regardless of how things are being signed behind the scenes. If tomorrow we need to use a browser wallet instead, we'd simply need to swap the `generateKeyPairSigner` function with the signer factory of our choice.
+
+This package offers a total of five different types of signers that may be used in combination when applicable. Three of them allow us to sign transactions whereas the other two are used for regular message signing.
+
+They are separated into three categories:
+
+-   **Partial signers**: Given a message or transaction, provide one or more signatures for it. These signers are not able to modify the given data which allows us to run many of them in parallel.
+-   **Modifier signers**: Given a message or transaction, potentially modify it before providing one or more signatures for it. When many signers are provided for a single entity, modifier signers should be used first and sequentially as they can modify the entity and therefore invalidate previous signatures.
```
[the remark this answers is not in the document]
lorisleiva: Updated `*ModifierSigners` and `*SenderSigners` to `*ModifyingSigners` and `*SendingSigners`.

packages/signers/src/__tests__/signable-message-test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    it('creates a SignableMessage from a UTF-8 string', () => {
+        // When we create a SignableMessage by providing a string.
+        const message = createSignableMessage('Hello world!');
+
+        // Then we expect this string to be UTF-8 encoded.
+        expect(message.content).toStrictEqual(new TextEncoder().encode('Hello world!'));
+        expect(message.signatures).toStrictEqual({});
+    });
+});
```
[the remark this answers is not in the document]
lorisleiva: I had the same thoughts. Initially I was planning on bringing `codecs-strings` but then thought let's at least narrow it to the TextEncoder impl package.

I'd say 99% of the time, when people sign messages, it's from a UTF-8 string so it would be a nice helper.

**I'm happy to make it only accept `Uint8Array`** but it kinda broke my heart when this piece of the documentation (the very first code snippet of the README) when from this:

```ts
const myMessage = createSignableMessage('Hello world!');
const [messageSignatures] = await mySigner.signMessage([myMessage]);
```

to this:

```ts
const myMessage = createSignableMessage(new TextEncoder().encode('Hello world!'));
const [messageSignatures] = await mySigner.signMessage([myMessage]);

// or even:
const myMessage = createSignableMessage(getUtf8Encoder().encode('Hello world!'));
const [messageSignatures] = await mySigner.signMessage([myMessage]);
```

What do you think?

packages/signers/src/__tests__/signable-message-test.ts
```diff
@@ -0,0 +1,42 @@
...
+            '1111': new Uint8Array([1, 1, 1, 1]),
+            '2222': new Uint8Array([2, 2, 2, 2]),
+        };
+
+        // When we create a SignableMessage using both the content and the existing signatures.
+        const message = createSignableMessage(content, signatures);
+
+        // Then we expect the message to contain both of these argument as-is.
+        expect(message.content).toBe(content);
+        expect(message.signatures).toBe(signatures);
```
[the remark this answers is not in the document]
lorisleiva: What about: not clone but ensure every `SignatureDictionary` is a readonly object?

packages/signers/src/signable-message.ts
lorisleiva: You can't do that natively on TypeScript but I found a solution from [this article](https://www.growingwiththeweb.com/2020/10/typescript-readonly-typed-arrays.html).

```ts
type TypedArrayMutableProperties = 'copyWithin' | 'fill' | 'reverse' | 'set' | 'sort';
interface ReadonlyUint8Array extends Omit<Uint8Array, TypedArrayMutableProperties> {
    readonly [n: number]: number;
}
```

However, this needs to be added to a lower-level library and used as a requirement for most functions because, rightfully, `Uint8Array` satisfies `ReadonlyUint8Array` but `ReadonlyUint8Array` does not satisfy `Uint8Array`.

In this particular case, I cannot use `ReadonlyUint8Array` because `signBytes` accepts a `Uint8Array` and not a `ReadonlyUint8Array`. I could type-cast it for now though. What do you think?

packages/signers/src/signable-message.ts
lorisleiva: Thanks!
- I marked everything as readonly. Even the transaction and signer interfaces themselves (let me know if that's too much).
- I froze the `SignatureDictionaries`, the `SignedMessages` and the created/generated `KeyPairSigner`.
- I added a `expect(x).toBeFrozenObject()` test for each `Object.freeze(x)`.
- I didn't assert the last point because I currently don't have a function that changes signatures, only functions that provide new frozen `SignatureDictionaries`.

lorisleiva:
Thanks for the review @steveluscher! I resolved most of the points and answered the others.

> Are we going to be able to eventually enforce things like ‘don't put a modifying signer after a regular signer’ in the type system, or in the runtime? Like, if we made the various signer types participate in discriminated unions (ie. by adding a  __type property that can be asserted on) would we be able to catch problems like that?

Before answering the question, I'd like to note that:
- In a way they are already discriminated by the existence of the function itself (e.g. `"modifyAndSignTransaction" in signer` which is what we use for the type guards).
- These interfaces are composable (e.g. `TransactionPartialSigner & TransactionModyfingSigner` is a valid use-case) so having an explicit `__type` discriminator would prevent composable signers.
- We should be careful in how much we control the signer execution because the signer itself will likely adapt based on the provided message or transaction.

Now, if we specifically wanted to ensure all `TransactionModyfingSigner` were used before any `TransactionPartialSigner` with TypeScript, we would need to keep track of that information on the transaction or message itself. Maybe add a `__partiallySigned` flag or something that would throw a TypeScript error when we try to run `modifyAndSign*` on it afterwards.

packages/signers/src/__tests__/signable-message-test.ts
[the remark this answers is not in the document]
buffalojoec: I vote bringing `codecs-strings` to keep the code looking like the first snippet.

packages/signers/README.md
```diff
@@ -0,0 +1,368 @@
...
+#### `createNoopSigner()`
+
+_Coming soon..._
+
+Creates a Noop (No-Operation) signer from a given address. It will return an implementation of both the `MessagePartialSigner` and `TransactionPartialSigner` interfaces that do not sign anything. Namely, signing a transaction or a message will return an empty `SignatureDictionary`.
+
+This signer may be useful:
+
+-   For testing purposes.
+-   For indicating that a given account is a signer and taking the responsibility to provide the signature for that account ourselves. For instance, if we need to send the transaction to a server that will sign it and send it for us.
```
buffalojoec, line 350: I know you haven't included this in the code, but I wanted to mention I don't really love the naming convention "no-op" here.

This sounds like the type of signer that would be used primarily with transaction requests, and I think it should have a name that's less off-putting for folks using Solana Pay, etc. and marking signers whose signatures might be provided by a QR-scanning wallet (for example).

My mind goes to "IOUSigner" (lol, you get the point), but how about something like `FutureSigner`?

packages/signers/src/message-partial-signer.ts
```diff
@@ -0,0 +1,29 @@
+import { Address } from '@solana/addresses';
+
+import { SignableMessage } from './signable-message';
+import { SignatureDictionary } from './types';
+
+/** Defines a signer capable of signing messages. */
+export type MessagePartialSigner<TAddress extends string = string> = Readonly<{
+    address: Address<TAddress>;
+    signMessage(messages: readonly SignableMessage[]): Promise<readonly SignatureDictionary[]>;
```
buffalojoec, line 9: nit: This was a little confusing at first, since it's called `signMessage` but it actually can sign multiple.

packages/signers/src/transaction-partial-signer.ts
```diff
@@ -0,0 +1,29 @@
+import { Address } from '@solana/addresses';
+import { CompilableTransaction } from '@solana/transactions';
+
+import { SignatureDictionary } from './types';
+
+/** Defines a signer capable of signing transactions. */
+export type TransactionPartialSigner<TAddress extends string = string> = Readonly<{
+    address: Address<TAddress>;
+    signTransaction(transactions: readonly CompilableTransaction[]): Promise<readonly SignatureDictionary[]>;
```
buffalojoec, line 9: nit: Same nit here on plurality.

packages/signers/README.md
```diff
@@ -0,0 +1,368 @@
...
+assertIsTransactionSigner({ address: myAddress, signAndSignTransaction: async () => {} }); // ✅ void
+```
+
+## Creating and generating KeyPair signers
+
+### Types
+
+#### `KeyPairSigner<TAddress>`
+
+Defines a signer that uses a `CryptoKeyPair` to sign messages and transactions. It implements both the `MessagePartialSigner` and `TransactionPartialSigner` interfaces and keeps track of the `CryptoKeyPair` instance used to sign messages and transactions.
```
buffalojoec, line 281: Do we not have any reason to roll `KeyPairModifyingSigner` and `KeyPairSendingSigner`?

buffalojoec: Even just as interfaces someone could implement on their own?

packages/signers/src/message-modifying-signer.ts
```diff
@@ -0,0 +1,30 @@
+import { Address, isAddress } from '@solana/addresses';
+
+import { SignableMessage } from './signable-message';
+
+/** Defines a signer capable of signing messages. */
+export type MessageModifyingSigner<TAddress extends string = string> = Readonly<{
+    address: Address<TAddress>;
+    modifyAndSignMessage(messages: readonly SignableMessage[]): Promise<readonly SignableMessage[]>;
```
buffalojoec, line 8: Would it make sense to return the new `SignableMessage` _and_ the new `SignatureDictionary`? Since the dictionary will now have one entry.

packages/signers/src/transaction-modifying-signer.ts
```diff
@@ -0,0 +1,29 @@
+import { Address } from '@solana/addresses';
+import { CompilableTransaction } from '@solana/transactions';
+
+/** Defines a signer capable of signing transactions. */
+export type TransactionModifyingSigner<TAddress extends string = string> = Readonly<{
+    address: Address<TAddress>;
+    modifyAndSignTransaction<TTransaction extends CompilableTransaction>(
```
buffalojoec, line 7: Same here on the dictionary return value.

buffalojoec (commented):
Just a few comments from my side, but overall I think the API is excellent. I also agree with Steve on tightening up some of the frozen-object stuff. This is almost ready to rip, though.

packages/signers/README.md
lorisleiva: I think signers should provide the most convenient interfaces they can provide. In the case of `CryptoKeyPairs`, we can provide the more efficient and flexible `*PartialSigner` interfaces, so I'm not sure why we'd want to offer more than that. Do you have any use-cases in mind?

packages/signers/src/message-partial-signer.ts
lorisleiva: Glad you mentioned it. It's also been on my mind.

This is the current API:

```ts
const [signatures] = await signer.signMessage([message]);
const [signaturesA, signaturesB] = await signer.signMessage([messageA, messageB]);
```

These are the alternatives:

```ts
// Option A: Use a variadic argument (this is what the wallet-standard does).
const [signatures] = await signer.signMessage(message);
const [signaturesA, signaturesB] = await signer.signMessage(messageA, messageB);

// Option B: Pluralize.
const [signatures] = await signer.signMessages([message]);
const [signaturesA, signaturesB] = await signer.signMessages([messageA, messageB]);
```

IMO, they each have pros and cons:
- **Current API**: most of the time we're only signing one message or transaction so `signMessage` reads better but there's confusion with the fact that we provide an array.
- **Option A**: Uses `signMessage` + no longer provides an array + consistent with the wallet standard. However, we loose the type symmetry between input and output. It feels a bit weird that you pass an object but get an array of objects.
- **Option B**: Most predictable and consistent API but we're having to use `signMessages` even when most of the time we want to sign single messages.

That being said, I prefer Option A or B over the current API. What about you?

packages/signers/src/transaction-partial-signer.ts
lorisleiva: Moving this conversation [here](https://github.com/solana-labs/solana-web3.js/pull/1710#discussion_r1389591624).

packages/signers/README.md
lorisleiva: Hmm I agree that `NoopSigner` feels low-level and not as inviting as it could be. However it best describes what the signer does: nothing.

I worry that with something like `FutureSigner` people will assume the signer actually does something.

packages/signers/src/message-modifying-signer.ts
lorisleiva: I'm not sure I understand the benefit of that since the the `SignableMessage` already contains a `SignatureDictionary`?

packages/signers/src/transaction-modifying-signer.ts
lorisleiva: Here as well the returned transaction already contains a `SignatureDictionary` in its `signatures` attribute.

packages/signers/src/__tests__/signable-message-test.ts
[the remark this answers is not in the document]
steveluscher: I'd still sort of like it frozen with `Object.freeze()` and we don't want to freeze the object that someone passed in because that's mean.

packages/signers/src/__tests__/signable-message-test.ts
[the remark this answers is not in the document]
steveluscher: I'm fine with that, it's just RIP now everyone has a peer dependency on `fastestsmallesttextencoderdecoder` for React Native. It won't get bundled, of course, but it will get downloaded.

packages/signers/src/signable-message.ts
steveluscher: Frustrating, but let's punt on it. It's a flourish.

packages/signers/src/message-partial-signer.ts
steveluscher: I worry that it won't be obvious enough to everyone that option A can be used to sign multiple messages in parallel if only you supply more than one message as an argument, so I prefer option B.

packages/signers/src/signable-message.ts
```diff
@@ -0,0 +1,20 @@
...
+}>;
+
+/**
+ * Creates a signable message from a provided content.
+ * If a string is provided, it will be UTF-8 encoded.
+ */
+export function createSignableMessage(content: string | Uint8Array, signatures: SignatureDictionary = {}) {
+    return Object.freeze({
+        content: typeof content === 'string' ? new TextEncoder().encode(content) : content,
+        signatures: Object.freeze(signatures),
```
steveluscher, line 18: This could be upsetting for the person who passed this object into this function. They may wonder why it's all of a sudden frozen.

steveluscher (approved):
[signaturedrake](https://app.graphite.dev/macros/solana-labs)

packages/signers/src/message-partial-signer.ts
buffalojoec: Yeah I’m also on team B!

packages/signers/src/message-modifying-signer.ts
buffalojoec: Oops my bad, that's cool then 😅

packages/signers/README.md
buffalojoec: Yes but it's doing nothing and is expecting resolution at some point. It's sort of like ear-marking a message/transaction for a later signature.

packages/signers/src/message-partial-signer.ts
lorisleiva: ✅ Updated the API using Option B.

packages/signers/src/__tests__/signable-message-test.ts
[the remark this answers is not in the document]
lorisleiva: How about `Object.freeze({ ...signatures })` instead so we only freeze a shallow clone of the provided signatures?

packages/signers/src/__tests__/signable-message-test.ts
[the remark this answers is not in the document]
lorisleiva: Right now, I'm not convince that `Record<Address, SignatureBytes>` is the best definition of a `SignatureDirectory` but I used it so it matches the definition of the `signatures` attribute of transactions.

I think `Record<Address, Signature>` may be more elegant since both the keys and values will be encoded as base 58 strings.

The reason I mention that here is that, if we make that change, we'll likely need `getBase58Encoder()` in this package anyway.

packages/signers/src/signable-message.ts
lorisleiva: Moving this conversation [here](https://github.com/solana-labs/solana-web3.js/pull/1710#discussion_r1390419403).

packages/signers/src/__tests__/keypair-signer-test.ts
```diff
@@ -0,0 +1,198 @@
...
+        expect(jest.mocked(getAddressFromPublicKey)).toHaveBeenCalledTimes(1);
+        expect(mySigner.address).toBe(myAddress);
+        expect(mySigner.keyPair).toBe(myKeyPair);
+
+        // And provided functions to sign messages and transactions.
+        expect(typeof mySigner.signMessage).toBe('function');
+        expect(typeof mySigner.signTransaction).toBe('function');
+
+        // And the signer is frozen.
+        expect(mySigner).toBeFrozenObject();
```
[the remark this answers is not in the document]
lorisleiva: ✅ Test refactored.

packages/signers/src/__tests__/signable-message-test.ts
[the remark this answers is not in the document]
lorisleiva: I updated the code and tests, let me know what you think.

packages/signers/README.md
lorisleiva: Let's move this conversation [here](https://github.com/solana-labs/solana-web3.js/pull/1842).

buffalojoec (approved):
ship it

packages/signers/src/__tests__/signable-message-test.ts
```diff
@@ -0,0 +1,73 @@
...
+        const signatures = {
+            '1111': new Uint8Array([1, 1, 1, 1]),
+            '2222': new Uint8Array([2, 2, 2, 2]),
+        };
+
+        // When we create a new SignableMessage using this signature directory.
+        const message = createSignableMessage('Hello world!', signatures);
+
+        // Then the signature directory is copied and frozen.
+        expect(message.signatures).not.toBe(signatures);
```
steveluscher, line 69: [nice](https://app.graphite.dev/macros/solana-labs)

steveluscher (approved):
[approvedarthvader](https://app.graphite.dev/macros/solana-labs)

PR #1722, opened 2023-10-11
refactor(experimental): add keys to compat

buffalojoec:
This PR adds `PublicKey` and `Keypair` compatibility helpers for the new API.

### DIFF
--- a/packages/compat/src/__tests__/keypair.ts
+++ b/packages/compat/src/__tests__/keypair.ts
@@ -1,0 +1,29 @@
+import { Keypair } from '@solana/web3.js';
+
+import { fromLegacyKeypair } from '../keypair';
+
+describe('keypair', () => {
+    const legacyKeypair = Keypair.generate();
+    let keyPair: CryptoKeyPair;
+    describe.each(['public', 'private'])('%s key', type => {
+        beforeEach(async () => {
+            keyPair = await fromLegacyKeypair(legacyKeypair);
+        });
+        it(`has the algorithm "Ed25519"`, () => {
+            expect(keyPair).toHaveProperty([`${type}Key`, 'algorithm', 'name'], 'Ed25519');
+        });
+        it('has the string tag "CryptoKey"', () => {
+            expect(keyPair).toHaveProperty([`${type}Key`, Symbol.toStringTag], 'CryptoKey');
+        });
+        it(`has the type "${type}"`, () => {
+            expect(keyPair).toHaveProperty([`${type}Key`, 'type'], type);
+        });
+    });
+    it.each([true, false])(
+        "sets the private key's `extractable` to `%s` when generating a key pair with the extractability `%s`",
+        async extractable => {
+            expect.assertions(1), (keyPair = await fromLegacyKeypair(legacyKeypair, extractable));
+            expect(keyPair.privateKey).toHaveProperty('extractable', extractable);
+        }
+    );
+});
--- a/packages/compat/src/keypair.ts
+++ b/packages/compat/src/keypair.ts
@@ -1,0 +1,61 @@
+import { Keypair } from '@solana/web3.js';
+
+function addPkcs8Header(bytes: Uint8Array): Uint8Array {
+    // prettier-ignore
+    return new Uint8Array([
+        /**
+         * PKCS#8 header
+         */
+        0x30, // ASN.1 sequence tag
+        0x2e, // Length of sequence (46 more bytes)
+
+            0x02, // ASN.1 integer tag
+            0x01, // Length of integer
+                0x00, // Version number
+
+            0x30, // ASN.1 sequence tag
+            0x05, // Length of sequence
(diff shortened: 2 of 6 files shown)


buffalojoec:
* master
  * **PR #1722** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1722" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1722?utm_source=stack-comment).

buffalojoec:
Still trying to figure out the best way to test the `CryptoKey` ones. Overall this addition came out pretty simple though

packages/compat/src/__tests__/keypair.ts
```diff
@@ -0,0 +1,4 @@
+// TODO: Not totally sure how to test this yet
+describe('keypair', () => {
+    it.todo('should convert from a Legacy Web3 JS Keypair to a CryptoKeyPair');
+});
```
mcintyre94, line 4: I'd probably just test the same as what the `generateKey` polyfill tests - basically that the properties on the key are set correctly
https://github.com/solana-labs/solana-web3.js/blob/5e1b2e5d451f4016b1d846d84b3fa3309e43e494/packages/webcrypto-ed25519-polyfill/src/__tests__/secrets-test.ts#L55-L83

packages/compat/src/keypair.ts
```diff
@@ -0,0 +1,63 @@
...
+    );
+    const privateKeyBytesPkcs8 = addPkcs8Header(keypair.secretKey);
+    const privateKey: CryptoKey = await crypto.subtle.importKey(
+        'pkcs8',
+        privateKeyBytesPkcs8,
+        {
+            name: 'EdDSA',
+            namedCurve: 'Ed25519',
+        },
+        false,
```
mcintyre94, line 56: With `generateKey` we allow the private key to have `extractable` set by the caller. I'm not 100% sure this is correct (https://github.com/solana-labs/solana-web3.js/issues/1723), but if it is then I wonder if we should have the same here?

packages/compat/src/keypair.ts
```diff
@@ -0,0 +1,63 @@
...
+ * @returns         A CryptoKeyPair
+ */
+export async function fromLegacyKeypair(keypair: Keypair): Promise<CryptoKeyPair> {
+    const publicKey: CryptoKey = await crypto.subtle.importKey(
+        'raw',
+        keypair.publicKey.toBytes(),
+        {
+            name: 'EdDSA',
+            namedCurve: 'Ed25519',
+        },
```
mcintyre94, line 44: This clashes with other code that uses the string `Ed25519` as the algorithm

https://github.com/solana-labs/solana-web3.js/blob/5e1b2e5d451f4016b1d846d84b3fa3309e43e494/packages/webcrypto-ed25519-polyfill/src/index.ts#L33

https://github.com/solana-labs/solana-web3.js/blob/5e1b2e5d451f4016b1d846d84b3fa3309e43e494/packages/addresses/src/public-key.ts#L7

https://github.com/solana-labs/solana-web3.js/blob/5e1b2e5d451f4016b1d846d84b3fa3309e43e494/packages/assertions/src/subtle-crypto.ts#L16

I'm not sure which is correct but I think we definitely need to consistently use the same algorithm in our code.

mcintyre94 (commented):
Nice! We have some other code that uses the string `Ed25519` as the algorithm though, and I think we need to make sure everything matches.

packages/compat/src/keypair.ts
buffalojoec: I was trying to dig for an answer to this issue, because I swear I read somewhere that it will reject you if you try to import a private key and mark it extractable, but I can't be sure. Maybe we just need to try it.

packages/compat/src/keypair.ts
buffalojoec: It's the one you linked haha throws an error for me with what I had!

packages/compat/src/keypair.ts
buffalojoec: added the latter, for review

mcintyre94 (approved):
LGTM! 

keys

packages/compat/src/__tests__/keypair.ts
```diff
@@ -0,0 +1,30 @@
...
+        });
+        it('has the string tag "CryptoKey"', () => {
+            expect(keyPair).toHaveProperty([`${type}Key`, Symbol.toStringTag], 'CryptoKey');
+        });
+        it(`has the type "${type}"`, () => {
+            expect(keyPair).toHaveProperty([`${type}Key`, 'type'], type);
+        });
+    });
+    it.each([true, false])(
+        "sets the private key's `extractable` to `false` when generating a key pair with the extractability `%s`",
```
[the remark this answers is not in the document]
buffalojoec: Ahh nice catch!

PR #1727, opened 2023-10-13
chore: make style:fix also fix package.json

mcintyre94:
I keep getting a mismatch between an editor setting and prettier styling for package.json

We run prettier on package.json so our `style:fix` should also fix it

Note that we don't run eslint on package.json and it makes a huge mess of it and complains about ordering of objects, so I've deliberately only added it to the prettier calls

### DIFF
--- a/packages/addresses/package.json
+++ b/packages/addresses/package.json
@@ -38,7 +38,7 @@
         "dev": "jest -c node_modules/test-config/jest-dev.config.ts --rootDir . --watch",
         "prepublishOnly": "version-from-git --no-git-tag-version --template experimental.short",
         "publish-packages": "pnpm publish --tag experimental --access public --no-git-checks",
-        "style:fix": "pnpm eslint --fix src/* && pnpm prettier -w src/*",
+        "style:fix": "pnpm eslint --fix src/* && pnpm prettier -w src/* package.json",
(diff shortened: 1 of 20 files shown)


mcintyre94:
* master
  * **PR #1727** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1727" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1727?utm_source=stack-comment).

buffalojoec (approved):
Nice, did you also happen to run it on all the packages? I think there's some that might be not alpha-ordered

mcintyre94:
> Nice, did you also happen to run it on all the packages? I think there's some that might be not alpha-ordered

Yep, the `test:prettier` step was already running on all the package.json files, so they were already being tested. This just means we autofix them too. I'm guessing prettier doesn't re-order dependencies in package.json, it's probably only doing a few checks. It does hate my editor either removing the blank line at the end though lol, that's what I kept getting caught on

buffalojoec:
> It does hate my editor either removing the blank line at the end though lol, that's what I kept getting caught on

Yep caught me before too

PR #1728, opened 2023-10-13
Add memoizeWithNoParams function to @solana/functional

mcintyre94:
This PR adds a new `memoizeWithNoParams` function to `@solana/functional`

It can be used to automatically memoize any function that does not require any params when called

### DIFF
--- a/packages/functional/src/__tests__/memoize-test.ts
+++ b/packages/functional/src/__tests__/memoize-test.ts
@@ -1,0 +1,18 @@
+import { memoizeWithNoParams } from "../memoize";
+
+describe('memoizeWithNoParams', () => {
+  it('returns a function that returns result of the memoized function', () => {
+    const fn = () => 1;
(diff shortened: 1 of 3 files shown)


mcintyre94:
* master
  * **PR #1728** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1728" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1728?utm_source=stack-comment).

mcintyre94:
I don't think this works the way we'd want to use it unfortunately

// this doesn't tree shake lol
```ts
const memoized = memoizeWithNoParams(() => fn())
```

// and this doesn't memoize
```ts
const memoized = () => memoizeWithNoParams(() => fn())();
```

I guess maybe the memoize state does have to be local to the file and abstracting it this way won't work

PR #1731, opened 2023-10-13
refactor(experimental): make rpc-graphql tree-shakable

buffalojoec:
This PR makes `@solana/rpc-graphql` tree-shakable, which is previously was not.

Basically, there appears to be _a lot_ of changes, but I've just converted static constants to arrow functions. Nothing else has been changed.

### DIFF
--- a/packages/rpc-graphql/src/rpc.ts
+++ b/packages/rpc-graphql/src/rpc.ts
@@ -22,14 +22,14 @@ export function createRpcGraphQL(rpc: Rpc<SolanaRpcMethods>): RpcGraphQL {
     const schema = new GraphQLSchema({
         query: new GraphQLObjectType({
             fields: {
-                ...accountQuery,
-                ...blockQuery,
-                ...programAccountsQuery,
-                ...transactionQuery,
+                ...accountQuery(),
+                ...blockQuery(),
+                ...programAccountsQuery(),
+                ...transactionQuery(),
             },
             name: 'RootQuery',
         }),
-        types: [...accountTypes, ...blockTypes, ...transactionTypes],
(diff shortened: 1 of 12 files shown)


mcintyre94:
@buffalojoec I added a bunch of memoisation here in a second commit

First I was trying to figure out what was likely to reduce the call stack, ie functions that call other functions and get called repeatedly themselves, and just memoise them. 

But once I got past the stack trace errors I started getting errors that the graphql schema had duplicate names. So everything with a `name` field has to be memoised to avoid that. Fixing all of those cases reduced the stack trace enough that the tests now pass locally. Performance should be better for end users too, since these functions will only be called once (per import) and will act more like the constants we had before (but tree shaken!)

buffalojoec:
> @buffalojoec I added a bunch of memoisation here in a second commit
> 
> First I was trying to figure out what was likely to reduce the call stack, ie functions that call other functions and get called repeatedly themselves, and just memoise them.
> 
> But once I got past the stack trace errors I started getting errors that the graphql schema had duplicate names. So everything with a `name` field has to be memoised to avoid that. Fixing all of those cases reduced the stack trace enough that the tests now pass locally. Performance should be better for end users too, since these functions will only be called once (per import) and will act more like the constants we had before (but tree shaken!)

Oh wow thanks for doing this!! I could have saved you a bit of trouble with my local changes, since I narrowed down where the stack trace was coming from. The `Account` interface has the `owner` field, which is _also_ of type `AccountInterface`. I was able to just pull that out of the interface and into the standalone implementations, but tbh that isn't the implementation I wanted, so if you managed to solve it differently I'm all for it.

buffalojoec (commented):
Awesome!! Thanks again, this is a huge lift! Seems to pass all tests.

PR #1730, opened 2023-10-13
refactor(experimental): use codecs-core for transaction version codec

mcintyre94:
This PR refactors the `TransactionVersion` serializer to use the codecs-core Encoder/Decoder/Codec data structures

The only behaviour change is that we no longer have the unimplemented runtime errors. Calling `encode` on the `Decoder`, or vice versa, is now a type error

PR #1729, opened 2023-10-13
refactor(experimental): refactor Address codec

mcintyre94:
- Use strings-codecs package
- Memoise created encoder/decoder

This should serve as a pattern for refactoring the other serializers related to transactions

PR #1726, opened 2023-10-13
refactor(experimental): refactor Address codec

mcintyre94:
- Use strings-codecs package
- Memoise created encoder/decoder

This should serve as a pattern for refactoring the other serializers related to transactions

PR #1732, opened 2023-10-13
refactor(experimental): use codecs packages for message header codec

mcintyre94:
- Refactor to use a memoized u8 encoder/decoder/codec

### DIFF
--- a/packages/transactions/src/serializers/header.ts
+++ b/packages/transactions/src/serializers/header.ts
@@ -1,51 +1,104 @@
-import { Serializer, struct, u8 } from '@metaplex-foundation/umi-serializers';
+import { Codec, Decoder, Encoder } from '@solana/codecs-core';
+import { getStructCodec, getStructDecoder, getStructEncoder } from '@solana/codecs-data-structures';
+import { getU8Codec, getU8Decoder, getU8Encoder } from '@solana/codecs-numbers';
 
 import { getCompiledMessageHeader } from '../compile-header';
 
 type MessageHeader = ReturnType<typeof getCompiledMessageHeader>;
 
-export function getMessageHeaderCodec(): Serializer<MessageHeader> {
-    return struct(
+let memoizedU8Encoder: Encoder<number> | undefined;
+function getMemoizedU8Encoder(): Encoder<number> {
+    if (!memoizedU8Encoder) memoizedU8Encoder = getU8Encoder();
+    return memoizedU8Encoder;
+}
+
+function getMemoizedU8EncoderDescription(description?: string): Encoder<number> {
+    const encoder = getMemoizedU8Encoder();
+    return {
+        ...encoder,
+        description: description ?? encoder.description,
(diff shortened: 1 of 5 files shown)


mcintyre94:
* master
  * **PR #1729** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1729" target="_blank">Graphite</a> 
    * **PR #1730** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1730" target="_blank">Graphite</a> 
      * **PR #1732** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1732" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1732?utm_source=stack-comment).

packages/transactions/src/serializers/header.ts
```diff
@@ -1,51 +1,104 @@
...
+const messageHeaderDescription = __DEV__
+    ? 'The transaction message header containing counts of the signer, readonly-signer, and readonly-nonsigner account addresses'
+    : undefined;
+
+export function getMessageHeaderEncoder(): Encoder<MessageHeader> {
+    return getStructEncoder(
+        [
+            ['numSignerAccounts', getMemoizedU8EncoderDescription(numSignerAccountsDescription)],
+            ['numReadonlySignerAccounts', getMemoizedU8EncoderDescription(numReadonlySignerAccountsDescription)],
+            ['numReadonlyNonSignerAccounts', getMemoizedU8EncoderDescription(numReadonlyNonSignerAccountsDescription)],
```
lorisleiva, line 72: Would it not be better to memoize the entire `MessageHeader(En|De)coder` instead of all their individual components?

lorisleiva (approved):

packages/transactions/src/serializers/header.ts
mcintyre94: So far I've taken the approach that the caller should `memoize` as it wants, so the caller of `getMessageHeaderEncoder` etc will memoize it in a following PR. So I'm kind of working my way up the stack adding memoization at the call sites.

I'm not completely sure if this is the best approach. But I thought it seemed wrong to add memoization to the `codecs-strings` functions for example, because I don't think that module should be concerned with memoization of the base codecs. Which leads to its caller in #1726 memoizing eg `getBase58Encoder` from that package. So I thought continuing that pattern was the best way to separate concerns, where the higher-level codecs memoize the lower level ones that they build on, but not themselves.

mcintyre94: Cool makes sense, I'll just add it where I think it makes sense/where we have more complexity. Agreed that we don't need it everywhere! 

> Ultimately function programming is bound to call lots of functions.

Agree but technically a bunch of functions could be consts if it wasn't for tree shaking 😄

PR #1735, opened 2023-10-16
createLookupTable should not require authority to be the signer

kirksgithub:
Authority should not be required as a signer for createLookupTable:
https://github.com/solana-labs/solana/pull/27248

Creation of a LookupTable via a Proposal using SPL Governance should allow for more flexibility as Slots are difficult to predict, ultimately Slots can easily be missed during instruction execution. Creation should be made by the payer which is flexible to set the authority as needed which in this case can be a Governance Treasury Wallet address that is not the same with the payer

### DIFF
--- a/packages/library-legacy/src/programs/address-lookup-table/index.ts
+++ b/packages/library-legacy/src/programs/address-lookup-table/index.ts
@@ -290,7 +290,7 @@ export class AddressLookupTableProgram {
       },
       {
         pubkey: params.authority,
-        isSigner: true,
+        isSigner: false,
         isWritable: false,
       },
       {


joncinque:
Thanks for your contribution! Since the underlying feature in https://github.com/solana-labs/solana/issues/27205 has not been enabled yet on any cluster, we'll need to wait to merge this. It'll be best to make the change once all networks have the change enabled.

joncinque:
I don't have write permissions on the repo, so someone else will need to reopen this PR -- we'll definitely want to land it once the underlying feature is enabled on mainnet. Considering there are currently 22 features in front of it at https://github.com/solana-labs/solana/wiki/Feature-Gate-Activation-Schedule, and we only enable 1 feature per epoch, this may take up to 2 months

PR #1738, opened 2023-10-18
refactor(experimental): fix test description for extractability test of `generateKey`

### LINKED ISSUE #1723: [Experimental]: Webcrypto Ed25519 Polyfill: should private key be able to be extractable?
## Overview

There's a mismatch between the test description and the code behaviour and I'm not sure which is correct. We need to either correct the test description, or change the code to match the test description behaviour.
https://github.com/solana-labs/solana-web3.js/blob/5e1b2e5d451f4016b1d846d84b3fa3309e43e494/packages/webcrypto-ed25519-polyfill/src/__tests__/secrets-test.ts#L70-L76

- the test description says `extractable` should always be false
- the test itself actually checks that it matches the `extractable` value that is passed in

The code itself also uses the `extractable` value:
https://github.com/solana-labs/solana-web3.js/blob/5e1b2e5d451f4016b1d846d84b3fa3309e43e494/packages/webcrypto-ed25519-polyfill/src/secrets.ts#L58-L63

While it explicitly overrides `extractable` for the public key:
https://github.com/solana-labs/solana-web3.js/blob/5e1b2e5d451f4016b1d846d84b3fa3309e43e494/packages/webcrypto-ed25519-polyfill/src/secrets.ts#L64-L69

My guess is that we just need to fix the test description, but I figure it's worth making sure there's not a code bug here. 

Also see https://github.com/solana-labs/solana-web3.js/pull/1681#discussion_r1355296887

steveluscher:
Closes #1723.

### DIFF
--- a/packages/webcrypto-ed25519-polyfill/src/__tests__/secrets-test.ts
+++ b/packages/webcrypto-ed25519-polyfill/src/__tests__/secrets-test.ts
@@ -68,7 +68,7 @@ describe('generateKeyPolyfill', () => {
         });
     });
     it.each([true, false])(
-        "sets the private key's `extractable` to `false` when generating a key pair with the extractability `%s`",
+        "sets the private key's `extractable` accordingly when generating a key pair with the extractability `%s`",
         extractable => {
             const { privateKey } = generateKeyPolyfill(extractable, ['sign', 'verify']);
             expect(privateKey).toHaveProperty('extractable', extractable);


steveluscher:
* master
  * **PR #1738** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1738" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1738?utm_source=stack-comment).

PR #1755, opened 2023-10-20
refactor(experimental): add Ed25519 importKey polyfill

mcintyre94:
This PR adds the `importKey` function to the webcrypto polyfill

It supersedes #1681, thanks to @ronanyeah for the contribution!

- when we import public key bytes we store them as-is. Since our existing secrets store stores secrets (from which we can derive a public key), we use a new store for these public key bytes which don't need to be converted to be used as a public key

- when we import private key bytes, we validate the PKCS8 header (we just compare to the known precise expected header), and strip it. The remaining private key bytes are stored in the secrets store

### DIFF
--- a/packages/webcrypto-ed25519-polyfill/src/__tests__/secrets-test.ts
+++ b/packages/webcrypto-ed25519-polyfill/src/__tests__/secrets-test.ts
@@ -1,7 +1,7 @@
 import {
     exportKeyPolyfill,
-    importKeyPolyfill,
     generateKeyPolyfill,
+    importKeyPolyfill,
     isPolyfilledKey,
     signPolyfill,
     verifyPolyfill,
@@ -21,7 +21,7 @@ const MOCK_PUBLIC_KEY_BYTES = new Uint8Array([
     166, 132, 114, 186, 49, 163, 23, 12, 11, 14, 119, 219, 102, 96, 26, 226, 91, 97, 238, 217, 236, 84, 232, 204, 62,
     212, 179, 252, 20, 37, 179, 52,
 ]);
-const MOCK_PKCS8_HEADER =
+const ED25519_PKCS8_HEADER =
     // prettier-ignore
     new Uint8Array([
         /**
@@ -52,7 +52,7 @@ const MOCK_PKCS8_HEADER =
 
             // Private key bytes as octet string
             0x04, // ASN.1 octet string tag
-            0x20  // String length (32 bytes)
+            0x20, // String length (32 bytes)
     ]);
 
 describe('exportKeyPolyfill', () => {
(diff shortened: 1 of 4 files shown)


packages/webcrypto-ed25519-polyfill/src/secrets.ts
```diff
@@ -131,10 +183,71 @@ export function verifyPolyfill(key: CryptoKey, signature: BufferSource, data: Bu
...
+
+        const cache = (publicKeyBytesStore ||= new WeakMap());
+        cache.set(publicKey, bytes);
+
+        return publicKey;
+    }
+
+    if (format === 'pkcs8') {
+        if (keyUsages.some(usage => usage === 'verify' || PROHIBITED_KEY_USAGES.has(usage))) {
+            throw new DOMException('Unsupported key usage for a Ed25519 key', 'SyntaxError');
```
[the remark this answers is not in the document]
mcintyre94: Nope, error in Node 18 is:
```
> await crypto.subtle.importKey('pkcs8', privateKeyBytes, 'Ed25519', true, ['verify'])
Uncaught DOMException [SyntaxError]: Unsupported key usage for a Ed25519 key
```

packages/webcrypto-ed25519-polyfill/src/__tests__/secrets-test.ts
```diff
@@ -44,6 +84,161 @@ describe('exportKeyPolyfill', () => {
...
 });
 
+describe('importKeyPolyfill', () => {
+    describe('when format is `raw`', () => {
+        it('allows importing valid public key bytes', () => {
+            expect(() => importKeyPolyfill('raw', MOCK_PUBLIC_KEY_BYTES, false, ['verify'])).not.toThrow();
+        });
+        it('allows importing with empty keyUsages', () => {
+            expect(() => importKeyPolyfill('raw', MOCK_PUBLIC_KEY_BYTES, false, [])).not.toThrow();
+        });
```
steveluscher, line 94: [orlyowl](https://app.graphite.dev/macros/solana-labs)

steveluscher: If someone imports with no usages, does this code still actually allow them to use it with `verify`? It shouldn't, if so.

packages/webcrypto-ed25519-polyfill/src/secrets.ts
```diff
@@ -131,10 +183,71 @@ export function verifyPolyfill(key: CryptoKey, signature: BufferSource, data: Bu
...
+        }
+        // 48 bytes: 16-byte PKCS8 header + 32 byte secret key
+        if (bytes.length !== 48) {
+            throw new DOMException('Invalid keyData', 'DataError');
```
steveluscher, line 229: <3

steveluscher (approved):

packages/webcrypto-ed25519-polyfill/src/__tests__/secrets-test.ts
mcintyre94: Yup! I was doing a pass to make sure my error messages all matched the real ones and realised this didn't actually error. I guess it makes sense because you could generate a keypair with just `["verify"]` and then if you exported and re-imported that private key it'd have empty usages.

PR #1774, opened 2023-10-24
refactor(experimental): Add docs for `@solana/transactions`

steveluscher:


steveluscher:
You can walk through each step of the docs in the ‘commits’ tab on Github, or you can just review the whole thing at once on Graphite.

buffalojoec (approved):
Nice!! Stealing some of this for the one-pager 😈

packages/transactions/README.md
```diff
@@ -14,12 +14,321 @@
...
+
+#### `setTransactionFeePayer()`
+
+Given a base58-encoded address of a system account, this method will return a new transaction having the same type as the one supplied plus the `ITransactionWithFeePayer` type.
+
+```ts
+import { address } from '@solana/addresses';
+import { setTransactionFeePayer } from '@solana/transactions';
+
+const myAddress = address('mpngsFd4tmbUfzDYJayjKZwZcaR7aWb2793J6grLsGu');
```
buffalojoec, line 75: Yeah, `address(..)` looks way better.

packages/transactions/README.md
```diff
@@ -14,12 +14,321 @@
...
+```ts
+import { setTransactionLifetimeUsingBlockhash } from '@solana/transactions';
+
+const { value: latestBlockhash } = await rpc.getLatestBlockhash().send();
+const txWithBlockhashLifetime = setTransactionLifetimeUsingBlockhash(latestBlockhash, tx);
+```
+
+#### `setTransactionLifetimeUsingDurableNonce()`
+
+Given a nonce, the account where the value of the nonce is stored, and the address of the account authorized to consume that nonce, this method will return a new transaction having the same type as the one supplied plus the `IDurableNonceTransaction` type. In particular, this method _prepends_ an instruction to the transaction designed to consume (or &lsquo;advance&rsquo;) the nonce in the same transaction whose lifetime is defined by it.
```
buffalojoec, line 121: Is the quote encoding something I’m only seeing because I’m on my phone or..? (`&lsquo`)

PR #1777, opened 2023-10-24
refactor(experimental): allow the nonce authority account to be writable

mcintyre94:
This PR updates the logic to determine whether a given instruction is a durable nonce transaction. Specifically, the nonce authority address can now be writeable. This is the case if the nonce authority is also the fee payer, for example in a transaction created with the CLI: https://explorer.solana.com/tx/2FZPN5ZJqi5LSMe18auxx59G1VZAP71p8KLCDvyH1WAFgqEXfag6xmC8KnzXQW3KSdbmEyWp8rdd1VnmxMsTrUdG?cluster=devnet 

I don't think any change is required for other accounts

- The nonce address is unlikely to need to be a signer, since it is already initialized 
- I think sysvar addresses are always readonly

### DIFF
--- a/packages/transactions/src/__tests__/durable-nonce-test.ts
+++ b/packages/transactions/src/__tests__/durable-nonce-test.ts
@@ -1,7 +1,7 @@
 import 'test-matchers/toBeFrozenObject';
 
 import { Base58EncodedAddress } from '@solana/addresses';
-import { AccountRole, ReadonlySignerAccount, WritableAccount } from '@solana/instructions';
+import { AccountRole, IInstruction, ReadonlySignerAccount, WritableAccount } from '@solana/instructions';
 
 import { Blockhash, ITransactionWithBlockhashLifetime } from '../blockhash';
 import {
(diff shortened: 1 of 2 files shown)


mcintyre94:
* master
    * **PR #1777** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1777" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1777?utm_source=stack-comment).

buffalojoec (approved):
Lgtm mate!

packages/transactions/src/durable-nonce.ts
```diff
@@ -101,7 +101,8 @@ export function isAdvanceNonceAccountInstruction(
         instruction.accounts[1].role === AccountRole.READONLY &&
         // Third account is nonce authority account
         instruction.accounts[2].address != null &&
-        instruction.accounts[2].role === AccountRole.READONLY_SIGNER
+        (instruction.accounts[2].role === AccountRole.READONLY_SIGNER ||
+            instruction.accounts[2].role === AccountRole.WRITABLE_SIGNER)
```
steveluscher, line 105: You can use `isSignerRole()` here if you feel like it.

packages/transactions/src/__tests__/durable-nonce-test.ts
```diff
@@ -129,6 +129,29 @@ describe('assertIsDurableNonceTransaction()', () => {
             assertIsDurableNonceTransaction({ ...durableNonceTx });
         }).not.toThrow();
     });
+    it('does not throw when the nonce authority is a writeable signer', () => {
```
steveluscher, line 132:
```suggestion
    it('does not throw when the nonce authority is a writable signer', () => {
```

steveluscher (changes requested):
> This is the case if the nonce authority is also the fee payer, for example in a transaction created with the CLI:

Hang on though; the way that transactions are compiled in the new web3.js is such that the `AccountRole` on any individual instruction has no bearing alone on what the `AccountRole` will be at the level of the entire transaction.

If the advance nonce instruction specifies account 'abc' as `READONLY_SIGNER` and the transaction fee payer is set to account 'abc', the transaction compiler will combine all of the `AccountRoles` of 'abc' in the transaction and compute the highest role (ie. `WRITABLE_SIGNER` in that case).

https://github.com/solana-labs/solana-web3.js/blob/df459651e9535bd3042807f10956329682e8c7b0/packages/instructions/src/roles.ts#L38-L50

It should be fine, therefore, to type the `AdvanceNonce` instruction with the _lowest_ `AccountRole` necessary to advance the nonce, and let the transaction compiler decide the ultimate role for every account at the transaction level.

packages/transactions/src/__tests__/durable-nonce-test.ts
mcintyre94: I don't even have the excuse of saying that's an English spelling!

mcintyre94:
Discussed on Slack, but just so it's here too - I forgot to mention that this is related to decompiling a transaction.

In a compiled transaction the account has had its roles merged, and when we apply the merged role to the instruction it may be different to what it originally had in that instruction (eg in this case the read-only signer is now a writable signer because of something else making it writable)

So basically it's something else that's lossy when we compile. We could special case the instruction on decompile but I think that would lead to more complexity.

PR #1779, opened 2023-10-24
refactor(experimental): fix bug in setTransactionLifetimeUsingDurableNonce

mcintyre94:
When the input transaction already has a durable nonce instruction, but
does not yet have the durable nonce lifetime constraint, we should not
add an additional identical instruction. We should just add the lifetime
constraint and leave the instructions alone.


mcintyre94:
* master
    * **PR #1777** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1777" target="_blank">Graphite</a> 
      * **PR #1779** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1779" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1779?utm_source=stack-comment).

packages/transactions/src/durable-nonce.ts
```diff
@@ -146,12 +165,38 @@ export function setTransactionLifetimeUsingDurableNonce<
...
+
+    // While the transaction is not already a durable nonce transaction, it may still have a durable nonce instruction
+    let newInstructions: [
+        AdvanceNonceAccountInstruction<TNonceAccountAddress, TNonceAuthorityAddress>,
+        ...IInstruction[]
+    ];
+    const firstInstruction = transaction.instructions[0];
+    if (!firstInstruction) {
+        newInstructions = [createAdvanceNonceAccountInstruction(nonceAccountAddress, nonceAuthorityAddress)];
+    } else if (isAdvanceNonceAccountInstruction(firstInstruction)) {
```
buffalojoec, line 177: You might be able to slide this handy `isAdvanceNonceAccountInstruction(..)` check a bit upward to check the initial transaction as well, since it's checking the first instruction already.

buffalojoec (commented):
Nice catch!

packages/transactions/src/durable-nonce.ts
mcintyre94: It's a bit tricky because it's part of the `isDurableNonceTransaction` check at the top, and we're in the branch where that was false. But we don't know if it's false because `isAdvanceNonceAccountInstruction` is false or because of something else, so I think we have to check it ourselves.

In the edge case I'm handling in this PR, you won't get to `isAdvanceNonceAccountInstruction` in the `isDurableNonceTransaction` code because `'lifetimeConstraint' in transaction` is false so it'll exit there.

buffalojoec: Oh yeah I meant the conditional on line 139. It looks like it does half the checks that `isAdvanceDurableNonceInstruction` does.

But it was already there I guess before this PR, so I'll just leave that up to you!

And yeah makes sense, that role check is still `READONLY_SIGNER` in this branch. You stackin' bro?

mcintyre94: Ah I see what you mean, we could potentially reverse the checks a bit and only do the `isDurableNonceTransaction` if `isAdvanceNonceAccountInstruction` is true, since it's a prerequisite for that. I'll have a look tomorrow and see if rejigging it a bit makes sense.

packages/transactions/src/__tests__/durable-nonce-test.ts
```diff
@@ -183,6 +183,32 @@ describe('setTransactionLifetimeUsingDurableNonce', () => {
             baseTx.instructions[0],
         ]);
     });
+    describe('given a transaction with an advance nonce account instruction but no nonce lifetime constraint', () => {
```
steveluscher, line 186: <3 the completionism.

steveluscher (approved):

packages/transactions/src/durable-nonce.ts
mcintyre94: Okay that's way better! I unrolled it, so it now checks if the first instruction is advance nonce, then if it's this exact one, then if it's already a durable nonce with this constraint. I think that makes it much clearer what the paths are here and how we want to update the instructions in each case.

Issue #1740, opened 2023-10-18
[experimental] Discussion: Extending an RPC with custom methods


steveluscher:
## Motivation

We've written a comprehensive typespec for the Solana JSON-RPC in `rpc-core`. If you're using an RPC that offers _more_ than the methods we've specified, you might like to extend your rpc object to include those extra methods.

## Example use case

Maybe you use QuickNode and you want to make use of the [qn_fetchNFTCollectionDetails_v2](https://www.quicknode.com/docs/ethereum/qn_fetchNFTCollectionDetails_v2) method. You can easily mix a typespec for `qn_fetchNFTCollectionDetails_v2()` into the existing Solana ones, just like we do here:

https://github.com/solana-labs/solana-web3.js/blob/fc4e94336e635b203374dd40cc62ed8248774115/packages/rpc-core/src/rpc-subscriptions/index.ts#L80-L86

## Problem

While our current implementation lets you:

* augment the TypeScript types with custom methods
* add pre/post processors to the inputs/results by supplying a concrete `api` implementation for your custom method (eg. you could call the QuickNode method above `fetchNFTCollectionDetails()` and have a concrete implementation substitute the JSON-RPC method name `qn_fetchNFTCollectionDetails_v2`).

…what we _can't_ do with the current implementation is to customize anything about the transport. This is a problem for, for instance, Helius. Helius doesn't implement custom APIs via proprietary JSON-RPC methods, but rather offers completely different endpoints for those APIs (eg. `https://api.helius.xyz/v0/addresses/<address>/balances?api-key=<your-key>`).

## Question

Does this matter? Is it a code smell to have `IRpcApi` know anything about transports? If an API is offered at a different endpoint should it by definition require that you create a new rpc object with a new `RpcTransport`?

buffalojoec:
Our spec still requires your alternate JSON RPC endpoint to adhere to this payload structure though, right?

```
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "getAccountInfo",
  "params": [
    "vines1vzrYbzLMRdu58ou5XTby4qAqVRLmqo36NKPTg",
    {
      "encoding": "base58"
    }
  ]
}
```

We're crafting this underneath `createJsonRpc(..)`:

https://github.com/solana-labs/solana-web3.js/blob/c7ef49cc49ee61422a4777d439a814160f6d7ce4/packages/rpc-transport/src/json-rpc.ts#L18

[...]

PR #1781, opened 2023-10-25
refactor(experimental): add generic `createJsonRpcApi` function for custom APIs

buffalojoec:
This PR is a concept for now. It serves to potentially address the discussion raised in #1740.

The goal of the collective `@solana/rpc-*` packages has been to allow complete customization, as long as their API adheres to the [official JSON RPC spec](https://www.jsonrpc.org/specification).

However, I couldn't find the best way someone would _actually_ go about doing this with our current implementation. Maybe I'm missing something.

In the actual `library` package, we're providing what's been dubbed the "default" Solana API. The function `createSolanaRpc(..)` allows you to provide your own transport, but it automatically uses the Solana RPC API, defined by `@solana/rpc-core` for the client API.

https://github.com/solana-labs/solana-web3.js/blob/589c379aa0ab495f50dac873f06573f52cdc9f98/packages/library/src/rpc.ts#L19-L24

If one wants to create their own RPC client manually, they can use the following code, comprised of `@solana/rpc-transport` and `@solana/rpc-core`, _not_ the main library.

```typescript
const api = createSolanaRpcApi();
    // ^ IRpcApi<SolanaRpcMethods>

const transport = createHttpTransport({ url: 'http://127.0.0.1:8899' });

const rpc = createJsonRpc<SolanaRpcMethods>({ api, transport });
    // ^ RpcMethods<SolanaRpcMethods>
```

You can see you can choose to define your API and provide it as a parameter to `createJsonRpc(..)`, however unless I'm missing something in our code, there's no generic API-creator.

This PR attempts to roll that generic API-creator as `createJsonRpcApi(..)`.

I envision this function being extremely useful for projects who wish to define their own `rpc-core` type-spec via interfaces, as we have with `SolanaRpcMethods`, and simply create it like so:

```typescript
// Define the method's response payload
type NftCollectionDetailsApiResponse = Readonly<{
    address: string;
    circulatingSupply: number;
    description: string;
    erc721: boolean;
    erc1155: boolean;
    genesisBlock: string;
    genesisTransaction: string;
    name: string;
    totalSupply: number;
}>;

// Set up an interface for the request method
interface NftCollectionDetailsApi {
    // Define the method's name, parameters and response type
    qn_fetchNFTCollectionDetails(args: { contracts: string[] }): NftCollectionDetailsApiResponse;
}

// Export the type spec for downstream users
export type QuickNodeRpcMethods = NftCollectionDetailsApi;

// Create the custom API
const api = createJsonRpcApi<QuickNodeRpcMethods>();

// Set up an HTTP transport
const transport = createHttpTransport({ url: 'http://127.0.0.1:8899' });

// Create the RPC client
const quickNodeRpc = createJsonRpc<QuickNodeRpcMethods>({ api, transport });
       // ^ RpcMethods<QuickNodeRpcMethods>
```

Of course you could _also_ combine your type spec with Solana's:

```typescript
export type QuickNodeSolanaRpcMethods = SolanaRpcMethods & NftCollectionDetailsApi;
```

Let me know any thoughts.

### DIFF
--- a/packages/rpc-transport/src/apis/methods/methods-api.ts
+++ b/packages/rpc-transport/src/apis/methods/methods-api.ts
@@ -1,0 +1,36 @@
+import { IRpcApi, RpcRequest } from '../../json-rpc-types';
+import { IRpcApiMethods, RpcApiConfig } from '../api-types';
+
+export function createJsonRpcApi<TRpcMethods extends IRpcApiMethods>(config?: RpcApiConfig): IRpcApi<TRpcMethods> {
+    return new Proxy({} as IRpcApi<TRpcMethods>, {
+        defineProperty() {
+            return false;
+        },
+        deleteProperty() {
+            return false;
+        },
+        get<TMethodName extends keyof IRpcApi<TRpcMethods>>(
+            ...args: Parameters<NonNullable<ProxyHandler<IRpcApi<TRpcMethods>>['get']>>
+        ) {
+            const [_, p] = args;
+            const methodName = p.toString() as keyof TRpcMethods as string;
+            return function (
+                ...rawParams: Parameters<
+                    TRpcMethods[TMethodName] extends CallableFunction ? TRpcMethods[TMethodName] : never
+                >
+            ): RpcRequest<ReturnType<TRpcMethods[TMethodName]>> {
+                const params = config?.parametersTransformer
+                    ? config?.parametersTransformer(rawParams, methodName)
+                    : rawParams;
+                const responseTransformer = config?.responseTransformer
+                    ? config?.responseTransformer<ReturnType<TRpcMethods[TMethodName]>>
+                    : (rawResponse: unknown) => rawResponse as ReturnType<TRpcMethods[TMethodName]>;
+                return {
+                    methodName,
+                    params,
+                    responseTransformer,
+                };
+            };
+        },
+    });
+}
--- a/packages/rpc-transport/src/json-rpc.ts
+++ b/packages/rpc-transport/src/json-rpc.ts
@@ -23,7 +23,9 @@ function createPendingRpcRequest<TRpcMethods, TResponse>(
             if ('error' in response) {
                 throw new SolanaJsonRpcError(response.error);
             } else {
-                return (responseTransformer ? responseTransformer(response.result) : response.result) as TResponse;
+                return (
+                    responseTransformer ? responseTransformer(response.result, methodName) : response.result
+                ) as TResponse;
             }
         },
     };
--- a/packages/rpc-core/src/rpc-methods/index.ts
+++ b/packages/rpc-core/src/rpc-methods/index.ts
@@ -1,4 +1,5 @@
-import type { IRpcApi, RpcRequest } from '@solana/rpc-transport';
+import { createJsonRpcApi } from '@solana/rpc-transport';
+import { IRpcApi } from '@solana/rpc-transport';
 
 import { patchParamsForSolanaLabsRpc } from '../params-patcher';
 import { patchResponseForSolanaLabsRpc } from '../response-patcher';
@@ -115,37 +116,17 @@ export type SolanaRpcMethods = GetAccountInfoApi &
     SimulateTransactionApi;
 
 export function createSolanaRpcApi(config?: Config): IRpcApi<SolanaRpcMethods> {
-    return new Proxy({} as IRpcApi<SolanaRpcMethods>, {
-        defineProperty() {
-            return false;
-        },
-        deleteProperty() {
-            return false;
-        },
-        get<TMethodName extends keyof IRpcApi<SolanaRpcMethods>>(
-            ...args: Parameters<NonNullable<ProxyHandler<IRpcApi<SolanaRpcMethods>>['get']>>
-        ) {
-            const [_, p] = args;
-            const methodName = p.toString() as keyof SolanaRpcMethods as string;
-            return function (
-                ...rawParams: Parameters<
-                    SolanaRpcMethods[TMethodName] extends CallableFunction ? SolanaRpcMethods[TMethodName] : never
-                >
-            ): RpcRequest<ReturnType<SolanaRpcMethods[TMethodName]>> {
-                const handleIntegerOverflow = config?.onIntegerOverflow;
-                const params = patchParamsForSolanaLabsRpc(
-                    rawParams,
-                    handleIntegerOverflow
-                        ? (keyPath, value) => handleIntegerOverflow(methodName, keyPath, value)
-                        : undefined,
-                );
-                return {
-                    methodName,
-                    params,
-                    responseTransformer: rawResponse => patchResponseForSolanaLabsRpc(rawResponse, methodName),
-                };
-            };
-        },
+    const handleIntegerOverflow = config?.onIntegerOverflow;
+    return createJsonRpcApi<SolanaRpcMethods>({
+        parametersTransformer: <T>(rawParams: T, methodName: string) =>
+            patchParamsForSolanaLabsRpc(
+                rawParams,
+                handleIntegerOverflow
+                    ? (keyPath, value) => handleIntegerOverflow(methodName, keyPath, value)
+                    : undefined,
+            ) as unknown[],
+        responseTransformer: <T>(rawResponse: unknown, methodName: string): T =>
+            patchResponseForSolanaLabsRpc(rawResponse, methodName as keyof SolanaRpcMethods),
     });
 }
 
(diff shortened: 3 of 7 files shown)


buffalojoec:
* master
  * **PR #1781** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1781" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1781?utm_source=stack-comment).

steveluscher (changes requested):
Hmm… interesting.

Backing way up, the point of the `IRpcApi` type is to say: ‘hey, if you wanna define an API for some JSON RPC, then give me something that conforms to this type’

```ts
(...params: WhateverParams[]) => {
    methodName: string;
    params: unknown[];
    responseProcessor?: (response: unknown) => WhateverResponse;
}
```

_One_ way of doing that is the fancy `Proxy` solution that I wrote for the Solana JSON-RPC API. That proxy solution works for Solana because all of the methods are conveniently named the same as one would like the JavaScript functions to be, and the params are mostly pass-throughs.

I designed the `IRpcApi` such that you could write a helper that produced it in some _other_ way, though. In QuickNode's case, for instance, it's very unlikely that they will want to mirror the method names as JavaScript function names.

```ts
// I doubt they will want to release this as a JavaScript API
await rpc.qn_fetchNFTCollectionDetails(...).send();

// It's more likely they'll want this
await rpc.fetchNFTCollectionDetails(...).send();
```

They could achieve that by releasing this:

```ts
function createQuickNodeRpcApi(): IRpcApi<QuickNodeRpcMethods> {
  return {
    fetchNFTCollectionDetails(...params) {
      return {
        methodName: 'qn_fetchNFTCollectionDetails',
        params,
      }
    },
  };
}
```

That conforms to the `IRpcApi` type, and they're good. That's not a _great_ design for an API creator (QuickNode, please don't do that) but it works.

The API creator in this PR adds a layer of indirection without actually letting QuickNode decouple the name of the JavaScript functions and the name of the JSON-RPC methods, so it's handy but maybe not useful.

We could solve this in two ways:

* do nothing and let people make their own API creators (fork ours if they like)
* double down on the ‘API creator in the proxy style’ helper you have here and let people transform the method names too.

buffalojoec:
> We could solve this in two ways:
> 
> * do nothing and let people make their own API creators (fork ours if they like)
> * double down on the ‘API creator in the proxy style’ helper you have here and let people transform the method names too.

Ok thanks for that insight.

My thinking was that our "here's my API's type spec, now make me an RPC API that conforms to the JSON RPC spec with them" approach was pretty slick. Then people can just fork `@solana/rpc-core` instead, replace our method types with their own, and plug them all in. IMO that's much easier.

One might think it introduces the idea of it being more restrictive, but if we split out `createJsonRpcApi(..)` into a generalized API-builder, anyone who _doesn't_ want to use the fancy `Proxy` setup can just... not. They could do the same thing you've described.

So I guess it comes down to: Should we be nice and de-couple the fancy`Proxy` from our API?

In my opinion, we have absolutely nothing to lose by doing so and we give developers better tooling for creating custom APIs, which is cool.

buffalojoec:
> Backing way up, the point of the `IRpcApi` type is to say: ‘hey, if you wanna define an API for some JSON RPC, then give me something that conforms to this type’
> 
> ```ts
> (...params: WhateverParams[]) => {
>     methodName: string;
>     params: unknown[];
>     responseProcessor?: (response: unknown) => WhateverResponse;
> }
> ```

I guess one follow-up question is how difficult is it to build an API spec with this _the right way_, and not use the `Proxy` stuff?

steveluscher:
Yeah! If we do offer it then we just need to add a method name transformer to the config, so that the method that gets called on the proxy can be different than the method that gets called on the actual RPC. Oh, and then you have to do this all over again for subscriptions, in case folks have custom subscription methods too.

> This example seems pretty simple tbh haha.

Yeah, the bad part of the simple example is ‘grows linearly with the number of methods’ whereas the `Proxy` solution stays the same size with an infinite number of methods.

buffalojoec:
> Oh, and then you have to do this all over again for subscriptions, in case folks have custom subscription methods too.

Yes I anticipated this. 🥞

> Yeah! If we do offer it then we just need to add a method name transformer to the config, so that the method that gets called on the proxy can be different than the method that gets called on the actual RPC.

I think this is valuable so I'm going to add it.

packages/rpc-transport/src/apis/methods/methods-api.ts
```diff
@@ -0,0 +1,36 @@
...
+                const params = config?.parametersTransformer
+                    ? config?.parametersTransformer(rawParams, methodName)
+                    : rawParams;
+                const responseProcessor = config?.responseTransformer
+                    ? config?.responseTransformer<ReturnType<TRpcMethods[TMethodName]>>
+                    : (rawResponse: unknown) => rawResponse as ReturnType<TRpcMethods[TMethodName]>;
+                return {
+                    methodName,
+                    params,
+                    responseProcessor,
```
steveluscher, line 31: I mean feel free to rename it, all the way down.

Also, in the READMEs.

packages/rpc-transport/src/json-rpc.ts
```diff
@@ -23,7 +23,9 @@ function createPendingRpcRequest<TRpcMethods, TResponse>(
             if ('error' in response) {
                 throw new SolanaJsonRpcError(response.error);
             } else {
-                return (responseTransformer ? responseTransformer(response.result) : response.result) as TResponse;
+                return (
+                    responseTransformer ? responseTransformer(response.result, methodName) : response.result
```
lorisleiva, line 27: - Why wouldn't we provide the entire response to the transformer?
- Would it be possible to wrap the method name, the payload and the URL in a request object so that the entire request can also be provided to the transformer?

buffalojoec: > * Why wouldn't we provide the entire response to the transformer?

We can, but I think we're only shedding `jsonrpc` and `id`.
Check out a [response payload](https://docs.solana.com/api/http#getaccountinfo) in the docs.

> * Would it be possible to wrap the method name, the payload and the URL in a request object so that the entire request can also be provided to the transformer?

I believe so, yes!

lorisleiva: The reason I'm saying all this is I think this should be comparable to [Axios interceptors](https://axios-http.com/docs/interceptors) (or equivalent) where we have no opinion over what could be transformed and we're just like "here's a request I'm about to send, do whatever you want with it" and same with the response. 🙂

buffalojoec: So, in my opinion I think `createJsonRpcApi(..)` should be designed to create APIs that adhere to the [JSON RPC spec](https://www.jsonrpc.org/specification), without deviation. So, to me, it makes sense to leave part of the request and response unchanged, since modifying `request.parameters` or `response.result` still adheres to the JSON RPC interface.

It seems like you probably have the Helius example in mind from #1740 but here, I think this function would _not_ be used to implement the custom transport you referenced there. What do you think?

lorisleiva: Yeah that's a fair point actually. So you're allowing the library to transform anything that can be transformed without affecting the JSON RPC specs. I like it!

Am I right to say that, in order to support a `urlTransformer` on the API side, we would need something like a `requestTransformer` on the Transport side? Similarly to the `responseTransformer`, the `requestTransformer` would only allow you to edit the components of the requests that don't affect the JSON RPC specs. Namely, the parameters and the URL.

buffalojoec: > Am I right to say that, in order to support a `urlTransformer` on the API side, we would need something like a `requestTransformer` on the Transport side? Similarly to the `responseTransformer`, the `requestTransformer` would only allow you to edit the components of the requests that don't affect the JSON RPC specs. Namely, the parameters and the URL.

We _could_ do this in theory, but I can't decide if I think it's something we _should_ support.

Are you thinking you'd be able to plug in a `requestTransformer` that, for example, takes all of the parameters and the method name from the POST body and maps them to URL parameters?

As an aside, I'm also not sure if the Helius endpoints for example are GETs instead of POSTs.

lorisleiva: Yeah that's what I'm thinking. I guess it all depends what is technically valid in the JSON RPC specs. Is it okay to use dynamic URLs? Is it okay to use different HTTP methods? If not, then maybe we shouldn't mess with it. If it is valid according to the specs though, then I strongly believe we should offer that level of flexibility.

buffalojoec: I can't see anything in the spec that says definitively one way or the other. But it does _seem_ like the intention is a request body.

steveluscher (approved):

buffalojoec:
### Merge activity

* **Dec 11, 8:31 PM**: @@buffalojoec started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1781).

PR #1789, opened 2023-10-27
refactor(experimental): optimise signTransaction to only encode the message once

### LINKED ISSUE #1788: [experimental] Performance issue on sign-heavy workloads
## Overview

I've been doing some performance testing of the new web3js on a sign-heavy workload. An example is create N signers, and then sign a transaction with all N signers M times.

At the extreme, there's a small performance regression between old web3js and new web3js:

Old:
```
Signed 100 transactions with 30 signers in 988 milliseconds
node index.mjs 30 100  1.38s user 0.07s system 106% cpu 1.361 total
```

New: 
```
Signed 100 transactions with 30 signers in 1251 milliseconds
node index.mjs 30 100  1.73s user 0.07s system 103% cpu 1.735 total
```

(the time in the first line is just the signing (measured with `performance.now` inside the script), the second line is the output of `time` when running the script, so it has a longer total time since it includes eg generating the keypairs)

With fewer signers, we're faster:

Old:
```
Signed 100 transactions with 5 signers in 184 milliseconds
node index.mjs 5 100  0.50s user 0.03s system 75% cpu 0.712 total
```
New:
```
Signed 100 transactions with 5 signers in 89 milliseconds
node index.mjs 5 100  0.33s user 0.04s system 44% cpu 0.804 total
```

This made me suspect that there's a performance issue in our logic for signing. It shouldn't take >10x longer to sign with 6x more signers per transaction.

The issue is here: https://github.com/solana-labs/solana-web3.js/blob/6910664bb9dd1e6b24c57a321b05fe49cc6815d8/packages/transactions/src/signatures.ts#L94-L101

For each keypair, we are calling `getAddressFromPublicKey(keyPair.publicKey)` and `getCompiledMessageSignature(compiledMessage, keyPair.privateKey)`

In `getCompiledMessageSignature` we encode the message:
https://github.com/solana-labs/solana-web3.js/blob/6910664bb9dd1e6b24c57a321b05fe49cc6815d8/packages/transactions/src/signatures.ts#L66-L70

This is the issue: we're re-encoding the transaction message for every signer every loop. 

If we modify this to call the `getCompiledMessageEncoder().encode(message)` in `signTransaction` only once, we see a massive performance improvement:

```
Signed 100 transactions with 30 signers in 120 milliseconds
node index.mjs 30 100  0.48s user 0.08s system 77% cpu 0.720 total
```

```
Signed 100 transactions with 5 signers in 44 milliseconds
node index.mjs 5 100  0.25s user 0.03s system 38% cpu 0.747 total
```

TLDR: We should refactor so that when you call `signTransaction` the `getCompiledMessageEncoder().encode(message)` call is only made once. This will unlock a massive performance improvement for sign-heavy workloads.

mcintyre94:
This greatly improves performance when signing a transaction with multiple keypairs

Fixes #1788

### DIFF
--- a/packages/transactions/src/signatures.ts
+++ b/packages/transactions/src/signatures.ts
@@ -4,7 +4,7 @@ import { isSignerRole } from '@solana/instructions';
 import { Ed25519Signature, signBytes } from '@solana/keys';
 
 import { ITransactionWithFeePayer } from './fee-payer';
-import { CompiledMessage, compileMessage } from './message';
+import { compileMessage } from './message';
 import { getCompiledMessageEncoder } from './serializers/message';
 
 export interface IFullySignedTransaction extends ITransactionWithSignatures {
@@ -63,12 +63,6 @@ export function isTransactionSignature(
     return true;
 }
 
-async function getCompiledMessageSignature(message: CompiledMessage, secretKey: CryptoKey) {
-    const wireMessageBytes = getCompiledMessageEncoder().encode(message);
-    const signature = await signBytes(secretKey, wireMessageBytes);
-    return signature;
-}
-
 export function getSignatureFromTransaction(
     transaction: ITransactionWithFeePayer & ITransactionWithSignatures
 ): TransactionSignature {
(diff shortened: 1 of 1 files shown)


mcintyre94:
* master
  * **PR #1789** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1789" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1789?utm_source=stack-comment).

lorisleiva:
Should we just remove the `getCompiledMessageSignature` helper since it's now just calling the `signBytes` function?

mcintyre94:
Oops yep, updated! Thanks!

lorisleiva (approved):

buffalojoec (approved):
Nice! This is a good performance enhancement.

steveluscher (commented):
giphy

PR #1790, opened 2023-10-27
refactor(experimental): parameterize account metas for instructions and instructions for transactions

lorisleiva:
This PR makes add the following type parameters:
- `TAccounts` on `IInstruction` so we can refine the account metas type without asserting that the accounts are provided (via the existing `IInstructionWithAccounts`). This is then being used to re-define the `LegacyTransaction` such that it excludes `IAccountLookupMeta` types.
- `TInstruction` on transaction types (i.e. `BaseTransaction`, `LegacyTransaction`, `V0Transaction` and `CompilableTransaction`). This enables functions to refine the instruction type that they support within a transaction. That way, we can create a `ILegacyInstruction` type for the `LegacyTransaction`.
- `TVersion` on transaction types to make it easier to define the `LegacyTransaction` and `V0Transaction` types.

On top of the nice refactoring we get, this PR allows the `signers` package to require the following type on its signing functions (notice the `IAccountSignerMeta` on the instruction meta that we now support):

```ts
type CompilableTransactionWithSignerMetas = CompilableTransaction<
    TransactionVersion,
    IInstruction<string, readonly (IAccountMeta | IAccountSignerMeta | IAccountLookupMeta)[]>
>;
```

PR #1784, opened 2023-10-26
refactor(experimental): export CompilableTransaction and CompiledTransaction types

lorisleiva:
This PR exports the `CompilableTransaction` and `CompiledTransaction` types as they define important Transaction states and will be required by the `signers` package.

PR #1782, opened 2023-10-25
refactor(experimental): add transaction decoder + codec

mcintyre94:
This PR adds `getTransactionDecoder`, which deserializes a compiled transaction and then decompiles it

It also adds `getTransactionCodec`, combining the existing encoder and this new decoder

This gives us the ability to deserialize from transaction bytes to a new transaction

PR #1780, opened 2023-10-25
refactor(experimental): add tests for decompiling durable nonce transaction

mcintyre94:
This PR adds unit tests for decompiling a durable nonce transaction

PR #1792, opened 2023-10-27
refactor(experimental): add IAccountSignerMeta type and signing helpers

lorisleiva:
This PR introduces a new account meta type `IAccountSignerMeta` which allows us to store `TransactionSigners` inside instructions and therefore inside transactions.

It then provides helper methods to:
- Extract signers from instructions and transactions.
- Deduplicate signers by addresses.
- Sign transactions (partially or fully) using signers extracted from the account metas (and an optional additional array of signers).
- Sign and send transactions using `TransactionSendingSigners` (also using extracted and manually provided signers).

Below is a copy/paste of the added documentation for more details.

---

## Storing transaction signers inside instruction account metas

This package defines an alternative definition for account metas that allows us to store `TransactionSigners` inside them. This means each instruction can keep track of its own set of signers and, by extension, so can transactions.

It also provides helper functions that deduplicate and extract signers from instructions and transactions which makes it possible to sign an entire transaction automatically as we will see in the next section.

#### `IAccountSignerMeta`

Alternative `IAccountMeta` definition for signer accounts that allows us to store `TransactionSigners` inside it.

```ts
const mySignerMeta: IAccountSignerMeta = {
    address: myTransactionSigner.address,
    role: AccountRole.READONLY_SIGNER,
    signer: myTransactionSigner,
};
```

#### `IInstructionWithSigners`

Extends the `IInstruction` type to allow `IAccountSignerMetas` to be used inside the instruction's `accounts` array.

```ts
const myInstructionWithSigners: IInstructionWithSigners = {
    programAddress: address('1234..5678'),
    accounts: [
        {
            address: myTransactionSigner.address,
            role: AccountRole.READONLY_SIGNER,
            signer: myTransactionSigner,
        },
    ],
};
```

#### `deduplicateSigners()`

Removes duplicated signers by comparing their addresses.

```ts
const mySignerA = { address: address('1111..1111'), signTransactions: async () => {} };
const mySignerB = { address: address('2222..2222'), signTransactions: async () => {} };
const mySignerC = { address: address('1111..1111'), signTransactions: async () => {} };
const myDeduplicatedSigners = deduplicateSigners([mySignerA, mySignerB, mySignerC]);
// ^ [mySignerA, mySignerB]
```

#### `getSignersFromInstruction()`

Extracts and deduplicates all signers stored inside the account metas of an instruction.

```ts
const mySignerA = { address: address('1111..1111'), signTransactions: async () => {} };
const mySignerB = { address: address('2222..2222'), signTransactions: async () => {} };
const mySignerC = { address: address('1111..1111'), signTransactions: async () => {} };
const myInstructionWithSigners: IInstructionWithSigners = {
    programAddress: address('1234..5678'),
    accounts: [
        { address: mySignerA.address, role: AccountRole.READONLY_SIGNER, signer: mySignerA },
        { address: mySignerB.address, role: AccountRole.WRITABLE_SIGNER, signer: mySignerB },
        { address: mySignerC.address, role: AccountRole.WRITABLE_SIGNER, signer: mySignerC },
    ],
};

const instructionSigners = getSignersFromInstruction(myInstructionWithSigners);
// ^ [mySignerA, mySignerB]
```

#### `getSignersFromTransaction()`

Similarly to `getSignersFromInstruction`, this function extracts and deduplicates all signers stored inside the account metas of all the instructions inside a transaction.

```ts
const transactionSigners = getSignersFromTransaction(myTransactionWithSigners);
```

## Signing transactions with signers

As we've seen in the previous section, we can store and extract `TransactionSigners` from instructions and transactions. This allows us to provide helper methods that sign transactions using the signers stored inside them.

#### `partiallySignTransactionWithSigners()`

Extracts all signers inside the provided transaction and uses them to sign it. It first uses all `TransactionModifyingSigners` sequentially before using all `TransactionPartialSigners` in parallel.

If a composite signer implements both interfaces, it will be used as a modifying signer if no other signer implements that interface. Otherwise, it will be used as a partial signer.

```ts
const mySignedTransaction = partiallySignTransactionWithSigners(myTransaction);
```

It also accepts an additional array of signers that will be merged with the ones extracted from the transaction, if any.

```ts
const mySignedTransaction = partiallySignTransactionWithSigners(myTransaction, [myOtherSigner]);
```

Finally, note that this function ignores `TransactionSendingSigners` as it does not send the transaction. See the `signAndSendTransactionWithSigners` function below for more details on how to use sending signers.

#### `signTransactionWithSigners()`

This function works the same as the `partiallySignTransactionWithSigners` function described above except that it also ensures the transaction is fully signed before returning it. An error will be thrown if that's not the case.

```ts
const mySignedTransaction = signTransactionWithSigners(myTransaction);

// With additional signers.
const mySignedTransaction = signTransactionWithSigners(myTransaction, [myOtherSigner]);

// We now know the transaction is fully signed.
mySignedTransaction satisfies IFullySignedTransaction;
```

#### `signAndSendTransactionWithSigners()`

Extracts all signers inside the provided transaction and uses them to sign it before sending it immediately to the blockchain. It returns the signature of the sent transaction (i.e. its identifier).

```ts
const myTransactionSignature = signAndSendTransactionWithSigners(myTransaction);

// With additional signers.
const myTransactionSignature = signAndSendTransactionWithSigners(myTransaction, [myOtherSigner]);
```

Similarly to the `partiallySignTransactionWithSigners` function, it first uses all `TransactionModifyingSigners` sequentially before using all `TransactionPartialSigners` in parallel. It then sends the transaction using the first `TransactionSendingSigner` it finds. Any other sending signer that does not implement another transaction signer interface will be ignored.

If no `TransactionSendingSigner` is extracted from the transaction or explicitly provided, the `fallbackSender` third argument will be used to send the transaction. If no fallback sender is provided and no sending signer is identified, an error will be thrown.

```ts
const fallbackSender = async (transaction: CompilableTransaction) => {
    const encodedTransaction = getBase64EncodedWireTransaction(transaction);
    const signature = await rpc.sendTransaction(encodedTransaction).send();
    return getBase58Encoder().encode(signature);
};

const myTransactionSignature = signAndSendTransactionWithSigners(myTransaction, [], fallbackSender);
```

Here as well, composite transaction signers are treated such that at least one sending signer is used if any. When a `TransactionSigner` implements more than one interface, use it as a:

-   `TransactionSendingSigner`, if no other `TransactionSendingSigner` exist.
-   `TransactionModifyingSigner`, if no other `TransactionModifyingSigner` exist.
-   `TransactionPartialSigner`, otherwise.


lorisleiva:
* master
    * **PR #1777** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1777" target="_blank">Graphite</a> 
      * **PR #1779** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1779" target="_blank">Graphite</a> 
        * **PR #1780** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1780" target="_blank">Graphite</a> 
          * **PR #1782** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1782" target="_blank">Graphite</a> 
            * **PR #1784** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1784" target="_blank">Graphite</a> 
              * **PR #1710** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1710" target="_blank">Graphite</a> 
                * **PR #1790** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1790" target="_blank">Graphite</a> 
                  * **PR #1792** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1792" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1792?utm_source=stack-comment).

packages/signers/README.md
```diff
@@ -359,10 +359,149 @@ const myNoopSigner = createNoopSigner(myAddress);
...
+#### `signAndSendTransactionWithSigners()`
+
+Extracts all signers inside the provided transaction and uses them to sign it before sending it immediately to the blockchain. It returns the signature of the sent transaction (i.e. its identifier).
+
+```ts
+const myTransactionSignature = signAndSendTransactionWithSigners(myTransaction);
+
+// With additional signers.
+const myTransactionSignature = signAndSendTransactionWithSigners(myTransaction, [myOtherSigner]);
+```
```
buffalojoec, line 487: Yeah, I like this API, it's clean!

packages/signers/src/sign-transaction.ts
```diff
@@ -0,0 +1,202 @@
...
+ * Signs and sends a transaction using any signers that may be stored in IAccountSignerMeta
+ * instruction accounts as well as any signers provided explicitly to this function.
+ * It will identify a single TransactionSendingSigners to use for sending the transaction, if any.
+ * Otherwise, it will send the transaction using the provided fallbackSender.
+ */
+export async function signAndSendTransactionWithSigners<
+    TTransaction extends CompilableTransactionWithSigners = CompilableTransactionWithSigners
+>(
+    transaction: TTransaction,
+    signers: readonly TransactionSigner[] = [],
```
buffalojoec, line 69: As per the documentation, wasn't this supposed to be optional?

packages/signers/src/sign-transaction.ts
```diff
@@ -0,0 +1,202 @@
...
+    fallbackSender?: (transaction: TTransaction & IFullySignedTransaction) => Promise<SignatureBytes>
```
buffalojoec, line 70: Should we opt for a "fallback" signer or a "default" signer? In other words, do I get to:
- provide the sender I want you to use if none of the signers are senders, or
- provide the sender I want you to use, period?

packages/signers/src/sign-transaction.ts
```diff
@@ -0,0 +1,202 @@
...
+    const sendingSigners = signers.filter(isTransactionSendingSigner);
+    if (sendingSigners.length === 0) return null;
+
+    // Prefer sending signers that do not offer other interfaces.
+    const sendingOnlySigners = sendingSigners.filter(
+        signer => !isTransactionModifyingSigner(signer) && !isTransactionPartialSigner(signer)
+    );
+    if (sendingOnlySigners.length > 0) {
+        return sendingOnlySigners[0];
+    }
```
buffalojoec, line 146: 🔥 🔥

packages/signers/src/sign-transaction.ts
```diff
@@ -0,0 +1,202 @@
...
+>(
+    transaction: TTransaction,
+    modifyingSigners: readonly TransactionModifyingSigner[] = [],
+    partialSigners: readonly TransactionPartialSigner[] = []
+): Promise<TTransaction & ITransactionWithSignatures> {
+    // Handle modifying signers sequentially.
+    const modifiedTransaction = await modifyingSigners.reduce(async (transaction, modifyingSigner) => {
+        const [tx] = await modifyingSigner.modifyAndSignTransactions([await transaction]);
+        return Object.freeze(tx);
+    }, Promise.resolve(transaction) as Promise<TTransaction>);
```
buffalojoec, line 183: I may be missing something, but how does this not void any signatures before the final `modifyAndSign`?

buffalojoec (commented):
This API looks great! I just want to understand how you managed multiple modifications with signatures.

packages/signers/src/sign-transaction.ts
lorisleiva: It is. See the ` = []` default value at the end. 🙂

packages/signers/src/sign-transaction.ts
lorisleiva: The first statement is true. It will only be used if no `TransactionSendingSigner` was provided.

packages/signers/src/sign-transaction.ts
buffalojoec: Ah, missed that!

packages/signers/src/sign-transaction.ts
lorisleiva: It may invalidate previous signatures but it's up to the signer itself to decide if they should modify when they encounter a message or a transaction that already contains signatures.

Here's a more detailed description of this from the internal document:

---

### Handling modifying conflicts

The first type of conflict occurs when multiple message or transaction modifying signers are provided.

For instance, say we have two `TransactionModifyingSigner` A and B for the same transaction. If A signs before B, B will potentially modify the transaction again which invalidates A’s signature. The same scenario occurs if B signs before A.

There are however two ways to get around these conflicts and, as such, we shouldn’t fail if we encounter this scenario.

- **The Signer may decide to not modify**. In our previous example, A or B could first check the state of the transaction before deciding if they want to modify the transaction or not. Typically, if other signatures exist on the given transaction, they will fallback to signing without modifying it.
- **The Signer may allow the app to decide**. A signer may implement both the `TransactionPartialSigner` and `TransactionModifyingSigner` interfaces (or equivalent for messages) and let the application decide which one they want to use for their specific use-cases. The tradeoff for the app would be that `TransactionPartialSigner` brings more flexibility and can be executed in parallel whereas `TransactionModifyingSigner` may bring additional features for the app.

Now, if none of these workarounds are applicable and both A and B *must* modify the transaction before signing it, then we should still not fail and let the second wallet itself throw an error explaining why it cannot successfully modify and sign the given transaction.

packages/signers/src/sign-transaction.ts
buffalojoec: I meant which would be more useful

packages/signers/src/sign-transaction.ts
```diff
@@ -0,0 +1,202 @@
...
+    modifyingSigners: readonly TransactionModifyingSigner[];
+    partialSigners: readonly TransactionPartialSigner[];
+    sendingSigner: TransactionSendingSigner | null;
+}> {
+    // Identify the unique sending signer that should be used.
+    const sendingSigner = identifySendingSigner ? identifyTransactionSendingSigner(signers) : null;
+
+    // Now, focus on the other signers.
+    // I.e. the modifying or partial signers that are not the identified sending signer.
+    // Note that any other sending only signers will be discarded.
```
steveluscher, line 117: Shouldn't we fatal if there are more than one sending signer? That would be extremely surprising to me for one of my senders to get chosen, for reasons that are inscrutable to me, and not the other(s).

packages/signers/src/sign-transaction.ts
```diff
@@ -0,0 +1,202 @@
...
+
+/** Identifies the best signer to use as a TransactionSendingSigner, if any */
+function identifyTransactionSendingSigner(signers: readonly TransactionSigner[]): TransactionSendingSigner | null {
+    // Ensure there are any TransactionSendingSigners in the first place.
+    const sendingSigners = signers.filter(isTransactionSendingSigner);
+    if (sendingSigners.length === 0) return null;
+
+    // Prefer sending signers that do not offer other interfaces.
+    const sendingOnlySigners = sendingSigners.filter(
+        signer => !isTransactionModifyingSigner(signer) && !isTransactionPartialSigner(signer)
```
steveluscher, line 142: Interesting choice. Wouldn't you expect that the most ‘potent’ signer be chosen (ie. the one that modifies _and_ sends)?

steveluscher (changes requested):
[infinitesunglasses](https://app.graphite.dev/macros/solana-labs) 

There's probably a bunch of stuff that I'm missing or not understanding, but in short my questions and concerns boil down to:

1. Choices of signers being circumstantially deterministic (eg. on the basis of the order they appear in the instructions)
2. The criteria, themselves, against which signers are chosen
3. The optionality of adopting this system (ie. the `fallbackSender` and the ‘extra’ signers)
4. The lack of a cancellation API when sending transactions
5. The lack of invariants against possibly bad combinations of signers (eg. having two sending signers)

packages/signers/src/index.ts
```diff
@@ -1,7 +1,10 @@
+export * from './account-signer-meta';
+export * from './deduplicate-signers';
```
steveluscher, line 2: y exported?

packages/signers/src/deduplicate-signers.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/** Removes all duplicated signers from a provided array by comparing their addresses. */
+export function deduplicateSigners<TSigner extends MessageSigner | TransactionSigner>(
+    signers: readonly TSigner[]
+): readonly TSigner[] {
+    const deduplicated: Record<Address, TSigner> = {};
+    signers.forEach(signer => {
+        if (!deduplicated[signer.address]) {
+            deduplicated[signer.address] = signer;
+        }
```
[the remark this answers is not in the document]
lorisleiva: We _could_ but I think this would be a premature optimisation.

Let me step back a bit.

The signing and sending flow implemented here is as follows:
1. Given a transaction with signers in its account metas (and optionally more signers provided manually), extract all these signers in an array.
2. Remove duplicates based on the address. **This is where you think we should have a strategy to figure out which ones to keep**.
3. From these deduplicates signers, categorise them. Composite signers (that implement more than one interface) should be chosen depending on what other signers can do. For instance, say Signer A is a `TransactionPartialSigner & TransactionModifyingSigner` and Signer B is a `TransactionModifyingSigner`. Then Signer A should be used as `TransactionPartialSigner` to [avoid modifying conflicts](https://github.com/solana-labs/solana-web3.js/pull/1792#discussion_r1393111926). However, if Signer B was a `TransactionPartialSigner`, then it would make more sense for Signer A to be used as a `TransactionModifyingSigner` as it would be the only one. **This is where I think it matters to have deterministic invariants**. As per our internal document, these invariants are:
    - `TransactionModifyingSigners` must be signed sequentially before any other type of signing.
    - `TransactionPartialSigners` must be signed in parallel.
    - Only one of the `TransactionSendingSigner` must be used last, if any.
    - Use provided fallback logic to send the transaction if no `TransactionSendingSigner` was provided.
    - When a `TransactionSigner` implements more than one interface, use it as a:
        - `TransactionSendingSigner`, if no other `TransactionSendingSigner` exist.
        - `TransactionModifyingSigner`, if no other `TransactionModifyingSigner` exist.
        - `TransactionPartialSigner`, otherwise.
4. Sign and send as per the invariants above.

Now, the reason why I think it is a premature optimisation to add complex invariants to point 2 is that, 99.9% of the time, we are only **filtering the same signer references**. Because it is likely that the same signer will be used multiple time throughout the transaction.

```ts
const mySigner: TransactionSigner = createMySigner();

const instructions = await Promise.all([
  transferSol({ source: mySigner, destination: alice, amount: sol(1.5) }),
  transferSol({ source: mySigner, destination: bob, amount: sol(1.5) }),
]);

const transaction = pipe(
  createTransaction({ version: 0 }),
  tx => appendTransactionInstructions(instructions, tx),
  // ...
);

const signers = getSignersFromTransaction(transaction);
// ^ [mySigner]
```

If we did have complex invariants when deduplicating signers we would be optimising for the following use-case:

"_I have two different signers for the same address and I want it to sign the first instruction differently than the second instruction._"

Which makes no sense at all.

```ts
const myWalletSigner = createWalletSigner(useWallet());
const myKeyPairSigner = generateKeyPairSigner();

const instructions = await Promise.all([
  transferSol({ source: myKeyPairSigner, destination: alice, amount: sol(1.5) }),
  transferSol({ source: myWalletSigner, destination: bob, amount: sol(1.5) }),
]);

const transaction = pipe(
  createTransaction({ version: 0 }),
  tx => appendTransactionInstructions(instructions, tx),
  // ...
);

const signers = getSignersFromTransaction(transaction);
// ^ ??? Why should we have complex invariant for such a non-sense scenario? Let's just pick one.
```

Now, if you decided to split these instructions into two transactions then, the "two signers for the same address" scenario makes more sense and the current implementation still works as expected.

```ts
const myWalletSigner = createWalletSigner(useWallet());
const myKeyPairSigner = generateKeyPairSigner();

const instructions = await Promise.all([
  transferSol({ source: myKeyPairSigner, destination: alice, amount: sol(1.5) }),
  transferSol({ source: myWalletSigner, destination: bob, amount: sol(1.5) }),
]);

const transactionA = pipe(
  createTransaction({ version: 0 }),
  tx => appendTransactionInstructions(instructions[0], tx),
  // ...
);

const transactionB = pipe(
  createTransaction({ version: 0 }),
  tx => appendTransactionInstructions(instructions[1], tx),
  // ...
);

const signersA = getSignersFromTransaction(transactionA);
// ^ [myKeyPairSigner]

const signersB = getSignersFromTransaction(transactionB);
// ^ [myWalletSigner]
```

packages/signers/src/__tests__/account-signer-meta-test.ts
```diff
@@ -0,0 +1,84 @@
...
+            createMockTransactionSendingSigner(addressA),
+        ];
+        const instruction = createMockInstructionWithSigners(signers);
+
+        // When we extract the signers from the instruction's account metas.
+        const extractedSigners = getSignersFromInstruction(instruction);
+
+        // Then we expect the extracted signers to be deduplicated.
+        expect(extractedSigners).toHaveLength(2);
+        expect(extractedSigners.map(signer => signer.address).sort()).toStrictEqual(['1111', '2222']);
```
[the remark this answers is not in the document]
lorisleiva: I actually wrote the test specifically so that it wouldn't matter _how_ the signer is deduplicated as long as there is only one left for each address.

See [my response here](https://github.com/solana-labs/solana-web3.js/pull/1792#discussion_r1394056836) and let me know if you still think I should be explicit about the method here.

packages/signers/src/__tests__/deduplicate-signers-test.ts
```diff
@@ -0,0 +1,45 @@
...
+describe('deduplicateSigners', () => {
+    it('removes duplicated signers by address', () => {
+        // Given a list of signers with some duplicates.
+        const addressA = '1111' as Address;
+        const addressB = '2222' as Address;
+        const signers = [
+            createMockTransactionPartialSigner(addressA),
+            createMockMessagePartialSigner(addressB),
+            createMockTransactionModifyingSigner(addressA),
+            createMockTransactionSendingSigner(addressA),
```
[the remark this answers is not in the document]
lorisleiva: Moving this conversation [here](https://github.com/solana-labs/solana-web3.js/pull/1792#discussion_r1394056836).

packages/signers/src/__tests__/sign-transaction-test.ts
```diff
@@ -0,0 +1,550 @@
...
+            '1111': '1111_signature',
+            '2222': '2222_signature',
+        });
+
+        // And the signers were called with the expected parameters.
+        expect(signerA.modifyAndSignTransactions).toHaveBeenCalledWith([transaction]);
+        expect(signerB.signTransactions).toHaveBeenCalledWith([modifiedTransaction]);
+    });
+
+    it('signs the transaction with extra signers', async () => {
```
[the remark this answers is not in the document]
lorisleiva: It enables anyone to use the Signer API without necessarily having constructed their transaction in a way that stores signers in the their account metas.

For instance, you're sending your partially signed transaction over the wire to your server. The server decodes the transaction and, of course, doesn't have signers attached to the account metas. Then the server can still use the same Signer API regardless like so.

```ts
await signTransactionWithSigners(decodedTransaction, [myServerSigner]);
```

Additionally, it allows more complex signing scenarios.

For instance, ephemeral signers.

```ts
// Some factory function offered by some walllet.
const myOriginalSigner = createSignerWithEphemeralSignersSupport();

// Let's imagine that the created signer offers an
// additional method to create these ephemeral signers.
const [
  myEphemeralSigner1, 
  myEphemeralSigner2
] = await myOriginalSigner.createEphemeralSigners(2);

const instruction = await someInstruction({
  signerAccount1: myEphemeralSigner1,
  signerAccount2: myEphemeralSigner2,
})

const transaction = pipe(
  createTransaction({ version: 0 }),
  tx => appendTransactionInstruction(tx, instruction),
  // ...
);
```

In this example, the ephemeral signers could be implemented in different ways:

- If the wallet decides to implement the ephemeral signers as `NoopSigner`, then the transaction won't be signed unless we provide the original signer which will provide signatures for all ephemeral signers.
  ```ts
  await signAndSendTransactionWithSigners(transaction, [myOriginalSigner]);
  ```
- If the wallet decides to implement the ephemeral signers such that they simply defer to the implementation of the original signer, then we don't need to provide an extra signer.
  ```ts
  await signAndSendTransactionWithSigners(transaction);
  ```

packages/signers/src/__tests__/sign-transaction-test.ts
```diff
@@ -0,0 +1,550 @@
...
+        expect(signerA.signTransactions).toHaveBeenCalledWith([transaction]);
+        expect(signerB.signAndSendTransactions).toHaveBeenCalledWith([
+            { ...transaction, signatures: { '1111': '1111_signature' } },
+        ]);
+
+        // And the returned signature matches the one returned by the sending signer.
+        expect(transactionSignature).toStrictEqual(new Uint8Array([1, 2, 3]));
+    });
+
+    it('sends the transaction using the provided fallback sender if no sending signer is used', async () => {
```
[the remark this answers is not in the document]
lorisleiva: Because they may not know whether or not a `TransactionSendingSigner` has been provided.

Sure, if the app knows the signers it's using then it can:

```ts
// Sign and send if we know there's a sending signer.
await signAndSendTransactionWithSigners(transaction);

// Sign THEN send if we know there's no sending signer.
const signedTransaction = await signTransactionWithSigners(transaction);
await rpc.sendTransaction(signedTransaction).send();
```

And that code works, as-is.

The reality is we may not know this in advance. For instance, a wallet abstraction such as wallet adapters or the wallet standard may offer a modifying signer or a sending signer or both. This `fallbackSender` enables us to fallback to sending the transaction however we want if no sending signer was identified or provided.

packages/signers/src/sign-transaction.ts
lorisleiva: Oh definitely the former. If there is a sending signer and you're not using it, chances are your transaction won't be fully signed.

Additionally, the latter can already be achieved using the `signTransactionWithSigners` function which will ignore any sending signers (that don't implement another interface).

```ts
const signedTransaction = await signTransactionWithSigners(transaction);
await rpc.sendTransaction(signedTransaction).send();
```

See [this related thread](https://github.com/solana-labs/solana-web3.js/pull/1792#discussion_r1394083824).

packages/signers/src/sign-transaction.ts
```diff
@@ -0,0 +1,202 @@
...
+ * Identifies each provided TransactionSigner and categorizes them into their respective types.
+ * When a signer implements multiple interface, it will try to used to most powerful interface
+ * but fallback to the least powerful interface when necessary.
+ * For instance, if a signer implements TransactionSendingSigner and TransactionModifyingSigner,
+ * it will be categorized as a TransactionSendingSigner if and only if no other signers implement
+ * the TransactionSendingSigner interface.
+ */
+function categorizeTransactionSigners(
+    signers: readonly TransactionSigner[],
+    identifySendingSigner: boolean = true
```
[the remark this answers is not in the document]
lorisleiva: Sure thing, will update.

packages/signers/src/sign-transaction.ts
```diff
@@ -0,0 +1,202 @@
...
+    );
+
+    const signedTransaction = await signModifyingAndPartialTransactionSigners(
+        transaction,
+        modifyingSigners,
+        partialSigners
+    );
+
+    if (sendingSigner) {
+        const [signature] = await sendingSigner.signAndSendTransactions([signedTransaction]);
```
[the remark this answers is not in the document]
lorisleiva: 100% agree. I was talking to Jordan about this yesterday and I think we should have a second optional `configs` argument that enables users to provide their abort signals.

I'll make a PR for it asap.

packages/signers/src/sign-transaction.ts
lorisleiva: We could fail but that will limit some of the advanced use-cases.

Again, the ephemeral signer example illustrates this nicely. Take [the example on this thread](https://github.com/solana-labs/solana-web3.js/pull/1792#discussion_r1394075990) and focus on the second bullet point, i.e. the ephemeral signers are implements such that they defer to the original signer. Now imagine that the original signer is a `TransactionSendingSigner`. All of the sudden you end up with multiple `TransactionSendingSigner` for different addresses (the ephemeral accounts) yet it does not matter which one is selected as they all execute the original signer's logic.

packages/signers/src/sign-transaction.ts
lorisleiva: I think there's a misunderstanding here.

No matter how many interfaces a signer implements, **only one of them will be used per address**. Providing multiple interfaces allows the app to decide which one to use.

Therefore, a signer that implements `TransactionModifyingSigner & TransactionSendingSigner` will either call `modifyAndSignTransactions` at the beginning OR `sendAndSendTransactions` at the end.

If a `TransactionSendingSigner` needs also needs to modify the transaction, they can do whatever they want since they are sending it anyway (as long as they don't invalidate previous signatures).

If a `TransactionSendingSigner` needs to modify the transaction whilst allowing others to partially sign it before it signs and sends it, then it simply need to offer a helper function that modify the transaction however it wishes. However a `modifyTransactions` function does not enter the realm of the Signer API as it does not sign anything.

packages/signers/src/index.ts
lorisleiva: Cause I documented it haha. I'll remove it from the `index.ts`. 👍

packages/signers/src/sign-transaction.ts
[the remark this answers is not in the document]
lorisleiva: **PR available here: https://github.com/solana-labs/solana-web3.js/pull/1847**

I found it more convenient to append the PR (instead of prepending it) so we can see the full abort signal changes in one go. I'll make sure both PRs are merged together when they are approved.

packages/signers/src/sign-transaction.ts
buffalojoec: I was thinking someone like a wallet may wish to forcibly use their sender over any others, regardless of whoever else in the transaction is a sender.

However, I think if a wallet wants to ensure they send every transaction, they'd probably use the snippet you provided.

I believe I was also considering a situation where `TransactionSendingSigner` has both `signTransaction` _and_ `signAndSendTransaction`. But that's not the case.

> If there is a sending signer and you're not using it, chances are your transaction won't be fully signed.

This would also happen if you had more than one sender, right? You'd be sending before the next one could sign.

packages/signers/src/sign-transaction.ts
buffalojoec: So you're basically saying the onus is on the second wallet for making sure they aren't modifying an already signed transaction? And if they are, but this transaction comes signed from some SDK's method, that's the fault of the first wallet, yeah? I just want to make sure we're not enabling this confusion/collison.

packages/signers/src/sign-transaction.ts
lorisleiva: > I was thinking someone like a wallet may wish to forcibly use their sender over any others

Interesting, I always see the Signer API being **used** by dApps and **implemented** by wallets but technically wallets could also use the Signer API and create a Signer API inception haha.

But yeah typically wallets would be implementing the `signAndSendTransactions` method themselves and if they use the new web3.js internally, then I think they'd be using lower-level functions from `@solana/keys` to sign and `@solana/rpc-*` to send.

> This would also happen if you had more than one sender, right? You'd be sending before the next one could sign.

That's right. That's the second type of "conflict" documented on Notion:

> ### Handling sender conflicts
> 
> The second type of conflict we’ve identified is when multiple transaction sending signers are provided.
> 
> For instance, say we have two `TransactionSendingSigner` A and B for a given transaction, one using Phantom and another using Solflare. How does one decide which one to use?
> 
> Here as well, **the Signer may allow the app** to decide by implementing the `TransactionSendingSigner` interface in conjunction with the `TransactionPartialSigner` or `TransactionModifyingSigner` interface.
> 
> However, if both signers *must* send the transaction, then there is nothing that can be done (with or without a Signer API) and we should simply choose one which _could_ end up in a transaction failure due to a lack of signer.

The reason it **could** and not **would** end up in a failure is that there can be some valid use-cases where multiple `TransactionSendingSigner` are provided and whichever one is chosen will sign for all other signers (so it doesn't matter which one we choose). An example of that is ephemeral signers.

packages/signers/src/sign-transaction.ts
lorisleiva: > So you're basically saying the onus is on the second wallet for making sure they aren't modifying an already signed transaction?

Yes, it's the responsibility of the subsequent wallets to decide whether they can also modify the transaction or not. If they _must_ modify the transaction and cannot then they can throw an error explaining why they cannot sign without invalidating previous signatures. But if _we_ throw an error on our helper methods, then we stop supporting valid use-cases. And our goal with this Signer API is that **it should not prevent wallets to do what they could do without**.

> that's the fault of the first wallet, yeah?

I don't think it's a wallet's fault in particular but just a combination of Signers that end up in an impossible scenario (with or without the Signer API). But it's the responsibility of the wallet that identifies the issue to fail accordingly.

packages/signers/src/deduplicate-signers.ts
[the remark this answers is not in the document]
steveluscher: Thanks for all of that explanation! It's awesome to have it down on paper so everyone can look back at it months from now.

I'm a ‘what can go wrong’ guy. Isn't it actually quite likely that someone might go:

1. Create the start of a transaction, somewhere, maybe save it away.
2. Add instructions to it with a `TransactionSigner`.
3. Time passes
4. Decide, later, how exactly to send it (think cluster switchers, endpoint switchers, etc)
5. Add some final instruction with the `TransactionSendingSigner`
6. Call `signAndSendTransactionWithSigners()` on it

…then be left wondering why it failed (ie. the deduper dropped their `TransactionSendingSigner`)

packages/signers/src/sign-transaction.ts
steveluscher: This will definitely get misused.

I also really don't like not being able to look at a transaction and predict which signer will send the transaction until you try.

packages/signers/src/sign-transaction.ts
steveluscher: > Therefore, a signer that implements `TransactionModifyingSigner & TransactionSendingSigner` will either call `modifyAndSignTransactions` at the beginning OR `sendAndSendTransactions` at the end.

Ah, this brought the API more into focus for me. Thanks.

steveluscher (changes requested):
This is a really cool API, I just keep having questions because I want to make sure it's as difficult for people as possible to misuse it / step on every rake.

packages/signers/src/__tests__/sign-transaction-test.ts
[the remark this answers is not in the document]
lorisleiva: The issue with that solution is it assumes the provided signer is present in the transaction. It sounds counter-intuitive but it may not be. If you see the example above with ephemeral signers, the `originalSigner` is signing on behalf of its ephemeral signers but is not itself part of the transaction (which would defeat the purpose of using ephemeral signers in the first place).

Another advantage of that optional `signers` array is we can make it so they override the extracted signer in case of duplicates.

If you _reaaaaally_ hate it, we can remove it and add it later if/when people need it but IMO it will be needed for some niche use-cases.

I even think @buffalojoec had some use-cases for that because I got some love when I mentioned I was gonna allow people to inject more signers. 😅 

CleanShot 2023-11-17 at 22 23 12@2x

packages/signers/src/deduplicate-signers.ts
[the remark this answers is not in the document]
lorisleiva: I see your point. It's kind of impossible to **find a heuristic that decides which signer to choose** though. You could argue the same the other way around — i.e. someone starts with a `TransactionSendingSigner`, time passes, adds a `TransactionPartialSigner` in another instruction and is left wondering why its own sending logic wasn't used.

I do agree that we could probably warn or even fail here to prevent that from happening in the first place. Meaning, the `deduplicate` function would **only remove duplicated references of signers**. If it finds two different signers object for the same `address` attribute, then it would fail and say something like: "we've identified two different signers for the same account".

What do you think?

packages/signers/src/sign-transaction.ts
lorisleiva: I'm happy for us to fail here but then we definitely cannot remove the optional `signers` array because we'd be making ephemeral signers impossible with the Signer API.

There are basically two ways the Signer API can currently support ephemeral signers depending on how the ephemeral signers themselves are implemented:

- **Option A: Delegate to the original signer.**
    
    One way to implement these ephemeral signers is to simply delegate to the `signAndSendTransactions` method of the original signer. The issue here is that we have 2 ephemeral signers. That being said, **if our sign and send helper doesn’t fail** and simply chooses one of the available `TransactionSendingSigner`, then that’ll work out of the box.
    
    ```tsx
    await signAndSendTransactionWithSigners(myTransaction);
    ```
    
- **Option B: Use noop signers.**
    
    Another way would be to make these ephemeral signers noop signers. Since we still need to use the original signer and it’s not part of the transaction, we would need to accept an optional array of additional signers to use in our sign and send helper function.
    
    ```tsx
    await signAndSendTransactionWithSigners(myTransaction, [myOriginalSigner]);
    ```

packages/signers/src/__tests__/sign-transaction-test.ts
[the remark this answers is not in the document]
lorisleiva: I'm happy with that actually. To make sure I'm getting this right, you're saying we should not accept a `fallbackSender` and just fail if no sending signer was identified (ideally with a coded error when we sort all our `// TODO: Coded error.` haha).

Would you not even be tempted to offer an additional helper like the following?

```ts
signAndSendTransactionWithSignersAndFallback(transaction, fallbackSender);
```

packages/signers/src/__tests__/sign-transaction-test.ts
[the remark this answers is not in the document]
buffalojoec: Yeah I felt like it was useful. Specifically I think I was again leaning toward flexibility on transaction requests, but perhaps it's not the only way that can be done.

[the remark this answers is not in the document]
lorisleiva: Basically we can either:
- Remove this option of manually provided signers (which I don't recommend) or
- Fail when more than one sending signer is provided.

But not both. Otherwise, we disable ephemeral signers (see https://github.com/solana-labs/solana-web3.js/pull/1792#discussion_r1397980537).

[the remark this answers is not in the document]
steveluscher: > …If you see the example above with ephemeral signers, the originalSigner is signing on behalf of its ephemeral signers but is not itself part of the transaction…

In the case that you didn't add the `Signer` in the first place, wouldn't you just `addSignersToTransaction()` with the ephemeral signer that, presumably, does match some address in the transaction, and delegates to `originalSigner` through 〰️magic〰️?

I think I don't properly understand what ephemeral signers are. I'm operating on the simp assumption that every signer is associated with an address and knows how to sign bytes, the end. I must be missing something.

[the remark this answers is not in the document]
lorisleiva: > delegates to originalSigner through 〰️magic〰️

Yes, but that won't work if there's more than one ephemeral signer for the same original signer and if that original signer is a sending signer. _Unless_, we allow multiple sending signers and choose one at random.

lorisleiva:
After a nice chat with @steveluscher discussing this PR in details, we've agreed to the following changes:

- **Remove the `fallbackSender` option** and provide a pair of `isSendableTransaction`/`assertIsSendableTransaction` helper that tells us if a sending signer is present in the transaction. This gives the end-user the opportunity to check beforehand if the transaction can be send via `signAndSendWithSigners`. Alternatively, they can try/catch to tackle that edge-case when it has actually happened.
- **Fail when more than one sending signer is present** in a transaction. We can fail when we try to `signAndSendWithSigners` but also when asserting that the transaction is a sendable transaction.
- **Offer a `addSignersToTransaction` helper** that goes through the transaction's account meta and match them with the provided set of signers. It then updates the account meta itself to attach the provided signers to the transaction.
- **Deduplicate signers by reference** only. Meaning, when extracting all signers in a transaction, we will remove duplicated reference of the same signer object. However, if we identity two different signer objects for the same address, we will fail as it is impossible to come up with a heuristic that works with every use-cases.

I will work on all these points in subsequent stacked PRs so they can be reviewed individually.

packages/signers/src/__tests__/account-signer-meta-test.ts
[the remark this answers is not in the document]
lorisleiva: Resolving this as a decision was made [here](https://github.com/solana-labs/solana-web3.js/pull/1792#issuecomment-1820487421).

packages/signers/src/__tests__/sign-transaction-test.ts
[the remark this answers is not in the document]
lorisleiva: Resolving this as decisions were made [here](https://github.com/solana-labs/solana-web3.js/pull/1792#issuecomment-1820487421).

Whilst these decisions make it more difficult for the specific use-case "ephemeral sending signer", a workaround is to make all ephemeral signers `NoopSigner` aside from one, which delegates to the original signer.

packages/signers/src/deduplicate-signers.ts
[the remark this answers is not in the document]
lorisleiva: Resolving this as a decision was made [here](https://github.com/solana-labs/solana-web3.js/pull/1792#issuecomment-1820487421).

packages/signers/src/sign-transaction.ts
lorisleiva: Resolving this as a decision was made [here](https://github.com/solana-labs/solana-web3.js/pull/1792#issuecomment-1820487421).

packages/signers/src/__tests__/sign-transaction-test.ts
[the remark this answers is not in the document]
lorisleiva: Resolving this as a decision was made [here](https://github.com/solana-labs/solana-web3.js/pull/1792#issuecomment-1820487421).

steveluscher:
giphy

steveluscher (approved):

lorisleiva:
### Merge activity

* **Nov 22, 6:19 AM**: @lorisleiva started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1792).

PR #1793, opened 2023-10-27
refactor(experimental): enhance null account info handling for GraphQL

buffalojoec:
This PR enables returning default data - including the account's address - when it's not found by the RPC call `getAccountInfo`. This allows chained/nested queries to be more reliable if the RPC encounters an account that does not exist.

### DIFF
--- a/packages/rpc-graphql/src/__tests__/account-test.ts
+++ b/packages/rpc-graphql/src/__tests__/account-test.ts
@@ -60,6 +60,24 @@ describe('account', () => {
                 },
             });
         });
+        it("can query an account's address", async () => {
+            expect.assertions(1);
+            const source = `
+            query testQuery($address: String!, $commitment: Commitment) {
+                account(address: $address, commitment: $commitment) {
+                    address
+                }
+            }
+        `;
+            const result = await rpcGraphQL.query(source, variableValues);
+            expect(result).toMatchObject({
+                data: {
+                    account: {
+                        address: 'AyGCwnwxQMCqaU4ixReHt8h5W4dwmxU7eM3BEQBdWVca',
+                    },
+                },
+            });
+        });
         it('can query multiple fields', async () => {
             expect.assertions(1);
             const source = `
@@ -95,6 +113,7 @@ describe('account', () => {
             query testQuery($address: String!, $commitment: Commitment) {
                 account(address: $address, commitment: $commitment) {
                     owner {
+                        address
                         executable
                         lamports
                         rentEpoch
@@ -107,6 +126,7 @@ describe('account', () => {
                 data: {
                     account: {
                         owner: {
+                            address: 'TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA',
                             executable: true,
                             lamports: expect.any(BigInt),
                             rentEpoch: expect.any(BigInt),
@@ -128,7 +148,9 @@ describe('account', () => {
             query testQuery($address: String!, $commitment: Commitment) {
                 account(address: $address, commitment: $commitment) {
                     owner {
+                        address
                         owner {
+                            address
                             executable
                             lamports
                             rentEpoch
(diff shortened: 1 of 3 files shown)


buffalojoec:
* master
  * **PR #1793** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1793" target="_blank">Graphite</a>  👈
    * **PR #1794** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1794" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1793?utm_source=stack-comment).

mcintyre94:
The code here looks good, but what's the use case where you want this instead of an account not found error if the account doesn't exist? I probably just don't get it because I don't use graphQL TBH! 

Also would it make sense to default to an all-zero address that (AFAIK) doesn't actually own any other addresses? It looks like this could be quite confusing as-is because the system program is actually the owner of a bunch of addresses so you couldn't distinguish between that and an error.

buffalojoec:
> this could be quite confusing as-is because the system program is actually the owner of a bunch of addresses so you couldn't distinguish between that and an error.

Yeah, this is a good point. However, I chose to follow the Explorer's pattern of showing the owner at System Program but the account doesn't actually exist (all null).

The main reason for this is to address the "quadruple nested query" test. If you try to query the native loader, it will throw null because it doesn't exist, and basically I was thinking "instead of null, why don't all roads just lead to the System Program?". Because as it turns out, the System Program is also owned by the native loader!

However, the more I think about it, the more this starts to seem like a code smell.

mcintyre94:
Does that nested null make the whole query fail, or do you just get a null and can't go any deeper? I think that'd be fine TBH, but I'm not really sure what the graphQL convention is here.

buffalojoec:
> Does that nested null make the whole query fail, or do you just get a null and can't go any deeper? I think that'd be fine TBH, but I'm not really sure what the graphQL convention is here.

You'd end up with this:

```typescript
account: {
   owner: {
      owner: null
   }
}
```

Which isn't the worst thing I guess!

mcintyre94:
Is there any relevant graphQL convention we could try to use as a guide?

buffalojoec:
> Is there any relevant graphQL convention we could try to use as a guide? 

I think the bull value for the field _is_ the convention lol. The rest of the query will work if you have other fields at different levels.

buffalojoec:
> > Does that nested null make the whole query fail, or do you just get a null and can't go any deeper? I think that'd be fine TBH, but I'm not really sure what the graphQL convention is here.
> 
> You'd end up with this:
> 
> ```ts
> account: {
>    owner: {
>       owner: null
>    }
> }
> ```
> 
> Which isn't the worst thing I guess!

Added changes to go back to this!

buffalojoec:
### Merge activity

* **Oct 27, 6:20 PM**: @buffalojoec started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1793).

steveluscher:
Late to the party here, but typically if a node is not found in a GraphQL query, the resolver should return null.

We should _probably_ do something similar here, though I get what this PR is aiming at.

* Return `null` if accounts are not rent-exempt
* Return a special type of `Account` (eg. `NonExistentAccount`) that has only the bare minimum of fields (eg. no `owner`)

> Yeah, this is a good point. However, I chose to follow the Explorer's pattern of showing the owner at System Program but the account doesn't actually exist (all null).

Explorer could just _do_ this without relying on the server to tell it. Like, if it queries for account `abc` and doesn't find it, it could just render an ‘account not found’ page like it does now with the owner hard coded to the system program. The RPC doesn't need to hold its hand to reach that conclusion.

buffalojoec:
> We should probably do something similar here, though I get what this PR is aiming at.
> 
> Return null if accounts are not rent-exempt, or
> Return a special type of Account (eg. NonExistentAccount) that has only the bare minimum of fields (eg. no owner)

The difficulty here is if you encounter an `owner` that does not exist - such as `Native1oader1111` or `Feature1111111` - you'll get `owner: null`. However, you _already_ had the address of the owner from the previous query layer, so IMO you shouldn't lose that address just because you tried to look it up for other fields. 

```typescript
const source = `
  myQuery {
    account(address: "xxx") {
      address
      owner {
        address
        lamports
      }
    }
  }
`;

const response = await rpcGraphQL.query(source);

const {
  data: {
    account: {
      address: accountAddress,
      owner: {
        address: ownerAddress,
        lamports: ownerLamports,
      }
    }
  }
} = response;

if (ownerLamports === null) {
  console.log(`Account ${ownerAddress} does not exist.`);
}
```

steveluscher:
Totally. Those accounts should probably materialize as a special kind of account in the union if possible account this field could materialize. `BuiltinProgramAccount` or `NativeProgramAccount` or some such thing.

Issue #1680, opened 2023-10-05
Suggestion: drop node 16 (and the node-fetch and ed25519 polyfills)


mikemaccana:
## Motivation

[node 16 is already end of life](https://nodejs.org/en/blog/announcements/nodejs16-eol). Current web3.js includes polyfills for `fetch` and for Ed25519 in `global.crypto`. Current supported versions of node have both `fetch` and (from SL on Twitter) Ed25519 support.

## Details

Downsides:
 - Will hurt people using an already end of life version of node. They'll receive the following message on install:

```
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: '@solana/web3.js@2.0.0',
npm WARN EBADENGINE   required: { node: '>=18.0.0' },
npm WARN EBADENGINE   current: { node: 'v16.6.0', npm: '9.8.1' }
npm WARN EBADENGINE }
``` 
 
Upsides:
 - Avoid [discussion of polyfills](https://x.com/steveluscher/status/1707409603577979179?s=20) that may or may not be necessary for the users version of node
 - Reduce additional dependencies
 - Reduce work by users to remove unnecessary dependencies (I made [just-use-native-fetch](https://github.com/mikemaccana/just-use-native-fetch)
 - Reduce support issues from users using an unsupported node, that probably have other issues
 
`engines` in the `package.json` can be used to specify minimum node versions.

```
  "engines": {
    "node": ">=18.0.0"
  },
```

steveluscher:
I'm a huge fan of making Node LTS+ the requirement, I added this in few packages but seem to have missed some others.

```json
"engine": {
  "node": ">=17.4"
},
```

I don't know that there's much benefit at this point to churning everyone on the current web3.js. Might as well churn the API and the Node requirement all at once.

mcintyre94:
What about browsers though?

On the latest Chrome this is an error:

```js
await window.crypto.subtle.generateKey('Ed25519', false, ['sign'])
Uncaught DOMException: Failed to execute 'generateKey' on 'SubtleCrypto': Algorithm: Unrecognized name
```

Seems to be behind a feature flag on Chrome: https://chromestatus.com/feature/4913922408710144

Caniuse doesn't seem to have this nuance, but from a quick check Firefox also doesn't work, Safari does.

steveluscher:
When you try to use `generateKeyPair()` in the new library, the error message tells you to install and import `@solana/webcrypto-ed25519-polyfill`.

[...]

PR #1802, opened 2023-10-28
refactor(experimental): Use the native `fetch` built into Node 18+

steveluscher:
# Summary

OK, so here's what I'm thinking.

`node-fetch` is in this weird state where:

- they refuse to build CommonJS bundles for the 3.x line
- the 2.x line supports `HttpAgent`
  - note that browsers will never support `HttpAgent`
- the 2.x line does not support HTTP/2

This means that we're taking on a huge dependency and exposing its config through web3.js, but will never end up in the ideal state (HTTP/2 support everywhere).

In this PR, we throw in the towel and make the _default_ transport that ships with web3.js API compatible with the `fetch` in both Node and browsers.

From that point:

- We could write our own HTTP/2 transport that's API compatible with `fetch()` and ship that
- Someone _else_ could write an all-singing, all-dancing transport _specifically_ for Node and open source it

Addresses #1680.


steveluscher:
* master
  * **PR #1802** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1802" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1802?utm_source=stack-comment).

buffalojoec (approved):
Well this ties directly into #1680. New Web3 JS is only supported on Node LTS? Ok with me.

steveluscher:
> Well this ties directly into #1680. New Web3 JS is only supported on Node LTS? Ok with me.
> 
> 

Yeah, I think this is the right thing to do so that we can go bare-bones for the default transport, and then start building it back up from there with transports that work well in Node.

PR #1807, opened 2023-11-02
docs(experimental): add custom retry logic example

quellen-sol:
Example shows a simple retrying transport that will retry up to `MAX_ATTEMPTS`. Would appreciate any feedback regarding implementation, as I wasn't too sure how much error handling would be best to show, for example.

Thanks! Super excited to start using this!

### DIFF
--- a/packages/library/README.md
+++ b/packages/library/README.md
@@ -364,7 +364,51 @@ Another example of a possible customization for RPC transports is sharding. Here
 The transport library can also be used to implement custom retry logic on any request:
 
 ```tsx
-// TODO: Your turn; send us a pull request with an example.
+import { createDefaultRpcTransport } from "@solana/web3.js";
+import { IRpcTransport } from "@solana/rpc-transport/dist/types/transports/transport-types";
+import { createJsonRpc } from "@solana/rpc-transport";
+import { createSolanaRpcApi } from "@solana/rpc-core";
+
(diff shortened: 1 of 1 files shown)


steveluscher:
Cool; I like it!

What do you think about this: often when we talk about retry we talk about smearing those retries across time. If you try 4 times in a loop, you might get the same result every time because the conditions haven't changed. On the other hand if you introduce a _time_ delay between retries, you might have more success.

Want to try to introduce a concept like that, and then we'll get this in?

Reading: https://dzone.com/articles/understanding-retry-pattern-with-exponential-back

steveluscher (changes requested):
Cool; I like it!

What do you think about this: often when we talk about retry we talk about smearing those retries across time. If you try 4 times in a loop, you might get the same result every time because the conditions haven't changed. On the other hand if you introduce a time delay between retries, you might have more success.

Want to try to introduce a concept like that, and then we'll get this in?

Reading: https://dzone.com/articles/understanding-retry-pattern-with-exponential-back

quellen-sol:
Hah! I was debating whether or not to use exponential backoff in the example. I definitely agree it'd be nice to introduce that concept, will do, ty!

steveluscher (approved):
Looks good! Only thing that's not handled here that you would want for a production version is cancellation (eg. listening to the `abortSignal`) but it's good enough for an example.

PR #1809, opened 2023-11-03
fix(experimental): parameter names in GetTokenAccountsByOwnerApi

### LINKED ISSUE #1808: [experimental] Fix Params Names in `getTokenAccountsByOwner` experimental
## Overview
Two small issues related to params of experimental version of `getTokenAccountsByOwner` method. 

1. The label is currently `program` but should be `owner`. The method runs fun but is incorrectly labelled. 
- Location of Issue: https://github.com/solana-labs/solana-web3.js/blob/fbd397436f622a9b87a7251c08bd9664599299ee/packages/rpc-core/src/rpc-methods/getTokenAccountsByOwner.ts#L58
- Source of expected parameter: https://github.com/solana-labs/solana/blob/662ac8bc8678beef560727bc9c6a0911784589ea/rpc/src/rpc.rs#L1887

2. The final overloading does not include an `encoding` type. Though it's not required in the config, I've been unable to get any valid response without explictly passing an `encoding`.  I don't think this is an issue w/ the web3js, but a question of the default encoding. 

## Steps to reproduce

For #2 - looking at [the tests](https://github.com/solana-labs/solana-web3.js/blob/fbd397436f622a9b87a7251c08bd9664599299ee/packages/rpc-core/src/rpc-methods/__tests__/get-token-accounts-by-owner-test.ts#L390), this is a known issue.

amilz:
This commit renames the parameter 'program' to 'owner' across all instances within the GetTokenAccountsByOwnerApi interface. This modification corrects the name, making the API more intuitive and consistent with the expectations of the method.

Fixes #1808. 

IMO, since every token account is > 128 bytes, I'd suggest removing the final overload of `GetTokenAccountsByOwnerApi` the one with no encoding. If the user does not pass an encoding value, the method returns an error. Happy to make that tweak, but I wanted to confirm it here first.

### DIFF
--- a/packages/rpc-core/src/rpc-methods/getTokenAccountsByOwner.ts
+++ b/packages/rpc-core/src/rpc-methods/getTokenAccountsByOwner.ts
@@ -55,7 +55,7 @@ export interface GetTokenAccountsByOwnerApi {
      * Returns all SPL Token accounts by token owner.
      */
     getTokenAccountsByOwner(
-        program: Base58EncodedAddress,
+        owner: Base58EncodedAddress,
         filter: AccountsFilter,
         config: GetTokenAccountsByOwnerApiCommonConfig &
             GetTokenAccountsByOwnerApiSliceableCommonConfig &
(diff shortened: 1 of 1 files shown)


buffalojoec (approved):
Ship it. I believe if you rebase the CI will pass. I pushed a lint fix.

Thanks!!

PR #1810, opened 2023-11-04
refactor(experimental): remove README from lint and prettier

buffalojoec:
This is causing some CI annoyance.


lorisleiva (approved):
I still don't understand why CI is being so annoying here but if that helps then I'm all for it.

PR #1814, opened 2023-11-07
refactor(experimental): rename `Base58EncodedAddress` to `Address`

steveluscher:
# Summary

To make this API more terse, we're going to take the _default_ type of addresses (the base-58 encoded one) and name it simply `Address`. Specific types of addresses will be given names like `AddressBytes` and `AddressBase64` but the conventional one will not carry a specifier.

This was basically a find-and-replace with some light spot checking for grammar (eg. ‘a `Base58EncodedAddress`’ => ‘an `Address`.’

This change was brought to you by bikeshedding on Twitter. https://twitter.com/lorismatic/status/1710595018170245603

# Test Plan

```
pnpm build
```

PR #1815, opened 2023-11-07
refactor(experimental): rename `Ed25519Signature` and `TransactionSignature` to `SignatureBytes` and `Signature`

### LINKED ISSUE #1709: [experimental] Unify `Base58EncodedTransactionSignature` and `TransactionSignature` types
## Overview

I suspect that these types should be the same. Both are implemented as `string & {readonly __brand: unique symbol;}`

`Base58EncodedTransactionSignature` is the output from `sendTransaction`: https://github.com/solana-labs/solana-web3.js/blob/278784a7b7799354f16f5132a614dd05e7f671e3/packages/rpc-core/src/rpc-methods/sendTransaction.ts#L13

It's defined in the rpc-core `common.ts`, these types are not exported

`TransactionSignature` is the input to `getTransaction`: https://github.com/solana-labs/solana-web3.js/blob/278784a7b7799354f16f5132a614dd05e7f671e3/packages/rpc-core/src/rpc-methods/getTransaction.ts#L164

It's defined in `@solana/transactions` and is exported

## Steps to reproduce

```ts
const sig = await rpc.sendTransaction(encodedTransaction).send();
// TS error here
const tx = await rpc.getTransaction(sig);
```

## Description of bug

Error:

```
Argument of type 'Base58EncodedTransactionSignature' is not assignable to parameter of type 'TransactionSignature'.
  Type 'Base58EncodedTransactionSignature' is not assignable to type '{ readonly __brand: unique symbol; }'.
    Types of property '__brand' are incompatible.
      Type 'typeof __brand' is not assignable to type 'typeof __brand'. Two different types with this name exist, but they are unrelated.
```

I don't know which name we should keep, but I think these should be the same type. We probably need to keep exporting it from `@solana/transactions` for compatibility with other functions.

steveluscher:
# Summary

This is a follow on to #1814 where we make the _common_ type terse, and any specific type suffixed with the shape of the data.

In this PR I _also_ remove any concept of ‘transaction’ from the `Signature` type. This _might_ be the wrong thing to do; you can no longer constrain a method's inputs based on knowing for sure that something is the signature of a _transaction_. If you think this is important, comment on this PR!

* `Ed25519Signature` => `SignatureBytes`
* `TransactionSignature` => `Signature`
* `Base58EncodedTransactionSignature` => `Signature`

Oh, also I moved all the types and methods into `@solana/keys`.

Resolves #1709.

# Test Plan

```
pnpm build
```


steveluscher:
* master
  * **PR #1815** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1815" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1815?utm_source=stack-comment).

lorisleiva (approved):
This is making me so happy. I think merging the two types is fine. I'd rather we add it back later if we really see a need for it than just add it preemptively.

PR #1817, opened 2023-11-09
refactor(experimental): revise accounts schema

buffalojoec:
This PR introduces a revised version of the account schema for `jsonParsed`
account data.

Typically we see JSON-parsed account data like this:

```javascript
{
  data: {
    parsed: {
      info: {
        /* data */
      },
      type: 'mint',
    },
    program: 'Tokenkegxxxxxxxx',
    space: 82n,
  },
}
```

However, when it comes to GraphQL, I've made an opinionated decision to move
away from the JSON-parsed format returned by the RPC, and instead opt for this
"optimized" format:

```javascript
{
  data: {
    /* data */
  },
  meta: {
    program: 'Tokenkegxxxxxxxx',
    space: 82n,
    type: 'mint',
  },
}

```

Since we know we'll always get `program`, `space`, and `type`, we can extract
those to be "parsed metadata" about an account and introduce the `meta` field.

### DIFF
--- a/packages/rpc-graphql/README.md
+++ b/packages/rpc-graphql/README.md
@@ -202,14 +202,10 @@ const source = `
         account(address: $address) {
             ... on MintAccount {
                 data {
-                    parsed {
-                        info {
-                            decimals
-                            isInitialized
-                            mintAuthority
-                            supply
-                        }
-                    }
+                    decimals
+                    isInitialized
+                    mintAuthority
+                    supply
                 }
             }
         }
@@ -222,9 +218,7 @@ for (const address of maybeMintAddresses) {
         const {
             data: {
                 account: {
-                    data: {
-                        parsed: { info: mintInfo },
-                    },
+                    data: mintInfo,
                 },
             },
         } = result;
@@ -251,26 +245,22 @@ const source = `
         account(address: $address) {
             ... on MintAccount {
                 data {
-                    parsed {
-                        info {
-                            decimals
-                            isInitialized
-                            supply
-                        }
-                        type
-                    }
+                    decimals
+                    isInitialized
+                    supply
+                }
+                meta {
+                    type
                 }
             }
             ... on TokenAccount {
                 data {
-                    parsed {
-                        info {
-                            isNative
-                            mint
-                            state
-                        }
-                        type
-                    }
+                    isNative
+                    mint
+                    state
+                }
+                meta {
+                    type
                 }
             }
         }
@@ -283,11 +273,9 @@ for (const address of mintOrTokenAccountAddresses) {
         const {
             data: {
                 account: {
-                    data: {
-                        parsed: {
-                            info: accountParsedInfo,
-                            type: accountType,
-                        }
+                    data: accountParsedInfo,
+                    meta: {
+                        type: accountType,
                     }
                 }
             }
@@ -427,13 +415,9 @@ const source = `
             ... on MintAccount {
                 address
                 data {
-                    parsed {
-                        info {
-                            mintAuthority {
-                                address
-                                lamports
-                            }
-                        }
+                    mintAuthority {
+                        address
+                        lamports
                     }
                 }
             }
@@ -453,13 +437,9 @@ data: {
     account: {
         address: 'AyGCwnwxQMCqaU4ixReHt8h5W4dwmxU7eM3BEQBdWVca',
         data: {
-            parsed: {
-                info: {
-                    mintAuthority: {
-                        address: 'DpfJkNonoVB3sor9H9ceajhex4XHVPrDAGAq2ahdG4JZ',
-                        lamports: 10290815n,
-                    }
-                }
+            mintAuthority: {
+                address: 'DpfJkNonoVB3sor9H9ceajhex4XHVPrDAGAq2ahdG4JZ',
+                lamports: 10290815n,
             }
         },
     },
@@ -524,26 +504,24 @@ const source = `
         programAccounts(programAddress: $address) {
             ... on MintAccount {
                 data {
-                    parsed {
-                        info {
-                            decimals
-                            isInitialized
-                            mintAuthority
-                            supply
-                        }
-                    }
+                    decimals
+                    isInitialized
+                    mintAuthority
+                    supply
+                }
+                meta {
+                    type
                 }
             }
             ... on TokenAccount {
                 data {
-                    parsed {
-                        info {
-                            isNative
-                            mint
-                            owner
-                            state
-                        }
-                    }
+                    isNative
+                    mint
+                    owner
+                    state
+                }
+                meta {
+                    type
                 }
             }
         }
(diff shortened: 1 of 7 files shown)


buffalojoec:
* master
  * **PR #1817** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1817?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1817?utm_source=stack-comment).

steveluscher (changes requested):
Let's go _further_! Free yourself _completely_ from the JSON-RPC responses, and imagine building a UI that displays a mint account. For sure you would want this:

```graphql
query {
  account(address: "abc") {
    # Stuff that every account has...
    address
    owner
    space
    ... on MintAccount {
      decimals
      isInitialized
      mintAuthority
      supply
    }
  }
}
```

No need for `data`, `meta`, or `type` fields.

Then in JS:

```ts
const account = ...; // Fetch the account
if (account.__type === 'MintAccount') {
  // TypeScript has now refined `account` to have the `MintAccount` fields
  console.table({
    decimals: account.decimals,
    supply: account.supply,
  });
}
```

The only thing that has to happen to make this work is that the fields have to be unique among all the types. You can't have a `MintAccount` that returns a `supply` that's an integer and a `FooAccount` that returns `supply` that's an object. In practice I haven't found this to be very difficult to avoid.

tl;dr collapse the account schema as flat as is practical and see what we end up with.

packages/rpc-graphql/README.md
```diff
@@ -975,54 +947,38 @@ const source = `
...
-                                                                                    decimals
-                                                                                }
-                                                                            }
-                                                                        }
-                                                                    }
-                                                                }
-                                                                owner {
+                                                        address
+                                                        mint {
+                                                            ... on MintAccount {
```
steveluscher, line 971: Isn't this _always_ of the `MintAccount` type? Shouldn't be any need to type the subselection.

packages/rpc-graphql/src/__tests__/account-test.ts
```diff
@@ -611,41 +595,39 @@ describe('account', () => {
...
+                                }
+                                priorVoters
+                                rootSlot
+                                votes {
+                                    confirmationCount
+                                    slot
+                                }
+                            }
+                            meta {
                                 program
```
steveluscher, line 628: What should we call this for real in this greenfield schema?

* Assigned Program id
* Owner
* Owning program

https://docs.solana.com/developing/programming-model/accounts#ownership-and-assignment-to-programs

packages/rpc-graphql/src/__tests__/account-test.ts
```diff
@@ -656,47 +638,45 @@ describe('account', () => {
...
-                                            epoch: expect.any(BigInt),
-                                            previousCredits: expect.any(String),
-                                        },
-                                    ]),
-                                    lastTimestamp: {
-                                        slot: expect.any(BigInt),
-                                        timestamp: expect.any(BigInt),
-                                    },
-                                    node: {
+                            authorizedVoters: expect.arrayContaining([
```
steveluscher, line 641: We need to talk about whether to make things like this _plural fields_ or _connections_ to prepare for a future where it's possible to paginate through them.

https://relay.dev/graphql/connections.htm

packages/rpc-graphql/src/resolvers/account.ts
```diff
@@ -0,0 +1,89 @@
...
+        return cached;
+    }
+
+    const account = await rpc
+        .getAccountInfo(address, requestConfig as Parameters<SolanaRpcMethods['getAccountInfo']>[1])
+        .send()
+        .then(res => res.value)
+        .catch(e => {
+            throw e;
+        });
```
steveluscher, line 53: Oh, I've gotta get you on the `DataLoader` train ASAP.

In particular, the architecture you've developed here won't actually dedupe requests for the same thing when multiple resolvers are working in parallel, because you don't _save_ the cache result until the request is over. All of them will kick off, think there's no cache, make _n_ queries, and then save the same (maybe) result in the cache at the end.

DataLoader:

1. Specify all of the resolvers in the form of dataloaders; bind the rpc at the very very top level context
2. Pass those dataloaders down to resolvers through context
3. Call `.load()` on them.

https://github.com/steveluscher/zero-to-graphql/blob/3daf2d8dc2ac9bc63cf16966d6489fe43cbddbe9/zero-node/index.js#L32-L48

steveluscher: @buffalojoec, to get you started: https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1827/refactor-experimental-use-DataLoader-to-coalesce-multiple-requests-for-the-same-block#

packages/rpc-graphql/README.md
buffalojoec: Yes it's always a mint account, and I could stick the `MintAccount` type in there (which would be pretty cool), but that eliminates the potential for someone to get the `mint` account as `base58`, etc. It would always be the parsed `MintAccount`.

I'm leaning towards it though, because having to specify the spread here stinks.

packages/rpc-graphql/src/__tests__/account-test.ts
buffalojoec: `ownerProgram` I think.

packages/rpc-graphql/src/resolvers/account.ts
buffalojoec: Love it. Thanks!

packages/rpc-graphql/src/__tests__/account-test.ts
buffalojoec: Pagination has been discussed for RPC v2, yeah? Probably good to do.

buffalojoec:
> Let's go _further_! Free yourself _completely_ from the JSON-RPC responses, and imagine building a UI that displays a mint account.

Yeah, I like this. The primary reason I went with `data` and `meta` was to isolate the field names, so any parsed accounts could be named whatever, without worry of collision.

> The only thing that has to happen to make this work is that the fields have to be unique among all the types. You can't have a `MintAccount` that returns a `supply` that's an integer and a `FooAccount` that returns `supply` that's an object. In practice I haven't found this to be very difficult to avoid.

Right, so we could do this with the JSON-parsed data no problem. It's all known parsed structures and we can rename certain fields however we want to make this work. _However_, I'm thinking about a future where I can plug my IDL or Kinobi AST into this API and get parsing of _my program's stuff_!

This might complicate that.

buffalojoec:
Merging this stack for now so we can get the new schema layout in #1819, which will allow us to much more easily discuss changes! I also won't have to make the changes twice.

Created #1831 for tracking, for now.

packages/rpc-graphql/README.md
steveluscher: One way we can do that is to let fields take arguments:

```graphql
source {
  mint(encoding: BASE_64) {
    data # Base64String
  }
}
```

steveluscher: My idea [here](https://github.com/solana-labs/solana-web3.js/pull/1837#discussion_r1407188113) is better.

PR #1827, opened 2023-11-11
refactor(experimental): use `DataLoader` to coalesce multiple requests for the same block

steveluscher:
# Summary

The DataLoader pattern is used to dedupe and cache requests for identical entities across a GraphQL query. Created at the top level, a DataLoader coalesces requests in the same tick, sends them as an array of keys to a batch loader function, and caches the results based on the key generated by `cacheKeyFn`.

With this change, this query will produce exactly one request to `getBlock()`

```graphql
query {
  block(slot: 123) {
    blockTime
  }
  transactionThatHappensToBeInBlock123: transaction(signature: '...') {
    block { # Cached from before
      blockTime
    }
  }
}
```

In this PR we replace the custom cache for resolving blocks with a dataloader. Once all of the resolvers are rewritten as dataloaders, you can delete the `GraphQLCache` completely.

# Test Plan

Didn't.

### DIFF
--- a/packages/rpc-graphql/src/context.ts
+++ b/packages/rpc-graphql/src/context.ts
@@ -1,5 +1,10 @@
 import { SolanaRpcMethods } from '@solana/rpc-core';
+import { GetBlockApi } from '@solana/rpc-core/dist/types/rpc-methods/getBlock';
 import { Rpc } from '@solana/rpc-transport/dist/types/json-rpc-types';
+import DataLoader from 'dataloader';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+import fastStableStringify from 'fast-stable-stringify';
 import { GraphQLResolveInfo } from 'graphql';
 
 import { createGraphQLCache, GraphQLCache } from './cache';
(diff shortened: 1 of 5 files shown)


steveluscher:
* master
  * **PR #1827** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1827?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1827?utm_source=stack-comment).

packages/rpc-graphql/src/context.ts
```diff
@@ -195,13 +189,11 @@ async function resolveTransaction(
 export function createSolanaGraphQLContext(rpc: Rpc<SolanaRpcMethods>): RpcGraphQLContext {
     const cache = createGraphQLCache();
     return {
+        block: new DataLoader(createBlockBatchLoadFn(rpc), { cacheKeyFn: fastStableStringify }),
```
steveluscher, line 192: This `cacheKeyFn` is actually insufficient, because you want to apply _defaults_ before creating the cache key.

For instance, these should produce the _same_ cache key:

* `{slot: 123n, commitment: 'finalized'}`
* `{slot: 123n}`

…because `finalized` is the server default.

I'll leave the creation of this `normalizeConfig` method to the reader (that's you!)

buffalojoec: Yeah, this is a great call-out. I'll add to my active stack!

buffalojoec:
Thanks for laying this up for me! I've reworked your changes into this new stack: #1832 

We can close this one.

Issue #1831, opened 2023-11-11
Expand on Accounts schema in GraphQL


buffalojoec:
Making this issue to track discussions on the new Accounts schema in GraphQL.

The revised Account schema should (potentially) make use of interfaces in the following manner - as discussed in [this comment](https://github.com/solana-labs/solana-web3.js/pull/1817#pullrequestreview-1725449282):

```graphql
query {
  account(address: "abc") {
    # Stuff that every account has...
    address
    owner
    space
    ... on MintAccount {
      decimals
      isInitialized
      mintAuthority
      supply
    }
  }
}
```

In doing so, we'll need to ensure field names don't collide. This means, for example, changing `owner` to `ownerProgram` in the root `Account` type.

Developers also shouldn't need to do spread syntax on nested account queries, but this will require careful consideration of encoded nested account data (`base58`, etc.). See [this comment](https://github.com/solana-labs/solana-web3.js/pull/1817#discussion_r1389809335) for more context.

buffalojoec:
Note, this PR continues talks from #1817, since schema modifications will be much easier to follow after #1819

PR #1837, opened 2023-11-12
refactor(experimental-graphql): revise accounts schema (again)

buffalojoec:
This PR attempts to revise the accounts schema, as per #1831.

### DIFF
--- a/packages/rpc-graphql/src/__tests__/account-test.ts
+++ b/packages/rpc-graphql/src/__tests__/account-test.ts
@@ -84,6 +84,7 @@ describe('account', () => {
                         executable
                         lamports
                         rentEpoch
+                        space
                     }
                 }
             `;
@@ -94,6 +95,7 @@ describe('account', () => {
                         executable: false,
                         lamports: 10290815n,
                         rentEpoch: 0n,
+                        space: 165n,
                     },
                 },
             });
@@ -105,16 +107,17 @@ describe('account', () => {
             address: 'AyGCwnwxQMCqaU4ixReHt8h5W4dwmxU7eM3BEQBdWVca',
             commitment: 'confirmed',
         };
-        it("can perform a nested query for the account's owner", async () => {
+        it("can perform a nested query for the account's ownerProgram", async () => {
             expect.assertions(1);
             const source = /* GraphQL */ `
                 query testQuery($address: String!, $commitment: Commitment) {
                     account(address: $address, commitment: $commitment) {
-                        owner {
+                        ownerProgram {
                             address
                             executable
                             lamports
                             rentEpoch
+                            space
                         }
                     }
                 }
@@ -123,11 +126,12 @@ describe('account', () => {
             expect(result).toMatchObject({
                 data: {
                     account: {
-                        owner: {
+                        ownerProgram: {
                             address: 'TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA',
                             executable: true,
                             lamports: expect.any(BigInt),
                             rentEpoch: expect.any(BigInt),
+                            space: 133352n,
                         },
                     },
                 },
@@ -140,18 +144,19 @@ describe('account', () => {
             address: 'AyGCwnwxQMCqaU4ixReHt8h5W4dwmxU7eM3BEQBdWVca',
             commitment: 'confirmed',
         };
-        it("can perform a double nested query for each account's owner", async () => {
+        it("can perform a double nested query for each account's ownerProgram", async () => {
             expect.assertions(1);
             const source = /* GraphQL */ `
                 query testQuery($address: String!, $commitment: Commitment) {
                     account(address: $address, commitment: $commitment) {
-                        owner {
+                        ownerProgram {
                             address
-                            owner {
+                            ownerProgram {
                                 address
                                 executable
                                 lamports
                                 rentEpoch
+                                space
                             }
                         }
                     }
@@ -161,13 +166,14 @@ describe('account', () => {
             expect(result).toMatchObject({
                 data: {
                     account: {
-                        owner: {
+                        ownerProgram: {
                             address: 'TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA',
-                            owner: {
+                            ownerProgram: {
                                 address: 'BPFLoader2111111111111111111111111111111111',
                                 executable: true,
                                 lamports: expect.any(BigInt),
                                 rentEpoch: expect.any(BigInt),
+                                space: 25n,
                             },
                         },
                     },
@@ -181,16 +187,16 @@ describe('account', () => {
             address: 'AyGCwnwxQMCqaU4ixReHt8h5W4dwmxU7eM3BEQBdWVca',
             commitment: 'confirmed',
         };
-        it("can perform a triple nested query for each account's owner", async () => {
+        it("can perform a triple nested query for each account's ownerProgram", async () => {
             expect.assertions(1);
             const source = /* GraphQL */ `
                 query testQuery($address: String!) {
                     account(address: $address) {
-                        owner {
+                        ownerProgram {
                             address
-                            owner {
+                            ownerProgram {
                                 address
-                                owner {
+                                ownerProgram {
                                     address
                                 }
                             }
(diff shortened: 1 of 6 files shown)


buffalojoec:
* master
  * **PR #1837** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1837?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1837?utm_source=stack-comment).

buffalojoec:
We're still going to want to sort out [this comment here](https://github.com/solana-labs/solana-web3.js/pull/1817#discussion_r1389809335) from the previous PR. I'm not sure if blocking the ability to specify encoding for these nested accounts is the way.

packages/rpc-graphql/src/__tests__/account-test.ts
```diff
@@ -288,37 +293,36 @@ describe('account', () => {
...
+                                address
                             }
-                            meta {
-                                program
-                                space
-                                type
+                            blockhash
+                            feeCalculator {
+                                lamportsPerSignature
                             }
```
steveluscher, line 322: Tangent; this is hella deprecated right? Should we do something else here?

packages/rpc-graphql/src/__tests__/account-test.ts
```diff
@@ -327,55 +331,44 @@ describe('account', () => {
...
+                            addresses
+                            authority {
+                                address
                             }
+                            deactivationSlot
+                            lastExtendedSlot
+                            lastExtendedSlotStartIndex
+                            programName
+                            accountSpace
+                            accountType
```
steveluscher, line 371: I think we should eliminate this. We can completely replace this with the canonical `__type` right?

packages/rpc-graphql/src/__tests__/account-test.ts
```diff
@@ -327,55 +331,44 @@ describe('account', () => {
...
+                        ... on LookupTableAccount {
+                            addresses
+                            authority {
+                                address
                             }
+                            deactivationSlot
+                            lastExtendedSlot
+                            lastExtendedSlotStartIndex
+                            programName
+                            accountSpace
```
steveluscher, line 370: This stuff should just be on the base type, right?

Also, why `account.accountSpace`. Were you worried of a collision with the name `account.space`?

packages/rpc-graphql/src/schema/account.ts
```diff
@@ -19,7 +19,7 @@ export const accountTypeDefs = /* GraphQL */ `
         encoding: String
         executable: Boolean
         lamports: BigInt
-        owner: Account
+        ownerProgram: Account
         rentEpoch: BigInt
```
steveluscher, line 23: <i class='graphite__hidden'>[Re: lines 18 to 23]<br/></i>

This is where I imagine `space` belongs. No?<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1837?utm_source=unchanged-line-comment">Graphite</a>.</span>

steveluscher (changes requested):
Returning this to your queue for some questions!

[queuepatrick](https://app.graphite.dev/macros/solana-labs)

packages/rpc-graphql/src/__tests__/account-test.ts
buffalojoec: Hah, I didn't check if it was deprecated 😅

Something different, like, take it out completely?

packages/rpc-graphql/src/__tests__/account-test.ts
buffalojoec: I thought `__type` was to return the schema of the current type being queried. This would be the `type` field that comes back from a JSON parsed account, ie:

```javascript
const {
  data: {
    parsed: {
      info: {
        /* data */
      }
      type: myAccountType
    }
  }
} = await rpc.getAccountInfo(/* jsonParsed /*).send();
```

packages/rpc-graphql/src/__tests__/account-test.ts
buffalojoec: > This stuff should just be on the base type, right?

Was it just for the older versions that `space` was only available with `jsonParsed`? Because I didn't see this in the default account type from the payloads I was checking out.

> Also, why `account.accountSpace`. Were you worried of a collision with the name `account.space`?

Yeah, wanted to make sure, although there's no other accounts in our lib that have `space`, if someone rolled pluggable parsing they didn't have issues. `space` is a pretty generic field name.

packages/rpc-graphql/src/schema/account.ts
buffalojoec: It probably does now, yeah. But I think before 1.16 it wasn't so, right?

packages/rpc-graphql/src/__tests__/account-test.ts
buffalojoec: I think I see what you meant. I took a stab at removing some unnecessary fields from the schema, including:
- `encoding`
- `accountType`

packages/rpc-graphql/src/__tests__/account-test.ts
buffalojoec: nvm, I'm going to cross this bridge if/when we get there. Reverted back to `space`.

buffalojoec:
> Returning this to your queue for some questions!

OK, I pushed up a revised version! Lmk what you think.

packages/rpc-graphql/src/__tests__/account-test.ts
```diff
@@ -304,14 +310,11 @@ describe('account', () => {
             const source = /* GraphQL */ `
                 query testQuery($address: String!) {
                     account(address: $address, encoding: PARSED) {
```
steveluscher, line 312: So, here's how I think about this.

People should just straight up not be bothered with these fetch modes. It should just never surface in the GraphQL API.

I should absolutely not give a shit about specifying an encoding, I should just do this.

```graphql
account(address: "123") {
  ... on MintAccount {
    supply
  }
}
```

…and if that requires that we ask the RPC to parse the data using the `jsonParsed` mode, so be it.

To be pedantic, if we detect the presence of subselections in the query that we can't possibly fulfil except by fetching the thing in `jsonParsed` mode, then we should fetch the account in `jsonParsed` mode. Otherwise, we shouldn't.

If I _really_ want the underlying buffer, then we can offer this:

```graphql
account(address: "123") {
  data(encoding: BASE64)
}
```

Then we fetch it with the base64 encoding.

What if someone does this _totally legal query_?

```graphql
account(address: "123") {
  compressedBase64Data: data(encoding: BASE64_ZSTD)
  base64Data: data(encoding: BASE64_ZSTD)
  ... on MintAccount {
    supply
  }
}
```

Yep. We have to make that underlying fetch _three times_ with the current JSON RPC API.

Except when we rewrite the RPC, we can make this more efficient, and get all of the data in one shot.

packages/rpc-graphql/src/__tests__/account-test.ts
steveluscher: Take it out if there's no non-deprecated way to get at it, or flatten it down to the same level as `blockhash` and give it a good name.

packages/rpc-graphql/src/__tests__/program-accounts-test.ts
```diff
@@ -167,18 +160,17 @@ describe('programAccounts', () => {
...
                 query testQuery($programAddress: String!, $commitment: Commitment, $encoding: AccountEncoding) {
                     programAccounts(programAddress: $programAddress, commitment: $commitment, encoding: $encoding) {
-                        encoding
+                        address
+                        executable
                         ... on AccountBase64Zstd {
-                            address
                             data
-                            executable
                         }
```
steveluscher, line 174: Much preferable to do `data(encoding: AccountEncoding)`.

steveluscher (approved):
I would _really_ eliminate `encoding` from the `account` field entirely, and bury all that complexity in the resolvers. Either way, you could rip out `encoding` in a separate PR.

Let me know if you want to do a call to talk about how this would actually be implemented.

buffalojoec:
### Merge activity

* **Nov 28, 8:18 AM**: @buffalojoec started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1837).

packages/rpc-graphql/src/__tests__/account-test.ts
steveluscher: image

PR #1843, opened 2023-11-14
refactor(experimental): make RPC methods extend `IRpcApiMethods`

buffalojoec:
This PR adds a type for `RpcApiMethods` and ensures all RPC methods defined in
`@solana/rpc-core` extend this type.

This relationship will allow use of these RPC method interfaces in our generic
`createJsonRpcApi(..)` function (next commit).

### DIFF
--- a/packages/rpc-transport/src/apis/api-types.ts
+++ b/packages/rpc-transport/src/apis/api-types.ts
@@ -1,0 +1,11 @@
+export type RpcApiConfig = Readonly<{
+    parametersTransformer?: <T>(params: T, methodName: string) => unknown[];
+    responseTransformer?: <T>(response: unknown, methodName: string) => T;
+}>;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type RpcMethod = (...args: any) => any;
+
+export interface IRpcApiMethods {
+    [methodName: string]: RpcMethod;
+}
--- a/packages/rpc-core/src/rpc-methods/getAccountInfo.ts
+++ b/packages/rpc-core/src/rpc-methods/getAccountInfo.ts
@@ -1,4 +1,5 @@
 import { Address } from '@solana/addresses';
+import type { IRpcApiMethods } from '@solana/rpc-transport';
 import { Commitment } from '@solana/rpc-types';
 
 import {
@@ -31,7 +32,7 @@ type GetAccountInfoApiSliceableCommonConfig = Readonly<{
     dataSlice?: DataSlice;
 }>;
 
-export interface GetAccountInfoApi {
+export interface GetAccountInfoApi extends IRpcApiMethods {
     /**
      * Returns all information associated with the account of provided public key
      */
--- a/packages/rpc-core/src/rpc-methods/getBalance.ts
+++ b/packages/rpc-core/src/rpc-methods/getBalance.ts
@@ -1,11 +1,12 @@
 import { Address } from '@solana/addresses';
+import type { IRpcApiMethods } from '@solana/rpc-transport';
 import { Commitment, LamportsUnsafeBeyond2Pow53Minus1 } from '@solana/rpc-types';
 
 import { RpcResponse, Slot } from './common';
 
 type GetBalanceApiResponse = RpcResponse<LamportsUnsafeBeyond2Pow53Minus1>;
 
-export interface GetBalanceApi {
+export interface GetBalanceApi extends IRpcApiMethods {
     /**
      * Returns the balance of the account of provided Pubkey
      */
--- a/packages/rpc-core/src/rpc-methods/getBlock.ts
+++ b/packages/rpc-core/src/rpc-methods/getBlock.ts
@@ -1,3 +1,4 @@
+import type { IRpcApiMethods } from '@solana/rpc-transport';
 import { Commitment, UnixTimestamp } from '@solana/rpc-types';
 import { Blockhash, TransactionVersion } from '@solana/transactions';
 
@@ -58,7 +59,7 @@ type GetBlockEncoding = 'base58' | 'base64' | 'json' | 'jsonParsed';
 // - These rules apply to both "accounts" and "full" transaction details.
 type GetBlockMaxSupportedTransactionVersion = Exclude<TransactionVersion, 'legacy'>;
 
-export interface GetBlockApi {
+export interface GetBlockApi extends IRpcApiMethods {
     /**
      * Returns identity and transaction information about a confirmed block in the ledger
      */
--- a/packages/rpc-core/src/rpc-methods/getBlockCommitment.ts
+++ b/packages/rpc-core/src/rpc-methods/getBlockCommitment.ts
@@ -1,3 +1,4 @@
+import type { IRpcApiMethods } from '@solana/rpc-transport';
 import { LamportsUnsafeBeyond2Pow53Minus1 } from '@solana/rpc-types';
 
 import { Slot } from './common';
@@ -7,7 +8,7 @@ type GetBlockCommitmentApiResponse = Readonly<{
     totalStake: LamportsUnsafeBeyond2Pow53Minus1;
 }>;
 
-export interface GetBlockCommitmentApi {
+export interface GetBlockCommitmentApi extends IRpcApiMethods {
     /**
      * Returns the amount of cluster stake in lamports that has voted on
      * a particular block, as well as the stake attributed to each vote account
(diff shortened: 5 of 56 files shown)


buffalojoec:
* master
  * **PR #1843** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1843?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #1781** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1781?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #1844** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1844?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #1845** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1845?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1843?utm_source=stack-comment).

packages/rpc-transport/src/apis/api-types.ts
```diff
@@ -0,0 +1,11 @@
+export type RpcApiConfig = Readonly<{
+    parametersTransformer?: <T>(params: T, methodName: string) => unknown[];
+    responseTransformer?: <T>(response: unknown, methodName: string) => T;
+}>;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type RpcMethod = (...args: any) => any;
+
+export interface IRpcApiMethods {
+    [methodName: string]: RpcMethod;
```
steveluscher, line 10: I don't think this actually limits the type as you wish. [Playground link](https://www.typescriptlang.org/play?#code/MYewdgzgLgBARgJwIZgCYwLwwMoE8C2cIANgBQCUA3AFDUCWYUApggGZLBMwCSASgA7AAskygALEOgDe1GHJgBtRClQBdAFwwwTAG4tqAX1oNmbDlwBiIECPGSYTAB7M0EHgOGiJ02fNbWKTR0QOlRDY0YWdk4YACEkBFtvB2cmV3dBJPsZeXgEwJhg0PDqUEhYaABXVlZMLUriYhgkN0qwAGswEAB3MGa3eMSvewAfGCsbYdQaKpqYCCQoOghWOiY3PkypmloAen2Dw73dhwhiEwBaVGWkOGImC+1nC-PtGAABKFx+deAEOn4UAu61eUF2XWBjn452AdCBKFw1C+PxgHiy6CwpAAdDiEgBzCCaBHkTAAPmaYFwOxMUXMGU8dlQAHlxCwYDl5Ap8FNNNB-mA8RpUVtGSUaWYYvEAF7olIuVAbNFTFliNkcuRwJBSgpFMJGeiRCVceJQWXqxTcxm8qD8wWaChkwohPW0UrgaAwcQMAl1MANJotGBtTo9PqB6WysYm9E0L0CtwLJYrNaKkXeFUsSgwfYwaz8CDUIA).

What you want is a ‘marker’ interface – an empty interface that does literally nothing and specifies literally nothing other than to make a bunch of implementers ‘participate’ in it.

packages/rpc-transport/src/apis/api-types.ts
```diff
@@ -0,0 +1,11 @@
+export type RpcApiConfig = Readonly<{
+    parametersTransformer?: <T>(params: T, methodName: string) => unknown[];
+    responseTransformer?: <T>(response: unknown, methodName: string) => T;
+}>;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type RpcMethod = (...args: any) => any;
+
+export interface IRpcApiMethods {
```
steveluscher, line 9: Singular? `IRpcApiMethod`?

steveluscher (changes requested):

packages/rpc-transport/src/apis/api-types.ts
buffalojoec: The interface itself can have a bunch of overloads. But if we consider those to be all one method (they are), then yeah I guess we can do singular.

packages/rpc-transport/src/apis/api-types.ts
buffalojoec: I hated this approach but I did try this initially 💀

packages/rpc-transport/src/apis/api-types.ts
buffalojoec: Actually it works if you just roll the whole type spec into one interface, too. So maybe plural is OK?

Screenshot 2023-11-15 at 1 35 27 PM

lorisleiva: It's a tricky one because `IRpcApiMethods` best describe the type (as you can bundle multiple methods in one API type) but we do want to encourage people to provide one type per method so users can cherry-pick the methods they require. I'm fine with either but I have a slight preference towards to the plural version.

packages/rpc-transport/src/apis/api-types.ts
steveluscher: Take a look at my playground. I added a solution there.

In particular, try removing `extends IRpcMethod` from `interface BarMethod extends IRpcMethod` and you'll see it raises a type error.

buffalojoec: Right I can see how you did it, and the interface I wrote works the same way. All of the methods don't _need_ to extend `IRpcApiMethods`, but it felt weird to just make the first one extend it and then none of the rest. I figured people would be like (..?).

Specifically, I can mimic your playground by _just_ making, for example, `GetAccountInfoApi` extend `IRpcApiMethods` and none of the rest. That results in a valid generic type arg `T` for `createJsonRpcApi<T>(..)`.

buffalojoec: I also don't want to just add some random interface that does nothing, because that results in a property on the API object, which does nothing, and that's not kewl.

steveluscher: > …it felt weird to just make the first one extend it and then none of the rest.

I'm confused. My playground is a bit incoherent, maybe. Let me explain.

There are two parts to it, separated by `//////////`. The bottom part is the bad example, the top the good.

The bad example shows that you can _forget_ to add the marker interface to the second typedef (`BatMethod`) and the thing will still typecheck. That's the thing we want to avoid. All of the typedefs _should_ extend the marker interface.

The good example shows that _if_ you remove the marker interface from even one of the typedefs, a type error is raised. This is what we want – to enforce that anything that participates in that union extends the marker interface `IRpcMethod`.

I'm not just making stuff up; marker interfaces are [a thing](https://en.wikipedia.org/wiki/Marker_interface_pattern).

> …that results in a property on the API object, which does nothing, and that's not kewl.

It doesn't!

buffalojoec: > In particular, try removing `extends IRpcMethod` from `interface BarMethod extends IRpcMethod` and you'll see it raises a type error.

Hey, unless I'm missing something, this isn't the actual behavior. [Playground link](https://www.typescriptlang.org/play?#code/MYewdgzgLgBARgJwIZgCYwLwwMoE8C2cIANgBQCUA3AFDUCWYUApggGZLBMwCSASgA7AAskygALEOgDe1GHJgBtRClQBdAFwwwTAG4tqAX1oNmbDlwBiIECPGSYTAB7M0EHgOGiJ02fNbWKTR0QOlRDWgB6CJgAAycXVAgYmAQmfBA9MJMWdk4YACEkBFtvGBl5eCLAmGDQ8OpQSFhoAFdWVkwtFuJiGCQ3FrAAazAQAHcwPrdC4q97ADIYKxs51BpW9pgIJCg6CFY6Jjc+QRLJShgomGt+CCA)

steveluscher: Uhhh. OK, there are a few things wrong with my sort of braindead playground.

1. Yeah obviously the intersection of the two satisfies `IRpcMethod` (duh); oops – my bad.
2. It's not actually interesting that the _intersection_ of the interfaces satisfies `IRpcMethod`, it's important that each one individually satisfies it.
3. If you change that `&` to a `|` you'll see the type error.

Anyway, this has gone on long enough for nothing. Let's pretend I never said anything; I don't think this typing is adding any value.

steveluscher (approved):

buffalojoec:
### Merge activity

* **Dec 11, 8:31 PM**: @@buffalojoec started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1843).

PR #1850, opened 2023-11-16
refactor(experimental): use new partiallySignTransaction in KeyPairSigner implementation

lorisleiva:
Fix the typecheck error of `@solana/signers` in CI/CD.

### Problem

The implementation of the `KeyPairSigner` uses the `signTransaction` function from `@solana/transactions` under the hood which was recently updated to also assert the transaction is fully signed. Therefore, the mocked return value was giving a type error because it did not provide a `IFullySignedTransaction` as expected.

### Solution

Use the new `partiallySignTransaction` function from `@solana/transactions` instead of `signTransaction` which does not assert the transaction is full signed (and is equivalent to the old behavior of the `signTransaction` function).

### DIFF
--- a/packages/signers/src/__tests__/keypair-signer-test.ts
+++ b/packages/signers/src/__tests__/keypair-signer-test.ts
@@ -2,7 +2,7 @@ import 'test-matchers/toBeFrozenObject';
 
 import { address, getAddressFromPublicKey } from '@solana/addresses';
 import { generateKeyPair, SignatureBytes, signBytes } from '@solana/keys';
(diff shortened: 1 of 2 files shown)


lorisleiva:
* master
  * **PR #1850** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1850?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1850?utm_source=stack-comment).

PR #1853, opened 2023-11-16
refactor(experimental): use Node's instance of `Uint8Array` in browser tests

steveluscher:
# Summary

1. When you're running browser tests, we use the JSDOM environment
2. JSDOM doesn't support `SubtleCrypto` yet, so we inject `node:crypto`
3. Now we're in a situation where `Uint8Array` comes from JSDOM but the `instanceof` checks inside `SubtleCrypto#digest` are made against the Node instance of `Uint8Array`. These checks never pass.

To solve this, we override the `Uint8Array` global in JSDOM with the one from Node. Oh well.

PR #1854, opened 2023-11-16
refactor(experimental): use `@noble/ed25519` in the polyfill

steveluscher:
# Summary

So OMG, the polyfill provides an implementation for the _async_ methods of `SubtleCrypto` but we were using the _sync_ methods of `@noble/curves/ed25519`, whose SHA512 implementation is implemented in _userspace_.

In this PR, we replace it with `@noble/ed25519` which drops the userspace implementation of SHA512 in favour of just using `SubtleCrypto#digest`.

The resulting bundle size of our polyfill?

| Before      | After             |
| ----------- | ----------------- |
| 15524 bytes | 6236 bytes (-59%) |

# Test Plan

```
cd packages/webcrypto-ed25519-polyfill
pnpm test:unit:browser
pnpm test:unit:node
```


steveluscher:
* master
  * **PR #1853** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1853?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1854** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1854?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1854?utm_source=stack-comment).

packages/test-config/jest-unit.config.common.ts
```diff
@@ -22,6 +22,7 @@ const config: Partial<Config.InitialProjectOptions> = {
             },
         ],
     },
+    transformIgnorePatterns: [],
```
steveluscher, line 25: This is here to make Jest work, because `@noble/ed25519` is an ESM-only package. No idea why it works.

steveluscher:
So, we might need to think about this a bit more. `@noble/curves` is an ESM-only module, so I think building CommonJS bundles of `@solana/webcrypto-ed25519-polyfill` is dead on arrival.

Question is, does this matter?

1. People running the new web3.js in tests are running their tests in Node, where the polyfill is not needed at all.
2. The only target that needs this polyfill is the browser, where you have two options
    a. Be a hipster and build your stuff with ESM
    b. Literally include the iife build using a `<script>` tag.

Thoughts?

steveluscher:
Alternate option: I could _inline_ `@noble/ed25519` into every built product.

Pro: No more import/require problems.
Con: If you _happen_ to use `@noble/ed25519` somewhere else in your project, you'll get it twice now.

I think I'm basically answering my own question here. Thoughts?

lorisleiva:
I'd vote for inlining it. People that start using the new web3.js from scratch should have no reason to import these libraries directly I'm guessing so the likelihood of it being imported twice should be low. Right?

buffalojoec:
> So, we might need to think about this a bit more. `@noble/curves` is an ESM-only module, so I think building CommonJS bundles of `@solana/webcrypto-ed25519-polyfill` is dead on arrival.

How have people been dealing with this then with the legacy library?

I don't really know the right answer for an approach, but with this:

> Alternate option: I could _inline_ `@noble/ed25519` into every built product.
> 
> Pro: No more import/require problems. Con: If you _happen_ to use `@noble/ed25519` somewhere else in your project, you'll get it twice now.

...we're left with "here's the new API which is designed to be the maximally smallest bundle possible, except for `@noble/curves` you get that twice".

Does rolling a `@solana/curves` library that just wraps `@noble/curves` in-lined and using it in `@solana/web3.js` help at all? Then we can at least advise use of that library and the redundancy becomes _a little_ more the developer's fault.

steveluscher:
> How have people been dealing with this then with the legacy library?

Whoops, I misspoke. I'd originally written that `@noble/curves` is ESM-only, but I meant to say that `@noble/ed25519` is ESM-only. It's the introduction of the latter that's causing the problem here.

steveluscher:
> People that start using the new web3.js from scratch should have no reason to import these libraries directly I'm guessing so the likelihood of it being imported twice should be low. Right?

Except that `@solana/web3.js-legacy-sham` will also use it, so the likelihood is pretty high.

steveluscher:
I'm going to inline this for anyone doing CJS builds. Oh well.

steveluscher:
### Merge activity

* **Nov 20, 2:23 PM**: @steveluscher started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1854).

steveluscher:
Awwww yissss. 
image

PR #1852, opened 2023-11-16
refactor(experimental-graphql): add simulate transaction query

buffalojoec:
This PR adds transaction simulation support to `@solana/rpc-graphql`. It allows
you to simulate an encoded wire transaction and use GraphQL to interpret the
response.

Note: although this may not be terribly useful at the moment, it can be expanded
on with custom codec plugins to make parsing return data even slicker.

PR #1848, opened 2023-11-15
refactor(experimental-graphql): roll some new type aliases

buffalojoec:
This PR adds some type aliases and other niceties to the GraphQL type spec!

PR #1841, opened 2023-11-13
refactor(experimental-graphql): revise transaction instructions schema (again)

buffalojoec:
This PR introduces similar changes to its parent commit, this time to
transaction instructions.

It also serves as a space to discuss a proper revision to the transaction
instruction schema.

PR #1858, opened 2023-11-17
refactor(experimental-graphql): introduce custom parsing extension

buffalojoec:
This PR is merely a concept for now, but it introduces the idea of custom
parsing "plugins" for the GraphQL API.

Namely, this change would allow users to plug their IDL or Kinobi Tree into
`createJsonRpcGraphQL(..)` in order to customize the GraphQL schema.

Once customized, the schema can support custom IDL-defined types in pure
GraphQL, both in queries _and_ in introspection!

This would be greatly convenient for users wishing to parse out specific program
data that isn't supported by the canonical `jsonParsed`.

---

One major drawback here, which we would need to sort out before
productionalizing such a feature, is the fact that decoding
account/instruction/return data is happening on the client, not the server, and
has an O(n^2) performance rating relative to the size or quantity of IDLs or
Kinobi Trees provided.

Specifically, the library would have to attempt to decode some data string using
each configured decoder until it finds success or reaches the end of the decoder
set.

Even more troublesome, it _could_ decode something incorrectly and just not
fail, meaning you'd have the wrong data.

---

Regardless, if the community finds this feature useful - such as explorers,
indexers, etc. - we can explore optimizations.

### DIFF
--- a/packages/rpc-graphql/src/schema/account.ts
+++ b/packages/rpc-graphql/src/schema/account.ts
@@ -3,6 +3,7 @@ import { Address } from '@solana/addresses';
 import { DataSlice, Slot } from '@solana/rpc-core/dist/types/rpc-methods/common';
 
 import { resolveAccount } from '../resolvers/account';
+import { AstConfig } from './ast';
 
 export type AccountQueryArgs = {
     address: Address;
@@ -12,7 +13,7 @@ export type AccountQueryArgs = {
     minContextSlot?: Slot;
 };
 
-export const accountTypeDefs = /* GraphQL */ `
+const accountTypeDefs = /* GraphQL */ `
     # Account interface
     interface Account {
         address: Address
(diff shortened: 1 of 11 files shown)


buffalojoec:
* master
  * **PR #1848** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1848?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1837** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1837?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #1841** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1841?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #1852** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1852?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
          * **PR #1858** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1858?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1858?utm_source=stack-comment).

packages/rpc-graphql/src/schema/account.ts
```diff
@@ -191,87 +192,104 @@ export const accountTypeDefs = /* GraphQL */ `
...
+                    }
+                    if (account.accountType === 'lookupTable' && account.programName === 'address-lookup-table') {
+                        return 'LookupTableAccount';
+                    }
+                    return resolveCustomAccountType(account);
                 }
-            }
-            return 'AccountBase64';
+                return 'AccountBase64';
+            },
```
buffalojoec, line 232: This whole `__resolveType(..)` block didn't change, I just added this injectable resolve function at the end of the conditional.

packages/rpc-graphql/src/schema/ast/index.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+function buildSchemaAstConfig(programAst: ProgramAstSource): SchemaAstConfig {
+    const parsedAst: ProgramAst = typeof programAst === 'string' ? JSON.parse(programAst) : programAst;
+    return {
+        accounts: {
+            resolvers: {},
+            typeDefs: parsedAst.accounts.map(buildAccountGraphQLTypeDef).join('\n'),
+            // TODO: Conditional function for mapping decoded data.
+            resolveTypeExtension: (_account): string => null as unknown as string,
```
buffalojoec, line 24: This would basically be that conditional mapping, taking a look at the parsed data and determining which type it is. Haven't 100% thought about how this would work, but ideally it should align with the base `jsonParsed` format before being transformed into it's proper GraphQL type.

packages/rpc-graphql/src/schema/instruction.ts
```diff
@@ -873,718 +874,736 @@ export const instructionTypeDefs = /* GraphQL */ `
...
+                            return 'VoteUpdateCommissionInstruction';
+                        }
+                        if (instruction.instructionType === 'voteSwitch') {
+                            return 'VoteVoteSwitchInstruction';
+                        }
+                        if (instruction.instructionType === 'authorizeChecked') {
+                            return 'VoteAuthorizeCheckedInstruction';
+                        }
+                    }
+                    return resolveCustomInstructionType(instruction);
```
buffalojoec, line 1162: Same here, this whole `__resolveType(..)` block didn't change, I just added this injectable resolve function at the end of the conditional.

buffalojoec (commented):
Again, just a concept, but here's a lot of the stuff I was chatting with @lorisleiva about!

If you take a look at `ast-test.ts`, you can see how this would look in action!

lorisleiva:
Nice! I think the defined AST is a very small subset of the Kinobi tree or even the IDL but it's a nice prototype.

I'm a bit confused by the following statement:

> Specifically, the library would have to attempt to decode some data string using
each configured decoder until it finds success or reaches the end of the decoder
set.

If we have a spec that tells us how each account, instruction and types are defined, why would we need to attempt to decode using multiple decoders?

buffalojoec:
> If we have a spec that tells us how each account, instruction and types are defined, why would we need to attempt to decode using multiple decoders?

Good call-out. For the standard queries, you're right. We'd have the information we need if we:
- Define instructions in GraphQL and use the specific account types where they pop up in instructions
  - This would allow the resolver to know exactly which codec to use
- Use the program's name, ID, etc. as context for things like gPA
  - Here we can likely also use the GpaBuilder stuff you shared with me

The "drawback" use-case I had in mind was "I've found this account in a System Program transfer instruction and I have no clue what type of account it is, or if it's even one of mine, so I'm going to query it and then the resolver is going to try a million times until one works" 💀"

This last bit is where I don't have a solution, and people will try it.

lorisleiva:
IMO, if we want to support that, we've got no choice but to have a centralised storage of IDLs / Kinobi trees or any schema standard we want to support. Which is totally doable.

buffalojoec:
> IMO, if we want to support that, we've got no choice but to have a centralised storage of IDLs / Kinobi trees or any schema standard we want to support. Which is totally doable.

I really thought your message was going to end with "which is insane". 😅

buffalojoec:
I just wanted there to be a record of this change and open it up for discussion. I can circle back to it in the future, but for now I don't think there's a need to add this.

PR #1859, opened 2023-11-17
refactor(experimental): introduce a shim library for `Connection`

buffalojoec:
This PR introduces a shim library for the legacy `Connection` class, which is
likely going to be one of the most difficult obstacles for dApp developers to
overcome when migrating from `@solana/web3.js` legacy to 2.0.0.

In this library, I've introduced a function `createLegacyConnectionShim(..)`,
which takes the new RPC client (`Rpc<SolanaRpcMethods>`) and likely some other
inputs and returns an object that implements the interface of the `Connection`
class.

Although this object is relatively just as large as the legacy `Connection`
object, it has the following bundling advantages:
- Zero dependencies from the legacy library come with this object
- It runs on the rails of the new RPC client, so it's really only just a shell

However, the main benefit here is the ability to preserve existing APIs while
migrating to the new Web3.js. Developers can _immediately_ start using many of
the goodies the new API brings, such as custom transports, failover, and
extended RPC APIs without having to gut their entire app (but they still should,
eventually).

**Root:** [`packages/library-legacy-shim/src/connection.ts`](https://github.com/solana-labs/solana-web3.js/pull/1859/files#diff-9c05f0bcefb2ed5fb99a2caef113a7fe76797da778735ee8a78ff9bc5321bd50)


buffalojoec:
* master
  * **PR #1859** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1859?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1859?utm_source=stack-comment).

buffalojoec:
cc @acheroncrypto

Although we discussed your release strategy, here's something you could use in the meantime if it was of interest to you.

lorisleiva:
Sorry if that's a noob question but I'm a little confused by how this looks for the end-user. It seems like the library is still importing and overriding the `Connection` prototype from the legacy library. Does this means that in order for this shim to work, they'll need to download both the legacy and the new web3.js?

packages/library-legacy-shim/src/connection.ts
```diff
@@ -0,0 +1,33 @@
...
+    const connectionPrototype = Connection.prototype;
+    // TODO: Also need to handle other keys, like fields.
+    const methodNames = Object.getOwnPropertyNames(connectionPrototype).filter(
+        name => typeof connectionPrototype[name] === 'function' && name !== 'constructor'
+    ) as (keyof Connection)[];
+    const shim = {};
+    methodNames.forEach(methodName => {
+        shim[methodName] = async (...params: unknown[]) => processRequest(rpc, methodName, params);
+    });
+    return shim as InstanceType<typeof Connection>;
```
acheroncrypto, line 32: Does this not result in both v1 and v2 getting included in the bundle?

acheroncrypto (commented):
> cc @acheroncrypto
>
> Although we discussed your release strategy, here's something you could use in the meantime if it was of interest to you.

Thanks, this is a great interest to me.

buffalojoec:
@lorisleiva @acheroncrypto replying to both of you, I think I just have it configured wrong at the moment, but I believe if you only use the **type** `Connection` to make this shim, legacy gets tree-shaken away.

This was the intention.

packages/library-legacy-shim/src/response.ts
```diff
@@ -0,0 +1,84 @@
...
+    literal,
+    nullable,
+    number,
+    optional,
+    string,
+    Struct,
+    type as pick,
+    union,
+    unknown,
+} from 'superstruct';
```
steveluscher, line 17: Can we just… not do this validation?

steveluscher:
Crap, I wish I'd published my stack sooner. The way we're going to have to do this is to provide a _completely_ interface compatible version of _parts_ of web3.js (just the parts that the dependencies we care about actually use) and the only thing we ask of devs is to add the sham to their package.json as an npm `override`. See the README in [this PR](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1861/refactor-experimental-create-a-sham-for-PublicKey).

As of right now the stack that PR is a part of is nearly a complete drop-in replacement for `@solana/wallet-adapter`. The `Connection` sham is what I was missing and working on next. If you want to work on it instead, that's fine!

packages/library-legacy-shim/src/response.ts
buffalojoec: We can!

buffalojoec:
Closing in favor of new efforts by @steveluscher on the "sham"!

PR #1856, opened 2023-11-17
refactor(experimental): rename codec options to codec config

lorisleiva:
Rename `options` to `config` in the `@solana/codecs-*` and `@solana/options` libraries to stay consistent with the rest of the library.

PR #1865, opened 2023-11-18
refactor(experimental): pre-build the byte array before encoding with codecs

lorisleiva:
_Note this is a draft PR focusing on `codecs-core` to gather feedback, I will continue to update the other libraries afterwards._

This PR updates the codecs API such that both encoding and decoding function **have access to the entire byte array**. Let’s first see the change this PR introduce and then see why this is a valuable change.

## API Changes

- **Encode**: The `Encoder` type contains a new function `write`. Contrary to the `encode` function which creates a new `Uint8Array` and returns it directly, the `write function` updates the provided `bytes` argument at the provided `offset`. It then returns the next offset that should be written to.

  ```ts
  // Before
  type Encoder<T> = {
    encode: (value: T) => Uint8Array;
    // ...
  };

  // After
  type Encoder<T> = {
    encode: (value: T) => Uint8Array;
    write: (value: T, bytes: Uint8Array, offset: Offset) => Offset;
    // ...
  };
  ```
  
  A new `createEncoder` function was provided to automatically fill the `encode` function from the `write` function.

  ```ts
  const myU8Encoder = createEncoder({
    fixedSize: 1,
    write: (value: number, bytes: Uint8Array, offset: Offset) {
      bytes.set(value, offset);
      return offset + 1;
    };
  });
  ```
  

- **Decode**: The `decode` function already following a similar approach by using offsets. The newly added function `read` takes over this responsibility. The only difference is we now make the offset a mandatory argument to stay consistent with the `write` function. The `decode` function now becomes syntactic sugar for accessing the value directly.

  ```ts
  // Before
  type Decoder<T> = {
    decode: (bytes: Uint8Array, offset?: Offset) => [T, Offset];
    // ...
  };

  // After
  type Decoder<T> = {
    decode: (bytes: Uint8Array, offset?: Offset) => T;
    read: (bytes: Uint8Array, offset: Offset) => [T, Offset];
    // ...
  };
  ```

  Similarly to the `Encoder` changes, a new `createDecoder` function is provided to fill the `decode` function using the `read` function.

  ```ts
  const myU8Decoder = createDecoder({
    fixedSize: 1,
    read: (bytes: Uint8Array, offset: Offset) {
      return [bytes[offset], offset + 1];
    };
  });
  ```

- **Sizes**: Because we now need to pre-build the entire byte array that will be encoded, we need a way to find the variable size of a given value. We introduce a new `variableSize` function and narrow the types so that it can only be provided when `fixedSize` is `null`.

  ```ts
  // Before
  type Encoder<T> = {
    fixedSize: number | null;
    maxSize: number | null;
  }

  // After
  type Encoder<T> = { ... } & (
    | { fixedSize: number; }
    | { fixedSize: null; variableSize: (value: T) => number; maxSize?: number }
  )
  ```

  We do something similar the `Decoder` except that this one doesn’t need to know about the variable size (it would make no sense as the type parameter `T` for decoder refers to the decoded type and not the type to encode).

  ```ts
  // Before
  type Decoder<T> = {
    fixedSize: number | null;
    maxSize: number | null;
  }

  // After
  type Decoder<T> = { ... } & (
    | { fixedSize: number; }
    | { fixedSize: null; maxSize?: number }
  )
  ```

- **Description**: This PR takes this refactoring opportunity to remove the `description` attribute from the codecs API which brings little value to the end-user.

## Why?

- **Consistent API**: The implementation of the `encode` / `decode` and `write` / `read` functions are now consistent with each other. Before one was using offsets to navigate through an entire byte array and the other was returning and merge byte arrays together. Now they both use offsets to navigate the byte array to encode or decode.

- **Performant API**: By pre-building the byte array once, we’re avoiding creating multiple instance of byte arrays and merging them together.

- **Non-linear serialisation**: The main reason why it’s important for the `encode` method to have access to the entire encoded byte array is that it allows us to offer more complex codec primitives that are able to jump back and forth within the buffer. Without it, we are locking ourselves to only supporting serialisation strategies that are read linearly which isn’t always the case. For instance, imagine we have the size of an array being stored at the very beginning of the account whereas the items themselves are being stored at the end. Because we now have the full byte array when encoding, we can push the size at the beginning whilst inserting all items at the requested offset. We could even offer a `getOffsetCodec` (or similar) that allows us to shift the offset forward or backward to compose more complex, non-linear data structures. This would be simply impossible with the current format of the `encode` function.

### DIFF
--- a/packages/codecs-core/src/__tests__/codec-test.ts
+++ b/packages/codecs-core/src/__tests__/codec-test.ts
@@ -1,75 +1,84 @@
-import { Codec, Decoder, Encoder } from '../codec';
+import { Codec, createCodec, createDecoder, createEncoder, Decoder, Encoder } from '../codec';
 
 describe('Encoder', () => {
     it('can define Encoder instances', () => {
-        const myEncoder: Encoder<string> = {
+        const myEncoder: Encoder<string> = createEncoder({
             description: 'myEncoder',
-            encode: (value: string) => {
-                const bytes = new Uint8Array(32).fill(0);
-                const charCodes = [...value.slice(0, 32)].map(char => Math.min(char.charCodeAt(0), 255));
-                bytes.set(new Uint8Array(charCodes));
-                return bytes;
-            },
             fixedSize: 32,
-            maxSize: 32,
-        };
+            write: (value: string, bytes, offset) => {
+                const charCodes = [...value.slice(0, 32)].map(char => Math.min(char.charCodeAt(0), 255));
+                bytes.set(charCodes, offset);
+                return offset + 32;
+            },
+        });
 
         expect(myEncoder.description).toBe('myEncoder');
         expect(myEncoder.fixedSize).toBe(32);
         expect(myEncoder.maxSize).toBe(32);
+        expect(myEncoder.getSize('hello')).toBe(32);
 
         const expectedBytes = new Uint8Array(32).fill(0);
         expectedBytes.set(new Uint8Array([104, 101, 108, 108, 111]));
         expect(myEncoder.encode('hello')).toStrictEqual(expectedBytes);
+
+        const writtenBytes = new Uint8Array(32).fill(0);
+        expect(myEncoder.write('hello', writtenBytes, 0)).toBe(32);
+        expect(writtenBytes).toStrictEqual(expectedBytes);
     });
 });
 
 describe('Decoder', () => {
     it('can define Decoder instances', () => {
-        const myDecoder: Decoder<string> = {
-            decode: (bytes: Uint8Array, offset = 0) => {
+        const myDecoder: Decoder<string> = createDecoder({
+            description: 'myDecoder',
+            fixedSize: 32,
+            read: (bytes: Uint8Array, offset) => {
                 const slice = bytes.slice(offset, offset + 32);
                 const str = [...slice].map(charCode => String.fromCharCode(charCode)).join('');
                 return [str, offset + 32];
             },
-            description: 'myDecoder',
-            fixedSize: 32,
-            maxSize: 32,
-        };
+        });
 
         expect(myDecoder.description).toBe('myDecoder');
         expect(myDecoder.fixedSize).toBe(32);
         expect(myDecoder.maxSize).toBe(32);
-        expect(myDecoder.decode(new Uint8Array([104, 101, 108, 108, 111]))).toStrictEqual(['hello', 32]);
+
+        expect(myDecoder.decode(new Uint8Array([104, 101, 108, 108, 111]))).toBe('hello');
+        expect(myDecoder.read(new Uint8Array([104, 101, 108, 108, 111]), 0)).toStrictEqual(['hello', 32]);
     });
 });
 
 describe('Codec', () => {
     it('can define Codec instances', () => {
-        const myCodec: Codec<string> = {
-            decode: (bytes: Uint8Array, offset = 0) => {
+        const myCodec: Codec<string> = createCodec({
+            description: 'myCodec',
+            fixedSize: 32,
+            read: (bytes: Uint8Array, offset) => {
                 const slice = bytes.slice(offset, offset + 32);
                 const str = [...slice].map(charCode => String.fromCharCode(charCode)).join('');
                 return [str, offset + 32];
             },
-            description: 'myCodec',
-            encode: (value: string) => {
-                const bytes = new Uint8Array(32).fill(0);
+            write: (value: string, bytes, offset) => {
                 const charCodes = [...value.slice(0, 32)].map(char => Math.min(char.charCodeAt(0), 255));
-                bytes.set(new Uint8Array(charCodes));
-                return bytes;
+                bytes.set(charCodes, offset);
+                return offset + 32;
             },
-            fixedSize: 32,
-            maxSize: 32,
-        };
+        });
 
         expect(myCodec.description).toBe('myCodec');
         expect(myCodec.fixedSize).toBe(32);
         expect(myCodec.maxSize).toBe(32);
+        expect(myCodec.getSize('hello')).toBe(32);
 
         const expectedBytes = new Uint8Array(32).fill(0);
         expectedBytes.set(new Uint8Array([104, 101, 108, 108, 111]));
         expect(myCodec.encode('hello')).toStrictEqual(expectedBytes);
-        expect(myCodec.decode(new Uint8Array([104, 101, 108, 108, 111]))).toStrictEqual(['hello', 32]);
+
+        const writtenBytes = new Uint8Array(32).fill(0);
+        expect(myCodec.write('hello', writtenBytes, 0)).toBe(32);
+        expect(writtenBytes).toStrictEqual(expectedBytes);
+
+        expect(myCodec.decode(new Uint8Array([104, 101, 108, 108, 111]))).toBe('hello');
+        expect(myCodec.read(new Uint8Array([104, 101, 108, 108, 111]), 0)).toStrictEqual(['hello', 32]);
     });
 });
--- a/packages/codecs-core/src/__tests__/codec-test.ts
+++ b/packages/codecs-core/src/__tests__/codec-test.ts
@@ -3,7 +3,6 @@ import { Codec, createCodec, createDecoder, createEncoder, Decoder, Encoder } fr
 describe('Encoder', () => {
     it('can define Encoder instances', () => {
         const myEncoder: Encoder<string> = createEncoder({
-            description: 'myEncoder',
             fixedSize: 32,
             write: (value: string, bytes, offset) => {
                 const charCodes = [...value.slice(0, 32)].map(char => Math.min(char.charCodeAt(0), 255));
@@ -12,10 +11,7 @@ describe('Encoder', () => {
             },
         });
 
-        expect(myEncoder.description).toBe('myEncoder');
         expect(myEncoder.fixedSize).toBe(32);
-        expect(myEncoder.maxSize).toBe(32);
-        expect(myEncoder.getSize('hello')).toBe(32);
 
         const expectedBytes = new Uint8Array(32).fill(0);
         expectedBytes.set(new Uint8Array([104, 101, 108, 108, 111]));
@@ -30,7 +26,6 @@ describe('Encoder', () => {
 describe('Decoder', () => {
     it('can define Decoder instances', () => {
         const myDecoder: Decoder<string> = createDecoder({
-            description: 'myDecoder',
             fixedSize: 32,
             read: (bytes: Uint8Array, offset) => {
                 const slice = bytes.slice(offset, offset + 32);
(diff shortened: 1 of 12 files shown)


lorisleiva:
* master
  * **PR #1856** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1856?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1865** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1865?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1865?utm_source=stack-comment).

packages/codecs-core/src/codec.ts
```diff
@@ -59,3 +79,26 @@ export type BaseCodecOptions = {
...
+ * Get the encoded size of a given value in bytes.
+ */
+export function getEncodedSize<T>(value: T, encoder: Encoder<T>): number {
+    return encoder.fixedSize !== null ? encoder.fixedSize : encoder.variableSize(value);
+}
+
+/**
+ * Use the provided Encoder to encode the given value to a `Uint8Array`.
+ */
+export function encode<T>(value: T, encoder: Encoder<T>): Uint8Array {
```
mcintyre94, line 93: I think my only concern would be that encode and decode are really generic terms for us to be exporting into people's apps, where stuff like getTransactionEncoder is less likely to clash with anything they're doing
They're definitely the right names to use for this though, so not really suggesting any change there!

mcintyre94 (commented):
Just a clarifying question, when you say "offer more complex codec primitives that are able to jump back and forth within the buffer" do we have a use case for that more complex implementation now? Or is it just something you've recognised as a limitation of the current ones?

Also am I right in thinking the performance improvement would increase based on how much encoders/decoders are composed? I think that's potentially a really important motivation too because things like the transaction decoder are pretty complex

packages/codecs-core/src/codec.ts
lorisleiva: Thanks for the review Callum!

**Regarding the `encode` and `decode` functions**: I agree they are pretty generic. I’d love to keep the `codec.encode` and `codec.decode` API which we could. It just comes with its own set of challenges as codecs will now need to provide 4 functions such that two of them are just delegating to the other two. On the plus side, we no longer have that annoying `[0]` suffix on the main API.

**Regarding offering more complex primitives**: The main issue here is that the current `encode` and `decode` logic are not symmetrical. Meaning there are things we can do with `decode` that we cannot do with `encode` because we do not have access to the entire buffer. I do have some examples of that. For instance, imagine an `offsetCodec(codec, relativeOffset)` helper that, given a codec, shifts its position on the buffer by the given relative offset. We can do that right now with `Decoders` but it’s impossible to do with `Encoders` because they use a different mechanism that rely on linear serialisation only. An example of why we would we need something like an `offsetCodec` helper is any array such that its size does not immediately prefix the items.

**Regarding performance improvements**: The decoding logic should stay pretty much the same because it already uses a single buffer but the encoding logic should drastically use less instances of `Uint8Array` which I believe will be a big performance boost.

buffalojoec:
> Another way to achieve the same thing but keep the previous `encoder.encode(value)` API is to introduce a new set of low-level methods such as `write` and `read` on the `Encoder` and `Decoder` types respectively. However, this means every single `Encoder` and `Decoder` needs to implement the same `encode` and `decode` helper method over and over again. It also makes it easy for people to misuse it and use the `encode` method instead of the `write` method when chaining encoders together.

I think this sounds like a better approach to me. This API aligns with `borsh` in Rust (not that we care to align with `borsh` necessarily) so it's not unfamiliar, and I would really hate to see any partial destruction of the simple JavaScript syntax that is:

```ts
const bytes: UInt8Array = getStructCodec(..).encode(myObject);
```

If the intention is to provide better low-level functionality, better to do that with `write` and let people still get the slick API of `encode`.

I wouldn't even hate just adding offset to the return type of `encode`, but that's tripped me up with `decode` before, too. I'd probably rather see the offset removed from `decode` to match `encode`, and instead used in `read`.

> * **Performant API**: By pre-building the byte array once, we’re avoiding creating multiple instance of byte arrays and merging them together.

I understand you're also going for performance here, and you're right it the current implementation does degrade performance/memory, however, I think having a simple, user-friendly API alongside a lower-level performant one is more than enough.

buffalojoec:
> Thanks Joe! I agree, it would make me sad loosing the `getStructCodec(..).encode(myObject)` API.
> 
> I also like the alternative API, it's just a little bit more of a pain to write codecs with this API. That being said, I can offer helper methods like `createEncoder` and `createDecoder` that accepts an `Encoder` or `Decoder` with `write` and `read` function and provide the `encode` and `decode` function for you.
> 
> So for instance, if I wanted to create a number encoder. instead of this:
> 
> ```ts
> const myEncoder: Encoder<number> = {
>   fixedSize: 1,
>   encode(value, bytes, offset): Offset {
>     bytes.set(value, offset);
>     return offset + 1;
>   }
> }
> ```
> 
> I'd be doing this:
> 
> ```ts
> const myEncoder: Encoder<number> = createEncoder({
>   fixedSize: 1,
>   write(value, bytes, offset): Offset {
>     bytes.set(value, offset);
>     return offset + 1;
>   }
> })
> 
> // `createEncoder` basically just adds the `encode` function for us:
> myEncoder.encode(42); // <- Uint8Array.
> ```
> 
> What do you think?

Yep, I think that's great! Lgtm 💪🏼

mcintyre94:
Agreed, that looks like a really nice API to me. Only question is how easy is it to compose eg a struct encoder using that? Would you be internally using eg the `write` function of the nested fields, and the `encode` function is just the external API?

lorisleiva:
> Agreed, that looks like a really nice API to me. Only question is how easy is it to compose eg a struct encoder using that? Would you be internally using eg the `write` function of the nested fields, and the `encode` function is just the external API? 

That's right, codecs that are composed from other codecs should only use the `write` and `read` functions internally. The `encode` and `decode` functions are syntactic sugar for the end-user.

packages/codecs-core/src/codec.ts
```diff
@@ -4,36 +4,63 @@
...
+          /** The fixed size of the encoded value in bytes, if applicable. */
+          readonly fixedSize: number;
+      }
+    | {
+          /** Otherwise, a null fixedSize indicates it's a variable size encoder. */
+          readonly fixedSize: null;
+          /** The maximum size an encoded value can be in bytes, if applicable. */
+          readonly maxSize?: number;
+          /** The total size of the encoded value in bytes. */
+          readonly variableSize: (value: T) => number;
```
steveluscher, line 33: This might confuse people because ‘variable’ is both a noun and a verb. `getSizeForValue()`?

packages/codecs-core/src/__tests__/codec-test.ts
```diff
@@ -1,75 +1,73 @@
...
-                const bytes = new Uint8Array(32).fill(0);
+        const myEncoder: Encoder<string> = createEncoder({
+            fixedSize: 32,
+            write: (value: string, bytes, offset) => {
                 const charCodes = [...value.slice(0, 32)].map(char => Math.min(char.charCodeAt(0), 255));
-                bytes.set(new Uint8Array(charCodes));
-                return bytes;
+                bytes.set(charCodes, offset);
+                return offset + 32;
             },
```
steveluscher, line 11: Is there any reason to make this context-y with `this`?
```suggestion
            write(value: string, bytes, offset) {
                const length = this.fixedSize;
                const charCodes = [...value.slice(0, length)].map(char => Math.min(char.charCodeAt(0), 255));
                bytes.set(charCodes, offset);
                return offset + length;
            },
```

packages/codecs-core/src/codec.ts
```diff
@@ -4,36 +4,63 @@
...
+ * An object that can encode a value to a `Uint8Array`.
+ */
+export type Encoder<T> = EncoderSize<T> & {
+    /** Encode the provided value and return the encoded bytes directly. */
+    readonly encode: (value: T) => Uint8Array;
+    /**
+     * Writes the encoded value into the provided byte array at the given offset.
+     * Returns the offset of the next byte after the encoded value.
+     */
+    readonly write: (value: T, bytes: Uint8Array, offset: Offset) => Offset;
```
steveluscher, line 16: You could achieve the context-y suggestion above by doing something like this:


```suggestion
    readonly write: (
        this: Readonly<{fixedSize: number | null}>,
        value: T,
        bytes: Uint8Array, offset: Offset,
    ) => Offset;
```

Then something like this later:

```ts
const context = {fixedSize};
return {
  // ...
  write(...args) { return encoder.write.apply(context, args); },
}
```

steveluscher (approved):

packages/codecs-core/src/codec.ts
lorisleiva: I'm okay with `getSizeForValue`, I'll do the change in a separate PR.

lorisleiva: See https://github.com/solana-labs/solana-web3.js/pull/1901

packages/codecs-core/src/codec.ts
```diff
@@ -4,36 +4,63 @@
...
-    encode: (value: T) => Uint8Array;
-};
+export type EncoderSize<T> =
+    | {
+          /** The fixed size of the encoded value in bytes, if applicable. */
+          readonly fixedSize: number;
+      }
+    | {
+          /** Otherwise, a null fixedSize indicates it's a variable size encoder. */
+          readonly fixedSize: null;
```
buffalojoec, line 29: It strikes me as a bit confusing, if these are being exported, to have the `fixedSize` field on a variable sized encoder.

I can see you're using it for some neat assertion management, but I wonder why you chose not to just check for `undefined` instead?

buffalojoec (commented):
Looking slick! One question from me popped up.

packages/codecs-core/src/codec.ts
lorisleiva: It's a fair point. It makes juggling with sizes a bit more complicated when creating composite codecs (like array, tuples, etc.) but at the benefit of offering a cleaner API for the end-user which IMO is more valuable.

I'll work on that in a separate PR. Thanks!

buffalojoec: Sweet, sounds good!

buffalojoec (approved):

mcintyre94 (approved):
This looks good to me now! I think the `createEncoder` and `createDecoder` and keeping the `encode`/`decode` is much nicer too.

lorisleiva:
### Merge activity

* **Dec 1, 10:16 AM**: @lorisleiva started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1865).

PR #1873, opened 2023-11-21
refactor(experimental): helper functions that add signers to instructions and transactions

lorisleiva:
This PR provides two helper methods `addSignersToInstruction` and `addSignersToTransaction` which, given an array of signers, go through the account metas of the provided object and attach signers to them when applicable.

For an account meta to match a provided signer it:

-   Must have a signer role (`AccountRole.READONLY_SIGNER` or `AccountRole.WRITABLE_SIGNER`).
-   Must have the same address as the provided signer.
-   Must not have an attached signer already.

```ts
const myInstruction: IInstruction = {
    accounts: [
        { address: '1111' as Address, role: AccountRole.READONLY_SIGNER },
        { address: '2222' as Address, role: AccountRole.WRITABLE_SIGNER },
    ],
    // ...
};

const signerA: TransactionSigner<'1111'>;
const signerB: TransactionSigner<'2222'>;
const myInstructionWithSigners = addSignersToInstruction([mySignerA, mySignerB], myInstruction);

// myInstructionWithSigners.accounts[0].signer === mySignerA
// myInstructionWithSigners.accounts[1].signer === mySignerB
```

It also redefines the `IInstructionWithSigners` type and introduce a new analogous `ITransactionWithSigners` type for transactions. Both of these types _add_ to the type definition of an instruction/transaction instead of defining the entire object like they were before.

Meaning you can now do `IInstruction & IInstructionWithSigners` or `BaseTransaction & ITransactionWithSigners` to compose types together just like you can with `ITransactionWithSignatures`.

This improvement was suggested by the reviews of https://github.com/solana-labs/solana-web3.js/pull/1792.

### DIFF
--- a/packages/signers/src/add-signers.ts
+++ b/packages/signers/src/add-signers.ts
@@ -1,0 +1,43 @@
+import { IInstruction, isSignerRole } from '@solana/instructions';
+import { BaseTransaction } from '@solana/transactions';
+
+import { IAccountSignerMeta, IInstructionWithSigners, ITransactionWithSigners } from './account-signer-meta';
+import { deduplicateSigners } from './deduplicate-signers';
+import { TransactionSigner } from './transaction-signer';
+
+/** Attaches the provided signers to the account metas of an instruction when applicable. */
+export function addSignersToInstruction<TInstruction extends IInstruction>(
+    signers: TransactionSigner[],
+    instruction: TInstruction | (TInstruction & IInstructionWithSigners)
+): TInstruction & IInstructionWithSigners {
+    if (!instruction.accounts || instruction.accounts.length === 0) {
+        return instruction as TInstruction & IInstructionWithSigners;
+    }
+
+    const signerByAddress = new Map(deduplicateSigners(signers).map(signer => [signer.address, signer]));
+    return Object.freeze({
+        ...instruction,
+        accounts: instruction.accounts.map(account => {
+            const signer = signerByAddress.get(account.address);
+            if (!isSignerRole(account.role) || 'signer' in account || !signer) {
+                return account;
+            }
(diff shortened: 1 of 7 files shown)


packages/signers/src/add-signers.ts
```diff
@@ -0,0 +1,42 @@
...
+/** Attaches the provided signers to the account metas of an instruction when applicable. */
+export function addSignersToInstruction<TInstruction extends IInstruction>(
+    signers: TransactionSigner[],
+    instruction: TInstruction | (TInstruction & IInstructionWithSigners)
+): TInstruction & IInstructionWithSigners {
+    if (!instruction.accounts || instruction.accounts.length === 0) {
+        return instruction as TInstruction & IInstructionWithSigners;
+    }
+
+    const signerByAddress = new Map(signers.map(signer => [signer.address, signer]));
```
steveluscher, line 16: What should we do if someone hands you >1 signer for the same address? This code will just choose the last one and drop the rest.

steveluscher (approved):
One nit about colliding signers that you might like to do something about (fatal?) and write a test to cover.

packages/signers/src/add-signers.ts
lorisleiva: How about we plug in the `deduplicateSigners` logic in there? That way it will deduplicate by reference but fail if distinct signers are provided for the same address.

lorisleiva: I updated the code and added a test for it. ✅

lorisleiva:
### Merge activity

* **Nov 22, 6:19 AM**: @lorisleiva started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1873).

PR #1875, opened 2023-11-21
refactor(experimental): export from messages in @solana/transactions

mcintyre94:
This allows consumers to perform signing logic manually:

```ts
const compiledMessage = compileMessage(transaction);
const bytes = getCompiledMessageEncoder().encode(compiledMessage);

// app logic
const signature = signTransaction(bytes);
```

This is probably not a super common use case, but it's necessary for eg a wallet that doesn't necessarily have `CryptoKeyPair` objects available for accounts it is signing for, and therefore can't use our `signTransaction` helper function.

### DIFF
--- a/packages/transactions/src/index.ts
+++ b/packages/transactions/src/index.ts
@@ -4,6 +4,7 @@ export * from './create-transaction';
 export * from './durable-nonce';
 export * from './fee-payer';
 export * from './instructions';
+export * from './message';
 export * from './serializers';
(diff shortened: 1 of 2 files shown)


lorisleiva (approved):
Yes, thank you! I actually would benefit from this for the KeyPair implementation of the Signer API. Because it's not exported right now, I'm having to use `signTransaction` and extract its signature.

PR #1879, opened 2023-11-22
refactor(experimental): improve robustness of publish-packages

### LINKED ISSUE #1402: Ignore already published NPM packages
Our CI script publishes NPM packages after running builds and tests.

# Problem

If a build fails halfway through publishing _some_ of the packages, you can't run it again. It will _always_ fail with:

> You cannot publish over the previously published versions

# Proposed solution

Figure out how not to fail the build in cases where an npm package has already been published from a prior run.

mcintyre94:
This PR reworks our `publish-packages` flow to hopefully be more robust and less annoying

Previously, we ran all the `build` and `publish-packages` jobs in parallel, which meant that you could have a build job fail (eg we have some flaky tests sometimes) after some packages had published. This doesn't sound like a problem, but if you re-ran the job then it would attempt to re-publish those packages and that would fail. This ultimately meant that if a `build` job failed you could probably never get to a point where all packages were published by re-running the CI.

This PR adds the `--continue` flag to the `turbo run publish-packages` line. This means that if a job fails then it is ignored and other independent jobs can continue. This means that if you need to publish 20 packages and 10 are already published, the unpublished ones can be published.

However we don't want to ignore build failures before publishing, so I've separated the build stage and run it without the `--continue` flag, before the `publish-packages` stage. This runs with `--force=true` to match the `cache: false` behaviour of the `publish-packages` turbo job. 

The result of this is likely that the publish workflow will be a bit slower (it no longer publishes packages while others build), but more predictable.

- If any build stage fails, no packages will be published. The pipeline can be safely retried
- If for some reason some packages fail to publish, we can re-run and those already published will be ignored instead of blocking publishing of the unpublished ones

Fixes #1402

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -38,7 +38,6 @@
         },
         "publish-packages": {
             "cache": false,
-            "dependsOn": ["build", "^publish-packages"],
             "outputs": []
         },
         "style:fix": {
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
         "build": "turbo run build",
         "compile": "turbo run compile:js compile:typedefs",
         "lint": "turbo run test:lint",
-        "publish-packages": "turbo run publish-packages",
+        "publish-packages": "turbo run --force=true build && turbo run --continue publish-packages",
         "style:fix": "turbo run style:fix",
         "test": "turbo run test:unit:browser test:unit:node",
         "test:live-with-test-validator": "turbo run test:live-with-test-validator",


buffalojoec (approved):
beautiful

turbo.json
```diff
@@ -38,7 +38,6 @@
         },
         "publish-packages": {
             "cache": false,
-            "dependsOn": ["build", "^publish-packages"],
```
steveluscher, line 41: Ooh, no. Publish really does have to depend on the upstream publishes, otherwise the wrong version numbers will get written into the package.jsons.

steveluscher (changes requested):
This works. Big hammer energy tho!

Some alternate designs:

1. Teach the publish to skip if the version already exists; some thing in the style of https://www.npmjs.com/package/publish-if-not-published
2. Shim the publish with a shell script; catch and log the error, but return a healthy status code unconditionally.

turbo.json
mcintyre94: In my testing it seemed that not having a `dependsOn` means that it runs all the upstream `publish-package` tasks, is that not the same as setting `"dependsOn": ["^publish-packages"]`?

mcintyre94:
I did consider using a shell script instead for the upstream `publish-packages` calls, but doing that for each package seemed annoying. I can give that a try though!

turbo.json
steveluscher: Oh, really? Oh in that case, ship it.

steveluscher (approved):

turbo.json
steveluscher: Oh, I see. It will.

mcintyre94: I think `pnpm turbo publish-packages` won't, but `pnpm publish-packages` will

PR #1883, opened 2023-11-27
refactor(experimental): add FixedSizeCodec and VariableSizeCodec types

lorisleiva:
This PR slightly refactors the type definitions of `Encoders`, `Decoders` and `Codecs` by providing a `FixedSize*` and `VariableSize*` type variant for each of them.

This allows us to provide type safety when codec factories expect fixed size or variable size codecs. For instance, `reverseEncoder({} as VariableSizeEncoder<string>)` will now throw a TypeScript error as the `reverseEncoder<T>()` function expects a `FixedSizeEncoder<T>`.


steveluscher (approved):

buffalojoec (approved):
This is pretty cool, I dig it!

packages/codecs-core/src/__typetests__/combine-codec-typetest.ts
```diff
@@ -0,0 +1,45 @@
...
+import { combineCodec } from '../combine-codec';
+
+{
+    // [combineCodec]: It keeps track of the fixed or variable size types.
+    combineCodec({} as FixedSizeEncoder<string>, {} as FixedSizeDecoder<string>) satisfies FixedSizeCodec<string>;
+    combineCodec(
+        {} as VariableSizeEncoder<string>,
+        {} as VariableSizeDecoder<string>
+    ) satisfies VariableSizeCodec<string>;
+    combineCodec({} as Encoder<string>, {} as Decoder<string>) satisfies Codec<string>;
```
mcintyre94, line 21: Is it valid to combine a `FixedSizeEncode` and a `VariableSizeDecoder` or vice versa? It looks like if you did that it'd hit this case and you'd just get a `Codec`, but it might be worth adding a typetest to be explicit that that's how it should work assuming that's correct.

mcintyre94 (approved):
Nice, always a fan of better typing! 🔥

packages/codecs-core/src/__typetests__/combine-codec-typetest.ts
lorisleiva: That's a good point, I'll add a test to make sure it's not type safe to do so.

lorisleiva: The conversation was [moved here](https://github.com/solana-labs/solana-web3.js/pull/1903#discussion_r1410943228).

PR #1888, opened 2023-11-27
refactor(experimental): make the polyfill set the `crypto` global in Node

steveluscher:
# Summary

Node only sets the `crypto` global when the `--experimental-global-webcrypto` is set. Let's teach the polyfill to also ensure that the global is set.


steveluscher:
Github is stuck. Shipped in 17a54d24aa331f074d04d83076b91c3de227f37f.

PR #1902, opened 2023-11-28
Refactor: Add .js extensions to relative imports

mcintyre94:
First thing's first: Sorry!

With that out the way, I will try to justify why I want to change over 400 files.

Currently, if you import the new web3js into a project (say, a secret project 🪽)  that uses [`"moduleResolution": "NodeNext"` in tsconfig ](https://www.typescriptlang.org/tsconfig#moduleResolution), Intellisense will refuse to function. It won't recognise imports from the library, and it won't give any autocomplete etc functionality when you use them. So far all developers who have experienced this (sample size: 1) were sad.

Screenshot 2023-11-28 at 14 44 00

Screenshot 2023-11-28 at 14 45 14

The reason this happens is that there are no extensions in the `node_modules/@solana/<lib>/dist/types/*.d.ts` files. For example, here's the `index.d.ts` for `codecs-core`:
```ts
export * from './assertions';
export * from './bytes';
export * from './codec';
export * from './combine-codec';
export * from './fix-codec';
export * from './map-codec';
export * from './reverse-codec';
```

Compare this to for example `wallet-adapter-base`:
```ts
export * from './adapter.js';
export * from './errors.js';
export * from './signer.js';
export * from './standard.js';
export * from './transaction.js';
export * from './types.js';
```

Without this .js extension in the generated types, `NodeNext` module resolution fails to resolve anything.

Thanks to the always excellent experience of web development and tooling, we achieve "adding the .js file extension" by, adding it ourselves to every relative import/export in every file. See eg the wallet-adapter base index.ts: https://github.com/solana-labs/wallet-adapter/blob/master/packages/core/base/src/index.ts 

Fortunately we have an eslint plugin that does this for us: https://github.com/solana-labs/eslint-plugin-require-extensions 

So this PR does the following:

- Adds the `require-extensions` plugin to `.eslintrc.json`
- Uses the `require-extensions` plugin to add .js to all the relative imports in all packages
- Updates the jest config to map these .js imports so tests keep working

The result is that the `dist/types/*.d.ts` files all have .js file extensions, and they are able to be resolved correctly

Screenshot 2023-11-28 at 14 57 27

Screenshot 2023-11-28 at 14 58 03

Going forward this will be enforced by the eslint plugin and `style:fix` will fix any issues, so this will keep working automatically for all libraries using the shared eslint config. 

Note that I haven't changed any of the tsconfig for web3js itself, since that doesn't seem to be necessary. I've only added the .js extensions to make its exports compatible with `NodeNext`.


buffalojoec:
Damn. Is this really the only way? I noticed we do this in the SPL JS libraries, but I figured it was dated or something.

mcintyre94:
It seems to be :( See https://github.com/solana-labs/wallet-adapter/issues/546 and the readme of the eslint plugin: https://github.com/solana-labs/eslint-plugin-require-extensions for some more details!

lorisleiva:
I _really_ dislike this and I think the reason `wallet-adapters` is having to do this is because it's compiling using `tsc` only which simply does not allow changing the extension when compiling. However, we're using Esbuild (via `tsup`) so I'm pretty sure this can be achieved at compile time.

steveluscher:
Whatever happens here, can you add the rule to @solana/eslint-config instead, and then bump the version?

lorisleiva:
Could we consider doing post-compiling renaming instead?

See https://stackoverflow.com/a/77150985/11440277

steveluscher:
Is this unobtrusive enough, and does it work with sourcemaps and watch mode? https://www.npmjs.com/package/tsc-alias

mcintyre94:
> Is this unobtrusive enough, and does it work with sourcemaps and watch mode? https://www.npmjs.com/package/tsc-alias

It sounds good, but I'm not sure how it's meant to work. If I run `pnpm tsc-alias --debug -p tsconfig.declarations.json` in `packages/addresses` with our current dist and dist/types it doesn't change anything

The debug output shows the right files, all 6 under dist/types/*.d.ts
But for each file it just logs "tsc-alias debug: replaced file without changes: '/home/sol/src/solana-web3.js/packages/addresses/dist/types/index.d.ts'"

In its readme it compares itself to https://github.com/dividab/tsconfig-paths which doesn't seem to be solving the problem as this PR but I'm not sure really!

steveluscher:
I just heard that it did this if you set `resolveFullPaths`.

https://stackoverflow.com/a/76678279/802047

mcintyre94:
Still doesn't seem to be working for me. I tried `pnpm tsc-alias --resolve-full-paths  --debug -p tsconfig.declarations.json` and again it didn't change anything

jordaaash:
> I _really_ dislike this

@lorisleiva can you say more about why? I didn't love it when I first started doing it, but now I do it on all projects and don't notice, as it's handled automatically by the IDE and linting with `--fix`.

lorisleiva:
@jordaaash I can understand doing something like this if you don't have a build step. In fact, it's pretty popular these days to write pure ESM with JS docs and export the code that you write directly. Why not!

But IMO, if you go through the effort of compiling your code for different environments and extracting your type definitions, then why would you fold here? It doesn't even make sense as you're importing `.ts` files under the `.js` extension (and folder imports get even uglier). The only reason this looks normal to people like us right now is because we know the quirks of this ecosystem.

Again, without a build step, that's what you've got to do and that's the challenge you accepted when taking that decision. However here we've got the luxury of changing things at compile time. Why not use it? I'm convinced shit like this will sort itself out eventually. Why make our code confusing in the meantime when we don't need to?

PR #1904, opened 2023-11-28
Use tsup to generate typedefs

mcintyre94:
This is an alternative solution to #1902, which has more context about the issue

Currently we use `tsup` for the `compile:js` stage, but we use `tsc` for the `compile:typedefs` stage. This PR changes all package.json files to use `tsup` for `compile:typedefs`

The reason to do this is that `tsup` compiles `d.ts` files differently, by putting all the exports into a single file.

For example you can see the dist/types for `codecs-core` here: https://www.npmjs.com/package/@mcintyre94/codecs-core/v/2.0.1-experimental?activeTab=code
Compare to current: https://www.npmjs.com/package/@solana/codecs-core/v/2.0.0-experimental.7d05b9c?activeTab=code

As another example, `packages/addresses/dist/types/index.d.ts` looks like:
<details>
<summary>index.d.ts</summary>

```ts
import { Encoder, Decoder, Codec } from '@solana/codecs-core';

type Address<TAddress extends string = string> = TAddress & {
    readonly __brand: unique symbol;
};
declare function isAddress(putativeAddress: string): putativeAddress is Address<typeof putativeAddress>;
declare function assertIsAddress(putativeAddress: string): asserts putativeAddress is Address<typeof putativeAddress>;
declare function address<TAddress extends string = string>(putativeAddress: TAddress): Address<TAddress>;
declare function getAddressEncoder(config?: Readonly<{
    description: string;
}>): Encoder<Address>;
declare function getAddressDecoder(config?: Readonly<{
    description: string;
}>): Decoder<Address>;
declare function getAddressCodec(config?: Readonly<{
    description: string;
}>): Codec<Address>;
declare function getAddressComparator(): (x: string, y: string) => number;

/**
 * An address derived from a program address and a set of seeds.
 * It includes the bump seed used to derive the address and
 * ensure the address is not on the Ed25519 curve.
 */
type ProgramDerivedAddress<TAddress extends string = string> = Readonly<[
    Address<TAddress>,
    ProgramDerivedAddressBump
]>;
/**
 * A number between 0 and 255, inclusive.
 */
type ProgramDerivedAddressBump = number & {
    readonly __brand: unique symbol;
};
/**
 * Returns true if the input value is a program derived address.
 */
declare function isProgramDerivedAddress<TAddress extends string = string>(value: unknown): value is ProgramDerivedAddress<TAddress>;
/**
 * Fails if the input value is not a program derived address.
 */
declare function assertIsProgramDerivedAddress<TAddress extends string = string>(value: unknown): asserts value is ProgramDerivedAddress<TAddress>;
type ProgramDerivedAddressInput = Readonly<{
    programAddress: Address;
    seeds: Seed[];
}>;
type SeedInput = Readonly<{
    baseAddress: Address;
    programAddress: Address;
    seed: Seed;
}>;
type Seed = string | Uint8Array;
declare function getProgramDerivedAddress({ programAddress, seeds, }: ProgramDerivedAddressInput): Promise<ProgramDerivedAddress>;
declare function createAddressWithSeed({ baseAddress, programAddress, seed }: SeedInput): Promise<Address>;

declare function getAddressFromPublicKey(publicKey: CryptoKey): Promise<Address>;

export { type Address, type ProgramDerivedAddress, type ProgramDerivedAddressBump, address, assertIsAddress, assertIsProgramDerivedAddress, createAddressWithSeed, getAddressCodec, getAddressComparator, getAddressDecoder, getAddressEncoder, getAddressFromPublicKey, getProgramDerivedAddress, isAddress, isProgramDerivedAddress };
```

</details>

So rather than adding .js to all the relative imports, this solves the problem by not having any relative imports in the `.d.ts` files

### The Downside

With this approach we no longer have the `*.d.ts.map` files in `dist/types`. These are TypeScript declaration maps (distinct from sourcemaps) that make go-to-definition work across packages in the monorepo. Without this it takes you to the d.ts file instead which is nowhere near as nice

Unfortunately [tsup can't generate these files](https://tsup.egoist.dev/#generate-typescript-declaration-maps--d-ts-map)

The documentation suggests using ` tsc` command to generate it, but they're documented how to use `tsc` to generate the whole type definitions, ie what we already do. I can't see a way to get the map file and the `tsup` type definitions. 

### Alternatives

- Do nothing. I don't think this is acceptable because `NodeNext` projects need to be able to use our libraries with good DevEx
- Tweak and merge #1902. This keeps the `tsc` typedefs and therefore the Typescript declaration maps, but at the cost of cluttering every relative import/export with the .js extension
- Do a [Babel post-compile](https://stackoverflow.com/a/77150985/1375972), suggested by Loris on #1902. I haven't looked into this properly but instinctively I'd like to keep not using Babel on the new libraries if possible.

I think on balance this is probably the least complex option, but the loss of those declaration maps sucks and will be annoying for us.


steveluscher:
> Without this it takes you to the d.ts file instead which is nowhere near as nice
 
giphy

lorisleiva:
Another alternative is to run a simple post-compile script that goes through the imports and add the extensions. We don't need to use Babel for something as trivial.

Something like this maybe: https://github.com/microsoft/TypeScript/issues/16577#issuecomment-310426634

buffalojoec:
> Another alternative is to run a simple post-compile script that goes through the imports and add the extensions. We don't need to use Babel for something as trivial.
> 
> Something like this maybe: [microsoft/TypeScript#16577 (comment)](https://github.com/microsoft/TypeScript/issues/16577#issuecomment-310426634)

This still lands us in "PR with 450 files changed" territory though, right?

lorisleiva:
> This still lands us in "PR with 450 files changed" territory though, right?

No because that will only change the compiled code inside the `dist` folder. So we can continue to code the usual way, yet our compiled types will please every environment.

mcintyre94:
I've pushed a new version which only generates the `tsup` types on publish. This means that in development in the monorepo we continue using `tsc` with the `d.ts.map` files and the best go-to-definition support. Then when we publish we swap the types out for the `tsup` ones that squash the relative imports and don't require the `.js` extension.

The downside now is:

- our published types aren't the exact same files that we test with locally. If there's a bug in `tsup` then we'll publish incorrect types without noticing locally. If we need to debug such an issue we'll need to run `pnpm publish:typedefs` to generate the published types locally. I think this is reasonably low risk given we already use `tsup` for the actual published code though. 

- we don't have the *.d.ts.map files in the published types. This means that if you go-to-definition from an app that uses our libraries, you will end up at the types file. This isn't unusual, it's the same for Next and Lodash libraries for example. But wallet-adapter libraries will take you to eg `node_modules/@solana/wallet-adapter-base/src/signer.ts` which is pretty slick. We won't be able to do that because we won't be publishing the map file. Note that for some reason this currently doesn't work with `@solana/web3.js@experimental` even though we publish the map. I haven't investigated why but in theory we could fix that currently, but with this change we couldn't because the map won't be published. Also FWIW the `tsup` docs say "They should not be included in a published NPM package and should not be confused with sourcemaps." but don't explain why they shouldn't be published.

mcintyre94:
Closing - superseded by #1907

PR #1907, opened 2023-11-29
refactor(experimental): add build script that adds .js to type imports

lorisleiva:
This is an alternative solution to #1902 and #1904.

It essentially goes through all the compiled `dist/**/**.d.ts` files and add the `.js` extension to relative import and exports.

For instance, it transforms `packages/addresses/dist/types/index.d.ts` from this:

```ts
export * from './address';
export * from './program-derived-address';
export * from './public-key';
```

To this:

```ts
export * from './address.js';
export * from './program-derived-address.js';
export * from './public-key.js';
```

It also checks whether the imported relative path is a folder and, when it is, adds `/index.js` instead of `.js`.

For instance, it transforms `packages/rpc-core/dist/types/index.d.ts` from this:

```ts
export * from './rpc-methods';
export * from './rpc-subscriptions';
```

To this:

```ts
export * from './rpc-methods/index.js';
export * from './rpc-subscriptions/index.js';
```

### DIFF
--- a/packages/accounts/package.json
+++ b/packages/accounts/package.json
@@ -34,7 +34,7 @@
     ],
     "scripts": {
         "compile:js": "tsup --config build-scripts/tsup.config.library.ts",
-        "compile:typedefs": "tsc -p ./tsconfig.declarations.json",
+        "compile:typedefs": "tsc -p ./tsconfig.declarations.json && node node_modules/build-scripts/add-js-extension-to-types.mjs",
         "dev": "jest -c node_modules/test-config/jest-dev.config.ts --rootDir . --watch",
         "prepublishOnly": "version-from-git --no-git-tag-version --template experimental.short",
         "publish-packages": "pnpm publish --tag experimental --access public --no-git-checks",
--- a/packages/addresses/package.json
+++ b/packages/addresses/package.json
@@ -34,7 +34,7 @@
     ],
     "scripts": {
         "compile:js": "tsup --config build-scripts/tsup.config.library.ts",
-        "compile:typedefs": "tsc -p ./tsconfig.declarations.json",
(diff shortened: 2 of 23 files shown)


lorisleiva:
* master
  * **PR #1907** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1907?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1907?utm_source=stack-comment).

mcintyre94:
I think this would be a nice solution, but when I asked Jordan if we could just do this at the top level he mentioned he tried that in Icarus and it partially fixed it but he still got some errors and needed to use the .js imports all the way through the codebase. @jordaaash is that accurate?

jordaaash:
Yes, all imports in the .d.ts files will need to be updated. Should be _okay_ with a recursive `find` or whatever the script needs to use. Also, just a heads up that using some shell scripts may cause this to fail on Windows. If using a regex though, there are likely to be some weird cases. What we do in the eslint plugin is actually check for the existence of the imported file, and also use this to fix directory index imports.

lorisleiva:
I updated to the code to support directory import paths — i.e. it now adds `/index.js` instead of `.js` for directories.

I also changed the script's language from shell to JavaScript since the latter is more portable and allows me to hook logic into the find and replace regex.

steveluscher:
Does running this with `cross-env` give you Windows compatibility for free somehow or no?

lorisleiva:
I didn't even think we'd need something like `cross-env` for it to work on Windows since it's using core Node libraries only but maybe. I'm not sure how to test that on a Windows machine.

buffalojoec:
> I'm not sure how to test that on a Windows machine.

I have a Windows box if you need me to.

PR #1908, opened 2023-11-29
refactor(experimental): add a function to assert a transaction has a blockhash lifetime

mcintyre94:
This PR adds a new function `assertIsBlockhashLifetimeTransaction` which narrows the type of a transaction to `BaseTransaction & ITransactionWithBlockhashLifetime`, if the transaction contains a valid blockhash lifetime constraint

It mirrors the behaviour of the existing `assertIsDurableNonceTransaction`

My specific use case for this is using the default transaction confirmer with a deserialized transaction. The default transaction confirmer requires the `lastValidBlockHeight` field of the lifetime constraint: https://github.com/solana-labs/solana-web3.js/blob/8f39c907a5e9d612111eeb029b27017b0693c6d6/packages/library/src/transaction-confirmation.ts#L40-L47

By asserting that the deserialized transaction has a valid blockhash lifetime constraint, we can satisfy this requirement and that of any other code that requires a blockhash lifetime on the transaction.


buffalojoec (commented):
Lgtm, but for consistency I think this should be called `assertIsTransactionWithBlockhashLifetime`. Wdyt?

packages/transactions/src/blockhash.ts
```diff
@@ -44,6 +44,26 @@ export function assertIsBlockhash(putativeBlockhash: string): asserts putativeBl
...
+
+export function assertIsBlockhashLifetimeTransaction(
+    transaction: BaseTransaction | (BaseTransaction & ITransactionWithBlockhashLifetime)
+): asserts transaction is BaseTransaction & ITransactionWithBlockhashLifetime {
+    if (!isBlockhashLifetimeTransaction(transaction)) {
+        // TODO: Coded error.
+        throw new Error('Transaction is not a blockhash transaction');
+    }
+    assertIsBlockhash(transaction.lifetimeConstraint.blockhash);
+}
```
lorisleiva, line 65: I'm not sure if it's on purpose but the logic of the `isX` and `assertIsX` functions is not the same. Passing a transaction with an invalid base58 blockhash will work for `isX` but fail for `assertIsIx`.

To make them the same, we could push that conditional to the `assertIsX` function and just do a try/catch on the `isX` function.

lorisleiva (approved):

packages/transactions/src/blockhash.ts
mcintyre94: Good point, thanks! I've refactored to move all the logic to the `isX` function, which makes them symmetrical as you say and also mirrors the durable nonce equivalent.

mcintyre94:
> Lgtm, but for consistency I think this should be called `assertIsTransactionWithBlockhashLifetime`. Wdyt?

Agreed, renamed to that!

buffalojoec (approved):

packages/transactions/src/blockhash.ts
```diff
@@ -44,6 +44,31 @@ export function assertIsBlockhash(putativeBlockhash: string): asserts putativeBl
     }
 }
 
+function isTransactionWithBlockhashLifetime(
+    transaction: BaseTransaction | (BaseTransaction & ITransactionWithBlockhashLifetime)
+): transaction is BaseTransaction & ITransactionWithBlockhashLifetime {
+    const lifetimeConstraintShapeMatches =
+        'lifetimeConstraint' in transaction &&
+        typeof transaction.lifetimeConstraint.blockhash === 'string' &&
```
steveluscher, line 52: You _could_ use `assertIsBlockhash` here, but meh.

PR #1924, opened 2023-12-03
chore: lock `@types/node` to the LTS version of the 18 line

steveluscher:
The thought here is that if Node 18 is what we support at minimum, then the types we use should be from that era.

### DIFF
--- a/packages/build-scripts/package.json
+++ b/packages/build-scripts/package.json
@@ -8,7 +8,7 @@
         "tsup.config.package.ts"
     ],
     "devDependencies": {
-        "@types/node": "^20",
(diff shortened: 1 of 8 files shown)


steveluscher:
* master
  * **PR #1924** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1924?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1924?utm_source=stack-comment).

PR #1931, opened 2023-12-05
chore: replace the Ed25519 implementation in the legacy library with the published version

steveluscher:
# Summary

Here we reimport the new ed25519 implementation package into the legacy library.

PR #1930, opened 2023-12-05
chore: a standalone package for the userspace Ed25519 implementation

steveluscher:
# Summary

This is the first module we'll extract from the monolithic legacy library. Here we create a module that encapsulates just the Ed25519 implementation.

PR #1929, opened 2023-12-05
chore: move legacy library into a sub-workspace

steveluscher:
# Summary

We're going to start to extract modules from the legacy library, publish them separately, then re-import them into the legacy library.

For example, imagine that we extract `PublicKey` and `Keypair` into a module called `@solana/keys` that's published on npm. Anyone who needs _only_ key utilties can now import that much smaller module instead of having to import all of `@solana/web3.js`.

For much more on where we're headed with this, read [this Discord message](https://discord.com/channels/428295358100013066/976527749973942374/1180727294457819147).

PR #1928, opened 2023-12-05
chore: stop building synthetic module typedef for legacy library

steveluscher:
# Summary

Many moons ago, we had a [situation](https://github.com/solana-labs/solana/issues/16207) where old versions of web3.js had a hand-hewn typedef which was clobbering the newly auto-generated one. In order to overcome this, we made the autogenerated one behave like the hand-hewn one so as to clobber it back.

In this PR, we stop doing that. It's unlikely that someone is mixing `@solana/web3.js@^1.1.0` with version 1.87.7.

PR #1932, opened 2023-12-05
chore: create custom publish strategy for legacy packages

steveluscher:
# Summary

Here we add a `prepare` step to the release script for the legacy library.

1. Build all of the dependencies (right now just `@solana/ed25519`)
2. Versiona and publish all of the dependencies to npm
3. Publish `@solana/web3.js` with all of those versions baked in

In doing so, _all_ of the legacy packages version together.

NOTE: If, for whatever reason, you run this locally, don't commit the result. It leaves changes in your working copy.


steveluscher:
* master
  * **PR #1928** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1928?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1929** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1929?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #1930** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1930?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #1931** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1931?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
          * **PR #1932** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1932?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1932?utm_source=stack-comment).

lorisleiva (approved):
It would be nice to use changeset for this in the future. We can set it up so that all legacy libraries are published together whereas the new libraries can simply be linked together.

steveluscher:
Totally. This was the smallest change I could make not to rock the boat too much, and even still it took me a week and like 20 hours of reading conventional changelog code to be able to do.

I think when we move to something else we're going to have a real, serious problem with the way that I've put multiple npm packages with the same name in the same workspace. I do this because I want to be able to pull modern packages into legacy and have them build with workspaces, but release management software is going to hate that. Gonna be a nightmare.

buffalojoec:
> I think when we move to something else we're going to have a real, serious problem with the way that I've put multiple npm packages with the same name in the same workspace. I do this because I want to be able to pull modern packages into legacy and have them build with workspaces, but release management software is going to hate that. Gonna be a nightmare.

Ummm, is this something we can ice now before it swells?

We definitely don't want dependency problems with JS out of the gate, considering the hell v1.16 put devs through on the Rust side of things. If dependency management, dependabot, `yarn upgrade`, etc. are shaky on/after release we are _toast_.

lorisleiva:
@steveluscher Changeset should work fine with multiple packages in the same namespace as long as they are all part of the same folder or explicitly whitelisted. What sort of issues do you anticipate?

steveluscher:
* The workspace will puke if there are two packages named `@solana/keys` (_maybe_ we can hack around this with a `.pnpmfile.cjs`)
* Changesets' CI script will not know wtf to do if I call a package `@solana/keys-legacy` but expect it to be published as `@solana/keys@1.x.x`.

lorisleiva:
Oh I see! So why don't we just not do this in the first place and use `legacy-` prefixes everywhere? 🙃

steveluscher:
> So why don't we just not do this in the first place and use legacy- prefixes everywhere?

I would feel like a clown releasing `@solana/legacy-keys` and `@solana/keys` rather than `@solana/keys` at versions 1 and 2.

steveluscher:
image

lorisleiva:
The fact is we're gonna have to do that for some libraries anyway (such as `legacy-utils`).

I think it's much cleaner to have a set of packages that share a consistent naming convention then having some living under `legacy-x`, some living under the latest version of `x` and some living under the previous major version of `x` like `keys`.

I see nothing wrong with releasing `legacy-keys` for the split up legacy library and `keys` for the library we want to maintain going forward.

I'm worried we're overcomplicating things tbh.

PR #1940, opened 2023-12-05
refactor(experimental): fix choppy validator connection on startup

buffalojoec:
This PR attempts to solve one of our many choppy CI problems.

Specifically, this change _should_ eliminate the all-too occasional
`ECONNREFUSED` error we get from the local RPC in CI.

My theory is that we're trying to hit the RPC before the server is online, since
a test validator usually takes a couple of seconds to boot up and begin
accepting connections to its RPC.

This change ensures we get a proper RPC response from the test validator before
exiting the `start-shared-test-validator.sh` script.

### DIFF
--- a/scripts/start-shared-test-validator.sh
+++ b/scripts/start-shared-test-validator.sh
@@ -11,6 +11,19 @@ TEST_VALIDATOR=$( cd "$(dirname "${BASH_SOURCE[0]}")/.." ; pwd -P )/.solana/acti
 TEST_VALIDATOR_LEDGER="$( cd "$(dirname "${BASH_SOURCE[0]}")/.." ; pwd -P )/test-ledger"
 FIXTURE_ACCOUNTS_DIR="$( cd "$(dirname "${BASH_SOURCE[0]}")/fixtures" ; pwd -P)"
 
+wait_for_rpc_server() {
+    while true; do
+        RESPONSE=$(curl http://127.0.0.1:8899 -s -X POST -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","id":1, "method":"getHealth"}')
+        if echo "$RESPONSE" | jq -e .result > /dev/null; then
+            echo "RPC server ready for connections"
+            break
+        else
+            echo "Waiting for the RPC server to come online..."
+            sleep 1 # Wait for 1 second before retrying
+        fi
+    done
+}
+
 mkdir -p $LOCK_DIR
 
 (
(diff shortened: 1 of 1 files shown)


buffalojoec:
* master
  * **PR #1940** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1940?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1940?utm_source=stack-comment).

scripts/start-shared-test-validator.sh
```diff
@@ -11,6 +11,19 @@ TEST_VALIDATOR=$( cd "$(dirname "${BASH_SOURCE[0]}")/.." ; pwd -P )/.solana/acti
 TEST_VALIDATOR_LEDGER="$( cd "$(dirname "${BASH_SOURCE[0]}")/.." ; pwd -P )/test-ledger"
 FIXTURE_ACCOUNTS_DIR="$( cd "$(dirname "${BASH_SOURCE[0]}")/fixtures" ; pwd -P)"
 
+wait_for_rpc_server() {
+    while true; do
+        RESPONSE=$(curl http://127.0.0.1:8899 -s -X POST -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","id":1, "method":"getHealth"}')
```
steveluscher, line 16: Fun fact, this is also available at `http://127.0.0.1:8899/health`.

buffalojoec: 🪦 RIP

steveluscher (commented):
This code is supposed to do that, for Jest:

https://github.com/solana-labs/solana-web3.js/blob/master/packages/test-config/test-validator-setup.js#L9-L17

And this `start-server-and-test` helper for everything outside of Jest:

https://github.com/solana-labs/solana-web3.js/blob/aa7fffbc6a064138d828e233cc06ffba852a1399/packages/library-legacy/package.json#L50

If those are failing to do their work, we need to figure out why.

I have had the passing thought that maybe Turborepo caching is to blame here (does something think the server is started because of some cached step, when it actually isn't?).

scripts/start-shared-test-validator.sh
```diff
@@ -23,6 +36,7 @@ mkdir -p $LOCK_DIR
       validator_pid=$!
       echo "Started test validator (PID $validator_pid)"
       wait
+      wait_for_rpc_server
```
steveluscher, line 39: I don't quite remember how this works, but I think this line will never run, because the `wait` above it is waiting for the `flock` to be released.

buffalojoec: I was curious if for some reason `flock` was releasing but then the validator was still undergoing startup after `flock` detached from the process.

PR #1943, opened 2023-12-06
fix: make custodian signer for authorize and withdraw

ochaloup:
The `StakeProgram` instructions that considers the custodian does not allow to use it as a signer. That's an issue.
When the `authorize` is called then for staker the custodian on `Lockup` is not needed but for withdrawer the `custodian` signature is needed.
For `withdraw` when the custodian is provided then the signature should be expected as well.

https://github.com/solana-labs/solana/blob/v1.17.7/programs/stake/src/stake_state.rs#L1110
https://github.com/solana-labs/solana/blob/v1.17.7/sdk/program/src/stake/state.rs#L326

This PR is a kind of issue report with suggestion to fix. I can adjust when demanded.

### DIFF
--- a/packages/library-legacy/src/programs/stake.ts
+++ b/packages/library-legacy/src/programs/stake.ts
@@ -734,7 +734,7 @@ export class StakeProgram {
     if (custodianPubkey) {
       keys.push({
         pubkey: custodianPubkey,
-        isSigner: false,
+        isSigner: true,
         isWritable: false,
       });
     }
@@ -776,7 +776,7 @@ export class StakeProgram {
     if (custodianPubkey) {
       keys.push({
         pubkey: custodianPubkey,
-        isSigner: false,
+        isSigner: true,
         isWritable: false,
       });
     }
@@ -903,7 +903,7 @@ export class StakeProgram {
     if (custodianPubkey) {
       keys.push({
         pubkey: custodianPubkey,
-        isSigner: false,
+        isSigner: true,
         isWritable: false,
       });
     }


packages/library-legacy/src/programs/stake.ts
```diff
@@ -776,7 +777,8 @@ export class StakeProgram {
     if (custodianPubkey) {
       keys.push({
         pubkey: custodianPubkey,
-        isSigner: false,
+        isSigner: stakeAuthorizationType.index
+            === StakeAuthorizationLayout.Withdrawer.index,
```
buffalojoec, line 781: This one looks like it's a signer whenever provided:
<https://github.com/solana-labs/solana/blob/fca44b78a681ac1abf7878079e8a7a319c31b23f/sdk/program/src/stake/instruction.rs#L618-L620>

packages/library-legacy/src/programs/stake.ts
```diff
@@ -903,7 +905,7 @@ export class StakeProgram {
     if (custodianPubkey) {
       keys.push({
         pubkey: custodianPubkey,
-        isSigner: false,
+        isSigner: true,
```
buffalojoec, line 908: Looks good!
<https://github.com/solana-labs/solana/blob/fca44b78a681ac1abf7878079e8a7a319c31b23f/sdk/program/src/stake/instruction.rs#L701-L703>

packages/library-legacy/src/programs/stake.ts
```diff
@@ -734,7 +734,8 @@ export class StakeProgram {
     if (custodianPubkey) {
       keys.push({
         pubkey: custodianPubkey,
-        isSigner: false,
+        isSigner: stakeAuthorizationType.index 
+            === StakeAuthorizationLayout.Withdrawer.index,
```
buffalojoec, line 738: Also looks like a signer whenever provided:
<https://github.com/solana-labs/solana/blob/fca44b78a681ac1abf7878079e8a7a319c31b23f/sdk/program/src/stake/instruction.rs#L567-L569>

buffalojoec (commented):
Hey thanks for PR'ing this in!

I've linked each instruction from the SDK to compare. It seems like whenever the custodian is provided it's a signer.

Here's the function each of the `authorize` instructions call into from the links that you provided in the description:
<https://github.com/solana-labs/solana/blob/fa251768b301f13535dec9ce663ec663bc746271/sdk/program/src/stake/state.rs#L303-L343>

packages/library-legacy/src/programs/stake.ts
buffalojoec: > But as the `rs` code sets it to `true` all the time then you would consider to have that consistent and set it up to `true` always, would that be so?

Yeah, ideally if the crate is building an instruction one way, we want to exactly mimic that in JS unless there's a bug report associated with it.

ochaloup: I see, thanks for clarification. I updated the PR to define the signer being `true` on all three places.

PR #1955, opened 2023-12-08
ci: add manual trigger for "publish packages"

buffalojoec:
This change adds a manual trigger for "publish packages" CI/CD, in case we ever want to trigger a "publish packages" CI run from GitHub manually.

<https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow>


lorisleiva (approved):
Thanks!

steveluscher (approved):

PR #1964, opened 2023-12-11
refactor(experimental): export the TransactionError type to consumers of @solana/rpc-core

mcintyre94:
This PR adds the `TransactionError` type to the exports of the `@solana/rpc-core` package. Note that this package is not currently re-exported by `@solana/web3.js`. 

This is useful for writing a wrapper function around `simulateTransaction` or `sendTransaction`, where you want to return a `TransactionError` if one is returned by the RPC calls.

### DIFF
--- a/packages/rpc-core/src/index.ts
+++ b/packages/rpc-core/src/index.ts
@@ -1,2 +1,3 @@
 export * from './rpc-methods';
 export * from './rpc-subscriptions';
+export * from './transaction-error';


buffalojoec (approved):
Only a matter of time until we need `InstructionError`, too or nah?

lorisleiva (approved):

mcintyre94:
> Only a matter of time until we need InstructionError, too or nah?

My gut is probably not, because it's just a nested type under `TransactionError` and isn't itself the return type of anything. You wouldn't really write code that might return an `InstructionError` without it also maybe returning a `TransactionError`, and at that point you might as well just return `TransactionError`. If it did become the return type of eg. some RPC call then we'd have to export it from `transaction-error.ts` and at that point it'd be exposed to consumers of the package too.

buffalojoec:
> My gut is probably not, because it's just a nested type under `TransactionError` and isn't itself the return type of anything. You wouldn't really write code that might return an `InstructionError` without it also maybe returning a `TransactionError`, and at that point you might as well just return `TransactionError`. I

What about `try ... catch`? I could see trying to catch certain instruction errors in some cases, especially for a program client.

PR #1966, opened 2023-12-13
You can now post-process params at selected keypaths

steveluscher:
# Summary

The goal of the next PR is to default the commitment to `confirmed` for all methods that involve commitment. This overrides the server default of `finalized`, which we believe not to be practical for most use cases.

This PR adds infrastructure to allow you to inject node transformers that operate at key paths. See how this is used in the next PR.

# Test plan

```
cd packages/rpc-core/
pnpm test:unit:browser
pnpm test:unit:node
```

### DIFF
--- a/packages/rpc-core/src/__tests__/params-patcher-test.ts
+++ b/packages/rpc-core/src/__tests__/params-patcher-test.ts
@@ -1,4 +1,5 @@
 import { patchParamsForSolanaLabsRpc } from '../params-patcher';
+import { KeyPath, KEYPATH_WILDCARD } from '../patcher-types';
 
 describe('patchParamsForSolanaLabsRpc', () => {
     [10, '10', null, undefined, Symbol()].forEach(input => {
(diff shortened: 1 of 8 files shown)


steveluscher:
* `master`
  * **PR #1966** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1966?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #1967** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1967?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1966?utm_source=stack-comment).

packages/rpc-core/src/__tests__/params-patcher-test.ts
```diff
@@ -53,30 +54,70 @@ describe('patchParamsForSolanaLabsRpc', () => {
...
+                patchParamsForSolanaLabsRpc(BigInt(Number.MAX_SAFE_INTEGER), { onIntegerOverflow });
                 expect(onIntegerOverflow).not.toHaveBeenCalled();
             });
         });
     });
+    describe.each`
+        keyPath               | params                                 | expectedTransformees | expectedPatchedParams
+        ${[KEYPATH_WILDCARD]} | ${[1, 2, 3]}                           | ${[1, 2, 3]}         | ${['🍕', '🍕', '🍕']}
+        ${[1, 1]}             | ${[1, [2, 22], 3]}                     | ${[22]}              | ${[1, [2, '🍕'], 3]}
+        ${['foo', 1, 'bar']}  | ${{ baz: 1, foo: [2, { bar: 3 }, 4] }} | ${[3]}               | ${{ baz: 1, foo: [2, { bar: '🍕' }, 4] }}
```
buffalojoec, line 87: Pizza tests!

buffalojoec (approved):
keyz

steveluscher:
I changed this to make it recursive-ish. The tests are now _really_ subtle. Can you take another look at this and see that I got it right?

lorisleiva (changes requested):
### Current Solution

I think there is a fundamental issue with this approach as **it reuses the existing recursive parameter transformer** whose purpose is to handle integer overflow.

The purpose of this PR is to eventually add another parameter transformer which will handle default commitment values. This one does not need to be recursive.

Moreover, this approach does not allow composing additional parameter transformers to the default ones. Composing transformers of any kind is an important aspect of building more complex RPC APIs and transports.

### Proposed Solution

What I suggest we do instead is **break up the current parameter transformers into two distinct transformers**:
- One recursive transformer that only handles integer overflow.
- One non-recursive transformer that accepts a `defaultCommitment` value and sets it on all relevant methods.

Then, we simply pipe the two together. That is, the transformed parameters of the former will be passed as arguments of the latter. That way, we end up with an array of parameters that have gone through all transformers without them needing to know about each other.

Additionally, we could accept an optional `extraParameterTransformer` (for lack of a better name) configuration that can be added at the end of the pipe should the end-user need to transform these parameters even further.

CleanShot 2023-12-14 at 10 22 39@2x

buffalojoec:
Thanks @lorisleiva for the detailed overview.

What if we instead just offered a recursive and non-recursive transformer? You can choose to define `n` number of operations to perform on recursive node visits _or_ on particular key paths.

This way, one could stack `integerOverflow` and some other operations on recursive visits, while `defaultCommitment` could be combined with something like `defaultAccountEncoding` or `defaultTransactionDetails`.

This way, we avoid having to define a new config field for each type of transformer.

lorisleiva:
@buffalojoec I think we probably need to have a quick chat about this because I'm a bit confused.

I think it's much cleaner to have one way of composing transformers and then offer many transformers that each have one specific responsibility.

If someone happens to have another param transformer that needs to be recursive, then they can offer their own recursive transformer. I also think trying to offer a one-size-fits-all recursive solution is a bad idea. It's always very much case-by-case.

buffalojoec:
> @buffalojoec I think we probably need to have a quick chat about this because I'm a bit confused.
> 
> I think it's much cleaner to have one way of composing transformers and then offer many transformers that each have one specific responsibility.
> 
> If someone happens to have another param transformer that needs to be recursive, then they can offer their own recursive transformer. I also think trying to offer a one-size-fits-all recursive solution is a bad idea. It's always very much case-by-case.

Something sort of like this was what I was thinking, but I'm not sure how difficult it would be to implement `getRecursiveTransformer(..)` as defined.

```ts
type Config = Readonly<{
    getNodeTransformersForKeyPaths?: (methodName: string) => /* ?? */;
    getRecursiveTransformer?: (methodName: string, keyPath: (number | string)[], ...args) => void;
}>;
```

lorisleiva:
Noooooo. I really think neither `getNodeTransformersForKeyPaths` nor `getRecursiveTransformer` should be part of the configuration. They can be helper method to helper people build their own transformers if we want to, but they are implementation details. They are not the reason why we are adding them to the default RPC API.

steveluscher:
Abandoned in favor of #1977.

PR #1978, opened 2023-12-18
refactor(experimental): add programs package

lorisleiva:
This PR adds a new `@solana/programs` package whose purpose is to:

1. Offer program types that generated JS client can use (See `Program` and `ProgramWithError`).
2. Offer a helper function that can resolve custom program errors from raw errors caused by transaction failures (See `resolveTransactionError`).


lorisleiva:
* `master`
  * **PR #1978** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1978?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1978?utm_source=stack-comment).

packages/programs/README.md
```diff
@@ -0,0 +1,79 @@
...
+        // ...
+    },
+};
+```
+
+## Functions
+
+### `resolveTransactionError()`
+
+This function takes a raw error caused by a transaction failure and attempts to resolve it into a custom program error.
```
buffalojoec, line 61: 🫶🏼

packages/programs/src/program.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+export type Program<TAddress extends string = string> = {
+    /**
+     * A unique name for the Program.
+     *
+     * To avoid conflict with other organizations, it is recommended
+     * to prefix the program name with a namespace that is unique to
+     * your organization. For instance, programs belonging to the
+     * Solana Program Library are prefixed with `spl` like so:
+     * `splSystem` or `splAddressLookupTable`.
```
buffalojoec, line 14: nit: These aren't SPL programs, so they don't actually have these prefixes. Maybe use SPL Memo or SPL Token?

buffalojoec (approved):
oh yeah

packages/programs/src/program.ts
lorisleiva: Good shout, I'll update first thing in the morning so it's less confusing.

packages/programs/src/program.ts
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * A unique name for the Program.
+     *
+     * To avoid conflict with other organizations, it is recommended
+     * to prefix the program name with a namespace that is unique to
+     * your organization. For instance, programs belonging to the
+     * Solana Program Library are prefixed with `spl` like so:
+     * `splMemo` or `splToken`.
+     */
+    name: string;
```
mcintyre94, line 16: What's this used for? Is it just intended for debugging/logging?

packages/programs/src/resolve-transaction-error.ts
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Resolves a custom program error from a transaction error
+ * with logs using the provided list of registered programs.
+ * The original error is returned if the error cannot be
+ * resolved from the given programs.
+ *
+ * @param error The raw error to resolve containing the program logs.
+ * @param transaction The transaction that caused the error.
+ * @param programs The list of registered programs.
```
mcintyre94, line 14: What does it mean for a program to be registered? The term is used a few times but it looks like you just need an instance of `Program`

mcintyre94 (approved):
Nice, looks great - I really like un-hexing the error codes especially! Just a couple of minor questions but LGTM!

packages/programs/src/program.ts
lorisleiva: Good question. I debated whether to keep that one but that's something we get for free from the IDL so it could be nice to keep in the definition.

The main reason I used the `name` attribute with the Umi framework is that it would allow you to search a program by name or program address. It's not uncommon for program developers to deploy their programs using different addresses on different clusters. In this case, a search by name would allow you to find all instances of them.

In the Umi framework though, [the program definition](https://github.com/metaplex-foundation/umi/blob/8bbb100378b578c108d7b0e49bc3f8f6773be02e/packages/umi/src/Program.ts#L58-L65) also includes an `isOnCluster` function that allows us to filter programs even further. So with Umi you could do `umi.programs.get('myProgramName', 'devnet')` and it would find you the correct program in the correct cluster even when program addresses differ on different clusters.

That being said, I'm happy to remove it from this definition of a `Program`. We could always add a `ProgramWithName` interface if required in the future.

packages/programs/src/resolve-transaction-error.ts
lorisleiva: That's a good point, I should probably change the wording around that.

Again, that's because, with Umi, you would _register_ your programs in a program repository by doing `umi.programs.add(myProgram)`.

Here though, we do not have control over how the end user keeps track of its programs which is why I accept an array of programs instead.

What do you think about the following?

```suggestion
 * @param programs The list of programs to go through when resolving the transaction error. They should ideally contain all programs the transaction is sending instructions to.
```

packages/programs/src/program.ts
mcintyre94: It definitely seems like it could be useful for debugging etc, much nicer than an address!

If we use the name from the IDL though, is that likely to have the prefix the comment mentions?

packages/programs/src/resolve-transaction-error.ts
mcintyre94: Ah I see, the suggested change looks good! :)

packages/programs/src/program.ts
lorisleiva: It depends on how the IDL is generated but with metaplex for instance, we used to always name our programs `mpl_my_program` so that worked well.

In any case, you can easily update the name of your programs in your `kinobi.cjs` file to whatever you want.

We might want to consider offering an even stronger naming convention while we're at it though. Maybe even a domain based on like `com.solana.system`?

lorisleiva:
@steveluscher Since this adds a new package to the monorepo, do you want to add your review or should I just merge and deploy this?

packages/programs/src/resolve-transaction-error.ts
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @param error The raw error to resolve containing the program logs.
+ * @param transaction The transaction that caused the error.
+ * @param programs The list of programs to go through when resolving the transaction error.
+ * They should ideally contain all programs the transaction is sending instructions to.
+ * @returns The resolved program error, or the original transaction error
+ * if the error cannot be resolved using the provided programs.
+ */
+export function resolveTransactionError(
+    error: Error & { logs?: string[] },
```
steveluscher, line 20: Nit: I'm obsessed with `Readonly<{logs?: readonly string[]}>`

packages/programs/src/resolve-transaction-error.ts
```diff
@@ -0,0 +1,53 @@
...
+    transaction: Transaction,
+    programs: Program[],
+): Error {
+    // Compute the full logs from which to parse the instruction index and error code.
+    const fullLogs = error.message + '\n' + (error.logs ?? []).join('\n');
+
+    // Parse the instruction number, or return the original error.
+    const instructionRegex = /Error processing Instruction (\d+)/;
```
steveluscher, line 28: Does this one need to be case-insensitive too?

packages/programs/src/resolve-transaction-error.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @returns The resolved program error, or the original transaction error
+ * if the error cannot be resolved using the provided programs.
+ */
+export function resolveTransactionError(
+    error: Error & { logs?: string[] },
+    transaction: Transaction,
+    programs: Program[],
+): Error {
+    // Compute the full logs from which to parse the instruction index and error code.
+    const fullLogs = error.message + '\n' + (error.logs ?? []).join('\n');
```
steveluscher, line 25: I guess the thinking here is that building a string and then running one `match` over it is faster than building a state machine?

```ts
const enum State {
  SEARCHING_FOR_INDEX,
  SEARCHING_FOR_CUSTOM_ERROR,
};

let state;
for (const log in logs) {
  switch (state.__type) {
    case State.SEARCHING_FOR_INDEX:
    {
      const index = log.match(/Error processing Instruction (\d+)/)?.[1];
      if (index != null) {
        state = {__type: State.SEARCHING_FOR_CUSTOM_ERROR, instructionIndex: index};
      }
      break;
    }
    case State.SEARCHING_FOR_CUSTOM_ERROR:
    {
      /* ... */
    }
  }
}
```

steveluscher (approved):

packages/programs/src/resolve-transaction-error.ts
lorisleiva: You need to find both of them in order to be able to resolve the error so it's not the case of switching between different strategies. There are situations were the index and the error code may sometimes be present in the message and not in the logs, and vice versa. By combining the whole thing together and searching that, we offer a more robust solution.

packages/programs/src/resolve-transaction-error.ts
lorisleiva: No that one is consistent for some reasons. Happy to make it case-insensitive for consistency with the other regex if you want.

steveluscher: ¯\\\_(°ペ)\_/¯

packages/programs/src/resolve-transaction-error.ts
steveluscher: That's not a strategy switch, it's a state machine. Iterate over logs until you find the instruction by index, transition to the next state where you search for the custom error, etc…

lorisleiva: Gotcha. I prefer leaving this as-is because IMO the invariant "the instruction index needs to appear before the error code" isn't necessary.

steveluscher (commented):
This is great! The one flaw in the design here is that you have to ship _every_ error message _to your app_ even if you never hit a single one.

The ideal situation is for these errors to live on some web server, like they do in React. All of the React error codes are stripped from the production build, and all that are left are the codes. Then the codes get plugged into the following formula:

```
https://react.dev/errors/{CODE}
```

Try it here: https://react.dev/errors/469

It would be AMAZING if Anchor programs could start to actually throw the URLs in logs.

Instead of:

```
Program returned error: "custom program error: 0x1e"
```

Do this:

```
Program returned error: "custom program error: 0x1e; visit https://errors.anchor-lang.com/{PUBKEY}/0x1e for the full message."
```

This is 100% automated when React is built. We could do the same at IDL generation time. See https://github.com/facebook/react/tree/main/scripts/error-codes

PR #1988, opened 2023-12-24
Update encoding in the transaction config

shiralwz:
**Issue:**
The intention here seems to be to fetch the transaction again with jsonParsed encoding if the requested encoding is not jsonParsed.  but the config object passed to rpc.getTransaction in this block is the same as the original call and will still contain the original encoding. It does not change to jsonParsed as intended.

**Fix:**
Created a new jsonParsedConfig object with encoding explicitly set to jsonParsed. This ensures the second call to rpc.getTransaction fetches the transaction with the jsonParsed encoding.


buffalojoec (approved):
Heh, nice catch! Thanks!

PR #1990, opened 2023-12-28
Detect block height exceedences using a combination of slot notifications and `getEpochInfo`

### LINKED ISSUE #1985: [experimental] Blockheight confirmation strategy compares block height to slot height, rendering it useless
The slot height and the block height are not the same thing. When they go far enough out of sync on any given cluster, this bug makes it impossible to confirm transactions.

https://github.com/solana-labs/solana-web3.js/blob/a1fafee840589e493b215dc2c9ec5ae9ba44ca13/packages/library/src/transaction-confirmation-strategy-blockheight.ts#L23

I wrote this code; mea culpa.

steveluscher:
# Summary

This PR repairs the block height exceedence detection in the transaction confirmer. Previously, we mistakenly compared slot height to the last valid block height. Those are not equivalent values.

Since there is no subscription API for block height, this implementation determines the difference between the slot height and the block height, then subscribes to the slot height. It waits until the slot when the block height is expected to be exceeded to be reached.

Finally, in consideration that one or more blocks might have been skipped since having started to wait, it rechecks the slot height / block height difference before failing.

# Test plan

```
cd packages/library/
pnpm turbo test:unit:browser
pnpm turbo test:unit:node
```

Fixes #1985.

### DIFF
--- a/packages/library/src/transaction-confirmation-strategy-blockheight.ts
+++ b/packages/library/src/transaction-confirmation-strategy-blockheight.ts
@@ -1,33 +1,73 @@
-import type { Slot, SlotNotificationsApi } from '@solana/rpc-core';
-import type { RpcSubscriptions } from '@solana/rpc-transport';
+import type { GetEpochInfoApi, SlotNotificationsApi } from '@solana/rpc-core';
+import type { Rpc, RpcSubscriptions } from '@solana/rpc-transport';
+import { Commitment } from '@solana/rpc-types';
 
 type GetBlockHeightExceedencePromiseFn = (config: {
     abortSignal: AbortSignal;
-    lastValidBlockHeight: Slot;
+    commitment?: Commitment;
+    lastValidBlockHeight: bigint;
 }) => Promise<void>;
 
-export function createBlockHeightExceedencePromiseFactory(
-    rpcSubscriptions: RpcSubscriptions<SlotNotificationsApi>,
-): GetBlockHeightExceedencePromiseFn {
-    return async function getBlockHeightExceedencePromise({ abortSignal: callerAbortSignal, lastValidBlockHeight }) {
+export function createBlockHeightExceedencePromiseFactory({
+    rpc,
+    rpcSubscriptions,
+}: Readonly<{
+    rpc: Rpc<GetEpochInfoApi>;
+    rpcSubscriptions: RpcSubscriptions<SlotNotificationsApi>;
+}>): GetBlockHeightExceedencePromiseFn {
+    return async function getBlockHeightExceedencePromise({
+        abortSignal: callerAbortSignal,
+        commitment,
+        lastValidBlockHeight,
+    }) {
         const abortController = new AbortController();
-        function handleAbort() {
+        const handleAbort = () => {
             abortController.abort();
-        }
+        };
         callerAbortSignal.addEventListener('abort', handleAbort, { signal: abortController.signal });
-        const slotNotifications = await rpcSubscriptions
-            .slotNotifications()
-            .subscribe({ abortSignal: abortController.signal });
+        async function getBlockHeightAndDifferenceBetweenSlotHeightAndBlockHeight() {
+            const { absoluteSlot, blockHeight } = await rpc
+                .getEpochInfo({ commitment })
+                .send({ abortSignal: abortController.signal });
+            return {
+                blockHeight,
+                differenceBetweenSlotHeightAndBlockHeight: absoluteSlot - blockHeight,
+            };
+        }
         try {
-            for await (const slotNotification of slotNotifications) {
-                if (slotNotification.slot > lastValidBlockHeight) {
-                    // TODO: Coded error.
-                    throw new Error(
-                        'The network has progressed past the last block for which this transaction ' +
-                            'could have committed.',
-                    );
+            const [slotNotifications, { blockHeight, differenceBetweenSlotHeightAndBlockHeight }] = await Promise.all([
+                rpcSubscriptions.slotNotifications().subscribe({ abortSignal: abortController.signal }),
+                getBlockHeightAndDifferenceBetweenSlotHeightAndBlockHeight(),
+            ]);
+            if (blockHeight <= lastValidBlockHeight) {
+                let lastKnownDifferenceBetweenSlotHeightAndBlockHeight = differenceBetweenSlotHeightAndBlockHeight;
+                for await (const slotNotification of slotNotifications) {
+                    const { slot } = slotNotification;
+                    if (slot - lastKnownDifferenceBetweenSlotHeightAndBlockHeight > lastValidBlockHeight) {
+                        // Before making a final decision, recheck the actual block height.
+                        const {
+                            blockHeight: currentBlockHeight,
+                            differenceBetweenSlotHeightAndBlockHeight: currentDifferenceBetweenSlotHeightAndBlockHeight,
+                        } = await getBlockHeightAndDifferenceBetweenSlotHeightAndBlockHeight();
+                        if (currentBlockHeight > lastValidBlockHeight) {
+                            // Verfied; the block height has been exceeded.
+                            break;
+                        } else {
+                            // The block height has not been exceeded, which implies that the
+                            // difference between the slot height and the block height has grown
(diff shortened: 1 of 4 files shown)


steveluscher:
* `master`
  * **PR #1990** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1990?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1990?utm_source=stack-comment).

packages/library/src/transaction-confirmation-strategy-blockheight.ts
```diff
@@ -1,35 +1,128 @@
...
+    commitment?: Commitment;
+    lastValidBlockHeight: ReturnType<GetBlockHeightApi['getBlockHeight']>;
 }) => Promise<void>;
 
-export function createBlockHeightExceedencePromiseFactory(
-    rpcSubscriptions: RpcSubscriptions<SlotNotificationsApi>,
-): GetBlockHeightExceedencePromiseFn {
-    return async function getBlockHeightExceedencePromise({ abortSignal: callerAbortSignal, lastValidBlockHeight }) {
-        const abortController = new AbortController();
+async function sleep(ms: number, abortSignal: AbortSignal) {
```
steveluscher, line 11: Review this first.

packages/library/src/transaction-confirmation-strategy-blockheight.ts
```diff
@@ -1,35 +1,128 @@
...
+            const timeSinceLastSuccessfulFetch =
+                timestampOfLastSuccessfulFetch == null ? 0 : Date.now() - timestampOfLastSuccessfulFetch;
+            return Math.max(minDelay, regularFetchDelay - timeSinceLastSuccessfulFetch);
+        } else {
+            return minDelay;
+        }
+    };
+}
+
+export function getDefaultSleepDelayBetweenBlockHeightFetchesGetter(msPerSlot: number) {
```
steveluscher, line 53: Review this, and its tests, second.

packages/library/src/transaction-confirmation-strategy-blockheight.ts
```diff
@@ -1,35 +1,128 @@
...
+        function handleTimerExpired() {
+            abortSignal.removeEventListener('abort', handleAbort);
+            resolve();
+        }
+        const timeoutId = setTimeout(handleTimerExpired, ms);
+        abortSignal.addEventListener('abort', handleAbort);
+    });
+}
+
+export function getDefaultSleepDelayBetweenBlockHeightFetchFailuresGetter(
```
steveluscher, line 33: Review this, and its tests, third.

packages/library/src/transaction-confirmation-strategy-blockheight.ts
```diff
@@ -1,35 +1,128 @@
...
+        lastFetchedBlockHeight: ReturnType<GetBlockHeightApi['getBlockHeight']> | undefined,
+        timestampOfLastSuccessfulFetch: number | undefined,
+    ) => number;
+    getSleepDelayBetweenBlockHeightFetches: (
+        lastValidBlockHeight: ReturnType<GetBlockHeightApi['getBlockHeight']>,
+        lastFetchedBlockHeight: ReturnType<GetBlockHeightApi['getBlockHeight']>,
+    ) => number;
+    rpc: Rpc<GetBlockHeightApi>;
+}>): GetBlockHeightExceedencePromiseFn {
+    return async function getBlockHeightExceedencePromise({ abortSignal, commitment, lastValidBlockHeight }) {
```
steveluscher, line 83: Review the changes to this, last.

packages/library/src/transaction-confirmation.ts
```diff
@@ -73,7 +77,18 @@ export function createDefaultRecentTransactionConfirmer({
     rpc,
     rpcSubscriptions,
 }: DefaultRecentTransactionConfirmerConfig) {
-    const getBlockHeightExceedencePromise = createBlockHeightExceedencePromiseFactory(rpcSubscriptions);
+    const getSleepDelayBetweenBlockHeightFetches = getDefaultSleepDelayBetweenBlockHeightFetchesGetter(
+        400 /* msPerSlot */,
```
steveluscher, line 81: I really hate hardcoding this, and I'm open to suggestions.

packages/library/src/transaction-confirmation.ts
```diff
@@ -73,7 +77,18 @@ export function createDefaultRecentTransactionConfirmer({
...
+    );
+    const getSleepDelayBetweenBlockHeightFetchFailures = getDefaultSleepDelayBetweenBlockHeightFetchFailuresGetter(
+        5000 /* minDelay */,
```
steveluscher, line 84: This is a *big* handwave. The thinking here was ‘if we failed to fetch a blockhash, we might as well be conservative and give the RPC a break.’ The worst case scenario for this code failing forever is your transaction confirmation never ‘blockheight-ing out.’

lorisleiva:
Might be a dump question but since we have a slot subscription API and a way to fetch block heights from a minimum slot, could we not construct something equivalent to a block height subscription that way?

That being said, if polling is the only way to solve this then, that's a great execution of that solution.

packages/library/src/transaction-confirmation.ts
lorisleiva: May I ask the rationale behind that number?

steveluscher: The block height can only possibly advance once per slot. 400ms is the current target latency for slot progression.

If any given slot is skipped, the block height will not advance. This is why the block height is nearly always lower than the slot height.

steveluscher:
> …could we not construct something equivalent to a block height subscription that way?

100%! That's what should happen.

It would take months to deploy though (land, testnet, devnet, mainnet, feature activations, version upgrades, et cetera). Even if we introduced a block height subscription, we _have_ to make the change in this PR now.

steveluscher:
Ohhhh… sorry; I understand your suggestion now, @lorisleiva. So the algorithm would be something like:

1. Subscribe to the slot height.
2. Fetch the block height for the first obtained slot.
3. Presume that the delta between the two will remain fixed for the duration of the transaction confirmation (ie. there will be no skipped slots while you're waiting)
4. Wait until the slot where the block height would be exceeded.
5. _Recheck_ the delta to confirm that the block height, indeed, has been exceeded (ie. that there were no skipped slots while you were waiting).

I sort of like that. Way fewer requests.

Should I do that?

lorisleiva:
Nice, I wasn't sure if this was possible but it sounds like you've got this. I think that would be a great solution. That way, the main "pulse" comes from a subscription as opposed to a poll of requests. Also, I'm guessing that removes the need to hardcode latency values.

steveluscher:
Oh, crap, actually `getBlockHeight` does not return the context with the slot in it, and `minContextSlot` is not sufficient to answer the question ‘what’s the block height for this slot?’

Damn.

Well, I guess the solution in the present PR works, and it’s as light on the RPC as possible.

¯\_(°ペ)_/¯

lorisleiva:
What about using `getBlock`? It accepts a slot number as a parameter and returns the block height within the object it returns. Or is it too much strain on the RPC to get the full block object at each pulse?

steveluscher:
Innnnnteresting. We could always use the `getBlock` API with `transactionDetails` set to `none`.

This may still be a non-starter because `getBlock` doesn't support `processed` commitment.

I'll think about it!

buffalojoec:
If only `blockSubscribe` was stable...

Regardless, I agree with the advice suggested by @lorisleiva on using subscriptions for this.

Why does `getBlockHeight` need to have the context slot?

steveluscher:
> Why does `getBlockHeight` need to have the context slot?

I just wanted something to answer ‘what's the block height for _this_ slot’ not ‘what's the current block height.’

Splitting hairs I guess.

steveluscher:
OK. Overhauled this completely. Turns out you can get a point-in-time difference between the slot height and the block height using `getEpochInfo`.

buffalojoec (approved):
epic

PR #1996, opened 2024-01-03
Add function to assert account(s) are decoded

mcintyre94:
This PR adds two new functions to the `accounts` package, `assertAccountDecoded` and `assertAccountsDecoded`.

These narrow `Account<TData | Uint8Array>` to `Account<TData>`

This is particularly useful for handling the result of `fetchJsonParsedAccount<TData>`. This returns a `MaybeAccount<TData | Uint8Array>`, ie the account might or might not exist, and if it does exist then the account might or might not have actually returned JSON parsed data.

Note that as with the rest of the `accounts` package, we're not asserting that the data matches the declared shape. The only check done here is whether it's a `Uint8Array`, if it's not then we narrow it to `TData`. 

We already had `assertAccountExists` to narrow to `Account<TData | Uint8Array>`, so this function adds the final step to narrow it to `Account<TData>`:

```ts
const account = await fetchJsonParsedAccount<MockData>(
    {} as unknown as Rpc<GetAccountInfoApi>,
    '1111' as Address,
    {},
);
account satisfies MaybeAccount<MockData | Uint8Array>;

assertAccountExists<MockData | Uint8Array>(account);
account satisfies Account<MockData | Uint8Array>;

assertAccountDecoded(account);
account satisfies Account<MockData>;
```

If you put this assertion first you also remove the need for that type parameter on the `assertAccountExists` call:

```ts
const account = await fetchJsonParsedAccount<MockData>(
    {} as unknown as Rpc<GetAccountInfoApi>,
    '1111' as Address,
    {},
);
account satisfies MaybeAccount<MockData | Uint8Array>;

assertAccountDecoded(account);
account satisfies MaybeAccount<MockData>;

assertAccountExists(account);
account satisfies Account<MockData>;
```

### DIFF
--- a/packages/accounts/README.md
+++ b/packages/accounts/README.md
@@ -267,3 +267,47 @@ const myDecoder: Decoder<MyAccountData> = getStructDecoder([
 const myDecodedAccount = decodeAccount(myAccount, myDecoder);
 myDecodedAccount satisfies Account<MyAccountData, '1234..5678'>;
 ```
+
+### `assertAccountDecoded()`
+
+This function asserts that an account stores decoded data, ie not a Uint8Array. Note that it does not check the shape of the data matches the decoded type, only that it is not a Uint8Array.
+
+```ts
+type MyAccountData = { name: string; age: number };
+
+const myAccount: Account<MyAccountData | Uint8Array, '1234..5678'>;
+assertAccountDecoded(myAccount);
+
+// now the account data can be used as MyAccountData
+account.data satisfies MyAccountData;
+```
+
+This is particularly useful for narrowing the result of fetching a JSON parsed account.
+
+
+```ts
+    const account: MaybeAccount<MockData | Uint8Array> = await fetchJsonParsedAccount<MockData>(
+        rpc, 
+        '1234..5678' as Address,
+    )
+
+    assertAccountDecoded(account);
+    // now we have a MaybeAccount<MockData>
+    account satisfies MaybeAccount<MockData>
+```
+
+### `assertAccountsDecoded`
+
+This function asserts that all input accounts store decoded data, ie not a Uint8Array. As with `assertAccountDecoded` it does not check the shape of the data matches the decoded type, only that it is not a Uint8Array.
+
+```ts
+type MyAccountData = { name: string; age: number };
+
+const myAccounts: Account<MyAccountData | Uint8Array, Address>[];
+assertAccountsDecoded(myAccounts);
+
+// now the account data can be used as MyAccountData
+for(const a of account) {
+    account.data satisfies MyAccountData;
+}
+```
--- a/packages/accounts/src/__tests__/decode-account-test.ts
+++ b/packages/accounts/src/__tests__/decode-account-test.ts
@@ -1,8 +1,10 @@
 import 'test-matchers/toBeFrozenObject';
 
-import { EncodedAccount } from '../account';
-import { decodeAccount } from '../decode-account';
-import { MaybeEncodedAccount } from '../maybe-account';
+import { Address } from '@solana/addresses';
+
+import { Account, EncodedAccount } from '../account';
+import { assertAccountDecoded, assertAccountsDecoded, decodeAccount } from '../decode-account';
+import { MaybeAccount, MaybeEncodedAccount } from '../maybe-account';
 import { getMockDecoder } from './__setup__';
 
 describe('decodeAccount', () => {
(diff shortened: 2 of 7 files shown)


mcintyre94:
* `master`
  * **PR #1996** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1996?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #1997** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1997?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1996?utm_source=stack-comment).

lorisleiva (approved):
That's a nice addition! 👌

packages/accounts/src/__tests__/maybe-account-test.ts
buffalojoec, line 1: smh @lorisleiva where's your linter?? 👊🏼

packages/accounts/src/__tests__/decode-account-test.ts
```diff
@@ -69,3 +71,125 @@ describe('decodeAccount', () => {
...
+        ];
+
+        // When we assert that the accounts are decoded
+        const fn = () => assertAccountsDecoded(accounts);
+
+        // Then we expect an error to be thrown
+        expect(fn).toThrow('Expected accounts [1111, 2222] to be decoded.');
+    });
+
+    it('does not throw if all of the provided accounts are encoded', () => {
```
buffalojoec, line 148: nit: `decoded`

packages/accounts/README.md
```diff
@@ -267,3 +267,47 @@ const myDecoder: Decoder<MyAccountData> = getStructDecoder([
...
+```ts
+    const account: MaybeAccount<MockData | Uint8Array> = await fetchJsonParsedAccount<MockData>(
+        rpc, 
+        '1234..5678' as Address,
+    )
+
+    assertAccountDecoded(account);
+    // now we have a MaybeAccount<MockData>
+    account satisfies MaybeAccount<MockData>
+```
```
buffalojoec, line 297: Wouldn't this make sense to return a `Account<MockData>`? If you can verify the data is not a `Uint8Array`, then you are also confirming its existence, no?

buffalojoec (commented):
Looks good! Curious on how to handle "maybe" accounts.

packages/accounts/README.md
lorisleiva: I think it makes more sense to let the two different type guards check their respective properties so they can be composed.

> If you can verify the data is not a Uint8Array, then you are also confirming its existence, no?

No, if the account does not exist it does not fail when asserting it's decoded.

mcintyre94: Yep, the main reason it works like this (if no data then assert is OK) is so you can compose the checks in either order.

It also mirrors the behaviour of `decodeAccount`, which will no-op on an account that doesn't exist, ie will convert `MaybeAccount<Uint8Array>` to `MaybeAccount<TData>`. I think it'd be really confusing if this threw after you call `decodeAccount`:

```ts
const account = <MaybeEncodedAccount>{address: '1111' as Address, exists: false};

// this already works
const decoder: Decoder<TData> = ...
const decodedAccount = decodeAccount(account, decoder)

// this should also
assertAccountDecoded(decodedAccount)
```

buffalojoec: Alrighty sounds good!

buffalojoec (approved):

packages/accounts/src/decode-account.ts
```diff
@@ -28,3 +28,41 @@ export function decodeAccount<TData extends object, TAddress extends string = st
...
+export function assertAccountsDecoded<TData extends object, TAddress extends string = string>(
+    accounts: (Account<TData | Uint8Array, TAddress> | MaybeAccount<TData | Uint8Array, TAddress>)[],
+): asserts accounts is (Account<TData, TAddress> | MaybeAccount<TData, TAddress>)[] {
+    const encoded = accounts.filter(a => accountExists(a) && a.data instanceof Uint8Array);
+    if (encoded.length > 0) {
+        const encodedAddresses = encoded.map(a => a.address).join(', ');
+        // TODO: Coded error.
+        throw new Error(`Expected accounts [${encodedAddresses}] to be decoded.`);
+    }
+}
```
steveluscher, line 68: I hate how TypeScript doesn't let you run an assertion function over all elements in an array to assert that the array contains * whatever it did before * but definitely not a `Uint8Array`.

If we could do that then you could pass _various_ accounts with _various_ data into this function and have them all come out clean (ie. without `Uint8Arrays` among them).

https://stackoverflow.com/a/77323905/802047

steveluscher (commented):
Is this semantically more like ‘parsed’ rather than ‘decoded?’ 

* `assertAccountIsParsed`
* `assertAccountsAreParsed`

PR #1997, opened 2024-01-03
Add support for deserializing a transaction that has address lookup tables

mcintyre94:
This PR updates the transaction decoder (ie converting from a Uint8Array to a Transaction) to support transactions that include address lookup tables

This involves changing the operation to be async: we need to fetch the lookup tables from on-chain accounts in order to convert from indices into them into addresses and ultimately `IAccountLookupMeta` objects. This is by design using information that is not available on the serialized transaction.

I have not made transaction encoding async. It's not clear how to do this with encoders, because the `unmap` function would need to convert from `Promise<Transaction>` to `Uint8Array`. It'd also be confusing because there is nothing that should be async about transaction encoding. Therefore I've removed the transaction codec - we now have a sync encode and async decode, which are incompatible.

I also haven't kept a sync version of decode transaction. Any transaction that doesn't have any address lookup tables can be fully decoded synchronously. But given only an encoded transaction it's non-trivial to know whether it has lookup tables. And I don't think we should make it easy to write code that doesn't work with most versioned transactions in the new web3js.


mcintyre94:
Closing this to rewrite. I think a better approach will be to keep it sync in the `transactions` package and pass in the lookup tables to use, and then write an async helper pulling in accounts + transactions in the `library` package as Loris suggested.

PR #1999, opened 2024-01-04
Update transaction decoder to use lookup tables

mcintyre94:
This PR updates `decompileTransaction` to support transactions that include address lookup tables. It is passed a dictionary of `Address -> Address[]` mapping each lookup table address to its addresses. These are used to generate `IAccountLookupMeta` objects for the instructions that use addresses from the lookup tables. An error is thrown for any lookup table required by the transaction that is missing from the mapping passed in.

The `getTransactionDecoder` and `getTransactionCodec` functions are both modified to receive this mapping of address lookups and pass it to `decompileTransaction`. It is the responsibility of the caller to fetch the required lookups and pass them in.

A follow up PR will add an async `decodeTransaction` helper function to the `library` package that uses an `RPC<GetMultipleAccountsApi>` to fetch them. This will be the recommended method of decoding transactions.

### DIFF
--- a/packages/transactions/src/decompile-transaction.ts
+++ b/packages/transactions/src/decompile-transaction.ts
@@ -59,15 +59,15 @@ function getAccountMetas(message: CompiledMessage): IAccountMeta[] {
     return accountMetas;
 }
 
-export type LookupTables = { [lookupTableAddress: Address]: Address[] };
+export type AddressesByLookupTableAddress = { [lookupTableAddress: Address]: Address[] };
 
 function getAddressLookupMetas(
-    transactionLookups: ReturnType<typeof getCompiledAddressTableLookups>,
-    knownLookups: LookupTables,
+    compiledAddressTableLookups: ReturnType<typeof getCompiledAddressTableLookups>,
+    addressesByLookupTableAddress: AddressesByLookupTableAddress,
 ): IAccountLookupMeta[] {
     // check that all message lookups are known
-    const transactionLookupAddresses = transactionLookups.map(l => l.lookupTableAddress);
-    const missing = transactionLookupAddresses.filter(a => knownLookups[a] === undefined);
+    const transactionLookupAddresses = compiledAddressTableLookups.map(l => l.lookupTableAddress);
+    const missing = transactionLookupAddresses.filter(a => addressesByLookupTableAddress[a] === undefined);
     if (missing.length > 0) {
         const missingAddresses = missing.join(', ');
         // TODO: coded error.
@@ -78,8 +78,8 @@ function getAddressLookupMetas(
     const writableMetas: IAccountLookupMeta[] = [];
 
     // we know that for each lookup, knownLookups[lookup.lookupTableAddress] is defined
-    for (const lookup of transactionLookups) {
-        const addresses = knownLookups[lookup.lookupTableAddress];
+    for (const lookup of compiledAddressTableLookups) {
+        const addresses = addressesByLookupTableAddress[lookup.lookupTableAddress];
 
         const highestIndex = Math.max(...lookup.readableIndices, ...lookup.writableIndices);
         if (highestIndex >= addresses.length) {
@@ -183,7 +183,7 @@ function convertSignatures(compiledTransaction: CompiledTransaction): ITransacti
 
 export function decompileTransaction(
     compiledTransaction: CompiledTransaction,
-    lookupTables: LookupTables,
+    addressesByLookupTableAddress: AddressesByLookupTableAddress,
     lastValidBlockHeight?: bigint,
 ): CompilableTransaction | (CompilableTransaction & ITransactionWithSignatures) {
     const { compiledMessage } = compiledTransaction;
@@ -197,7 +197,7 @@ export function decompileTransaction(
         'addressTableLookups' in compiledMessage &&
         compiledMessage.addressTableLookups !== undefined &&
         compiledMessage.addressTableLookups.length > 0
-            ? getAddressLookupMetas(compiledMessage.addressTableLookups, lookupTables)
+            ? getAddressLookupMetas(compiledMessage.addressTableLookups, addressesByLookupTableAddress)
             : [];
     const transactionMetas = [...accountMetas, ...accountLookupMetas];
 
--- a/packages/transactions/src/decompile-transaction.ts
+++ b/packages/transactions/src/decompile-transaction.ts
@@ -66,12 +66,12 @@ function getAddressLookupMetas(
     addressesByLookupTableAddress: AddressesByLookupTableAddress,
 ): IAccountLookupMeta[] {
     // check that all message lookups are known
-    const transactionLookupAddresses = compiledAddressTableLookups.map(l => l.lookupTableAddress);
-    const missing = transactionLookupAddresses.filter(a => addressesByLookupTableAddress[a] === undefined);
+    const compiledAddressTableLookupAddresses = compiledAddressTableLookups.map(l => l.lookupTableAddress);
+    const missing = compiledAddressTableLookupAddresses.filter(a => addressesByLookupTableAddress[a] === undefined);
     if (missing.length > 0) {
         const missingAddresses = missing.join(', ');
         // TODO: coded error.
-        throw new Error(`Transaction includes missing address lookup tables: [${missingAddresses}]`);
+        throw new Error(`Addresses not provided for lookup tables: [${missingAddresses}]`);
     }
 
     const readOnlyMetas: IAccountLookupMeta[] = [];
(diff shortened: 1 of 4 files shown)


mcintyre94:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1999?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `master`
  * **PR #1996** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1996?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1999** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1999?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1999?utm_source=stack-comment).

packages/transactions/src/serializers/transaction.ts
```diff
@@ -50,15 +50,17 @@ export function getTransactionEncoder(): VariableSizeEncoder<
 }
 
 export function getTransactionDecoder(
+    lookupTables: LookupTables,
```
buffalojoec, line 53: Should we make this an optional param? For UX.

ie: `lookupTables: LookupTables = {},`

packages/transactions/src/decompile-transaction.ts
```diff
@@ -1,10 +1,11 @@
 import { Address, assertIsAddress } from '@solana/addresses';
 import { pipe } from '@solana/functional';
-import { AccountRole, IAccountMeta, IInstruction } from '@solana/instructions';
+import { AccountRole, IAccountLookupMeta, IAccountMeta, IInstruction } from '@solana/instructions';
 import { SignatureBytes } from '@solana/keys';
 
 import { Blockhash, setTransactionLifetimeUsingBlockhash } from './blockhash';
 import { CompilableTransaction } from './compilable-transaction';
+import { getCompiledAddressTableLookups } from './compile-address-table-lookups';
```
buffalojoec, line 8: I'm not sure if it does a whole lot, but you can `import type` this.

packages/transactions/src/decompile-transaction.ts
```diff
@@ -58,6 +59,54 @@ function getAccountMetas(message: CompiledMessage): IAccountMeta[] {
...
+    knownLookups: LookupTables,
+): IAccountLookupMeta[] {
+    // check that all message lookups are known
+    const transactionLookupAddresses = transactionLookups.map(l => l.lookupTableAddress);
+    const missing = transactionLookupAddresses.filter(a => knownLookups[a] === undefined);
+    if (missing.length > 0) {
+        const missingAddresses = missing.join(', ');
+        // TODO: coded error.
+        throw new Error(`Transaction includes missing address lookup tables: [${missingAddresses}]`);
+    }
```
buffalojoec, line 75: I stared at this for a while, thinking about eliminating the additional iteration(s). However, I actually like how you've opted for a better error here, one that describes exactly which tables are missing. Nice.

buffalojoec (approved):
I think this looks great! Nice suggestion @lorisleiva

packages/transactions/src/serializers/transaction.ts
mcintyre94: I thought about it, but I think it might be better to make it required as a subtle hint that you really shouldn't be using this without knowing which lookup tables you need. Outside of tests you really shouldn't be hardcoding this unless you know exactly which transactions you're dealing with. Not a strong preference either way though TBH!

packages/transactions/src/decompile-transaction.ts
mcintyre94: Thanks good spot! We should probably lint for this somehow

packages/transactions/src/serializers/transaction.ts
buffalojoec: Fair point, but I vote it should be optional. You shouldn't ever need to provide a `{}` to a function in my opinion!

lorisleiva: The only downside I see to making this argument optional is it "hides" the need to provide the lookup tables when it is actually a requirement for a significant portions of transactions. That being said, I also cringe at passing `{}` as an argument so I'm happy with either.

If we're making this optional though, we may consider transforming this into a config object to provide a more explicit DX:

```ts
// Before.
getTransactionDecoder({}, 100n);

// After.
getTransactionDecoder({ lastValidBlockHeight: 100n });
```

lorisleiva (approved):
This is nice! Love the thorough tests. 🤤

packages/transactions/src/decompile-transaction.ts
```diff
@@ -58,6 +59,54 @@ function getAccountMetas(message: CompiledMessage): IAccountMeta[] {
...
 }
 
+export type LookupTables = { [lookupTableAddress: Address]: Address[] };
+
+function getAddressLookupMetas(
+    transactionLookups: ReturnType<typeof getCompiledAddressTableLookups>,
+    knownLookups: LookupTables,
+): IAccountLookupMeta[] {
+    // check that all message lookups are known
+    const transactionLookupAddresses = transactionLookups.map(l => l.lookupTableAddress);
```
steveluscher, line 69: Nit: confusing naming here. What's a ‘transaction lookup?’ Next it's called a ‘message lookup.’

Maybe just call it what it is.

```ts
function getAddressLookupMetas(
    compiledAddressTableLookups: ReturnType<typeof getCompiledAddressTableLookups>,
    addressesByLookupTableAddress: AddressesByLookupTableAddress,
): IAccountLookupMeta[] {
```

packages/transactions/src/decompile-transaction.ts
```diff
@@ -58,6 +59,54 @@ function getAccountMetas(message: CompiledMessage): IAccountMeta[] {
...
+    const missing = transactionLookupAddresses.filter(a => knownLookups[a] === undefined);
+    if (missing.length > 0) {
+        const missingAddresses = missing.join(', ');
+        // TODO: coded error.
+        throw new Error(`Transaction includes missing address lookup tables: [${missingAddresses}]`);
```
steveluscher, line 74: Nit: it's not that the _transaction_ is ‘missing address lookup tables,‘ it's that the developer didn't _supply_ the table to the decoder, right?

packages/transactions/src/decompile-transaction.ts
```diff
@@ -58,6 +59,54 @@ function getAccountMetas(message: CompiledMessage): IAccountMeta[] {
...
+    const writableMetas: IAccountLookupMeta[] = [];
+
+    // we know that for each lookup, knownLookups[lookup.lookupTableAddress] is defined
+    for (const lookup of transactionLookups) {
+        const addresses = knownLookups[lookup.lookupTableAddress];
+
+        const highestIndex = Math.max(...lookup.readableIndices, ...lookup.writableIndices);
+        if (highestIndex >= addresses.length) {
+            // TODO coded error
+            throw new Error(`Cannot look up index ${highestIndex} in lookup table ${lookup.lookupTableAddress}`);
```
steveluscher, line 87: Nit: keep going with this? Explain that there being an index greater than the one in the `addressesByLookupTableAddress` might imply that they supplied an _old_ version of the lookup table – that it might since have been extended with _more_ addresses?

packages/transactions/src/serializers/transaction.ts
steveluscher: I'm in the same boat: cringe at creating an empty object. I like @lorisleiva's idea of making it a config object.

steveluscher (approved):

packages/transactions/src/serializers/transaction.ts
mcintyre94: Thanks everyone, I've refactored to use a config object, that's definitely better!

packages/transactions/src/decompile-transaction.ts
mcintyre94: Reworded to `Addresses not provided for lookup tables: [1111, 2222]`. Does that make more sense?

packages/transactions/src/decompile-transaction.ts
mcintyre94: Thanks! I added this error mostly as a just-in-case and wasn't sure what would cause it, so this is really helpful. I've added ` The lookup table may have been extended since the addresses provided were retrieved.` to the end.

PR #2000, opened 2024-01-04
Fix bug where decompileTransaction would return empty signatures instead of excluding the field

mcintyre94:
This PR fixes a bug where decompiling a transaction that had no real signatures would result in returning `signatures: {}` instead of excluding that field. This meant that compiling and then decompiling such a transaction would not give an identical result when it should.

When we compile a transaction, we insert an all-zero 64-byte Uint8Array in place of all missing signatures.

We already had logic to exclude these signatures when we decompile a transaction, since we don't store them on our `Transaction` model. They're simply an artifact of compiling a transaction.

But the logic that determined whether to include the `signatures` field used the input `compiledTransaction.signatures` instead of our filtered and reshaped signatures to determine whether there were any signatures. This meant that it included the field even if there were only all-zero signatures that we removed.

### DIFF
--- a/packages/transactions/src/__tests__/decompile-transaction-test.ts
+++ b/packages/transactions/src/__tests__/decompile-transaction-test.ts
@@ -316,6 +316,51 @@ describe('decompileTransaction', () => {
                 lastValidBlockHeight: 100n,
             });
         });
+
+        it('excludes the signatures field if the transaction has no signatures', () => {
+            const compiledTransaction: CompiledTransaction = {
+                compiledMessage: {
+                    header: {
+                        numReadonlyNonSignerAccounts: 0,
+                        numReadonlySignerAccounts: 0,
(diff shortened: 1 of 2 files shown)


mcintyre94:
* `master`
  * **PR #2000** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2000?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2000?utm_source=stack-comment).

buffalojoec (approved):

lorisleiva (approved):

steveluscher (approved):

PR #2002, opened 2024-01-05
feat: specify which account is missing/has an invalid signature on signature verification failure

cryptopapi997:
In my experience, an annoying error to get because it doesn't give you the information you really want (which account has the problem) even though it clearly can. This PR fixes this.

Also seems I'm not the only that finds this annoying: https://twitter.com/redacted_noah/status/1742943786299191741


steveluscher (changes requested):
Awesome!

This code used to, and still does, have bail-out behaviour – it returns upon finding the first bad signature.

Let's go further and make it all-the-way useful. Can you refactor this PR so that it _unconditionally_ verifies _all_ of the signatures, and then logs the addresses of _all_ the failed signers?

steveluscher:
cat-cute

steveluscher (approved):

PR #2005, opened 2024-01-05
Add async function to decode a transaction using an RPC to fetch lookup tables

mcintyre94:
This PR adds a new helper function `decodeTransaction`. It takes as input encoded transaction bytes and a function to get multiple accounts. It:

- Decodes the transaction to an intermediate `CompiledTransaction`
- Inspects that transaction, and if it contains any address lookup tables it fetches them using `fetchJsonParsedAccounts`
- Decompiles the `CompiledTransaction` using the fetched lookup tables

This gives us the ability to decode an arbitrary transaction, whether or not it has address lookup tables

### DIFF
--- a/packages/library/src/decode-transaction.ts
+++ b/packages/library/src/decode-transaction.ts
@@ -1,0 +1,75 @@
+import {
+    assertAccountsDecoded,
+    assertAccountsExist,
+    type FetchAccountsConfig,
+    fetchJsonParsedAccounts,
+} from '@solana/accounts';
+import { Address } from '@solana/addresses';
+import type { GetMultipleAccountsApi } from '@solana/rpc-core';
+import type { Rpc } from '@solana/rpc-transport';
+import {
+    type AddressesByLookupTableAddress,
+    type CompilableTransaction,
+    decompileTransaction,
+    getCompiledTransactionDecoder,
+    type ITransactionWithSignatures,
+} from '@solana/transactions';
+
+let compiledTransactionDecoder: ReturnType<typeof getCompiledTransactionDecoder> | undefined = undefined;
+
+type FetchedAddressLookup = {
+    addresses: Address[];
+};
+
+async function fetchLookupTables(
+    lookupTableAddresses: Address[],
+    rpc: Rpc<GetMultipleAccountsApi>,
+    config?: FetchAccountsConfig,
+): Promise<AddressesByLookupTableAddress> {
+    const fetchedLookupTables = await fetchJsonParsedAccounts<FetchedAddressLookup[]>(
+        rpc,
+        lookupTableAddresses,
+        config,
+    );
+    assertAccountsDecoded(fetchedLookupTables);
+    assertAccountsExist(fetchedLookupTables);
+
+    return fetchedLookupTables.reduce<AddressesByLookupTableAddress>((acc, lookup) => {
+        return {
+            ...acc,
+            [lookup.address]: lookup.data.addresses,
+        };
+    }, {});
+}
+
+type DecodeTransactionConfig = FetchAccountsConfig & {
+    lastValidBlockHeight?: bigint;
+};
+
+export async function decodeTransaction(
+    encodedTransaction: Uint8Array,
+    rpc: Rpc<GetMultipleAccountsApi>,
+    config?: DecodeTransactionConfig,
+): Promise<CompilableTransaction | (CompilableTransaction & ITransactionWithSignatures)> {
+    const { lastValidBlockHeight, ...fetchAccountsConfig } = config ?? {};
+
+    if (!compiledTransactionDecoder) compiledTransactionDecoder = getCompiledTransactionDecoder();
+    const compiledTransaction = compiledTransactionDecoder.decode(encodedTransaction);
+    const { compiledMessage } = compiledTransaction;
+
+    const lookupTables =
+        'addressTableLookups' in compiledMessage &&
+        compiledMessage.addressTableLookups !== undefined &&
+        compiledMessage.addressTableLookups.length > 0
+            ? compiledMessage.addressTableLookups
+            : [];
+    const lookupTableAddresses = lookupTables.map(l => l.lookupTableAddress);
+
+    const fetchedLookupTables =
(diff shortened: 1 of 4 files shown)


mcintyre94:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2005?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `master`
  * **PR #1996** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1996?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1999** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/1999?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #2000** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2000?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #2005** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2005?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2005?utm_source=stack-comment).

packages/library/src/decode-transaction.ts
```diff
@@ -0,0 +1,75 @@
...
+type FetchedAddressLookup = {
+    addresses: Address[];
+};
+
+async function fetchLookupTables(
+    lookupTableAddresses: Address[],
+    rpc: Rpc<GetMultipleAccountsApi>,
+    config?: FetchAccountsConfig,
+): Promise<AddressesByLookupTableAddress> {
+    const fetchedLookupTables = await fetchJsonParsedAccounts<FetchedAddressLookup[]>(
```
steveluscher, line 29: Now that I'm seeing this, all-up, I feel like we're missing the ability to pass an `AbortSignal` here, but I guess that's a limitation of `@solana/codecs-*`. There's no way to pass in per-decode context like that. cc/ @lorisleiva

steveluscher (approved):

packages/library/src/decode-transaction.ts
mcintyre94: You can pass an abort signal into this function, that `config: FetchAccountsConfig` includes it: https://github.com/solana-labs/solana-web3.js/blob/59e58931e1183c45f7ca084bb939b36906d88c5f/packages/accounts/src/fetch-account.ts#L45-L49

The `decodeTransaction` function takes a config object that's just `FetchAccountsConfig` with an extra optional `lastValidBlockHeight` field, and everything from `FetchAccountsConfig` is passed through to `fetchJsonParsedAccounts`

buffalojoec, line 29: Heh, to use the generated Address Lookup Table client here, or `jsonParsed`?? In the future, that is.

buffalojoec (approved):

packages/library/src/decode-transaction.ts
```diff
@@ -0,0 +1,75 @@
...
+    }, {});
+}
+
+type DecodeTransactionConfig = FetchAccountsConfig & {
+    lastValidBlockHeight?: bigint;
+};
+
+export async function decodeTransaction(
+    encodedTransaction: Uint8Array,
+    rpc: Rpc<GetMultipleAccountsApi>,
```
lorisleiva, line 51: Nit: As a convention, I've been putting the `rpc` argument first whenever it is required. We might want to do the same here for consistency.

packages/library/src/decode-transaction.ts
```diff
@@ -0,0 +1,75 @@
...
+        compiledMessage.addressTableLookups !== undefined &&
+        compiledMessage.addressTableLookups.length > 0
+            ? compiledMessage.addressTableLookups
+            : [];
+    const lookupTableAddresses = lookupTables.map(l => l.lookupTableAddress);
+
+    const fetchedLookupTables =
+        lookupTableAddresses.length > 0 ? await fetchLookupTables(lookupTableAddresses, rpc, fetchAccountsConfig) : {};
+
+    return decompileTransaction(compiledTransaction, {
```
lorisleiva, line 71: I was thinking of pushing slightly more logic to the lower-lever `decodeTransaction` function so that its high-level version would literally just be: a) fetch LUTs b) delegate to lower-level function.

In other words, I think we should add the fetched LUT array as an argument of the lower-level function (so we no longer throw an error when a transaction uses LUTs) but it's still synchronous and doesn't add a dependency circle.

lorisleiva (approved):
Looks good! Just a couple comments. ☺️

packages/library/src/decode-transaction.ts
mcintyre94: Probably the generated client because it'll have better typing and we won't need that `FetchedAddressLookup` type param.

packages/library/src/decode-transaction.ts
mcintyre94: I did it this way so that the lower level function is less opinionated and only takes the input that it needs to care about, which is just the addresses per lookup table.

A bit speculative but that would allow fetching the addresses from somewhere else other than an RPC call if you want, like a database or cache that doesn't store LUT objects. My guess is that a use case like Helius, where you're providing ~real time parsed transactions, wouldn't be possible without caching LUTs.

With this implementation you have a lot of flexibility. If you cache LUTs as-is then you can provide an `Rpc<GetMultipleAccountsApi>` implementation that returns values from your cache and use this higher-level function. But if you want to store something more optimised then you have that option too using the lower-level function, as long as you can get a map of addresses per LUT from somewhere.

lorisleiva: Oh sorry I was being stupid. I checked the low-level function on the wrong branch for comparison and missed all the nice changes from the previous branches of the stack. Ignore me. 😬

PR #2014, opened 2024-01-09
refactor(experimental): add bigint support to `fast-stable-stringify`

### LINKED ISSUE #1986: [graphql] BigInt variables cause runtime fatal when passed to `isFinite()`
This query:

```graphql
query GetTransactionsInBlock($block: BigInt!) {
  block(slot: $block) {
    blockhash
    blockHeight
    blockTime
    parentSlot
    previousBlockhash
  }
}
```

Generates this error

```
{
  "errors": [
    {
      "message": "Cannot convert a BigInt value to a number",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "block"
      ]
    }
  ],
  "data": {
    "block": null
  }
}
```

[Repro link](https://graphql.solana.com/#q:0:0:I4VwpgTgngBA4mALgFQgQwHYGc0GNECWA9tgJIYBCANkbgNYAUAJAEY30BcMFBA5uYgCEAShgBvAFAwYbWoyw1EXVuzqjJ06bPoALNFh1TN2ugAkwfHYiNbVyAgFswNmAAc0EMBkQBlRS9dPADdiECxqOT0DIwBfCRigA:N4KABGBEBGA2D2BjA1pAXFATAZgOyYFYCBOY3SEAXyA).

buffalojoec:
This PR adds custom `bigint` support to the cache key functions used by the
resolvers within `DataLoader`.

The library used to create these cache keys - `fast-stable-stringify` - does not
support `bigint`, so I've added a custom `cacheKey` function to convert `bigint`
to `string` in all parameter objects before handing the object over to
`fast-stable-stringify`.

This somewhat defeats the "fast" purpose of `fast-stable-stringify`, but as
mentioned in the function's comments, considering the size of these parameter
objects, the effect should be negligible.

Closes #1986

### DIFF
--- a/packages/rpc-graphql/src/__tests__/block-tests.ts
+++ b/packages/rpc-graphql/src/__tests__/block-tests.ts
@@ -31,17 +31,72 @@ describe('block', () => {
         rpcGraphQL = createRpcGraphQL(rpc);
     });
 
+    // The `block` query takes a `BigInt` as a parameter. We need to test this
+    // for various input types that might occur outside of a JavaScript
+    // context, such as string or number.
+    describe('bigint parameter', () => {
+        const source = /* GraphQL */ `
+            query testQuery($block: BigInt!) {
+                block(slot: $block) {
+                    blockhash
+                }
+            }
+        `;
+        1;
+        it('can accept a bigint parameter', async () => {
+            expect.assertions(2);
+            fetchMock.mockOnce(JSON.stringify(mockRpcResponse(mockBlockNone)));
+            const variables = { block: 511226n };
+            const result = await rpcGraphQL.query(source, variables);
+            expect(result).not.toHaveProperty('errors');
+            expect(result).toMatchObject({
+                data: {
+                    block: {
+                        blockhash: expect.any(String),
+                    },
+                },
+            });
+        });
+        it('can accept a number parameter', async () => {
+            expect.assertions(2);
+            fetchMock.mockOnce(JSON.stringify(mockRpcResponse(mockBlockNone)));
+            const variables = { block: 511226 };
+            const result = await rpcGraphQL.query(source, variables);
+            expect(result).not.toHaveProperty('errors');
+            expect(result).toMatchObject({
+                data: {
+                    block: {
+                        blockhash: expect.any(String),
+                    },
+                },
+            });
+        });
+        it('can accept a string parameter', async () => {
+            expect.assertions(2);
+            fetchMock.mockOnce(JSON.stringify(mockRpcResponse(mockBlockNone)));
+            const variables = { block: '511226' };
+            const result = await rpcGraphQL.query(source, variables);
+            expect(result).not.toHaveProperty('errors');
+            expect(result).toMatchObject({
+                data: {
+                    block: {
+                        blockhash: expect.any(String),
+                    },
+                },
+            });
+        });
+    });
     describe('basic queries', () => {
         it("can query a block's block time", async () => {
             expect.assertions(1);
             fetchMock.mockOnce(JSON.stringify(mockRpcResponse(mockBlockFull)));
             const source = /* GraphQL */ `
-                query testQuery {
-                    block(slot: ${defaultSlot}) {
-                        blockTime
+                    query testQuery {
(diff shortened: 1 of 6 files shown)


buffalojoec:
* `master`
  * **PR #2014** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2014?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2014?utm_source=stack-comment).

steveluscher (changes requested):
Oooh, hang on though. I wonder if this is our chance to fork `fast-stable-stringify` and just add a `case` to the `switch (typeof val)` in there, handle bigints, and get out. Thoughts?

buffalojoec:
> Oooh, hang on though. I wonder if this is our chance to fork `fast-stable-stringify` and just add a `case` to the `switch (typeof val)` in there, handle bigints, and get out. Thoughts?

I was close to doing that, tbh. Do we want to just put it in the Web3.js monorepo?

buffalojoec:
In addition to the fact it hasn't had a new version in 6 years, the repository is mostly perf testing. We could probably slim it down quite a bit and add it as a package in Web3.js. Seems like it would need minimal support.

steveluscher:
I wonder how fast an object hash would be in JS if you just walked an object depth first, turned everything you find into a byte buffer, and the called `crypto.subtle.digest()` on the whole thing.

```ts
function hashObject(obj) {
  const data = new Uint8Array((function*() {
    // Walk the object in the usual ways...
    Object.keys(obj).forEach(key => {
      const value = obj[key];
      switch (typeof value) {
        case 'string':
          yield* stringToBytes(value);
          break;
        case 'number': {
          const floatArray = new Float64Array(1);
          floatArray[0] = value;
          yield* new Uint8Array(floatArray.buffer);
        /* etc etc etc */
      }
    });
  })());
  const hash = await crypto.subtle.digest('SHA-256', data);
  return bytesToHex(hash);
}
```

steveluscher:
It would be async though.

steveluscher:
Somewhat related (not really though).

https://github.com/GoogleChromeLabs/jsbi/issues/30#issuecomment-521460510

lorisleiva:
How about using the battle tested original [`json-stable-stringify` package](https://www.npmjs.com/package/json-stable-stringify) from which the "fast" one forked from.

It is downloaded over 3 million times a week and was last published 2 months ago as opposed to all the "fast" versions that were published 6 years ago. I'm pretty sure the main package would have had some extra optimisations by now.

buffalojoec:
> How about using the battle tested original [`json-stable-stringify` package](https://www.npmjs.com/package/json-stable-stringify) from which the "fast" one forked from?
> 
> It is downloaded over 3 million times a week and was last published 2 months ago as opposed to all the "fast" versions that were published 6 years ago. I'm pretty sure the main package would have had some extra optimisations by now.

This one looks like it might work. I can see support for setting the global `BigInt.toJson` or defining a `replacer`.

steveluscher:
> I can see support for setting the global `BigInt.toJson`…

Definitely definitely don't modify prototypes. web3.js will break in the first app that protects against monkeypatched builtins.

> …or defining a `replacer`.

A replacer doesn't do what we want. What we want is this:

```ts
stringify({foo: 1n}); // {"foo":1n}
```

The _most_ a replacer will let us do ([demo](https://stackblitz.com/edit/typescript-ue5rcm?devToolsHeight=33&file=index.ts)) is this:

```ts
stringify({foo: 1n}); // {"foo":"1n"}
```

…which is a hash collision with this:

```ts
stringify({foo: '1n'}); // {"foo":"1n"} 
```

The whole root of the problem here is that we dgaf about JSON – we just need a hash – but all these libraries have as a goal the production of valid JSON.

steveluscher:
Also, it's fucking huge. https://bundlephobia.com/package/json-stable-stringify@1.1.0 
image

PR #2017, opened 2024-01-09
refactor(experimental): provide account encoding on the data field

buffalojoec:
This PR refactors the accounts schema to allow for providing specific account
data encoding _only_ when requesting the `data` field, which is always going to
be a string of encoded data.

Accounts are queried from the RPC as `jsonParsed` by default, so when this field
is provided with an explicit `encoding`, it will return the account's data in
the requested encoding in the `data` field.

Note this means it's now possible to query multiple data encodings on accounts,
as depicted in the tests.

For more context, here's a review comment by @steveluscher on a previous PR:

https://github.com/solana-labs/solana-web3.js/pull/1837#discussion_r1407188113

**Important Note:** This change will cause `programAccounts` queries to spam
the RPC for each account, as the resolver is called from each account's context.
This is remedied in the next commit in the stack.

### DIFF
--- a/packages/rpc-graphql/src/resolvers/account.ts
+++ b/packages/rpc-graphql/src/resolvers/account.ts
@@ -13,3 +13,19 @@ export const resolveAccount = (fieldName: string) => {
     ) =>
         parent[fieldName] === null ? null : context.loaders.account.load({ ...args, address: parent[fieldName] }, info);
 };
+
+export const resolveAccountData = () => {
+    return (
+        parent: { address: Address },
+        args: { encoding: AccountLoaderArgs['encoding']; dataSlice?: AccountLoaderArgs['dataSlice'] },
+        context: RpcGraphQLContext,
+        info: GraphQLResolveInfo | undefined,
+    ) => {
+        return (
+            context.loaders.account
+                .load({ ...args, address: parent.address }, info)
+                // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                .then((account: any) => account.data)
+        );
+    };
+};
--- a/packages/rpc-graphql/src/__tests__/account-test.ts
+++ b/packages/rpc-graphql/src/__tests__/account-test.ts
@@ -228,15 +228,12 @@ describe('account', () => {
             // See scripts/fixtures/gpa1.json
             const variableValues = {
                 address: 'CcYNb7WqpjaMrNr7B1mapaNfWctZRH7LyAjWRLBGt1Fk',
-                encoding: 'BASE_58',
             };
             const source = /* GraphQL */ `
-                query testQuery($address: String!, $encoding: AccountEncoding) {
-                    account(address: $address, encoding: $encoding) {
+                query testQuery($address: String!) {
+                    account(address: $address) {
                         address
-                        ... on AccountBase58 {
-                            data
-                        }
+                        dataBase58: data(encoding: BASE_58)
                     }
                 }
             `;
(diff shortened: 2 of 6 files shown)


buffalojoec:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2017?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `master`
  * **PR #2014** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2014?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #2015** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2015?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #2016** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2016?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #2017** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2017?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2017?utm_source=stack-comment).

packages/rpc-graphql/src/resolvers/account.ts
```diff
@@ -13,3 +13,19 @@ export const resolveAccount = (fieldName: string) => {
     ) =>
         parent[fieldName] === null ? null : context.loaders.account.load({ ...args, address: parent[fieldName] }, info);
```
steveluscher, line 14: Maybe this is what you mean by ‘This is remedied in the next commit in the stack,’ but what should happen here is:

1. The account resolver introspects the query itself, and looks into its own subselection for _direct_ child `data` fields.
    a. If there are no data fields, fetch the account with whatever encoding is lightest on the RPC.
    b. If there _are_ data fields, fetch the account with one of those encodings.

steveluscher (changes requested):
I _think_ what this code is doing is:

* unconditionally doing one `account.load` with _no_ `encoding`
* doing `account.loads` with whatever `encoding` is on zero or more `data` fields in the subselection

What I hope we can make happen is stuff like this:

* If there are no `data` fields, fetch the account with the fastest-to-fetch encoding
* If there is one `data` field set to Y encoding, then _both_ fetches in `resolveAccount` _and_ in `resolveAccountData` should use Y encoding

Back to your queue for comments!

packages/rpc-graphql/src/resolvers/account.ts
buffalojoec: Yeah, I was going to tackle this in the next commit, but I guess now that I think of it there's no reason for this one then. 😅

buffalojoec:
> What I hope we can make happen is stuff like this:
> 
> * If there are no `data` fields, fetch the account with the fastest-to-fetch encoding
> * If there is one `data` field set to Y encoding, then _both_ fetches in `resolveAccount` _and_ in `resolveAccountData` should use Y encoding

I think this is probably a good approach. I can latch into this to help with `programAccounts`, which would be complete spam city with this implementation. I might walk this one back.

buffalojoec:
Closing to pull this off the current stack and introduce a new stack for schema revisions.

PR #2020, opened 2024-01-09
refactor(experimental): add support for variable-sized items in remainder-sized codecs

lorisleiva:
This PR allows `Array`, `Map` and `Set` codecs of `"remainder"` size to have variable-size items (as [recently updated on Umi's serializers](https://github.com/metaplex-foundation/umi/pull/107)).

These array-like codecs support the following three configurable size strategies:

- `{ size: 42 }` means a fixed-size array of 42 items.
- `{ size: getU32Codec() }` means the array is prefixed by a `u32` number storing its number of items.
- `{ size: "remainder" }` means the remainder of the bytes should be used to encode and/or decode the items inside the array.

Prior to this PR, we were enforcing arrays of `"remainder"` size to have fixed-size items. This allowed us to pre-compute the size of the array by dividing the size of what's left in the bytes by the fixed size of its item codec. If the modulo of that division was not zero, we would throw an error.

However, this restriction is not actually necessary and prevents valid use cases. This PR lifts this restriction and allows arrays of `"remainder"` size to have variable-size items. When decoding the bytes, instead of pre-computing the size of the array, we simply decode the rest of the bytes continuously until we've reached the end of the bytes. If anything goes wrong during that process, the item codec will throw an error as it won't get the expected number of bytes.

### DIFF
--- a/packages/codecs-data-structures/src/__tests__/array-test.ts
+++ b/packages/codecs-data-structures/src/__tests__/array-test.ts
@@ -79,15 +79,15 @@ describe('getArrayCodec', () => {
         expect(array(string({ size: 1 }), remainder).encode(['a', 'b'])).toStrictEqual(b('6162'));
         expect(array(string({ size: 1 }), remainder).read(b('6162'), 0)).toStrictEqual([['a', 'b'], 2]);
 
+        // Variable sized items.
+        expect(array(string({ size: u8() }), remainder).encode(['a', 'bc'])).toStrictEqual(b('0161026263'));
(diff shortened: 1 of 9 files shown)


lorisleiva:
* `master`
  * **PR #2020** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2020?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2020?utm_source=stack-comment).

buffalojoec (approved):
Nice! This sheds a lot of code!

snip

PR #2025, opened 2024-01-10
Re-export @solana/accounts from @solana/web3.js

mcintyre94:
Same logic as https://github.com/solana-labs/solana-web3.js/pull/2007

`accounts` is a relatively high level package and you shouldn't need to install it separately from `@solana/web3.js` if you're using the main package.


lorisleiva (approved):
🫶

PR #2026, opened 2024-01-10
Add @solana/rpc-json-parsed and move TokenAccount there

mcintyre94:
This PR adds a new package @solana/rpc-json-parsed, which will contain types for the RPC's JSON parsed encoding

I don't currently expect to add anything except types to this package, but it can be expanded later if needed. 

We already had `TokenAccount` in rpc-core which is one of these types, so I've moved that to the new package

I also moved `TokenBalance` from rpc-core to rpc-types, because it is used by RPC methods that don't have a JSON parsed encoding, as well as by `TokenAccount`

As part of this work I noticed that the current implementations of eg `getTokenAccountsByDelegate` are screwy, but I'll fix them in a follow up PR because this one is big enough!


mcintyre94:
* `master`
  * **PR #2026** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2026?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2026?utm_source=stack-comment).

packages/rpc-json-parsed/package.json
```diff
@@ -0,0 +1,95 @@
...
+        "@typescript-eslint/eslint-plugin": "^6.13.2",
+        "@typescript-eslint/parser": "^6.3.0",
+        "agadoo": "^3.0.0",
+        "build-scripts": "workspace:*",
+        "eslint": "^8.45.0",
+        "eslint-plugin-jest": "^27.4.2",
+        "eslint-plugin-sort-keys-fix": "^1.1.2",
+        "jest": "^29.7.0",
+        "jest-environment-jsdom": "^29.7.0",
+        "jest-fetch-mock-fork": "^3.0.4",
```
steveluscher, line 77: When there are no tests, I usually try to nuke these.

packages/rpc-json-parsed/src/token-accounts.ts
```diff
@@ -0,0 +1,17 @@
...
+    mint: Address;
+    owner: Address;
+    tokenAmount: TokenAmount;
+    delegate?: Address;
+    state: TokenAccountState;
+    isNative: boolean;
+    rentExemptReserve?: TokenAmount;
+    delegatedAmount?: TokenAmount;
+    closeAuthority?: Address;
+    extensions?: unknown[];
```
steveluscher, line 16: Why is my beloved key sort linter not failing on this? :)

steveluscher (approved):
Sounds good! Slight preference for calling it something more understandable to non-insiders, like maybe `@solana/rpc-parsed-account-types` or `@solana/rpc-parsed-accounts`.

packages/rpc-json-parsed/src/token-accounts.ts
buffalojoec: I hate that thing.

buffalojoec:
> Sounds good! Slight preference for calling it something more understandable to non-insiders, like maybe `@solana/rpc-parsed-account-types` or `@solana/rpc-parsed-accounts`.

+1 on this, except this will likely have parsed instructions, too. Maybe just `@solana/rpc-parsed-types`?

mcintyre94:
> +1 on this, except this will likely have parsed instructions, too. Maybe just @solana/rpc-parsed-types?

I'm happy with any of these names, but I'm not sure this will include parsed instructions. AFAIK the RPC doesn't do anything clever with those, so you can generate them based just on an IDL, ie Loris' client generation will work there. The only reason these accunt types are different is because the RPC has custom logic for them. I'm not aware of anywhere else that actually happens, but might be wrong.

packages/rpc-json-parsed/src/token-accounts.ts
mcintyre94: The place I copied it from has it in that order too so I think that's been ignoring things for a while, not sure why!

mcintyre94:
Actually I take that back, a `getTransaction` call returns `meta` with:

```json
            "innerInstructions": [
                {
                    "index": 1,
                    "instructions": [
                        {
                            "parsed": {
                                "info": {
                                    "lamports": 1614720,
                                    "newAccount": "HWxduLa57B5KUP36vFy9Rk9H3qRHwUmqfky8SAKP5b9o",
                                    "owner": "4R3gSG8BpU4t19KYj8CfnbtRpnT8gtk4dvTHxVRwc2r7",
                                    "source": "GZctHpWXmsZC1YHACTGGcHhYxjdRqQvTpYkb9LMvxDib",
                                    "space": 104
                                },
                                "type": "createAccount"
                            },
                            "program": "system",
                            "programId": "11111111111111111111111111111111",
                            "stackHeight": 2
                        }
                    ]
                }
            ],
```

And that maps to stuff like `ParsedInstruction` and a familiar enum of `ParsableProgram` in the RPC. Basically the same logic as accounts.

I'll go with `@solana/rpc-parsed-types` for now!

packages/rpc-json-parsed/src/token-accounts.ts
lorisleiva: IIRC, that linter rule only checks "real" JavaScript objects, not object types.

PR #2027, opened 2024-01-10
Fix return type of getTokenAccountsByDelegate/Owner

mcintyre94:
These data stuctures were messed up, with the `TokenAccount` nested under the `program` instead of `data`

I've added typetests to verify the fix. I don't think we had these when we first wrote most of these methods, but they're better than the unit tests we have for rpc-methods TBH. Probably worth adding them at least where we find bugs!

PR #2028, opened 2024-01-10
Add some JSON parsed RPC types

mcintyre94:
Creating the PR at a somewhat arbitrary point to avoid it becoming too big. Will follow up with more PRs for different types, splitting if they become big

- Add address lookup table JSON parsed account types
- Add BPF upgradeable loader JSON parsed account types (as part of this I moved encoded bytes types like Base64EncodedBytes from rpc-core to rpc-types)
- Add Config JSON parsed account types

Note that we already have fixtures for most of these, I've used these to create typetests to validate that the data from eg `getAccountInfo` matches these types


mcintyre94:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2028?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `master`
  * **PR #2026** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2026?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #2027** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2027?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #2028** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2028?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2028?utm_source=stack-comment).

lorisleiva (approved):

packages/rpc-parsed-types/src/config-accounts.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export type ConfigProgramAccount =
+    | Readonly<{
+          info: StakeConfigAccount;
+          type: 'stakeConfig';
+      }>
+    | Readonly<{
+          info: ValidatorInfoAccount;
+          type: 'validatorInfo';
+      }>;
```
buffalojoec, line 24: Wdyt about this?

```ts
type RpcParsedType<T> = Readonly<{
    info: T,
    type: string
}>;

type StakeConfigAccount = Readonly<{
    slashPenalty: number;
    warmupCooldownRate: number;
}>;

// stake config account
{
    const account = {
        info: {
            slashPenalty: 12,
            warmupCooldownRate: 0.25,
        },
        type: 'stakeConfig' as const,
    };
    account satisfies RpcParsedType<StakeConfigAccount>;
}
```

[Playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBASmDGAFAhgJwM4QCYBVwQB4cA+KAXlgmSwHsA7AGxAIG8AoKTqASzoDMaALig4ANBy6hIw9MFS8A5mwC+xANxs2U6AGVgyANYQAwvT7cFAQXjwaAVzrBylavSasJndA2ToAFogQdMgMoMJ0dgC2AEYQqBpcUADuaJF2YKY0DLRJdDDIwBDhUbHxKuqaAPSVULKG0Lb8FlDINvaObOyJjbItbQ5OFF2JXLwCwsMjI96+AUEhYVAAjABM4lNTKahpGTRZOXkFRVAADAB0KwCs6xvKN1PawgDkdUamTQpPLehQPcD3UGUCS4rVsA1qBW46HMEB+cCQaEwuHwBD09Xe5is-UcFWUQA)

buffalojoec: Ohhh I see you are fixing the string value as part of the type. Hmmm. Maybe we can do both somehow?

lorisleiva: This should work.

```ts
type RpcParsedType<TType extends string, TInfo> = Readonly<{
    info: TInfo,
    type: TType
}>;


export type ConfigProgramAccount =
    | RpcParsedType<"stakeConfig", StakeConfigAccount>
    | RpcParsedType<"validatorInfo", ValidatorInfoAccount>;
```

mcintyre94: Nice, that definitely cleans it up a bit!

packages/rpc-parsed-types/src/common.ts
```diff
@@ -0,0 +1 @@
+export type Slot = bigint;
```
buffalojoec, line 1: `Slot` should already be in our other `rpc-types`, no?

lorisleiva: Sadly, it currently lives in `rpc-core`. That's one of the suggested change I mentioned following the dependency graph.

mcintyre94: It's in `rpc-core` but it's also not a branded type, just an alias. So I just copied it here instead of refactoring all its uses

No issue with doing that (and rpc-types is the best place I agree), I just didn't think it was important enough to add to this PR

lorisleiva: +1 on "not a concern for this PR".

Would there be any benefit in making this a branded type? Or is there not much point adding validation to this `bigint` since any `bigint` could in theory be a valid `Slot`?

mcintyre94: I think it would be worth adding for `Slot` actually because it's used as an input type to RPC methods (`minContextSlot` comes to mind, there are probably others). Not critical but I think it'd add value. 

Probably not worth writing an assertion, we could bound check against u64 I guess but I think that's pretty low value

But having type safety from RPC method returning a slot -> RPC method taking a slot as input is probably worthwhile. Or having to explicitly cast a bigint as a `Slot` to use it as one

mcintyre94: I've moved this alias to `rpc-types` and imported it from there in `rpc-parsed-types`, just to reduce the future work.

PR #2029, opened 2024-01-11
The coalescing transport no longer double-throws `AbortError` when the last subscriber aborts

steveluscher:
# Summary

The coalescing transport uses _one_ underlying transport to serve one or more subscribers. When the last subscriber aborts, the underlying transport is aborted.

The problem is that the `AbortError` that results when the underlying transport is aborted was bubbling up through people's programs, uncaught. In this PR, we catch and ignore it, since aborting the underlying transport is a mere implementation detail of how the coalescing transport works.

# Test plan

The test plan is ‘upon making that test synchronous, the test no longer blows up.’ Without the change in behaviour, the Jest test would not even finish.

The edge case in transaction confirmation was confirmed using this:

```diff
diff --git a/packages/library/src/transaction-confirmation-strategy-recent-signature.ts b/packages/library/src/transaction-confirmation-strategy-recent-signature.ts
index 6f5529778..c01d4ea8f 100644
--- a/packages/library/src/transaction-confirmation-strategy-recent-signature.ts
+++ b/packages/library/src/transaction-confirmation-strategy-recent-signature.ts
@@ -26,17 +26,23 @@ export function createRecentSignatureConfirmationPromiseFactory(
         /**
          * STEP 1: Set up a subscription for status changes to a signature.
          */
+        console.log('me subscribing');
         const signatureStatusNotifications = await rpcSubscriptions
             .signatureNotifications(signature, { commitment })
             .subscribe({ abortSignal: abortController.signal });
+        console.log('me subscribed');
         const signatureDidCommitPromise = (async () => {
+            console.log('wiaiting for notifs');
             for await (const signatureStatusNotification of signatureStatusNotifications) {
+                console.log('got notif', signatureStatusNotification);
                 if (signatureStatusNotification.value.err) {
+                    console.log('me bad');
                     // TODO: Coded error.
                     throw new Error(`The transaction with signature \`${signature}\` failed.`, {
                         cause: signatureStatusNotification.value.err,
                     });
                 } else {
+                    console.log('me snagged it');
                     return;
                 }
             }
@@ -46,20 +52,29 @@ export function createRecentSignatureConfirmationPromiseFactory(
          *         This will only yield a result if the signature is still in the status cache.
          */
         const signatureStatusLookupPromise = (async () => {
-            const { value: signatureStatusResults } = await rpc
-                .getSignatureStatuses([signature])
-                .send({ abortSignal: abortController.signal });
-            const signatureStatus = signatureStatusResults[0];
-            if (
-                signatureStatus &&
-                signatureStatus.confirmationStatus &&
-                commitmentComparator(signatureStatus.confirmationStatus, commitment) >= 0
-            ) {
-                return;
-            } else {
-                await new Promise(() => {
-                    /* never resolve */
-                });
+            console.log('getting one shot');
+            try {
+                const { value: signatureStatusResults } = await rpc
+                    .getSignatureStatuses([signature])
+                    .send({ abortSignal: abortController.signal });
+                console.log('got one shot', signatureStatusResults);
+                const signatureStatus = signatureStatusResults[0];
+                if (
+                    signatureStatus &&
+                    signatureStatus.confirmationStatus &&
+                    commitmentComparator(signatureStatus.confirmationStatus, commitment) >= 0
+                ) {
+                    console.log('me good');
+                    return;
+                } else {
+                    console.log('me forever');
+                    await new Promise(() => {
+                        /* never resolve */
+                    });
+                }
+            } catch (e) {
+                console.error('the one shot blew up', e);
+                throw e;
             }
         })();
         try {
diff --git a/packages/library/test.js b/packages/library/test.js
new file mode 100644
index 000000000..7c241289f
--- /dev/null
+++ b/packages/library/test.js
@@ -0,0 +1,56 @@
+import {
+    address,
+    appendTransactionInstruction,
+    createDefaultAirdropRequester,
+    createDefaultRpcSubscriptionsTransport,
+    createDefaultRpcTransport,
+    createDefaultTransactionSender,
+    createSolanaRpc,
+    createSolanaRpcSubscriptions,
+    createTransaction,
+    generateKeyPair,
+    getAddressFromPublicKey,
+    getSignatureFromTransaction,
+    lamports,
+    pipe,
+    setTransactionFeePayer,
+    setTransactionLifetimeUsingBlockhash,
+    signTransaction,
+} from './dist/index.node.js';
+
+// Create an RPC and associated utilities.
+const CLUSTER = 'testnet';
+const rpc = createSolanaRpc({
+    transport: createDefaultRpcTransport({
+        url: 'https://api.' + CLUSTER + '.solana.com',
+    }),
+});
+const rpcSubscriptions = createSolanaRpcSubscriptions({
+    transport: createDefaultRpcSubscriptionsTransport({
+        url: 'wss://api.' + CLUSTER + '.solana.com',
+    }),
+});
+const requestAirdrop = createDefaultAirdropRequester({
+    rpc,
+    rpcSubscriptions,
+});
+const sendTransaction = createDefaultTransactionSender({
+    rpc,
+    rpcSubscriptions,
+});
+
+// Create and fund an account.
+const keypair = await generateKeyPair();
+const myAddress = await getAddressFromPublicKey(keypair.publicKey);
+console.log('Created an account with address', myAddress);
+console.log('Requesting airdrop');
+await requestAirdrop({
+    commitment: 'confirmed',
+    lamports: lamports(1_000_000n),
+    recipientAddress: myAddress,
+});
+console.log('Airdrop confirmed');
+
+setTimeout(() => {
+    console.log('we are still alive!');
+}, 5000);
```

The case you have to trigger is the _subscription_ getting wind of the signature _before_ the one-shot RPC call even finishes. With this PR, that works. Without it, the script fatals on the double-`AbortError`.

### DIFF
--- a/packages/library/src/rpc-request-coalescer.ts
+++ b/packages/library/src/rpc-request-coalescer.ts
@@ -29,13 +29,28 @@ export function getRpcTransportWithRequestCoalescing(
         }
         if (coalescedRequestsByDeduplicationKey[deduplicationKey] == null) {
             const abortController = new AbortController();
+            const responsePromise = (async () => {
+                try {
+                    return await transport<TResponse>({
+                        ...config,
+                        signal: abortController.signal,
+                    });
+                } catch (e) {
+                    if (e && typeof e === 'object' && 'name' in e && e.name === 'AbortError') {
+                        // Ignore `AbortError` thrown from the underlying transport behind which all
+                        // requests are coalesced. If it experiences an `AbortError` it is because
+                        // we triggered one when the last subscriber aborted. Letting the underlying
+                        // transport's `AbortError` bubble up from here would cause runtime fatals
+                        // where there should be none.
+                        return;
+                    }
+                    throw e;
+                }
+            })();
             coalescedRequestsByDeduplicationKey[deduplicationKey] = {
                 abortController,
                 numConsumers: 0,
-                responsePromise: transport<TResponse>({
-                    ...config,
-                    signal: abortController.signal,
-                }),
+                responsePromise,
             };
         }
         const coalescedRequest = coalescedRequestsByDeduplicationKey[deduplicationKey];
--- a/packages/library/src/__tests__/rpc-request-coalescer-test.ts
+++ b/packages/library/src/__tests__/rpc-request-coalescer-test.ts
@@ -9,7 +9,7 @@ describe('RPC request coalescer', () => {
     beforeEach(() => {
         jest.useFakeTimers();
         hashFn = jest.fn();
-        mockTransport = jest.fn();
+        mockTransport = jest.fn().mockResolvedValue(null);
         coalescedTransport = getRpcTransportWithRequestCoalescing(mockTransport, hashFn);
     });
     describe('when requests produce the same hash', () => {
@@ -88,12 +88,16 @@ describe('RPC request coalescer', () => {
             beforeEach(() => {
                 abortControllerA = new AbortController();
                 abortControllerB = new AbortController();
-                mockTransport.mockImplementation(
-                    () =>
-                        new Promise(resolve => {
-                            transportResponsePromise = resolve;
-                        }),
-                );
+                mockTransport.mockImplementation(async ({ signal }) => {
+                    signal?.throwIfAborted();
+                    return await new Promise((resolve, reject) => {
+                        transportResponsePromise = resolve;
+                        signal?.addEventListener('abort', (e: AbortSignalEventMap['abort']) => {
+                            const abortError = new DOMException((e.target as AbortSignal).reason, 'AbortError');
+                            reject(abortError);
+                        });
+                    });
+                });
                 responsePromiseA = coalescedTransport({ payload: null, signal: abortControllerA.signal });
                 responsePromiseB = coalescedTransport({ payload: null, signal: abortControllerB.signal });
             });
@@ -110,13 +114,12 @@ describe('RPC request coalescer', () => {
                 abortControllerA.abort('o no');
                 await expect(responsePromiseA).rejects.toThrow(/o no/);
             });
-            it('aborts the transport when all of the requests abort', async () => {
-                expect.assertions(1);
+            it('aborts the transport when all of the requests abort', () => {
                 abortControllerA.abort('o no A');
                 abortControllerB.abort('o no B');
                 // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
                 const transportAbortSignal = mockTransport.mock.lastCall![0].signal!;
-                await expect(transportAbortSignal.aborted).toBe(true);
+                expect(transportAbortSignal.aborted).toBe(true);
             });
             it('does not abort the transport if fewer than every request aborts', async () => {
                 expect.assertions(1);


steveluscher:
* `master`
  * **PR #2029** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2029?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2029?utm_source=stack-comment).

packages/library/src/rpc-request-coalescer.ts
```diff
@@ -29,13 +29,28 @@ export function getRpcTransportWithRequestCoalescing(
...
+                        // Ignore `AbortError` thrown from the underlying transport behind which all
+                        // requests are coalesced. If it experiences an `AbortError` it is because
+                        // we triggered one when the last subscriber aborted. Letting the underlying
+                        // transport's `AbortError` bubble up from here would cause runtime fatals
+                        // where there should be none.
+                        return;
+                    }
+                    throw e;
+                }
+            })();
```
buffalojoec, line 49: I may be missing something, but wouldn't we want this to just be the `AbortSignal` provided to the transport itself? What's the reason we create a new one here?

steveluscher: The coalescing transport works by:

* reference counting all callers with the ‘same request’ (as defined by having the same deduplication key)
* creating an ‘underlying’ request when the number of callers goes from 0 to 1
* aborting the ‘underlying’ request when the number of callers goes from 1 to 0

When you have 2 or more callers and one is aborted, you want:

* the `AbortError` to be thrown at the callsite of the caller that was aborted
* the ‘underlying’ request to keep on chugging

So no; each caller gets a ‘synthetic’ promise for the ‘underlying’ request's response. When that caller aborts, we only want that synthetic to be aborted. We _don't_ want the ‘underlying’ request to be aborted until there are no more callers waiting on it, at which point we should abort it quietly and without impact to the rest of the app.

packages/library/src/rpc-request-coalescer.ts
```diff
@@ -29,13 +29,28 @@ export function getRpcTransportWithRequestCoalescing(
...
         if (coalescedRequestsByDeduplicationKey[deduplicationKey] == null) {
             const abortController = new AbortController();
+            const responsePromise = (async () => {
+                try {
+                    return await transport<TResponse>({
+                        ...config,
+                        signal: abortController.signal,
+                    });
+                } catch (e) {
+                    if (e && typeof e === 'object' && 'name' in e && e.name === 'AbortError') {
```
lorisleiva, line 39: I think it would be nice to have this condition in a helper method like `isAbortError(e)` because catching them must be quite common.

Just double checking: Is this name condition true for all environments?

lorisleiva (approved):

packages/library/src/rpc-request-coalescer.ts
steveluscher: It is!

I'll make this a helper in a future PR. I think I do this in a few places already.

PR #2031, opened 2024-01-11
Add Nonce and Token JSON parsed account types

mcintyre94:
Most of the changes in this PR come from moving `Blockhash` and `assertIsBlockhash` from `@solana/transactions` to `@solana/rpc-types`. This is done because `Blockhash` is used in the parsed Nonce account. And I think it makes sense anyway!

- Add nonce JSON parsed account type
- Add remaining JSON parsed token accounts (mint, multisig)


mcintyre94:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2031?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `master`
  * **PR #2026** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2026?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #2027** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2027?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #2028** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2028?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #2031** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2031?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2031?utm_source=stack-comment).

packages/library/package.json
```diff
@@ -73,12 +73,12 @@
         "@solana/keys": "workspace:*",
         "@solana/rpc-core": "workspace:*",
         "@solana/rpc-transport": "workspace:*",
-        "@solana/rpc-types": "workspace:*",
         "@solana/transactions": "workspace:*",
         "fast-stable-stringify": "^1.0.0"
     },
     "devDependencies": {
         "@solana/eslint-config-solana": "^1.0.2",
+        "@solana/rpc-types": "workspace:*",
```
lorisleiva, line 81: I like this change — i.e. using type-only dependencies as devDependencies — and the fact that it's consistent with the rest of the library. However, are we sure that this works for TypeScript end-users?

What I mean by that is: say I have package A which only needs types from package B and therefore imports it as a devDependency. Now a user installs package A and starts interacting with the types it provides. When it comes across a type defined in package B, does the user have that information? Or do they need to explicitly add package B as a devDependency as well?

The reason I'm asking this is, I initially wanted Umi to use the `Address` type of the new web3.js library when I refactored that bit. So I did exactly that, imported the `Address` type but imported `@solana/addresses` as a devDependency. Then, when I tested it in a sandbox application, TypeScript was complaining that the `Address` type had no definition.

I think in our case, it's less of an issue because the main library package exports `@solana/rpc-types` anyway so it's unlikely to be missing. I just have this internal fear that it's gonna fail for someone at some point and wanted to raise that so someone can reassure me haha.

packages/rpc-types/package.json
```diff
@@ -92,5 +92,8 @@
                 "path": "./dist/index*.js"
             }
         ]
+    },
+    "dependencies": {
+        "@solana/codecs-strings": "workspace:2.0.0-development"
     }
```
lorisleiva, line 98: Hmmm this no longer makes the `@solana/rpc-types` package a dependency-free package which is a shame considering it is supposed to be this low-level bag of types that people can import from stress free of circular dependencies.

That being said, I'm okay with taking the stance that **codec packages are even lower-level packages** since they enable encoding and decoding any types.

I think this PR is a good change and therefore it's a good tradeoff but I just wanted to flag that so we make an explicit decision on this dependency tree.

lorisleiva (approved):
Just a couple of rants but I love this! ☺️

packages/library/package.json
buffalojoec: I am not the guy to do the reassuring here, but I will add another question!

If you have package B full of types, and inside your main package A you use none of the types, but you re-export B from A, does B tree-shake if a downstream user importing A doesn't use the types from B either?

lorisleiva: I'm 98.7% sure it does tree-shake re-exported items. Otherwise the main library — [which re-exports most of the sub-libraries](https://github.com/solana-labs/solana-web3.js/blob/62b6bd6381e231918ea8666922341c404ac03527/packages/library/src/index.ts#L1-L7) — would not be tree-shakeable at all and I can't imagine Steven letting that happen! 😅

packages/rpc-types/package.json
buffalojoec: I feel similar sentiment @lorisleiva, although I don't have a hard stance on either. I think we should just iron out what this library and the types within it should do. Are they merely layouts, and thus should all be types with no dependencies? Or are they types *with assertions*?

Again on the topic of tree-shaking, if the assertion functions aren't used downstream, codecs should tree-shake out, right? If so it's prob not a big deal.

I do think codecs are probably our lowest-level considering their generic use and lack of dependencies on other packages in our library.

If we went the layout route, where would it make sense for `Blockhash` to live @mcintyre94?

lorisleiva: Yeah they would all tree-shake anyway. I'm more concerned about creating an explicit pattern with our package dependencies. Breaking things up in micro-modules is now but comes at the price of being careful with how we manage these dependencies.

My stance on this is:
- **Codecs are the lowest level packages of the whole library**. We should see this as low-level as using a `Uint8Array`. Similarly to how legacy web3.js would use borsh layouts or bs58. That means, codecs should never start importing stuff like `@solana/transaction`. Never!
- **Type packages** such as `rpc-types` should be **repositories of re-usable types with the helper functions** attached to them. Think of them like tiny type modules. We've even got a nice convention there for these helpers — e.g. `MyType`, `getMyTypeEncoder()`, `getMyTypeDecoder()`, `getMyTypeCodec()`, `isMyType()`, `assertIsMyType()` and even `myType()` to create such type when applicable.
- Everything else is just packages built on top of that in a careful hierarchy to avoid circular dependencies, creating more dedicated type packages when required.

P.S.: just pasting the current web3.js dependency graph here as well so it's publicly accessible.

Web3 js dependency graph

buffalojoec: Yep, I agree with this approach!

Maybe before we add types to `@solana/rpc-parsed-types`, we hash out the orange arrows in your diagram and follow the above proposed plan?

mcintyre94: I had the same thought about adding a dependency here.

The other option here is to leave `assertIsBlockhash` with `transactions`, or move it somewhere else. But `rpc-types` does have the other assert functions colocated with the types it defines, and I thought that consistency was more important than avoiding a codecs dependency.

IMO this colocation makes the most sense. I agree with Loris' comment above that these packages should include their helper functions in general. I don't really mind thought because if we move them somewhere else and just make sure it's exported from the main library, it'll be indistinguishable to most devs. We should be consistent though.

packages/library/package.json
mcintyre94: I'll move this back to a normal dependency in `library`, I don't think I meant to specifically change that one. I think I probably ran a `pnpm add --save-dev` without realising it was already a dependency and it got moved, or I accidentally moved it myself. I don't know exactly how these dependencies/re-exporting etc link together so I won't fiddle with it!

packages/rpc-types/package.json
lorisleiva: @mcintyre94 Agree with everything. I don't even think `rpc-parsed-types` having actual logic is an issue as long as its dependencies can only be at most: `codecs` and `rpc-types`.

Just let me know when you're happy with me detangling these types from `rpc-core` and `rpc-transport` because I expect if I do that now, we'll have lots of annoying merge conflicts to deal with haha.

mcintyre94: @lorisleiva I think I can probably finish these parsed account types today

I think the best thing is probably if I keep doing the minimum changes necessary for these (basically move stuff these need to rpc-types), and then you take over with the wider changes after that? I don't think anything I do will add work to that refactor!

lorisleiva: Perfect plan! 🙇‍♂️

packages/library/package.json
lorisleiva: @steveluscher ICYMI, I'd love your thoughts on the OP of this thread.

Winds126 (approved):

packages/library/package.json
lorisleiva: My fears have been confirmed:

https://stackoverflow.com/a/46011417/11440277

PR #2035, opened 2024-01-12
Add Stake, Sysvar, Vote JSON parsed types

mcintyre94:
This PR adds the remaining JSON parsed account types

- Add stake JSON parsed account types

- Add sysvar JSON parsed account types

- Add votes JSON parsed account types

Note that I've skipped the Epoch Reward sysvar, which seems to be removed from all networks. If someone corrects me I'll add it: https://solana.stackexchange.com/questions/9250/is-the-epoch-rewards-sysvar-deprecated

Edit: Corrected, it's upcoming! Added it based on validator source code.


mcintyre94:
* `master`
  * **PR #2035** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2035?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2035?utm_source=stack-comment).

buffalojoec (commented):
Looks good but do you have any concern with confusing people by using `Program` in the names? Particularly `SysvarProgram`? Technically it's not a program.

The others make sense, people are just hyper-used to `StakeAccount`, `VoteAccount`, etc. Not opposed to those though!

lorisleiva:
@buffalojoec +1 that also confused me. What is the difference between the `StakeAccount` and the `StakeProgramAccount` types. Why does the exported one need to be named like that?

If it's arbitrary, I think it'd be best to call the exported one `StakeAccount` and the internal one something like `StakeAccountData`.

mcintyre94:
Fair point! The main reason I added Program to them was because of Token. That one has multiple account types, and one of them is of type `token`. So `TokenProgramAccount` covers TokenAccount, MintAccount and MultisigAccount in that case. And we already had `TokenAccount` as the one I copied from rpc-core, and that one needs to keep being exported so I didn't want to mess with its name. 

So I needed something to distinguish those. `TokenAccountData` vs `TokenAccount` works reasonably enough though. It also makes the exports consistent between the ones with multiple types and the ones with just one.

mcintyre94:
I've renamed them to remove the `ProgramAccount`, I think that's definitely better overall

My only concern is that I'm not sure what people expect `TokenAccount` to mean. I think in many cases people mean what I've called `TokenAccountData`, ie the account info of type `token` owned by a Token/Token22 program. `StakeAccount` and `VoteAccount` are easier because they're always the same data shape. But I think we need some level of consistency here and this is probably the least annoying option.

lorisleiva:
Oh I see, the program in the name was to indicate that it's a union of all accounts available in said program.

Do we need that union type in the first place though? Could we not just export `TokenAccount`, `MintAccount`, etc. as standalone types and if someone want to support both the Token account and the Mint account for example, they'd just need to do the union themselves `TokenAccount | MintAccount`?

mcintyre94:
The union type is really about strongly typing the `type` field

An example of how I want to use this:

I fetch accounts with `jsonParsed` encoding, knowing nothing about them except their address. Based on the owner I cast to eg `StakeAccount` or `TokenAccount`. Or if I know I'm fetching account info for eg a token account, I can use `fetchJsonParsedAccount<TokenAccount>`

Now I can use the `type` field to get type safety (and TBH more usefully, typescript intellisense) on the fields, for example I can do this:

```ts
// account is TokenAccount
if(account.type === 'token') {
  const owner = account.info.owner
}
```

So it's less about somebody supporting these as types, and more about making it easy to work with the account data that you've fetched.

buffalojoec:
It sounds to me, from your description, we should have fixed `type` field types for `MintAccount` and `TokenAccount` and the `jsonParsed` RPC call for gPA on Token returns `type TokenProgramAccount = `MintAccount | TokenAccount`?

mcintyre94:
> The part I'm still confused about is when exactly would a user end up using something like AnyAccountFromTokenProgram?

> Surely when I fetch an account I'm not like "hey this could be any type of account from that program I don't care yolo", right? 😅

Here's a hypothetical 🪽

You want to understand what a transaction might do. You have a deserialized transaction, which lists a bunch of writable addresses. At this point you know nothing about any of those addresses, only that this transaction marks them writable.

You can use `getMultipleAccounts` with a JSON parsed encoding to fetch them. Some of them might return nicely parsed `data`, some might not. Those that will do that have known owners, because that's how the RPC parse account logic works.

So you can check if the owner is, say, the token program address. If it is then you know that you got parsed data for it, but you don't know exactly what type of token program related account/data it is. At this point I propose web3js steps in to help you: you can cast it to `AnyAccountFromTokenProgram`. It's not that you don't care yolo, it's that your code has no way of narrowing it any further than that yet.

Once you've done that cast, you can check the type, and narrow it in your code to a `TokenAccount`, or a `MintAccount`. And then you can figure out based on those known data structures whether you need to care about that account being writable in the transaction.

I don't think this is an overly niche situation though. Explorers face a similar challenge when you search for an arbitrary account address. Any attempt to parse a transaction where you don't have an IDL for every instruction leaves you in the same situation of sometimes only knowing an address.

mcintyre94:
I agree, but I think `TokenProgramAccount` as the typed union for `TokenAccount | MintAccount | MultisigAccount` probably works? For sysvar I'll just use `SysvarAccount` because there's no risk of a clash there.

I can also add `JsonParsed` prefixes per https://github.com/solana-labs/solana-web3.js/pull/2037#issuecomment-1889743200 - agree with that commment

lorisleiva:
@mcintyre94 I like that! 👌

mcintyre94:
Cool this should be good to review now! :)

lorisleiva (approved):

packages/rpc-parsed-types/src/bpf-upgradeable-loader-accounts.ts
```diff
@@ -3,16 +3,16 @@ import type { Base64EncodedDataResponse, Slot } from '@solana/rpc-types';
 
 import type { RpcParsedType } from './rpc-parsed-type';
 
-type ProgramAccount = Readonly<{
+type JsonParsedProgramAccount = Readonly<{
     programData: Address;
 }>;
 
-type ProgramDataAccount = Readonly<{
+type JsonParsedProgramDataAccount = Readonly<{
```
buffalojoec, line 10: `JsonParsedBpfUpgradeableLoaderProgramAccount` & `JsonParsedBpfUpgradeableLoaderProgramDataAccount`?

It's a mouthful, but these might confuse people into thinking they're the parent types of `JsonParsedXProgramAccount`, or for a generic `ProgramAccount` altogether.

packages/rpc-parsed-types/src/sysvar-accounts.ts
```diff
@@ -0,0 +1,79 @@
...
+    lamportsPerSignature: StringifiedBigInt;
+}>;
+
+type JsonParsedClockAccount = Readonly<{
+    slot: Slot;
+    epoch: Epoch;
+    epochStartTimestamp: UnixTimestamp;
+    leaderScheduleEpoch: Epoch;
+    unixTimestamp: UnixTimestamp;
+}>;
```
buffalojoec, line 15: These made me think of possibly rolling a helper somewhere, maybe in the main library, for `getClock()`. Thoughts?

packages/rpc-parsed-types/src/sysvar-accounts.ts
```diff
@@ -0,0 +1,79 @@
...
+    | RpcParsedType<'clock', JsonParsedClockAccount>
+    | RpcParsedType<'epochSchedule', JsonParsedEpochScheduleAccount>
+    | RpcParsedType<'fees', JsonParsedFeesAccount_DEPRECATED>
+    | RpcParsedType<'recentBlockhashes', JsonParsedRecentBlockhashesAccount_DEPRECATED>
+    | RpcParsedType<'rent', JsonParsedRentAccount>
+    | RpcParsedType<'slotHashes', JsonParsedSlotHashesAccount>
+    | RpcParsedType<'slotHistory', JsonParsedSlotHistoryAccount>
+    | RpcParsedType<'stakeHistory', JsonParsedStakeHistoryAccount>
+    | RpcParsedType<'lastRestartSlot', JsonParsedLastRestartSlotAccount>
+    | RpcParsedType<'epochRewards', JsonParsedEpochRewardsAccount>;
```
buffalojoec, line 79: Maybe also `getSysvar(sysvar: Sysvar)` (re: previous comment).

packages/rpc-parsed-types/src/token-accounts.ts
```diff
@@ -27,14 +27,14 @@ export type MintAccount = Readonly<{
...
 }>;
 
-export type TokenProgramAccount =
-    | RpcParsedType<'account', TokenAccount>
-    | RpcParsedType<'mint', MintAccount>
-    | RpcParsedType<'multisig', MultisigAccount>;
+export type JsonParsedTokenProgramAccount =
+    | RpcParsedType<'account', JsonParsedTokenAccount>
+    | RpcParsedType<'mint', JsonParsedMintAccount>
+    | RpcParsedType<'multisig', JsonParsedMultisigAccount>;
```
buffalojoec, line 40: 💪🏼 💪🏼

packages/rpc-parsed-types/src/sysvar-accounts.ts
mcintyre94: I think maybe a `@solana/sysvars` package with `getClock()`, `getEpochSchedule()`, etc might be useful?
Just encapsulates the sysvar addresses + `fetchJsonParsedAccount` + these types.

packages/rpc-parsed-types/src/bpf-upgradeable-loader-accounts.ts
mcintyre94: `JsonParsedBpfUpgradeableLoaderProgramAccount` clashes with the tagged union type

I'll just add `Bpf` to these to make sure it's clear what they come from. These nested types aren't exported currently, so I'd rather lose a bit of consistency there than on the exported tagged union type.

packages/rpc-parsed-types/src/sysvar-accounts.ts
buffalojoec: I dig it!

packages/rpc-parsed-types/src/sysvar-accounts.ts
buffalojoec: #2041

PR #2044, opened 2024-01-18
Change publish-packages to handle already published without --continue

mcintyre94:
This PR is another attempt to fix our publish-packages job

I've only updated the script for accounts, but it'd be the same change for all packages. Let's nail it down and then I can copy it to the rest

`publish-packages` now runs `npm view` with the name + version to be deployed. If this has an exit code of 0 (ie success, the package exists) then it exits without attempting to re-publish

If it has a non-zero exit code (ie the package does not exist), then it publishes it

This means that we no longer fail `publish-packages` if a package is already published. It will only fail if it is not published, and publishing fails. Therefore we remove the `--continue` flag, meaning we no longer publish a package if any of its dependencies failed to publish (and are not already published)

Closes: #1402


buffalojoec (commented):
Does this account for the commit hash, too? Since our versions have the commit hashes in them when they're published, I wonder if this is _never_ going to find the matching version, because a commit hash will always be new, even if the package in question isn't changed.

mcintyre94:
Yep at the point `publish-packages` runs the version that `$npm_package_version` is referring to is eg `2.0.0-experimental.909038e`

The only time it should find a matching version is when we're re-running the CI on a commit that has already been run, and some packages were published while others weren't for some reason. In that case it will be running with the exact same commit hash and will find the published version for the packages that published successfully.

We publish every package even if it hasn't changed, that doesn't change here. If it's a new commit being published for the first time, every package will (attempt to) be published.

buffalojoec:
Ahhhh I think I conflated the issue you were trying to solve here with a different one. Nice! Seems like it will work to me!

lorisleiva:
I can't wait for us to use `changeset`. 🥹

steveluscher (approved):
This looks awesome. Ship it.

And maybe don't modify those line endings :)

packages/library-legacy/package.json
```diff
@@ -43,7 +43,7 @@
     "build:fixtures": "set -ex; ./test/fixtures/noop-program/build.sh",
     "clean": "rimraf ./doc ./declarations ./lib",
     "dev": "cross-env NODE_ENV=development rollup -c --watch",
-    "publish-packages": "semantic-release --repository-url git@github.com:solana-labs/solana-web3.js.git",
+    "publish-packages": "npm view $npm_package_name@$npm_package_version > /dev/null 2>&1 || semantic-release --repository-url git@github.com:solana-labs/solana-web3.js.git",
```
mcintyre94, line 46: Question: does this look right for library-legacy? I'm guessing `semantic-release ...` is kinda the same as `pnpm publish ...`, so I can do the same for it?

packages/library/package.json
```diff
@@ -40,7 +40,8 @@
         "compile:typedefs": "tsc -p ./tsconfig.declarations.json && node node_modules/build-scripts/add-js-extension-to-types.mjs",
         "dev": "jest -c node_modules/test-config/jest-dev.config.ts --rootDir . --watch",
         "prepublishOnly": "version-from-git --no-git-tag-version --template experimental.short",
-        "publish-packages": "sed -i 's/@solana\\/web3\\.js-experimental/@solana\\/web3\\.js/g' package.json && sed -i 's/@solana\\/web3\\.js/@solana\\/web3\\.js-bak/g' ../library-legacy/package.json && pnpm publish --tag experimental --access public --no-git-checks && git reset --hard",
+        "publish-impl": "npm view $npm_package_name@$npm_package_version > /dev/null 2>&1 || pnpm publish --tag experimental --access public --no-git-checks",
+        "publish-packages": "sed -i 's/@solana\\/web3\\.js-experimental/@solana\\/web3\\.js/g' package.json && sed -i 's/@solana\\/web3\\.js/@solana\\/web3\\.js-bak/g' ../library-legacy/package.json && pnpm publish-impl && git reset --hard",
```
mcintyre94, line 44: Since `publish-packages` does a bit more here I've swapped out the `pnpm publish ...` command for `pnpm publish-impl`, which does the `npm view ... || pnpm publish ...` piece

packages/library-legacy/package.json
buffalojoec: I think this is OK. `semantic-release` will handle incrementing the packages versions beyond just the commit hash, but if the commit message doesn't warrant a new patch/minor/major, then this will perform the intended check.

buffalojoec (approved):
Ship it

steveluscher (approved):
[beautiful](https://app.graphite.dev/macros/solana-labs)

PR #2052, opened 2024-01-19
refactor(experimental): rpc-core: add master union types to API definitions

buffalojoec:
Our RPC API is very difficult to use when attempting to define types and/or use
IntelliSense on the parameters of any of the defined methods.

As an example, I've defined a type that uses
`Parameters<GetAccountInfoApi['getAccountInfo']` and you can see that in my
first attempt - without making any changes to the library - the type inferred
from that declaration is missing a few fields, particularly `encoding` and
`dataSlice`.

Screenshot 2024-01-19 at 12 04 52 PM

Because of the way our overloads are structured, TypeScript can't seem to
understand all possible variations of the arguments. Instead, it just takes the
last overload as the type.

This change proposes we add a master union function declaration at the end of
every RPC Core API method, so that IntelliSense and TypeScript can interpret the
full union for parameters or return types.

Screenshot 2024-01-19 at 12 06 06 PM

Edit: After further review, this change in fact does preserve type inference based on inputs for all API interfaces. However, `sendTransaction` and `simulateTransaction` (later in this stack), as well as any other overloads that are non-exhaustive, will need special handling. The rest are good to go.

### DIFF
--- a/packages/rpc-graphql/src/loaders/transaction.ts
+++ b/packages/rpc-graphql/src/loaders/transaction.ts
@@ -20,14 +20,7 @@ async function loadTransaction(
     rpc: Rpc<GetTransactionApi>,
     { signature, ...config }: TransactionLoaderArgs,
 ): Promise<TransactionLoaderValue> {
-    // @ts-expect-error FIX ME: https://github.com/solana-labs/solana-web3.js/pull/2052
-    return await rpc
-        .getTransaction(
-            signature,
-            // @ts-expect-error FIX ME: https://github.com/solana-labs/solana-web3.js/pull/2052
-            config,
-        )
-        .send();
+    return await rpc.getTransaction(signature, config).send();
 }
 
 function createTransactionBatchLoadFn(rpc: Rpc<GetTransactionApi>) {
--- a/packages/rpc-core/src/rpc-methods/__typetests__/common.ts
+++ b/packages/rpc-core/src/rpc-methods/__typetests__/common.ts
@@ -1,0 +1,4 @@
+export function assertNotAProperty<T extends object, TPropName extends string>(
+    _: { [Prop in keyof T]: Prop extends TPropName ? never : T[Prop] },
+    _propName: TPropName,
+): void {}
--- a/packages/rpc-core/src/rpc-methods/__typetests__/get-account-info-typetest.ts
+++ b/packages/rpc-core/src/rpc-methods/__typetests__/get-account-info-typetest.ts
@@ -1,0 +1,114 @@
+import { Address } from '@solana/addresses';
+import {
+    Base58EncodedBytes,
+    Base58EncodedDataResponse,
+    Base64EncodedDataResponse,
+    Base64EncodedZStdCompressedDataResponse,
+    Commitment,
+    Rpc,
+    U64UnsafeBeyond2Pow53Minus1,
+} from '@solana/rpc-types';
+
+import { GetAccountInfoApi } from '../getAccountInfo';
+
+const rpc = null as unknown as Rpc<GetAccountInfoApi>;
+const address = 'Joe11111111111111111111111111111' as Address<'Joe11111111111111111111111111111'>;
+
+// Parameters
+const params = null as unknown as Parameters<GetAccountInfoApi['getAccountInfo']>[1];
+params satisfies { commitment?: Commitment } | undefined;
+params satisfies { dataSlice?: { length: number; offset: number } } | undefined;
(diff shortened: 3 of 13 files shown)


buffalojoec:
* `master`
  * **PR #2052** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2052?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2052?utm_source=stack-comment).

lorisleiva:
I think this makes sense for devs building on top of these RPC methods since they can just delegate to the function type and handle each possible return values accordingly. It's also type accurate since, by providing a looser type, you're getting a looser response which you have to narrow yourself.

Even internally, like in `@solana/accounts` I've had to tell TypeScript `as OneOfTheOverloads` just to avoid having a switch/case that does the same thing for all of the overloaded definitions.

buffalojoec:
> I think this makes sense for devs building on top of these RPC methods since they can just delegate to the function type and handle each possible return values accordingly. It's also type accurate since, by providing a looser type, you're getting a looser response which you have to narrow yourself.
> 
> Even internally, like in `@solana/accounts` I've had to tell TypeScript `as OneOfTheOverloads` just to avoid having a switch/case that does the same thing for all of the overloaded definitions.

OK nice. I either have to revisit this PR to see if I fat-fingered the `getBlock` one, or this creates an issue. I think some type tests are needed (⚰️ ⚰️), which I can look into (🪦).

But if we think it's a worthy endeavor I'd like to get this sorted, since anything built on the new library(ies) is going to have this problem. I'm seeing it in `@solana/rpc-graphql` and I've littered the codebase with `@ts-expect-error FIXME`'s.

lorisleiva (approved):
Awesome! That's a shit lot of work haha. I massively appreciate the effort that went into making sure the union overload includes everything. 🙏

buffalojoec:
> Awesome! That's a shit lot of work haha. I massively appreciate the effort that went into making sure the union overload includes everything. 🙏

I'm going to turn this into a stack with the other methods!


pain

steveluscher (changes requested):
@lorisleiva just caught me up on what this stack is all about. So, I totally get it. I think that what you want to catch is the case where someone has some inputs that may _actually_ be compatible, but we don't know that statically. Something like this.

```ts
// Against the legacy `Connection` class, for instance...
const whoKnowsTx = null as unknown as (VersionedTransaction | Transaction);
const whoKnowsSecondArg = null as unknown as (Array<Signer> | SendOptions)

(null as unknown as Connection).sendTransaction(
  whoKnowsTx,
  whoKnowsSecondArg,
);
```

This, even on the legacy API with ‘normal’ method overloads is a type error.

I think this is a Good Thing Akshually™.  If you have a bag of values with unknown types you should _have_ to assert on their types before trying  to use them with a method that requires certain valid combinations. Not doing so is how you get runtime errors in production.

I probably wouldn't continue with this stack, and instead either do nothing, or create a few more kid-glove assertion methods that people can use to refine their types.

packages/rpc-graphql/src/loaders/transaction.ts
```diff
@@ -20,14 +20,7 @@ async function loadTransaction(
     rpc: Rpc<GetTransactionApi>,
     { signature, ...config }: TransactionLoaderArgs,
 ): Promise<TransactionLoaderValue> {
-    // @ts-expect-error FIX ME: https://github.com/solana-labs/solana-web3.js/pull/2052
-    return await rpc
-        .getTransaction(
-            signature,
-            // @ts-expect-error FIX ME: https://github.com/solana-labs/solana-web3.js/pull/2052
```
steveluscher, line 27: I think that the real FIXME is https://github.com/microsoft/TypeScript/issues/43187

> Generic inference to an overloaded function always chooses the last signature.

buffalojoec: Yup. Smh.

buffalojoec: #2133

buffalojoec:
I noticed this issue while building on top of `@solana/rpc-core` within `@solana/rpc-graphql`. However, after further discussion, it seems that it might be best to discard this stack.

If users run into similar problems in the future and need a solution, we can possibly revisit this stack. Closing all children.

PR #2053, opened 2024-01-21
refactor(experimental): add cluster level API for transports

buffalojoec:
This change introduces cluster-level transports as well as cluster-level JSON RPCs
that infer the cluster from the provided transport.

### DIFF
--- a/packages/rpc-transport/src/json-rpc.ts
+++ b/packages/rpc-transport/src/json-rpc.ts
@@ -1,8 +1,15 @@
-import { PendingRpcRequest, Rpc, RpcRequest, SendOptions } from '@solana/rpc-types';
+import { IRpcApi, PendingRpcRequest, Rpc, RpcRequest, SendOptions } from '@solana/rpc-types';
 
 import { RpcConfig } from './json-rpc-config';
 import { SolanaJsonRpcError } from './json-rpc-errors';
 import { createJsonRpcMessage } from './json-rpc-message';
+import { RpcDevnet, RpcFromTransport, RpcMainnet, RpcTestnet } from './json-rpc-types';
+import {
+    IRpcTransport,
+    IRpcTransportDevnet,
+    IRpcTransportMainnet,
+    IRpcTransportTestnet,
+} from './transports/transport-types';
 
 interface IHasIdentifier {
     readonly id: number;
@@ -54,6 +61,32 @@ function makeProxy<TRpcMethods>(rpcConfig: RpcConfig<TRpcMethods>): Rpc<TRpcMeth
     }) as Rpc<TRpcMethods>;
 }
 
-export function createJsonRpc<TRpcMethods>(rpcConfig: RpcConfig<TRpcMethods>): Rpc<TRpcMethods> {
-    return makeProxy(rpcConfig);
+export function createJsonRpc<TRpcMethods>(
+    rpcConfig: Readonly<{
+        api: IRpcApi<TRpcMethods>;
+        transport: IRpcTransportDevnet;
+    }>,
+): RpcDevnet<TRpcMethods>;
+export function createJsonRpc<TRpcMethods>(
+    rpcConfig: Readonly<{
+        api: IRpcApi<TRpcMethods>;
+        transport: IRpcTransportTestnet;
+    }>,
+): RpcTestnet<TRpcMethods>;
+export function createJsonRpc<TRpcMethods>(
+    rpcConfig: Readonly<{
+        api: IRpcApi<TRpcMethods>;
+        transport: IRpcTransportMainnet;
+    }>,
+): RpcMainnet<TRpcMethods>;
+export function createJsonRpc<TRpcMethods>(
+    rpcConfig: Readonly<{
+        api: IRpcApi<TRpcMethods>;
+        transport: IRpcTransport;
+    }>,
+): Rpc<TRpcMethods>;
+export function createJsonRpc<TRpcMethods, TConfig extends RpcConfig<TRpcMethods>>(
+    rpcConfig: TConfig,
+): RpcFromTransport<TRpcMethods, TConfig['transport']> {
+    return makeProxy(rpcConfig) as RpcFromTransport<TRpcMethods, TConfig['transport']>;
 }
--- a/packages/rpc-transport/src/transports/http/http-transport.ts
+++ b/packages/rpc-transport/src/transports/http/http-transport.ts
@@ -1,6 +1,7 @@
+import { ClusterUrl } from '@solana/rpc-types';
 import fetchImpl from 'fetch-impl';
 
-import { IRpcTransport } from '../transport-types';
+import { IRpcTransport, IRpcTransportFromClusterUrl } from '../transport-types';
 import { SolanaHttpError } from './http-transport-errors';
 import {
     AllowedHttpRequestHeaders,
@@ -8,12 +9,15 @@ import {
     normalizeHeaders,
 } from './http-transport-headers';
 
-type Config = Readonly<{
+type Config<TClusterUrl extends ClusterUrl> = Readonly<{
     headers?: AllowedHttpRequestHeaders;
-    url: string;
+    url: TClusterUrl;
 }>;
 
-export function createHttpTransport({ headers, url }: Config): IRpcTransport {
+export function createHttpTransport<TClusterUrl extends ClusterUrl>({
+    headers,
+    url,
+}: Config<TClusterUrl>): IRpcTransportFromClusterUrl<TClusterUrl> {
     if (__DEV__ && headers) {
         assertIsAllowedHttpRequestHeaders(headers);
     }
@@ -43,5 +47,5 @@ export function createHttpTransport({ headers, url }: Config): IRpcTransport {
             });
         }
         return (await response.json()) as TResponse;
-    };
+    } as IRpcTransportFromClusterUrl<TClusterUrl>;
 }
--- a/packages/rpc-transport/src/__typetests__/json-rpc-typetest.ts
+++ b/packages/rpc-transport/src/__typetests__/json-rpc-typetest.ts
@@ -1,0 +1,48 @@
+import { devnet, IRpcApiMethods, mainnet, Rpc, testnet } from '@solana/rpc-types';
+
+import { createJsonRpcApi } from '../apis/methods/methods-api';
+import { createJsonRpc } from '../json-rpc';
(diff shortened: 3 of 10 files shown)


buffalojoec:
* `master`
  * **PR #2053** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2053?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #2054** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2054?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #2055** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2055?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2053?utm_source=stack-comment).

lorisleiva:
I like this stack a lot. I love that only with types, we can cherry-pick the RPC methods available in different clusters.

The only API change I want to put on the table is, instead of _requesting_ a second `TCluster` type parameter which makes it tricky when you don't need to offer the first one, why don't we _infer_ that type from the config object provided. IIRC @steveluscher mentioned something like this a while ago. For instance, we could attach that piece of information on the URL directly.

```ts
type MainnetUrl = string & { '~cluster': 'mainnet' }
type DevnetUrl = string & { '~cluster': 'devnet' }
type TestnetUrl = string & { '~cluster': 'testnet' }
type ClusterUrl = string | MainnetUrl | DevnetUrl | TestnetUrl

function mainnet(putativeString: string): MainnetUrl { return putativeString as MainnetUrl }
function devnet(putativeString: string): DevnetUrl { return putativeString as DevnetUrl }
function testnet(putativeString: string): TestnetUrl { return putativeString as TestnetUrl }

// Devnet example.
const devnetTransport = createDefaultRpcTransport({ url: devnet('https://api.devnet.solana.com') });
createSolanaRpc({ transport: devnetTransport });
//  ^ RpcDevnet<DevnetSolanaRpcMethods> 

// Generic example.
const genericTransport = createDefaultRpcTransport({ url: 'http://127.0.0.1:8899' });
createSolanaRpc({ transport: genericTransport });
//  ^ Rpc<SolanaRpcMethods> 
```

Wdyt?

buffalojoec:
> The only API change I want to put on the table is, instead of _requesting_ a second `TCluster` type parameter which makes it tricky when you don't need to offer the first one, why don't we _infer_ that type from the config object provided. 

Interesting. Just thinking about your suggestion out loud here:

For just `@solana/rpc-transport`, we have two methods:
- `createJsonRpcApi<TMethods>(..)`: This one _requires_ the API generic, and we can optionally pass `TCluster` after it, or provide the cluster within the config as you've suggested.
- `createJsonRpc(.)`: This one does not take a generic, and we infer the cluster from it's config arg.

In either case, now that I look at it some more, when it comes to the main library in `@solana/web3.js`, I definitely can refactor the `createDefaultRpcTransport(..)` to apply the cluster argument there and eliminate the need for it - in either form - on `createSolanaRpc(..)`.

Personally I think using the word, for example `devnet`, in the generic argument versus the config object is not all that aesthetically different. However, one thing I do really like about your suggestion is the declaration of the cluster is _right there_ next to the URL string itself (lol) so it would be hard to fat-finger. Also, I think it may slightly clean up some of the new types I rolled here.

I'm going to give it a whirl!

lorisleiva:
What if we don't tag the API methods with a cluster though? What if we just have various API methods types available like `SolanaRpcMethodsMainnet`, `SolanaRpcMethodsDevnet` that includes `requestAirdrop`, etc.

You can see them as various libraries of RPC methods available in different clusters.

Then we select the one we need based on the given transport URL when we create the RPC (transport + api). Is this making any sense? 😅

buffalojoec:
> What if we don't tag the API methods with a cluster though? What if we just have various API methods types available like `SolanaRpcMethodsMainnet`, `SolanaRpcMethodsDevnet` that includes `requestAirdrop`, etc.
> 
> You can see them as various libraries of RPC methods available in different clusters.
> 
> Then we select the one we need based on the given transport URL when we create the RPC (transport + api). Is this making any sense? 😅

I had them isolated into per-cluster sets when I originally created this stack (or one before it I forget). Is this what you're thinking?

```ts
createJsonRpc({
    api: createJsonRpcApi<SolanaRpcMethodsDevnet>(),
    transport: createHttpTransport({ url: devnet('https://api.devnet.solana.com') }),
}); // RpcDevnet<SolanaRpcMethodsDevnet>

createJsonRpc({
    api: createJsonRpcApi<SolanaRpcMethodsTestnet>(),
    transport: createHttpTransport({ url: testnet('https://api.testnet.solana.com') }),
}); // RpcTestnet<SolanaRpcMethodsTestnet>

createJsonRpc({
    api: createJsonRpcApi<SolanaRpcMethodsMainnet>(),
    transport: createHttpTransport({ url: mainnet('https://api.mainnet-beta.solana.com') }),
}); // RpcMainnet<SolanaRpcMethodsMainnet>
```

lorisleiva:
Yes, exactly that. Then it's the responsibility of `createSolanaRpc` to provide the right default `api` based on the provided `transport`. Was there any issues with this approach?

buffalojoec:
> Yes, exactly that. Then it's the responsibility of `createSolanaRpc` to provide the right default `api` based on the provided `transport`. Was there any issues with this approach?

No I think that will work fine. Anyone who creates their own RPC client from our libraries would just have to make sure they align the proper API based on cluster as well, similar to what I just did above.

For our Solana-implemented helpers, we can wrap that.

packages/rpc-transport/src/json-rpc.ts
```diff
@@ -54,6 +68,32 @@ function makeProxy<TRpcMethods>(rpcConfig: RpcConfig<TRpcMethods>): Rpc<TRpcMeth
...
+export function createJsonRpc<TRpcMethods>(
+    rpcConfig: Readonly<{
+        api: IRpcApi<TRpcMethods>;
+        transport: IRpcTransport;
+    }>,
+): Rpc<TRpcMethods>;
+export function createJsonRpc<TRpcMethods, TConfig extends RpcConfig<TRpcMethods>>(
+    rpcConfig: TConfig,
+): RpcFromTransport<TRpcMethods, TConfig['transport']> {
+    return makeProxy(rpcConfig) as RpcFromTransport<TRpcMethods, TConfig['transport']>;
```
buffalojoec, line 98: I went with overloads here because we've designed this API for users to provide their RPC methods. It can be removed, but this function would then take 2 generic arguments, and the moment you provide your RPC methods it will complain you didn't provide `TConfig` as well.

packages/rpc-transport/src/transports/http/http-transport.ts
```diff
@@ -10,10 +10,13 @@ import {
 
 type Config = Readonly<{
     headers?: AllowedHttpRequestHeaders;
-    url: string;
+    url: ClusterUrl;
 }>;
 
-export function createHttpTransport({ headers, url }: Config): IRpcTransport {
+export function createHttpTransport<TConfig extends Config>({
```
buffalojoec, line 16: Introduced a generic arg here since we never had one before and it can be inferred.

packages/rpc-types/src/rpc-api.ts
```diff
@@ -12,6 +13,17 @@ type RpcTransportConfig = Readonly<{
 export interface IRpcTransport {
     <TResponse>(config: RpcTransportConfig): Promise<TResponse>;
 }
+export type IRpcTransportDevnet = IRpcTransport & { '~cluster': 'devnet' };
+export type IRpcTransportTestnet = IRpcTransport & { '~cluster': 'testnet' };
+export type IRpcTransportMainnet = IRpcTransport & { '~cluster': 'mainnet' };
+export type IIRpcTransport = IRpcTransport | IRpcTransportDevnet | IRpcTransportTestnet | IRpcTransportMainnet;
```
lorisleiva, line 19: `IICaptain` 🫡

This double `I` feels very strange to me. 😅

We may not even need this `IIRpcTransport` type since `IRpcTransportDevnet`, for instance, already extends `IRpcTransport`.

Also, I'm still making my way up the stack but I still feel like these types could be part of `@solana/rpc-transports`.

buffalojoec: > We may not even need this IIRpcTransport type since IRpcTransportDevnet, for instance, already extends IRpcTransport.

Yeah I thought about this, too, but to me it felt easier to read/follow the type conditionals with these explicit.

packages/rpc-transport/src/transports/http/http-transport.ts
lorisleiva: We could also make `Config` generic.

```ts
type Config<TClusterUrl extends ClusterUrl> = Readonly<{
    headers?: AllowedHttpRequestHeaders;
    url: TClusterUrl;
}>;

export function createHttpTransport<TClusterUrl extends ClusterUrl>({
    headers, 
    url 
}: Config<TClusterUrl>): IRpcTransportFromClusterUrl<TClusterUrl> {...}
```

packages/rpc-types/src/rpc-api.ts
lorisleiva: How bout `IRpcTransportWithCluster`?

buffalojoec: But it's a union that includes one transport without a cluster, so that could be a bit confusing. Unless I use `IRpcTransport | IRpcTransportWithCluster` everywhere?

lorisleiva: `IRpcTransport | IRpcTransportWithCluster` Yes, where applicable.

Personally though, I'd remove `IIRpcTransport` altogether. I can't see a valid reason to keep it by looking at this PR only (but maybe I'm missing some stack context).

packages/rpc-transport/src/transports/http/http-transport.ts
buffalojoec: Hah, this was a great call. Slick!

lorisleiva (approved):
Love it!

buffalojoec:
### Merge activity

* **Feb 2, 11:35 AM EST**: @buffalojoec started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2053).

PR #2065, opened 2024-01-24
SyntaxError due to unsupported ||= operator in @solana/web3.js/lib/index.cjs.js

NikhilWaghmode09:
I discovered a bug in the @solana/web3.js library related to the use of the logical OR assignment (||=) operator in the @solana/web3.js/lib/index.cjs.js file. 

The logical OR assignment (||=) operator is not supported in certain environments, leading to a SyntaxError. The suggested fix is to replace the ||= operator with the equivalent logical OR (||) and assignment (=) syntax, as shown below: 
Before: 
operator1 ||= operator2; 
After:
operator1 = operator1 || operator2;

### DIFF
--- a/packages/addresses/src/program-derived-address.ts
+++ b/packages/addresses/src/program-derived-address.ts
@@ -88,7 +88,7 @@ async function createProgramDerivedAddress({ programAddress, seeds }: ProgramDer
     }
     let textEncoder: TextEncoder;
     const seedBytes = seeds.reduce((acc, seed, ii) => {
-        const bytes = typeof seed === 'string' ? (textEncoder ||= new TextEncoder()).encode(seed) : seed;
+        const bytes = typeof seed === 'string' ? (textEncoder = textEncoder || new TextEncoder()).encode(seed) : seed;
         if (bytes.byteLength > MAX_SEED_LENGTH) {
             // TODO: Coded error.
             throw new Error(`The seed at index ${ii} exceeds the maximum length of 32 bytes`);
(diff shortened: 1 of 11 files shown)


NikhilWaghmode09:
**For more context this was the error I faced after running my typescript file:**

SyntaxError: Unexpected token '||='
    at wrapSafe (internal/modules/cjs/loader.js:1001:16)
    at Module._compile (internal/modules/cjs/loader.js:1049:27)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
    at Module.load (internal/modules/cjs/loader.js:950:32)
    at Function.Module._load (internal/modules/cjs/loader.js:790:12)
    at ModuleWrap.<anonymous> (internal/modules/esm/translators.js:199:29)
    at ModuleJob.run (internal/modules/esm/module_job.js:183:25)
    at async Loader.eval (internal/modules/esm/loader.js:170:24)
    at async internal/process/execution.js:49:24
    at async loadESM (internal/process/esm_loader.js:68:5)

steveluscher:
Thanks for the note! You can read more about the change that caused you to encounter this error, here: https://github.com/solana-labs/solana-web3.js/pull/1877

In short, we're now building web3.js to the _highest_ ECMAScript target possible, which is certainly makes it incompatible with lower targets like the one you're trying to run it on.

The idea here is that everyone should build their app for their target JS engine. We could replace the invocation with different syntax, as you've suggested, but a) there are a _ton_ of other language features that we use that would still fail in some environments, and b) the lower a target we compile down to the greater that penalizes apps that want to run at higher targets, cutting them off from performance and bundle size wins.

NikhilWaghmode09:
Great thanks for clarifying! Also my problem still continues when I import the "@solana/web3.js" library in my local directory using the command "npm install @solana/web3.js". Can you suggest a solution for that...

steveluscher:
This is a good question for https://stackoverflow.com/questions/ask. Tell folks what JavaScript engine you want to run your code on, what language features are not supported (eg. `||=`), and what build system you're currently using to take your source code and turn it into the thing that you actually execute.

PR #2068, opened 2024-01-27
refactor(experimental): graphql: revise accounts schema for data inputs

### LINKED ISSUE #1831: Expand on Accounts schema in GraphQL

buffalojoec:
This commit re-works the schema to deal with `encoding` and `dataSlice` directly
on the `data` field, rather than an `account` query at the top-level.

With this particular commit, the number of RPC calls to resolve a wide range of
queries is likely increased - such as multiple `data` fields on `programAccounts`.
However, the subsequent commits in this stack will seek to remedy this.

Specifically, the next commit introduces better account resolvers, which may
lessen the RPC calls for `account` queries by way of cache hits, but the commit
after that will introduce the revamped batch loader, which provides significant
optimizations to the number and timing of RPC calls.

Further up the stack are more changes to ensure the other queries, like
`programAccounts`, get to share in the benefits of similar optimizations.

Closes #1831


buffalojoec:
* `master`
  * **PR #2068** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2068?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #2069** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2069?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2068?utm_source=stack-comment).

packages/rpc-graphql/src/resolvers/account.ts
```diff
@@ -1,141 +1,196 @@
...
 }
 
-export function transformLoadedAccount({
-    account,
-    address,
-    encoding = 'jsonParsed',
-}: {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    account: any;
+export type AccountResult = {
```
buffalojoec, line 35: The diff here is terrible, but basically what I've done is laid out a structured object for an "account result", which contains a little dictionary of encoded-sliced data so that the `resolveAccountData` can just grab the proper slice from the result's dictionary.

packages/rpc-graphql/src/resolvers/resolve-info/account.ts
```diff
@@ -0,0 +1,95 @@
...
+    args: {
+        address: Address;
+        commitment?: Commitment;
+        minContextSlot?: Slot;
+    },
+    info?: GraphQLResolveInfo,
+): AccountLoaderArgs[] {
+    const argSet: AccountLoaderArgs[] = [args];
+
+    if (info?.fieldNodes[0]?.selectionSet) {
```
steveluscher, line 77: Starting with the selection set, you do a bunch of spelunking into nodes and child nodes. This seems like the time to write this as an AST visitor, right?

You wrote a method called `getAllFieldNodesByName()`. The visitor version of that:

```ts
const argSet = [args];
visit(fieldNodes[0], {
  Field(node) {
    if (node.parentNode !== fieldNodes[0]) {
      
    }
    if (node.name.value !== 'data') {
      return;
    }
    const { arguments } = node;
    const encodingArg = arguments.find(({name}) => name.value === 'encoding');
    // TODO: Assert `encodingArg` exists.
    const encoding = encodingArg.value.value;
    const dataSliceArg = arguments.find(({name}) => name.value === 'dataSlice');
    // etc...
  }
  InlineFragment(node) {
    // Determine whether it should cause a `jsonParsed` fetch or not
  }
  FragmentSpread(node) {
    // Lookup the definition and continue visiting the nodes of the spread fragment
  }
});
```

Very important: here's all the stuff your visitor can do by returning different values: https://github.com/DxCx/graphql-js/blob/async/src/language/visitor.js#L77-L81

packages/rpc-graphql/src/resolvers/resolve-info/account.ts
```diff
@@ -0,0 +1,95 @@
...
+        const { variableValues } = info;
+        const selections = info.fieldNodes[0].selectionSet.selections;
+
+        if (isInlineFragmentPresent(selections)) {
```
steveluscher, line 81: Isn't this _technically_ insufficient to determine if the account should be fetched with `jsonParsed` or not? Couldn't there be some kind of type in this union that doesn't actually require that we fetch parsed data? It would be more airtight to actually check the fragment's `typeCondition` against a known list of types.

¯\\_(°ペ)\_/¯

packages/rpc-graphql/src/resolvers/resolve-info/account.ts
```diff
@@ -0,0 +1,95 @@
...
+    } else {
+        return undefined;
+    }
+}
+
+/**
+ * Build a set of account loader args by inspecting which fields have
+ * been requested in the query (ie. `data` or inline fragments).
+ */
+export function buildAccountLoaderArgSetFromResolveInfo(
```
steveluscher, line 67: Right now, this misses all of the fields inside spread fragments, both ones that necessitate `jsonParsed` fetching _and_ ones that fetch data. You have to write your node visitor and have it _recurse_ into fragment spreads (non-inline fragments).

https://astexplorer.net/#/gist/a0ce58c7813b11e20472fc6d6bf5a6f2/ca707e66ebe2eec7fa4a5cf176d30f8881b0ab5d

packages/rpc-graphql/src/resolvers/account.ts
steveluscher: This will break down when you do the ‘fetch the aggregate range and I'll dole out the right slices to those who asked for them’ logic right? You'll have to replace this with a system that actually decodes the data, stores it keyed on the account, then slices and re-encodes it for each consumer.

That actually kind of sucks and I'm starting to regret it. Base58/64 is not _that_ bad, but if the user requests zstd compressed data and you fetch an aggregate of two slices, now you have to include a zstd compressor/decompressor _in the bundle_. And that thing is wasm/async too.

Oof. Regrets.

packages/rpc-graphql/src/resolvers/account.ts
```diff
@@ -1,141 +1,196 @@
...
+    insert(data: string, dataSlice: DataSlice | null, encoding: Encoding) {
+        this[encoding].push({ data, dataSlice });
+    }
+
+    get(dataSlice: DataSlice | null, encoding: Encoding): string | undefined {
+        return this[encoding].find(entry => {
+            if (dataSlice === null) {
+                return entry.dataSlice === null;
+            }
+            return entry.dataSlice?.length === dataSlice.length && entry.dataSlice?.offset === dataSlice.offset;
```
steveluscher, line 30: I feel like you could just do this with an object whose keys are `base58:0:10` (encoding/offset/length).

steveluscher (changes requested):
I think this is the time to write visitors. It'll become much more tractable to handle all of the many many edge cases:

```graphql
{
  coolAccount: account {
    data(encoding: BASE58)
    ... on TokenAccount {
      mint
    }
    owner {
      dataYouDoNotWantYourVisitorToProcessYet: data(encoding: BASE64)
    }
    ...SpecialAccountsFragment
    ...CompressedDataFragment
  }
}

fragment SpecialAccountsFragment on Account {
  ... on VoteAccount {
    authority
  }
}

fragment CompressedDataFragment on Account {
  compressedData: data(encoding: BASE64ZSTD) 
}
```

Oooh. I think I found that ‘merge the type info into the visitor’ thing I was looking for before.

https://github.com/DxCx/graphql-js/blob/async/src/language/visitor.js#L321-L351

packages/rpc-graphql/src/resolvers/account.ts
```diff
@@ -1,141 +1,196 @@
...
-                    return 'NonceAccount';
-                }
-                if (account.accountType === 'mint' && account.programName === 'spl-token') {
-                    return 'MintAccount';
-                }
-                if (account.accountType === 'account' && account.programName === 'spl-token') {
-                    return 'TokenAccount';
+
+            const argsSet = buildAccountLoaderArgSetFromResolveInfo({ ...args, address }, info);
+            const loadedAccounts = await Promise.all(await context.loaders.account.loadMany(argsSet));
```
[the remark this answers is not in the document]
buffalojoec: Yep, resolved!

packages/rpc-graphql/src/resolvers/account.ts
buffalojoec: Nice call, I implemented this!

packages/rpc-graphql/src/resolvers/account.ts
buffalojoec: Actually the batch loader handles this properly by slicing the data before resolving the promise. So, by the time data comes back to the resolver from the loader, it's got the right indices and the requested data slice arguments.

With that being said, I'll need to handle `base64+zstd` separately on the batch loader itself, so that it's always treated with an isolated request (yuck).

Regardless, this shouldn't break down since the loader does all of that slicing & dicing.

packages/rpc-graphql/src/resolvers/resolve-info/account.ts
buffalojoec: Done.

packages/rpc-graphql/src/resolvers/resolve-info/account.ts
buffalojoec: Done.

packages/rpc-graphql/src/resolvers/resolve-info/account.ts
buffalojoec: aaaand done!

packages/rpc-graphql/src/resolvers/resolve-info/visitor.ts
```diff
@@ -0,0 +1,95 @@
...
+    visit(root, {
+        Field(node, key, parent, path, ancestors) {
+            if (!parentIsRoot(ancestors)) return;
+            return operations.fieldNodeOperation(info, node, key, parent, path, ancestors);
+        },
+        FragmentSpread(node, key, parent, path, ancestors) {
+            const fragmentDefinition = info.fragments[node.name.value];
+            const fragmentType = schema.getType(fragmentDefinition.typeCondition.name.value)?.name;
+            let rootType;
+            if (root.kind === 'Field') {
```
steveluscher, line 53: God I wish this thing let you write AST _selectors_. https://medium.com/dailyjs/eslint-selectors-just-got-super-powers-402ab942def5

Then you could do things like:

```ts
visit(node, {
    'FragmentSpread > Field': (node, key, parent, path, ancestors) => {
        // Only finds `Field` nodes who are the children of `FragmentSpread` nodes.
    },
});
```

If only there were like a TSQuery for GraphQL? https://medium.com/@phenomnominal/easier-typescript-tooling-with-tsquery-d74f04f2b29d

steveluscher (changes requested):
K, I'm suspicious that there might be a bug here somewhere because I would not expect the test with no fragment specific fields to fetch with `jsonParsed` encoding. I think we need a few kinds of tests, multiplied by the three ways of making subselections (fields, inline fragments, and fragment spreads).

* Subselections that don't require a fetch at all (ie. just fetching `address`)
* Subselections that require `jsonParsed` (ie. literally requesting `__typename`, subtype-specific fields like `MintAccount#mint`)
* Subselections that require baseX `data`

…plus a combination of each.

packages/rpc-graphql/src/resolvers/__tests__/account-resolver-test.ts
```diff
@@ -104,4 +104,198 @@ describe('account resolver', () => {
...
+            expect(rpc.getAccountInfo).toHaveBeenCalledWith(
+                'AyGCwnwxQMCqaU4ixReHt8h5W4dwmxU7eM3BEQBdWVca',
+                {
+                    encoding: 'jsonParsed',
+                }
+            );
+            expect(rpc.getAccountInfo).toHaveBeenCalledWith(
+                'AyGCwnwxQMCqaU4ixReHt8h5W4dwmxU7eM3BEQBdWVca',
+                {
+                    encoding: 'base58',
```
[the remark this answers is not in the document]
buffalojoec: Renamed accordingly!

packages/rpc-graphql/src/resolvers/__tests__/account-resolver-test.ts
```diff
@@ -104,4 +104,198 @@ describe('account resolver', () => {
...
+                }
+            );
+        });
+        it('can properly resolve inline fragments with data when _no_ `jsonParsed` fields are requested', async () => {
```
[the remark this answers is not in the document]
buffalojoec: Added some tests for this.

packages/rpc-graphql/src/resolvers/__tests__/account-resolver-test.ts
```diff
@@ -102,6 +178,247 @@ describe('account resolver', () => {
...
+            // FIXME: Prefer async version of this timer runner. See https://github.com/jestjs/jest/issues/14549
+            await Promise.resolve();
+            jest.runAllTimers();
+            expect(rpc.getAccountInfo).toHaveBeenCalledTimes(2); // Fixed with batch loader
+            expect(rpc.getAccountInfo).toHaveBeenCalledWith('AyGCwnwxQMCqaU4ixReHt8h5W4dwmxU7eM3BEQBdWVca', {
+                encoding: 'base58',
+            });
+            expect(rpc.getAccountInfo).not.toHaveBeenCalledWith('AyGCwnwxQMCqaU4ixReHt8h5W4dwmxU7eM3BEQBdWVca', {
+                encoding: 'jsonParsed',
+            });
```
buffalojoec, line 298: @steveluscher It seems to me that the resolver is working correctly. I'm wondering if, in regards to your comments, you may have overlooked the `.not` here?

steveluscher: 🪦 OMG I missed that. Nice work!

buffalojoec:
### Merge activity

* **Feb 22, 3:57 PM EST**: @buffalojoec started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2068).

PR #2069, opened 2024-01-27
refactor(experimental): graphql: expand on batch loader for accounts

buffalojoec:
So.... this batch loader works great!

However, my tests are super flaky with the timer manipulation.
Every 5-6 attempts, they all pass, but the other times I get a few failures.
It's always the same tests that fail, and they seem to be 1 less `fetchMock`
invocation than expected.
I think I'm missing something.

### DIFF
--- a/packages/rpc-graphql/src/index.ts
+++ b/packages/rpc-graphql/src/index.ts
@@ -1,12 +1,9 @@
 import { makeExecutableSchema } from '@graphql-tools/schema';
-import type { GetAccountInfoApi, GetBlockApi, GetProgramAccountsApi, GetTransactionApi, Rpc } from '@solana/rpc';
 import { graphql } from 'graphql';
 
 import { createSolanaGraphQLContext } from './context';
 import { createSolanaGraphQLResolvers } from './resolvers';
 import { createSolanaGraphQLTypeDefs } from './schema';
-
-type RpcMethods = GetAccountInfoApi & GetBlockApi & GetProgramAccountsApi & GetTransactionApi;
 
 export interface RpcGraphQL {
     query(
@@ -15,14 +12,21 @@ export interface RpcGraphQL {
     ): ReturnType<typeof graphql>;
 }
 
-export function createRpcGraphQL(rpc: Rpc<RpcMethods>): RpcGraphQL {
+export function createRpcGraphQL(
+    rpc: Parameters<typeof createSolanaGraphQLContext>[0],
+    config?: Partial<Parameters<typeof createSolanaGraphQLContext>[1]>,
+): RpcGraphQL {
+    const rpcGraphQLConfig = {
+        maxDataSliceByteRange: config?.maxDataSliceByteRange ?? 200,
+        maxMultipleAccountsBatchSize: config?.maxMultipleAccountsBatchSize ?? 100,
+    };
     const schema = makeExecutableSchema({
         resolvers: createSolanaGraphQLResolvers(),
         typeDefs: createSolanaGraphQLTypeDefs(),
     });
     return {
         async query(source, variableValues?) {
-            const contextValue = createSolanaGraphQLContext(rpc);
+            const contextValue = createSolanaGraphQLContext(rpc, rpcGraphQLConfig);
             return graphql({
                 contextValue,
                 schema,
--- a/packages/rpc-graphql/src/loaders/__tests__/account-loader-test.ts
+++ b/packages/rpc-graphql/src/loaders/__tests__/account-loader-test.ts
@@ -10,7 +10,13 @@ import type {
 import { createRpcGraphQL, RpcGraphQL } from '../../index';
 
 describe('account loader', () => {
-    let rpc: Rpc<GetAccountInfoApi & GetBlockApi & GetMultipleAccountsApi & GetProgramAccountsApi & GetTransactionApi>;
+    let rpc: {
+        getAccountInfo: jest.MockedFunction<Rpc<GetAccountInfoApi>['getAccountInfo']>;
+        getBlock: jest.MockedFunction<Rpc<GetBlockApi>['getBlock']>;
+        getMultipleAccounts: jest.MockedFunction<Rpc<GetMultipleAccountsApi>['getMultipleAccounts']>;
+        getProgramAccounts: jest.MockedFunction<Rpc<GetProgramAccountsApi>['getProgramAccounts']>;
+        getTransaction: jest.MockedFunction<Rpc<GetTransactionApi>['getTransaction']>;
+    };
     let rpcGraphQL: RpcGraphQL;
     // Random address for testing.
     // Not actually used. Just needed for proper query parsing.
@@ -24,7 +30,11 @@ describe('account loader', () => {
             getProgramAccounts: jest.fn(),
             getTransaction: jest.fn(),
         };
-        rpcGraphQL = createRpcGraphQL(rpc);
+        rpcGraphQL = createRpcGraphQL(
+            rpc as unknown as Rpc<
+                GetAccountInfoApi & GetBlockApi & GetMultipleAccountsApi & GetProgramAccountsApi & GetTransactionApi
+            >,
+        );
     });
     describe('cached responses', () => {
         it('coalesces multiple requests for the same account into one', async () => {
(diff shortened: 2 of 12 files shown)


buffalojoec:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2069?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `master`
  * **PR #2068** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2068?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #2069** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2069?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2069?utm_source=stack-comment).

packages/rpc-graphql/src/loaders/account.ts
```diff
@@ -1,47 +1,319 @@
...
+                                });
+                            });
+                        } catch (e) {
+                            chunk.forEach(address => {
+                                addressCallbacks[address].callbacks.forEach(({ callback }) => {
+                                    callback.reject(e);
+                                });
+                            });
+                        }
+                    }
```
steveluscher, line 307: So, one of the problems that you're having so much trouble with your tests and the timers therein is that you're resolving these calls _in series_ in different runloops. You want to split the requests up into batches and fire _all_ the requests off in the same tick.

This is the classic ‘`await` in a loop’ problem. Wherever you notice an `await` in a loop, be suspicious. There's even [a linter](https://eslint.org/docs/latest/rules/no-await-in-loop) for this.

```suggestion
                    await Promise.all(
                        Array.from({ length: Math.ceil(addresses.length / chunkSize) }, async (_, ii) => {
                            const startIndex = ii * chunkSize;
                            const endIndex = startIndex + chunkSize;
                            const chunk = addresses.slice(startIndex, endIndex);
                            try {
                                const results = await resolveMultipleAccountsUsingRpc({
                                    addresses: chunk,
                                    ...args,
                                });
                                chunk.forEach((address, ii) => {
                                    const result = results[ii];
                                    addressCallbacks[address].callbacks.forEach(({ callback, dataSlice }) => {
                                        callback.resolve(sliceData(result, dataSlice, args.dataSlice));
                                    });
                                });
                            } catch (e) {
                                chunk.forEach(address => {
                                    addressCallbacks[address].callbacks.forEach(({ callback }) => {
                                        callback.reject(e);
                                    });
                                });
                            }
                        }),
                    );
```

packages/rpc-graphql/src/loaders/account.ts
```diff
@@ -1,47 +1,319 @@
...
+            /**
+             * For each set of accounts related to some common args, fetch them in the fewest number
+             * of network requests.
+             */
+            Object.values(accountFetchesByArgsHash).forEach(async ({ args, addresses: addressCallbacks }) => {
+                const addresses = Object.keys(addressCallbacks) as Address[];
+                if (addresses.length === 1) {
+                    const address = addresses[0];
+                    try {
+                        const result = await resolveAccountUsingRpc({ address, ...args });
```
steveluscher, line 277: Same problem as below. This is an `await` in a loop – the loop started here:

```ts
Object.values(accountFetchesByArgsHash).forEach(...);
```

You want to set up all the promises and launch them at once.

```ts
await Promise.all(
    Object.values(accountFetchesByArgsHash).map(...),
);
```

steveluscher (changes requested):
You've made it _really_ hard on yourself with these tests. They engage the entire pipeline _all_ the way down to `fetch()`. This is too far.

What you really want to test is that the account loader drives the _RPC_ in the way that you expect. You absolutely don't care what happens below that.

Instead of _creating_ an RPC.

```ts
rpc = createJsonRpc<
    GetAccountInfoApi & GetBlockApi & GetMultipleAccountsApi & GetProgramAccountsApi & GetTransactionApi
>({
    api: createSolanaRpcApi(),
    transport: createHttpTransport({ url: 'http://127.0.0.1:8899' }),
});
```

…create a mock one.

```ts
rpc = {
    getAccountInfo: jest.fn(),
    getBlock: jest.fn(),
    getMultipleAccounts: jest.fn(),
    getProgramAccounts: jest.fn(),
    getTransaction: jest.fn(),
};
```

Then use your normal assertions:

```ts
expect(rpc.getMultipleAccounts).toHaveBeenCalledTimes(2);
```

This is _much_ closer to what this code actually aims to do. After all, the RPC transport could do _further_ batching, which is _definitely_ not what you want to test here.

packages/rpc-graphql/src/loaders/__tests__/account-loader-test.ts
```diff
@@ -27,6 +27,10 @@ describe('account loader', () => {
         });
         rpcGraphQL = createRpcGraphQL(rpc);
     });
+    afterEach(async () => {
+        // await jest.runAllTimersAsync();
+        jest.runAllTicks();
+    });
```
steveluscher, line 33: Added via Giphy

packages/rpc-graphql/src/loaders/__tests__/account-loader-test.ts
```diff
@@ -13,10 +13,10 @@ import fetchMock from 'jest-fetch-mock-fork';
 import { createRpcGraphQL, RpcGraphQL } from '../../index';
 
 describe('account loader', () => {
+    jest.useFakeTimers();
```
steveluscher, line 16: Move this into the `beforeEach()`. I don't know why, just feels safer.

packages/rpc-graphql/src/loaders/account.ts
```diff
@@ -1,45 +1,334 @@
...
+                    });
+                    promiseCallbacksForAddress.push({ callback: orphanPromiseCallback });
+                });
+            });
+
+            /**
+             * For each set of accounts related to some common args, fetch them in the fewest number
+             * of network requests.
+             */
+            await Promise.all(
```
steveluscher, line 277: FWIW I don't think my [original take](https://github.com/solana-labs/solana-web3.js/pull/2063/files) had `await` in the `finally` block and yours doesn't have to either. All you want to do is make sure all the requests are fired off. You don't actually need to `await Promise.all()` from the perspective of `createAccountBatchLoadFn()()`, which has already returned.

packages/rpc-graphql/src/loaders/__tests__/account-loader-test.ts
```diff
@@ -66,4 +66,1085 @@ describe('account loader', () => {
...
+                    }),
+                }));
+
+                rpc = {
+                    getAccountInfo,
+                    getBlock: jest.fn(),
+                    getMultipleAccounts,
+                    getProgramAccounts: jest.fn(),
+                    getTransaction: jest.fn(),
+                };
```
steveluscher, line 218: You don't have to recreate and regenerate the `rpcGraphQL` inside the test. You can ‘reach back’ into the existing resolvers and make them do new stuff.

```ts
it('coalesces multi-layered multiple requests into `getMultipleAccounts` requests', () => {
  rpc.getMultipleAccounts // <- already a mock function
    .mockReturnValue({
      send: jest.fn()
        .mockReturnValueOnce(/* first return value */)
        .mockReturnValueOnce(/* second return value */),
    });
});
```

Or you could do this work ahead of time.

```ts
let getMultipleAccountsSender;
beforeEach(() => {
  getMultipleAccountsSender = jest.fn();
  rpc = {
    getMultipleAccounts: jest.fn()
      .mockReturnValue({send: getMultipleAccountsSender}),
  };
});
it('coalesces multi-layered multiple requests into `getMultipleAccounts` requests', () => {
  rpc.getMultipleAccountsSender
    .mockReturnValueOnce(/* first return value */)
    .mockReturnValueOnce(/* second return value */);
});
```

packages/rpc-graphql/src/loaders/__tests__/account-loader-test.ts
```diff
@@ -66,4 +66,1085 @@ describe('account loader', () => {
...
+                        'TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA',
+                        'Vote111111111111111111111111111111111111111',
+                    ],
+                    { encoding: 'base64' }, // GraphQL client prefers `base64`
+                );
+                // Evaluate layer 3
+                await jest.advanceTimersToNextTimerAsync();
+                jest.runAllTimers();
+                expect(rpc.getAccountInfo).toHaveBeenCalledTimes(1); // Single account remaining
+                expect(rpc.getMultipleAccounts).toHaveBeenCalledTimes(2); // Hasn't been called again
```
steveluscher, line 284: You can do this, but you can also ‘clear’ the mocks at each layer so that you can _actually_ write `.not.toHaveBeenCalled()` as you mean here, so that you don't have to comment the expectation to explain what's going on.

```ts
// Evaluate layer 2
/* ... */
jest.clearAllMocks();
// Evaluate layer 3
/* ... */
expect(rpc.getMultipleAccounts).not.toHaveBeenCalled();
```

This keeps the implementation, but [resets the call counts and stuff](https://jestjs.io/docs/mock-function-api#mockfnmockclear).

You can also call `rpc.getMultipleAccounts.mockClear()`.

You can also straight up reset all the mocks and prepare the `mockResolvedValue()` at every step.

Too many ways to skin this cat.

packages/rpc-graphql/src/loaders/__tests__/account-loader-test.ts
```diff
@@ -66,4 +66,1085 @@ describe('account loader', () => {
...
+                            }
+                        }
+                    `;
+                    rpcGraphQL.query(source);
+                    // FIXME: Prefer async version of this timer runner. See https://github.com/jestjs/jest/issues/14549
+                    await Promise.resolve();
+                    jest.runAllTimers();
+                    expect(rpc.getAccountInfo).toHaveBeenCalledTimes(3);
+                    expect(rpc.getAccountInfo).toHaveBeenCalledWith('Gh9ZwEmdLJ8DscKNTkTqPbNwLNNBjuSzaG9Vp2KGtKJr', {
+                        encoding: 'base58', // No `dataSlice` arg since one field asked for the whole data
```
steveluscher, line 658: 👍🏻

packages/rpc-graphql/src/loaders/__tests__/account-loader-test.ts
```diff
@@ -66,4 +66,1085 @@ describe('account loader', () => {
...
+                    jest.runAllTimers();
+                    expect(rpc.getMultipleAccounts).toHaveBeenCalledTimes(1);
+                    expect(rpc.getMultipleAccounts).toHaveBeenCalledWith(
+                        [
+                            'Gh9ZwEmdLJ8DscKNTkTqPbNwLNNBjuSzaG9Vp2KGtKJr',
+                            '2KAARoNUYTddAChEdWb21bdKH6dWu51AAPFjSjRmzsbb',
+                            '4rFV8bvFpacLkvxTFuVN4pqe5s7CTyEkmYvPpu45779u',
+                        ],
+                        {
+                            dataSlice: { length: 40, offset: 0 }, // Coalesced into one slice
```
steveluscher, line 958: Nice.

packages/rpc-graphql/src/loaders/account.ts
```diff
@@ -1,45 +1,334 @@
...
+             * Maximum number of acceptable bytes to waste before splitting two
+             * `dataSlice` requests into two requests.
+             */
+            const maxByteRange = 200;
+
+            /**
+             * Maximum number of accounts to fetch in a single batch.
+             * See https://docs.solana.com/api/http#getmultipleaccounts.
+             */
+            const maxMultipleAccountsBatchSize = 100;
```
steveluscher, line 136: Should probably make these into arguments that bubble up all the way up to `createGraphQLRpc()`, so that if you have an RPC URL that imposes a _lower_ `gMA` limit, or one with tolerance for a higher byte window, the RPC-bringer can configure that.

packages/rpc-graphql/src/loaders/account.ts
```diff
@@ -1,45 +1,334 @@
...
+                        args,
+                    });
+                    const { callbacks: promiseCallbacksForAddress } = (accountFetches.addresses[address] ||= {
+                        callbacks: [],
+                    });
+                    promiseCallbacksForAddress.push({ callback: promiseCallback, dataSlice: args.dataSlice ?? null });
+                });
+            });
+
+            // Place the orphans
```
steveluscher, line 240: Do what now? :)

packages/rpc-graphql/src/loaders/__tests__/account-loader-test.ts
```diff
@@ -66,4 +66,1085 @@ describe('account loader', () => {
             expect(rpc.getAccountInfo).toHaveBeenCalledTimes(2);
         });
     });
+    describe('batch loading', () => {
```
steveluscher, line 69: I didn't super double check, but do these (awesome) tests include the case where

* one thing got the default commitment of `confirmed` because it didn't specify a commitment, and
* another otherwise identical fetch explicitly stated `commitment: 'confirmed'`

We should have a test that asserts that two such requests merge into one.

steveluscher (changes requested):
Added via Giphy

packages/rpc-graphql/src/loaders/__tests__/account-loader-test.ts
buffalojoec: Done!

packages/rpc-graphql/src/loaders/__tests__/account-loader-test.ts
buffalojoec: I've added this test, and unfortunately it wasn't handled! I've updated the batch loader to coalesce this.

packages/rpc-graphql/src/loaders/account.ts
buffalojoec: Done!

packages/rpc-graphql/src/index.ts
```diff
@@ -1,29 +1,32 @@
...
 }
 
-export function createRpcGraphQL(rpc: Rpc<RpcMethods>): RpcGraphQL {
+export function createRpcGraphQL(
+    rpc: Parameters<typeof createSolanaGraphQLContext>[0],
+    config?: Parameters<typeof createSolanaGraphQLContext>[1]
+): RpcGraphQL {
+    const rpcGraphQLConfig = {
+        maxDataSliceByteRange: config?.maxDataSliceByteRange ?? 200,
+        maxMultipleAccountsBatchSize: config?.maxMultipleAccountsBatchSize ?? 100,
```
steveluscher, line 21: 👍🏻

packages/rpc-graphql/src/loaders/__tests__/account-loader-test.ts
```diff
@@ -66,4 +74,1264 @@ describe('account loader', () => {
...
+                    expect(rpc.getAccountInfo).toHaveBeenCalledWith('Gh9ZwEmdLJ8DscKNTkTqPbNwLNNBjuSzaG9Vp2KGtKJr', {
+                        commitment: 'confirmed',
+                        encoding: 'base58', // No `dataSlice` arg since one field asked for the whole data
+                    });
+                    expect(rpc.getAccountInfo).toHaveBeenCalledWith('Gh9ZwEmdLJ8DscKNTkTqPbNwLNNBjuSzaG9Vp2KGtKJr', {
+                        commitment: 'confirmed',
+                        encoding: 'base64', // No `dataSlice` arg since one field asked for the whole data
+                    });
+                });
+                it('always uses separate requests for `base64+zstd` no matter the data slice', async () => {
```
steveluscher, line 886: 👍🏻

packages/rpc-graphql/src/loaders/__tests__/account-loader-test.ts
```diff
@@ -66,4 +74,1264 @@ describe('account loader', () => {
             expect(rpc.getAccountInfo).toHaveBeenCalledTimes(2);
         });
     });
+    describe('batch loading', () => {
```
steveluscher, line 77: I think you still need a test to exercise multiple batches being split off when the number of fetches exceeds the max batch size. 

Maybe create an RPC with the `maxMultipleAccountsBatchSize` hardcoded to 3 and then try to make 4 requests.

steveluscher (approved):
Added via Giphy

packages/rpc-graphql/src/loaders/__tests__/account-loader-test.ts
buffalojoec: I have one! It's generating a huge query with 150 keys. Now that I have the input argument at the `createRpcGraphQL` level, I'm going to rework this to use the input arg as you said.

https://github.com/solana-labs/solana-web3.js/blob/024daa22af032636dd7975c535cdaf36f1b9580b/packages/rpc-graphql/src/loaders/__tests__/account-loader-test.ts#L500

buffalojoec: Done!

buffalojoec:
### Merge activity

* **Feb 22, 3:57 PM EST**: @buffalojoec started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2069).

PR #2072, opened 2024-01-29
refactor(experimental): add `info` field to JSON parsed types

mcintyre94:
Some of the programs that own JSON parseable accounts, such as token, have accounts with multiple types. They have both `info` and `type` fields, and we have these typed so that you can discriminate using the `type` field.

But some other programs, such as vote and address lookup tables, have only one type of account. They don't have a `type` field but do have the `info` field, with the same shape for all their accounts.

Since I didn't need to discriminate them, I exported these types without the`info` field too. This PR changes this so that their type includes the `info` field, with all their properties nested inside it.

I found that it's annoying to use these as-is because you end up with some types having an `info` field containing their properties, and others not. But if the data comes from `getAccountInfo` etc, that `info` field is always there, so it makes sense for the types to always include it too. This reduces special cases in code using these types.


steveluscher (approved):
Added via Giphy

Issue #1133, opened 2020-12-07
[web3.js][request] Throw custom Solana Errors instead of generic Errors


paul-schaaf:
#### Problem
In some places in the web3 js library, errors are thrown that could be extended with more info for downstream users of the library that want to handle and parse the error, instead of having to write code to parse the error msg for required info.

e.g. here

https://github.com/solana-labs/solana/blob/2f374df494f6a5ca8e5cbe8801b61e07a0540247/web3.js/src/connection.js#L2227

The tx signature is in the string but it could also be added to the error itself as a property.

#### Proposed Solution

Different exceptions could result in different custom errors
something like this could work for a transaction exception.

```javascript
class TransactionError extends Error {
   signature: TransactionSignature;
   commitment: Commitment;

   constructor(msg: string, signature: TransactionSignature, commitmentUsed: Commitment) {
       super(msg);
       this.signature = signature;
       this.commitment = commitmentUsed;
   }
}
```

titouanfreville:
Hello, any news/info arround this issue ? Current errors does not seems designed to be handled as we need to parse the message to find what happen and that's a bit frustrating :) 

I am not used to JS way to handle errors so both throw custom errors or respond with dedicated messages will do, but in both case it would be nice if you provided a way to identify the errors easily. 

Thanks a lot for your work on the lib. :)

steveluscher:
`SolanaError` lays this foundation. Follow along in #2118.

PR #2074, opened 2024-01-31
A coded error class for all Solana modules

steveluscher:
# Summary

These two things are bad:

- Adding kilobytes of error message strings to a JavaScript bundle
- Taking kilobytes of hard dependencies on `Error` subclasses just to be able to use them to do `instanceof` checks on thrown errors

There's another way.

In this PR we create Solana's ‘coded error’ class. This is a _single_ class that represents _all_ errors across _every_ module, using:

- Integer error codes
- TypeScript types

Any time you need to throw an error, throw an instance of this very small `SolanaError` class, constructing it using the right `SolanaErrorCode` and some optional context as provided by the TypeScript type associated with that error code.

```ts
throw new SolanaError({__code: /* SOME CODE */, extra: 'context!'});
```

At the catch-site, you can assert that a thrown error is a particular type of `SolanaError` using `isSolanaError()`

```ts
try {
  /* ... */
} catch (e) {
  if (isSolanaError(e /* SOME CODE */)) {
    console.log(e.context.extra); // -> 'context!'
  }
}
```

Addresses #1133.

# Test Plan

```shell
cd packages/errors/
pnpm turbo test:unit:browser
pnpm turbo test:unit:node
```


packages/errors/src/error.ts
```diff
@@ -0,0 +1,37 @@
...
+type SolanaErrorConfig = Readonly<{
+    [P in SolanaErrorCode]: Readonly<
+        (SolanaErrorContext[P] extends undefined ? object : SolanaErrorContext[P]) & { __code: P }
+    >;
+}>;
+
+export class SolanaError<TErrorCode extends SolanaErrorCode = SolanaErrorCode> extends Error {
+    // This gives us a way for the `isSolanaError()` function to identify a `SolanaError` without
+    // having to import the class for use in an `instanceof` check.
+    readonly [SOLANA_ERROR_TAG] = 1;
```
lorisleiva, line 30: Aren't global symbols just as susceptible to dual package hazards as `instanceof` checks?

An alternative would be to use something like `readonly __solanaError = 1` and then have an opaque type with a TypeScript `unique symbol` to ensure it's coming from this specific class.

packages/errors/src/message-formatter.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export function getHumanReadableErrorMessage<TErrorCode extends SolanaErrorCode>(
+    code: TErrorCode,
+    context: object = {},
+): string {
+    const messageFormatString = SolanaErrorMessages[code];
+    const message = messageFormatString.replace(/(?<!\\)\$(\w+)/g, (substring, variableName) =>
+        variableName in context ? `${context[variableName as keyof typeof context]}` : substring,
+    );
+    return message;
```
lorisleiva, line 12: I think this is great and will be enough 95% of the time.

However, how would you feel about, additionally, **allowing `SolanaErrorMessages[code]` to be a function** of type `(context: object) => string`. That way, if `typeof SolanaErrorMessages[code] === "string"`, we keep the same logic, otherwise, we simply call the custom function to convert a context from a specific error into a message.

This feature can be added later on without breaking changes so probably best to not future-proof too early but I do think this could be useful in some situations.

lorisleiva (approved):
Love it!

Am I right to assume that every single error we throw on the other packages of this monorepo has the right to register an error code here?

I'm thinking particularly about low-level packages such as codecs. Would we have error codes like `SOLANA_ERROR__CODECS__EMPTY_BUFFER`?

packages/errors/src/codes.ts
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * One of the goals of this library is to enable people to use some or none of it without having to
+ * bundle all of it.
+ *
+ * If we made the set of error codes an enum then anyone who imported it (even if to only use a
+ * single error code) would be forced to bundle every code and its label.
+ *
+ * Const enums appear to solve this problem by letting the compiler inline only the codes that are
+ * actually used. Unfortunately exporting ambient (const) enums from a library like `@solana/errors`
+ * is not safe, for a variety of reasons covered here: https://stackoverflow.com/a/28818850
```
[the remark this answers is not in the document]
steveluscher: I'm so mad about this still.

packages/errors/src/error.ts
steveluscher: Interesting.

The motivation behind this was never to avoid dual package hazards, it was to make it so that you could import `isSolanaError()` without taking a dependency on `class SolanaError`.

Alternatives that eliminate the dual-package hazard, but open the possibility of ‘spoofing’ some other object as a `SolanaError`:

* Instead of a unique symbol as I've used here, use the global `Symbol.for('sOlAnAeRrOr')` or whatever that can be recovered across packages
* Even easier, just check `e.name === 'SolanaError'`

God I hate classes.

packages/errors/src/message-formatter.ts
steveluscher: What you've described is 1000% the way I wrote it before this revision, and it was typesafe, and it was beautiful.

Then I realized that if they're functions:

* The function implementation could, itself, fatal, depending on what inputs you send it, which is really really bad when you're trying to throw an error.
* This makes it essentially impossible to serialize the inputs for use with the error decoder two PRs later in this stack. If you have, say, a `CryptoKey` in your `context` and your error message does something useful and informative with it, then you go to production and your error gets minified, there's nothing I can do now to serialize that for decoding by `npx @solana/errors decode ...`

So I rolled that beautiful idea back and made them Just Format Strings™.

steveluscher (commented):
> Am I right to assume that every single error we throw on the other packages of this monorepo has the right to register an error code here?

Yes, absolutely.

packages/errors/src/message-formatter.ts
lorisleiva: Ah yeah that's fair enough. I'm happy to leave things as-is and if we ever encounter a scenario where templated strings is not enough, we can revisit at this point. 👌

packages/errors/src/error.ts
lorisleiva: I see you went with `e.name === 'SolanaError'` which I quite like tbh. 👌 

Is it possible to add a TypeScript opaque attribute to a class so, at least, TypeScript prevents the spoofing of the class?

steveluscher: > Is it possible to add a TypeScript opaque attribute…

It wont be useful for any purpose, because JS errors thrown [are always](https://www.typescriptlang.org/play?#code/C4JwngBA3gUBEHoBUEB06JITAvhAxgIbD4AWEAFAKYCU0c8jCCEAegPy5A) `unknown`.

```ts
try {
  /* ... */
} catch (e) {
      // ^? e: unknown
}
```

steveluscher:
### Merge activity

* **Feb 1, 11:07 AM PST**: @steveluscher started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2074).

PR #2075, opened 2024-01-31
Convert two errors in `@solana/transactions` to the new coded error system

steveluscher:
# Summary

In this PR we make first use of the new `SolanaError` class to convert two regular errors into coded errors.

# Test Plan

```shell
cd packages/transactions/
pnpm turbo test:unit:browser
pnpm turbo test:unit:node
```

### DIFF
--- a/packages/errors/src/codes.ts
+++ b/packages/errors/src/codes.ts
@@ -1,4 +1,15 @@
 /**
+ * To add a new error, follow the instructions at
+ * https://github.com/solana-labs/solana-web3.js/tree/master/packages/error#adding-a-new-error
+ *
+ * WARNING:
+ *   - Don't remove error codes
+ *   - Don't change or reorder error codes.
+ */
+export const SOLANA_ERROR__TRANSACTION_MISSING_SIGNATURES = 1 as const;
+export const SOLANA_ERROR__TRANSACTION_SIGNATURE_NOT_COMPUTABLE = 2 as const;
+
+/**
  * A union of every Solana error code
  *
  * You might be wondering why this is not a TypeScript enum or const enum.
@@ -13,4 +24,6 @@
  * actually used. Unfortunately exporting ambient (const) enums from a library like `@solana/errors`
  * is not safe, for a variety of reasons covered here: https://stackoverflow.com/a/28818850
  */
-export type SolanaErrorCode = never;
+export type SolanaErrorCode =
+    | typeof SOLANA_ERROR__TRANSACTION_MISSING_SIGNATURES
+    | typeof SOLANA_ERROR__TRANSACTION_SIGNATURE_NOT_COMPUTABLE;
--- a/packages/library/src/__tests__/transaction-confirmation-test.ts
+++ b/packages/library/src/__tests__/transaction-confirmation-test.ts
@@ -1,4 +1,5 @@
 import { Address } from '@solana/addresses';
+import { SOLANA_ERROR__TRANSACTION_SIGNATURE_NOT_COMPUTABLE, SolanaError } from '@solana/errors';
 import { AccountRole, ReadonlySignerAccount, WritableAccount } from '@solana/instructions';
 import { Signature, SignatureBytes } from '@solana/keys';
 import type { Blockhash } from '@solana/rpc-types';
@@ -143,8 +144,7 @@ describe('waitForDurableNonceTransactionConfirmation', () => {
             transaction: transactionWithoutFeePayerSignature,
         });
         await expect(commitmentPromise).rejects.toThrow(
-            "Could not determine this transaction's signature. Make sure that the transaction " +
-                'has been signed by its fee payer.',
+            new SolanaError(SOLANA_ERROR__TRANSACTION_SIGNATURE_NOT_COMPUTABLE),
         );
     });
     it('resolves when the signature confirmation promise resolves despite the block height exceedence promise having thrown', async () => {
@@ -260,8 +260,7 @@ describe('waitForRecentTransactionConfirmation', () => {
             transaction: transactionWithoutFeePayerSignature,
         });
         await expect(commitmentPromise).rejects.toThrow(
-            "Could not determine this transaction's signature. Make sure that the transaction " +
-                'has been signed by its fee payer.',
+            new SolanaError(SOLANA_ERROR__TRANSACTION_SIGNATURE_NOT_COMPUTABLE),
         );
     });
     it('resolves when the signature confirmation promise resolves despite the block height exceedence promise having thrown', async () => {
--- a/packages/errors/src/__tests__/error-test.ts
+++ b/packages/errors/src/__tests__/error-test.ts
@@ -7,8 +7,8 @@ describe('SolanaError', () => {
     let error123: SolanaError;
     beforeEach(() => {
         error123 = new SolanaError(
-            123,
             // @ts-expect-error Mock error codes don't conform to `SolanaErrorCode`
+            123,
             { foo: 'bar' },
         );
     });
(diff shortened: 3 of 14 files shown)


steveluscher:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2075?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `master`
  * **PR #2074** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2074?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #2075** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2075?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
      * **PR #2076** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2076?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #2077** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2077?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2075?utm_source=stack-comment).

packages/errors/src/codes.ts
```diff
@@ -1,3 +1,14 @@
+/**
+ * To add a new error, follow the instructions at
+ * https://github.com/solana-labs/solana-web3.js/tree/master/packages/error#adding-a-new-error
+ *
+ * WARNING:
+ *   - Don't remove error codes
+ *   - Don't change or reorder error codes.
+ */
+export const SOLANA_ERROR__TRANSACTION_MISSING_SIGNATURES = 69420 as const;
+export const SOLANA_ERROR__TRANSACTION_SIGNATURE_NOT_COMPUTABLE = 69421 as const;
```
lorisleiva, line 10: nit: I appreciate the joke and the fact that it's good not to start with 0 or 1 but I think it'd be best to start with a different "random" number because a user seeing an error like `Solana error #69420` bubbling up in production might not make the dApp look very professional.

packages/errors/package.json
```diff
@@ -64,6 +64,9 @@
     "engine": {
         "node": ">=17.4"
     },
+    "dependencies": {
+        "@solana/addresses": "workspace:*"
+    },
```
lorisleiva, line 69: Wait what? Is this allowed?

I understand this is needed to define the context of some errors but IMO, **the `@solana/errors` package should be a leaf dependency** to avoid any circular dependencies. If we start adding other packages from the monorepo as dependencies here we're shooting ourselves in the foot because just about every single package of the monorepo will end-up depending on `@solana/errors`.

I wonder if maybe there's a case for each packages to define their own context then. 🤔

packages/library/src/__tests__/transaction-confirmation-test.ts
```diff
@@ -143,8 +144,7 @@ describe('waitForDurableNonceTransactionConfirmation', () => {
             transaction: transactionWithoutFeePayerSignature,
         });
         await expect(commitmentPromise).rejects.toThrow(
-            "Could not determine this transaction's signature. Make sure that the transaction " +
-                'has been signed by its fee payer.',
+            new SolanaError({ __code: SOLANA_ERROR__TRANSACTION_SIGNATURE_NOT_COMPUTABLE }),
```
lorisleiva, line 147: nit: could be nice to offer an alternative way of constructing these if no context is required.

```ts
new SolanaError(SOLANA_ERROR__TRANSACTION_SIGNATURE_NOT_COMPUTABLE)
```

lorisleiva (changes requested):

packages/errors/package.json
steveluscher: I don't know. Does it still count as a circular dependency (in the sense that it's consequential) if the error module is only pulling in _types_ from all the other packages, and not implementations? I didn't actually try to add an error to `@solana/addresses` yet, which would answer my question.

Two options here: 

1. Make the context structurally similar (ie. `string[]` instead of `Address[]`). This would make the context less rich and less useful in a `catch` block (ie. you'd get an array of `string` for a `TRANSACTION_SIGNATURE_MISSING` instead of an array of `Address`
2. Evict the `Address` type to some other package.

Thoughts?

packages/errors/src/codes.ts
steveluscher: ~For the culture!~

Added via Giphy

packages/library/src/__tests__/transaction-confirmation-test.ts
steveluscher: Interesting. The reason everything was balled up into a single object had to do with the way that TypeScript refines generics, but maybe if you don't have any context this would be enough. I can go back into the trenches and try something like this.

steveluscher (commented):
Will update.

packages/library/src/__tests__/transaction-confirmation-test.ts
steveluscher: HA.

```ts
constructor(
    ...[code, context]: SolanaErrorContext[TErrorCode] extends undefined
        ? [code: TErrorCode]
        : [code: TErrorCode, context: SolanaErrorContext[TErrorCode]]
)
```

packages/errors/package.json
steveluscher: Yeah, you're 100% correct this is a problem.

```shell
~/src/solana-web3.js-git$ pnpm turbo compile:js compile:typedefs test:typecheck test:unit:node
 ERROR  run failed: Invalid package dependency graph: cyclic dependency detected:
        @solana/errors, @solana/addresses
```

steveluscher: OK. I tried some more stuff but there's no way out. Should we throw in the towel and make this `string[]`, move `Address` somewhere else, or a third thing that I haven't thought of yet?

lorisleiva: I'm not totally disgusted by `string[]` instead of `Address[]`. I also think that by ~~forcing~~ encouraging JS primitives to be used in the context, it makes its serialization more robust and it makes it less likely that templated strings for error message won't be sufficient.

steveluscher: Yeah, at one point I thought about somehow constraining the type of `SolanaErrorContext` such that it has to be made up of primitives or objects/arrays of primitives. This makes context _so_ much less useful though, because now you can't do this sort of thing:

```ts
try {
  /* ... */
} catch (e) {
    if (isSolanaError(e, SOLANA_ERROR__TRANSACTION_SIGNATURE_MISSING) {
        const noopSigners = e.context.addresess.map(createNoopSigner);
        // Retry…
    }
}
```

Because `e.context.addresess` is `string[]` instead of `Address[]`. You'd have to do extra work like…

```ts
if (isSolanaError(e, SOLANA_ERROR__TRANSACTION_SIGNATURE_MISSING) {
    const noopSigners = e.context.addresess.map(putativeAddress => {
        assertIsAddress(putativeAddress);
        return createNoopSigner(address);
    });
    // Retry…
}
```

Maybe that's a good thing.

steveluscher:
### Merge activity

* **Feb 1, 11:07 AM PST**: @steveluscher started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2075).

PR #2090, opened 2024-02-03
Convert `SolanaRpcIntegerOverflowError` to a coded error

steveluscher:
# Summary

A whole lot of code here, just to evict the error message from the bundle. Saved 88 whole gzipped bytes.

# Test plan

```shell
cd packages/library/
pnpm turbo test:unit:browser test:unit:node
cd ../errors/
pnpm turbo test:unit:browser test:unit:node
```


steveluscher:
* `master`
  * **PR #2090** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2090?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2090?utm_source=stack-comment).

packages/library/src/rpc-integer-overflow-error.ts
```diff
@@ -1,45 +1,45 @@
...
+    const error = new SolanaError(SOLANA_ERROR__RPC_INTEGER_OVERFLOW, {
+        argumentLabel,
+        keyPath: keyPath as readonly (string | number | symbol)[],
+        methodName,
+        optionalPathLabel: path ? ` at path \`${path}\`` : '',
+        value,
+        ...(path !== undefined ? { path } : undefined),
+    });
+    if ('captureStackTrace' in Error && typeof Error.captureStackTrace === 'function') {
+        Error.captureStackTrace(error, createSolanaJsonRpcIntegerOverflowError);
```
steveluscher, line 42: This might be interesting to you, viewer. Lets you truncate the stack trace at {some method} so that it doesn't _appear_ in the stack trace. It's not interesting that the error happened to be `newed` up in `createSolanaJsonRpcIntegerOverflow()` – all the user cares about is where the error was thrown.

packages/library/src/rpc-integer-overflow-error.ts
```diff
@@ -1,45 +1,45 @@
...
+            ? keyPath
+                  .slice(1)
+                  .map(pathPart => (typeof pathPart === 'number' ? `[${pathPart}]` : pathPart))
+                  .join('.')
+            : undefined;
+    const error = new SolanaError(SOLANA_ERROR__RPC_INTEGER_OVERFLOW, {
+        argumentLabel,
+        keyPath: keyPath as readonly (string | number | symbol)[],
+        methodName,
+        optionalPathLabel: path ? ` at path \`${path}\`` : '',
```
lorisleiva, line 37: So I guess that's how we go around some of the more complex error messages without using callbacks. 👍

lorisleiva (approved):

packages/library/src/rpc-integer-overflow-error.ts
steveluscher: Yeah, I'm super sad about pushing this error-formatting code into the _production_ bundle but I think this error is important enough to collect this amount of detail, _particularly_ in production, where it will occur most insidiously.

PR #2091, opened 2024-02-04
Fix TypeScript index signature

ghepesdoru:
# Summary

Fixing an issue related to **ScalarEnum** type definition not complying with the TypeScript signature parameter types (cannot be union types).

# Heads up
This is my very first attempt to contribute to this project and I might not be entirely up to speed with the project's conventions. In the case I have missed anything, first I am sorry, and second let me know so I have a chance to fix it. Thank you.

# Details
When including @solana-web3.js inside a TypeScript project an error is raised against `type ScalarEnum<T>` more specifically it's signature parameter type is using a type union. 

The specific error is: `An index signature parameter type cannot be a union type. Consider using a mapped object type instead.` and can be reproduced quite easily by just including the dependency inside a TypeScript 4+ project.

## Mitigation
Separating the type union and discriminating the key as being in the union's types while casting it to a constant string should eliminate the issue.

### DIFF
--- a/packages/codecs-data-structures/src/scalar-enum.ts
+++ b/packages/codecs-data-structures/src/scalar-enum.ts
@@ -23,6 +23,8 @@ import {
     NumberEncoder,
 } from '@solana/codecs-numbers';
 
+type ScalarEnumKeyType = number | string;
+
 /**
  * Defines a scalar enum as a type from its constructor.
  *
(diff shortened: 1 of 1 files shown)


steveluscher:
I'll leave this up to @lorisleiva to take a look at, but this library is _definitely_ not compatible with TypeScript 4 and anything we do to try to make it so is a losing battle.

lorisleiva:
Hi there, thank you for raising this.

It's in my to-do list to try and simplify the definition of a Scalar Enum — i.e. recognising its "lookup object". When I do, this might be more TS4 friendly but, as Steven pointed out, this isn't a version of TS we support so it won't be my main goal.

Unfortunately, I'll be closing this PR in the meantime as this is not the solution I'm looking for.

ghepesdoru:
Thank you for your explanation @lorisleiva. 

Guess for the time being setting`"skipLibCheck": true` inside one's tsconfig.json `compilerOptions` is the way to go for integration with TypeScript 4+ contexts and Rollup based/inspired bundlers.

lorisleiva:
@ghepesdoru Not sure if you spotted that PR but I recently refactored the scalar enum codecs: https://github.com/solana-labs/solana-web3.js/pull/2100.

There's a chance this fixes your issue.

PR #2092, opened 2024-02-04
Update Readme broken link to rpc docs

SkyHustle:
Someone's gotta do the hard stuff 😄


steveluscher (changes requested):
Sweet. Can you also make the change in `packages/library-legacy/`, then I'll jam it in?

PR #2111, opened 2024-02-09
refactor(experimental): store fee payer Signers in transactions

lorisleiva:
### Context

The `@solana/signers` package allows us to store `TransactionSigners` inside instructions' account meta which can then be extracted in order to sign the transaction they belong to.

### Problem

Signing transactions that way currently doesn't take the fee payer into account. It works most of the time as it is likely that the fee payer will end up being required as a signer inside one of the instructions. However, when that's not the case, the fee payer will have to manually sign the transaction on top of using the `signTransactionWithSigners` function.

### Solution

This PR fixes that by storing the fee payer signer inside the transaction object, much like we do with the instruction account metas.

It offers a new `setTransactionFeePayerSigner` which, on top of adding the `feePayer` address, also add a `feePayerSigner` attribute that stores the fee payer as a `TransactionSigner`.

That way, when using the `@solana/signers` package, you get to attach all relevant signers to a transaction and then use `signTransactionWithSigners` to handle the whole signing process, including the fee payer.

### DIFF
--- a/packages/signers/src/__tests__/account-signer-meta-test.ts
+++ b/packages/signers/src/__tests__/account-signer-meta-test.ts
@@ -1,6 +1,8 @@
 import { Address } from '@solana/addresses';
+import { createTransaction } from '@solana/transactions';
 
 import { getSignersFromInstruction, getSignersFromTransaction } from '../account-signer-meta';
+import { setTransactionFeePayerSigner } from '../fee-payer-signer';
 import {
     createMockInstructionWithSigners,
     createMockTransactionModifyingSigner,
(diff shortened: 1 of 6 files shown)


lorisleiva:
* `master`
  * **PR #2111** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2111?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2111?utm_source=stack-comment).

steveluscher (approved):
Nice. Totally get why this is needed. And the reason that no changes to `signTransaction` are needed is because `CompilableTransaction` implies that you have a fee payer, and `IFullySignedTransaction` implies that they signed.

Added via Giphy

PR #2112, opened 2024-02-10
refactor(experimental): add createKeyPairFromBytes helper

lorisleiva:
Adds a helper function that creates a crypto KeyPair from a 64-bytes secret key.

### DIFF
--- a/packages/compat/src/keypair.ts
+++ b/packages/compat/src/keypair.ts
@@ -1,4 +1,4 @@
-import { createPrivateKeyFromBytes } from '@solana/keys';
+import { createKeyPairFromBytes } from '@solana/keys';
 import { Keypair } from '@solana/web3.js';
 
 /**
@@ -7,12 +7,8 @@ import { Keypair } from '@solana/web3.js';
  * @returns         A CryptoKeyPair
  */
 export async function fromLegacyKeypair(keypair: Keypair, extractable?: boolean): Promise<CryptoKeyPair> {
-    const [publicKey, privateKey] = await Promise.all([
-        crypto.subtle.importKey('raw', keypair.publicKey.toBytes(), 'Ed25519', true, ['verify']),
-        createPrivateKeyFromBytes(keypair.secretKey.slice(0, 32), extractable),
-    ]);
-    return {
-        privateKey,
-        publicKey,
-    } as CryptoKeyPair;
+    const bytes = new Uint8Array(64);
+    bytes.set(keypair.secretKey);
+    bytes.set(keypair.publicKey.toBytes(), /* offset */ 32);
+    return createKeyPairFromBytes(bytes, extractable);
 }
--- a/packages/errors/src/messages.ts
+++ b/packages/errors/src/messages.ts
@@ -1,4 +1,5 @@
 import {
+    SOLANA_ERROR__INVALID_KEYPAIR_BYTES,
     SOLANA_ERROR__RPC_INTEGER_OVERFLOW,
     SOLANA_ERROR__TRANSACTION_MISSING_SIGNATURES,
     SOLANA_ERROR__TRANSACTION_SIGNATURE_NOT_COMPUTABLE,
@@ -17,11 +18,12 @@ export const SolanaErrorMessages: Readonly<{
     // TypeScript will fail to build this project if add an error code without a message.
     [P in SolanaErrorCode]: string;
 }> = {
+    [SOLANA_ERROR__INVALID_KEYPAIR_BYTES]: 'Key pair bytes must be of length 64, got $byteLength.',
     [SOLANA_ERROR__RPC_INTEGER_OVERFLOW]:
         'The $argumentLabel argument to the `$methodName` RPC method$optionalPathLabel was ' +
         '`$value`. This number is unsafe for use with the Solana JSON-RPC because it exceeds ' +
-        '`Number.MAX_SAFE_INTEGER`',
-    [SOLANA_ERROR__TRANSACTION_MISSING_SIGNATURES]: 'Transaction is missing signatures for addresses: $addresses',
+        '`Number.MAX_SAFE_INTEGER`.',
+    [SOLANA_ERROR__TRANSACTION_MISSING_SIGNATURES]: 'Transaction is missing signatures for addresses: $addresses.',
     [SOLANA_ERROR__TRANSACTION_SIGNATURE_NOT_COMPUTABLE]:
         "Could not determine this transaction's signature. Make sure that the transaction has " +
         'been signed by its fee payer.',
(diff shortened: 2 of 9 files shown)


lorisleiva:
* `master`
  * **PR #2112** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2112?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #2113** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2113?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2112?utm_source=stack-comment).

packages/keys/src/key-pair.ts
```diff
@@ -9,3 +11,15 @@ export async function generateKeyPair(): Promise<CryptoKeyPair> {
     );
     return keyPair as CryptoKeyPair;
 }
+
+export async function createKeyPairFromBytes(bytes: Uint8Array, extractable?: boolean): Promise<CryptoKeyPair> {
+    if (bytes.length !== 64) {
+        // TODO: Coded error.
+        throw new Error('KeyPair bytes must be of length 64');
```
steveluscher, line 18: Coded errors are gtg if you want to write your first one! Just land the whole thing in the same PR. Add to `./codes.ts`, `./messages.ts` and `./context.ts` if you want to add some.

packages/keys/src/key-pair.ts
```diff
@@ -9,3 +11,15 @@ export async function generateKeyPair(): Promise<CryptoKeyPair> {
     );
     return keyPair as CryptoKeyPair;
 }
+
+export async function createKeyPairFromBytes(bytes: Uint8Array, extractable?: boolean): Promise<CryptoKeyPair> {
+    if (bytes.length !== 64) {
```
steveluscher, line 16: I've been used to writing `byteLength`. Is that literally just an alias for `length` or is there something more subtle?

I think it happens to be ‘the same’ particularly for `Uint8Array` (ie. the number of elements in the array happens to also be the number of bytes) but it's not true for multibyte integers.

```ts
new Uint16Array([1]).byteLength; // 2
new Uint16Array([1]).length; // 1
```

packages/keys/src/key-pair.ts
```diff
@@ -9,3 +11,15 @@ export async function generateKeyPair(): Promise<CryptoKeyPair> {
...
+        // TODO: Coded error.
+        throw new Error('KeyPair bytes must be of length 64');
+    }
+    const [publicKey, privateKey] = await Promise.all([
+        crypto.subtle.importKey('raw', bytes.slice(32), 'Ed25519', true, ['verify']),
```
steveluscher, line 21: For the next poor person who has to read this, can you comment the inscrutable positional args…
```suggestion
        crypto.subtle.importKey(
            'raw',
            bytes.slice(32),
            'Ed25519',
            /* extractable */ true,
            ['verify'],
        ),
```

steveluscher (approved):
For completeness, you might like to import this new implementation into `fromLegacyKeypair` to eliminate some duplication.

steveluscher (commented):
Also, add to `keys/README.md`?

packages/keys/src/key-pair.ts
lorisleiva: I'll change it to `byteLength` then for consistency.

packages/keys/src/key-pair.ts
lorisleiva: Done!

packages/compat/src/keypair.ts
```diff
@@ -7,12 +7,8 @@ import { Keypair } from '@solana/web3.js';
...
-    const [publicKey, privateKey] = await Promise.all([
-        crypto.subtle.importKey('raw', keypair.publicKey.toBytes(), 'Ed25519', true, ['verify']),
-        createPrivateKeyFromBytes(keypair.secretKey.slice(0, 32), extractable),
-    ]);
-    return {
-        privateKey,
-        publicKey,
-    } as CryptoKeyPair;
+    return createKeyPairFromBytes(
+        new Uint8Array([...keypair.secretKey.slice(0, 32), ...keypair.publicKey.toBytes()]),
```
steveluscher, line 11: Technically you can just do this, since the `secretKey` is the private scalar (32 bytes) plus the public key (32 bytes).
```suggestion
        keypair.secretKey,
```

packages/compat/src/keypair.ts
```diff
@@ -7,12 +7,8 @@ import { Keypair } from '@solana/web3.js';
...
+        extractable,
+    );
```
steveluscher, line 13: Either way, slight preference for:

```ts
const bytes = new Uint8Array(64);
bytes.set(keypair.secretKey);
bytes.set(keypair.publicKey.toBytes(), 32 /* offset */);
```

packages/keys/src/key-pair.ts
```diff
@@ -9,3 +12,14 @@ export async function generateKeyPair(): Promise<CryptoKeyPair> {
     );
     return keyPair as CryptoKeyPair;
 }
+
+export async function createKeyPairFromBytes(bytes: Uint8Array, extractable?: boolean): Promise<CryptoKeyPair> {
+    if (bytes.byteLength !== 64) {
+        throw new SolanaError(SOLANA_ERROR__INVALID_KEYPAIR_BYTES);
```
steveluscher, line 18: Is it worth adding stuff to the context, like the actual length, or the actual bytes? Maybe this is where we start needing ‘private’ context properties (ie. ones that won't get serialized by the context serializer).

packages/errors/src/messages.ts
```diff
@@ -17,6 +18,7 @@ export const SolanaErrorMessages: Readonly<{
     // TypeScript will fail to build this project if add an error code without a message.
     [P in SolanaErrorCode]: string;
 }> = {
+    [SOLANA_ERROR__INVALID_KEYPAIR_BYTES]: 'KeyPair bytes must be of length 64',
```
steveluscher, line 21: Slight preference for…
```suggestion
    [SOLANA_ERROR__INVALID_KEYPAIR_BYTES]: 'Key pair bytes must be of length 64',
```

steveluscher (approved):
Added via Giphy

packages/compat/src/keypair.ts
lorisleiva: Yeah I like this better to explicitly use the `publicKey` attribute of the `KeyPair`.

packages/keys/src/key-pair.ts
lorisleiva: Happy to add that info to the context. Is it really an attack vector to expose the length of the provided bytes though? I think a private context property might be a bit of a stretch here.

lorisleiva:
### Merge activity

* **Feb 16, 3:52 PM EST**: @lorisleiva started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2112).

PR #2138, opened 2024-02-19
Fix the CI but for real this time

mcintyre94:
It turns out that https://github.com/solana-labs/solana-web3.js/pull/2044 doesn't work properly. The `npm version` change from `version-from-git` wasn't running before `publish-packages` as I thought. So the `npm view $npm_package_name@$npm_package_version`  used `2.0.0-development` as the version.

I've done some testing with a modified CI here: https://github.com/solana-labs/solana-web3.js/pull/2135
You can see in the logs there eg:
```
  > echo "running publish-packages" && echo "$npm_package_name@$npm_package_version" && npm view $npm_package_name@$npm_package_version || echo "would publish $npm_package_name@$npm_package_version"
  
  running publish-packages
  @solana/accounts@2.0.0-development
  npm WARN Ignoring workspaces for specified package(s) 
  
  @solana/accounts@2.0.0-development | MIT | deps: 4 | versions: 82
...
```

In practice most packages don't have a `2.0.0-development` version published, but some do, and therefore don't publish with the previous CI change.

This PR updates the logic so that we run `version-from-git` (via `prepublishOnly`) before we do the `npm view` check. 

I've split this so that `publish-packages` runs `pnpm prepublishOnly && pnpm publish-impl`, where `publish-impl` runs `npm view ... || pnpm publish ...`. This split is necessary because the `$npm_package_version` is set when a package.json script begins, which means that if you do `version-from-git && npm view package@$npm_package_version` then it'll use the version before it was updated. This also means that we can't just put the `npm view` check into `prepublishOnly` after `version-from-git`.

I've kept the `version-from-git` in `prepublishOnly`, so that it's still used if anyone does some sort of manual publishing. It'll just be a no-op when it's called for the publish in `publish-impl` from `publish-packages` because it'll have already set the version accordingly. `version-from-git` prints an error message when it doesn't update the version, but still uses exit code 0 so the publish should work correctly.  

Special cases:

- `library-legacy` has the same logic as other packages, but uses `semantic-release ...` instead of `pnpm publish ...` in `publish-impl`
- `library` also now has the same logic (it already had this split into 2 commands), and just runs its extra sed commands before `pnpm prepublishOnly && pnpm publish-impl`. It also still does `git reset --hard` at the end, which I didn't add to the other commands because they should run concurrently.


steveluscher (approved):
Yeah, this is _also_ good because if package A has already published (ie. skips publishing) but package B that depends on A hasn't, package A's version will have been updated for use in the publish step by package B. Ship it.

Issue #2124, opened 2024-02-15
[experimental] Reshuffle `@solana/rpc-*` packages


steveluscher:
Expect code to move between these packages to untangle semantics and guard against circular dependencies.

lorisleiva:
Fixed by https://github.com/solana-labs/solana-web3.js/pull/2168

PR #2144, opened 2024-02-20
refactor(experimental): create @solana/rpc-spec package

lorisleiva:
_Dear reviewer,_

_Welcome to the RPC refactoring stack! 🙌 🥞_ 

_You are about to embark a new journey where very little change yet so much does. The main goal of this adventure is to untangle semantics and guard against circular dependencies as previously documented in #2124. But who knows what you might come across? Will these changes trigger some mind-blowing ideas for our RPC heroes? Let's find out!_

---

Before we jump into the first PR of our stack, let's see the overall plan and what it leads to. Here are the main goals we are aiming to reach by the end of this stack:

* **Separation of RPCs and RPC Subscriptions**. You will be able to import one without the other.
* **Separation of Solana-specific RPC designs and generic JSON RPC specifications**. Using the `*-spec` package suffix, we will be able to distinguish between the two more easily.
* **Separation of Transports and API definitions**. Whilst this is somewhat the case already, the two are still romantically tangled inside `rpc-core`. We aim to cruelly separate them and make API definitions standalone type-only packages.
* **Better abstraction of transports**. The `RpcTransport` and `RpcSubscriptionsTransport` types will respectively abstract the RPC and RPC Subscriptions transport layers. The language and package names will distinguish between the two in a more explicit and consistent manner.
* **Gentle API polishing**. We'll take this opportunity to iron out some of the RPC types for consistency. For instance, we'll remove the non-distinguishing `I` prefixes in types and uses a consistent language between RPC and RPC Subscription types (e.g. `RpcRequest` and `RpcSubscriptionRequest`).

And here's how we'll get there. We will start by creating new packages gradually picking code from the exciting RPC packages — such as `rpc-core` and `rpc-transport`. Then we'll update consumers of these packages to use our newly created packages. Finally, we'll remove the old RPC packages without looking back.

The following diagram describes this plan in more details. You can see the current state of RPC packages and the one we aim to reach by the end of this stack. Packages in the "**After**" section are labelled with their package description to quickly grasp their purpose.

RPC packages reshuffle@2x

Note that each of these new packages will require new READMEs. I will tackle this in a follow-up PR to this stack as soon as reviewers are happy with the direction of this refactoring.

Throughout this stack, I will provide an updated version of the diagram above that highlights the part(s) of the plan the current PR focuses on.

Which leads us to our first PR!

---

In this first PR and the next few ones, we focus on defining the generic part of JSON RPCs and their subscriptions. Meaning, it could in theory be used to create RPC implementations that have nothing to do with Solana.

This first PR focuses on the RPC calls (as opposed to RPC subscriptions) and defines core types such as `Rpc`, `RpcTransport` or `RpcApi`. It also offers a `createRpc` function that, given an api and a transport, returns an RPC implementation that uses proxy. Nothing new here in terms of logic, just a new home for our generic JSON RPC implementation.

RPC packages reshuffle@2x (1)


lorisleiva:
* **#2146** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2146?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2145** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2145?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2144** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2144?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2144?utm_source=stack-comment).

steveluscher (approved):
Love it. I have to say that I don't _completely_ get the `-spec` suffix, but I really don't feel strongly about it.

Added via Giphy

lorisleiva:
I'm happy to take other suggestions. Initially I had `-proxy` but I wasn't too happy about that one.

packages/rpc-spec/src/rpc-message.ts
```diff
@@ -0,0 +1,15 @@
...
+}
+
+export function createRpcMessage<TParams>(method: string, params: TParams) {
+    return {
+        id: getNextMessageId(),
+        jsonrpc: '2.0',
+        method,
+        params,
+    };
+}
```
buffalojoec, line 15: 🔥 🔥

buffalojoec (approved):
I think this looks good, but I'm wondering why we are dropping the `json-` prefix? I realize you're splitting methods and notifications, but after all this is _still_ a JSON RPC specification implementation library. I lean towards it being named as such.

buffalojoec:
> Love it. I have to say that I don't _completely_ get the `-spec` suffix, but I really don't feel strongly about it.

I like spec!

lorisleiva:
Prior to this stack, we had a mixture of `json-` prefixed names and non-prefixed names such as `createJsonRpcMessage` and `Rpc`.

When trying to find a good convention I could have done something like this:

| Non-Solana specific | Solana specific |
| - | - |
| `JsonRpc` | `SolanaRpc` |
| `JsonRpcApi` | `SolanaRpcApi` |
| `JsonRpcTransport` | `SolanaRpcTransport` |

But then it feels like the `SolanaRpc` isn't a JSON one, in comparison to the `JsonRpc` types.

I could have also used types like `SolanaJsonRpc` but then it adds verbosity to all our RPC types when actually, within the scope of this library, we only care about JSON RPCs. So that's why I went with the following instead:


| Non-Solana specific | Solana specific |
| - | - |
| `Rpc` | `SolanaRpc` |
| `RpcApi` | `SolanaRpcApi` |
| `RpcTransport` | `SolanaRpcTransport` |

lorisleiva:

PR #2146, opened 2024-02-20
refactor(experimental): extract common spec logic into new @solana/rpc-spec-types package

lorisleiva:
This PR extracts common types and helpers from `rpc-spec` and `rpc-subscriptions-spec` into a new `rpc-spec-types` package.

RPC packages reshuffle@2x (3)

PR #2145, opened 2024-02-20
refactor(experimental): create @solana/rpc-subscriptions-spec package

lorisleiva:
In this PR, we continue our efforts to create generic JSON RPC packages, this time, focusing on RPC Subscriptions.

Notice how some of the types and helpers have been renamed to stay consistent with the `rpc-spec` types and helpers. For instance, we have:

| `rpc-spec` | `rpc-subscriptions-spec` |
|-|-|
| `Rpc` | `RpcSubscriptions` |
| `RpcApi` | `RpcSubscriptionsApi` |
| `RpcTransport` | `RpcSubscriptionsTransport` |
| `RpcRequest` | `RpcSubscriptionsRequest` |
| `PendingRpcRequest` | `PendingRpcSubscriptionsRequest` |
| `createRpc()` | `createSubscriptionRpc()` |
| `createRpcApi()` | `createRpcSubscriptionsApi()` |

Even the folder structure of these packages were made consistent.

Notice a bit of redundancy between `rpc-spec` and `rpc-subscriptions-spec`? The next PR solves this.

RPC packages reshuffle@2x (2)

PR #2147, opened 2024-02-20
refactor(experimental): create @solana/rpc-transport-http package

lorisleiva:
This PR extract the HTTP implementation of the `RpcTransport` into its own package.

Notice that it does not include any Solana-specific types or helpers such as `ClusterUrl` since this HTTP transport should be used as a standalone HTTP transport of the `@solana/rpc-spec` package. It is the responsibility of the Solana-specific `@solana/rpc` package — created later on in the stack — to wrap this transport and introduce these concepts.

RPC packages reshuffle@2x (4)


lorisleiva:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2147?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2147** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2147?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2146** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2146?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2145** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2145?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2144** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2144?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2147?utm_source=stack-comment).

steveluscher (approved):
This is fine, but one problem about replicating the code here and deleting the other stuff later is that we'll lose the entire git history. If you can nuke it and add it in the same PR so that git recognizes the move, that would be better.

[queuepatrick](https://app.graphite.dev/macros/solana-labs)

lorisleiva:
@steveluscher The only issue is `rpc-transport` contains more than just this code so I'd need to be merging like 4 PRs together. It also means that this combined PR would also need to include the re-wiring of all consumers which would be massive (library, accounts, graphql, etc). I tried to split the stack into more digestible narratives but yeah it's a shame about the git history. 😞

steveluscher:
We can always squash the stack into a single commit when we jam this in.

lorisleiva:
Perfect, let's do that! 👌

packages/rpc-transport-http/src/http-transport-errors.ts
```diff
@@ -0,0 +1,16 @@
...
+    readonly statusCode: number;
+    constructor(details: SolanaHttpErrorDetails) {
+        super(`HTTP error (${details.statusCode}): ${details.message}`);
+        Error.captureStackTrace(this, this.constructor);
+        this.statusCode = details.statusCode;
+    }
+    get name() {
+        return 'SolanaHttpError';
+    }
+}
```
buffalojoec, line 16: Should these be here? Or should they be called `Solana`?

From PR description:

> Notice that it does not include any Solana-specific types

buffalojoec (commented):
Ohhh I'm starting to like where this is going.

packages/rpc-transport-http/src/http-transport-errors.ts
lorisleiva: I was going to replace that with the `SolanaError` from `@solana/errors` but we decided to do that outside of this PR so I just left it as-is for now.

lorisleiva:

PR #2157, opened 2024-02-20
refactor(experimental): use new RPC libraries in @solana/accounts

lorisleiva:
This PR rewires the RPC dependencies of `@solana/accounts` to use the new RPC packages.

RPC packages reshuffle@2x (14)


packages/accounts/src/__tests__/__setup__.ts
```diff
@@ -1,16 +1,15 @@
...
     AccountInfoWithBase58EncodedData,
     AccountInfoWithBase64EncodedData,
-    GetAccountInfoApi,
-    GetMultipleAccountsApi,
-    JsonParsedDataResponse,
-} from '../rpc-api';
+    SolanaRpcResponse,
+} from '@solana/rpc-types';
+
+import type { GetAccountInfoApi, GetMultipleAccountsApi, JsonParsedDataResponse } from '../rpc-api';
```
buffalojoec, line 12: Remind me again why we have to re-implement these in this package? It seems like through your refactoring, `@solana/rpc` doesn't need `@solana/accounts`, neither does `@solana/rpc-api`.

lorisleiva:

packages/accounts/src/__tests__/__setup__.ts
buffalojoec: Yeahhhh I think we should hold off on that. That's like 60 packages! 🫠

lorisleiva: Agreed! 😅

PR #2177, opened 2024-02-24
Ignore any test files ending with `-test.node.ts` or `-test.browser.ts` in other environments

steveluscher:

PR #2176, opened 2024-02-24
Remove `node-fetch` external from build script

steveluscher:

PR #2178, opened 2024-02-24
Allow devs to supply a dispatcher to the Node implementation of `fetch`

### LINKED ISSUE #2126: [experimental] Figure out recommended way to override `fetch` dispatcher in Node
The current version of `@solana/web3.js` lets people supply their own `HTTPAgent` (often [at their own peril](https://github.com/solana-labs/solana/pull/29130)) to be able to put a keep-alive config on the wire. We'll need a way to do that in the new web3.js without bothering people who use it on the web / React Native.

Options:

1. Create a transport for use in Node only. Plumb `dispatcher` all the way down to `fetch()`. See [typedef](https://github.com/nodejs/undici/blob/cdc6d5ca45e64a0cf21b7ee0f13aa993568cb428/types/fetch.d.ts#L117).
2. Recommend that people override the global Unidici dispatcher on which `fetch()` depends. See [this Stack Overflow question](https://stackoverflow.com/questions/77381314/is-it-possible-to-set-a-global-dispatcher-for-the-built-in-fetch-in-node).

steveluscher:
# Summary

The current version of web3.js allows folks to supply what's called an ‘HTTP Agent.’ This is a class that mediates how network requests should be fetched. One example is to establish a connection pool for a given URL, and keep connections open for a time so that TLS/SSL negotiation doesn't have to be done on every request.

The new web3.js up until this point has punted on how to reintroduce this concept. In this PR we bring it back in the form of allowing callers to create transports with custom Undici `Dispatchers`.

It only works in Node. If you supply the `dispatcher_NODE_ONLY` property in non-Node environments you'll get a console warning that your config has been ignored.

# Alternate designs considered

I desperately wanted to avoid the situation in this PR, which was to create a config property on transports that's only usable in Node.

Alternate approaches considered:

1. Create a special transport for use in Node. Callers would have to `createDefaultNodeRpcTransport()` and/or `solanaRpcFactoryForNode()` which I thought was a bridge too far.
2. Encourage people to [inject a global dispatcher](https://stackoverflow.com/a/77526783/802047). This is a terrible idea for all of the reasons that global state is terrible:
    * Applies to all connections in the process
    * Your global dispatcher can be unset by someone _else_
    * There might already be a global dispatcher installed (eg. a `ProxyAgent`) and you might unknowingly replace it
3. Export a `Dispatcher` setter that lets you install a dispatcher accessible _only_ to `@solana/fetch-impl`. Besides having most of the bad features of #&#8203;2, this would not work. We inline `@solana/fetch-impl` in the build, meaning there's no longer anywhere to export such a method from.

# Test Plan

```shell
cd packages/fetch-impl/
pnpm test:unit:node
cd ../rpc-transport-http/
pnpm test:unit:node
```

See benchmark tests in next PR.

Closes #2126.

### DIFF
--- a/packages/build-scripts/getBaseConfig.ts
+++ b/packages/build-scripts/getBaseConfig.ts
@@ -43,6 +43,8 @@ export function getBaseConfig(platform: Platform, formats: Format[], _options: O
                           external: [
                               // Despite inlining `@solana/text-encoding-impl`, do not recursively inline `fastestsmallesttextencoderdecoder`.
                               'fastestsmallesttextencoderdecoder',
+                              // Despite inlining `@solana/fetch-impl`, do not recursively inline `undici`.
+                              'undici',
                               // Despite inlining `@solana/ws-impl`, do not recursively inline `ws`.
                               'ws',
                           ],
--- a/packages/fetch-impl/package.json
+++ b/packages/fetch-impl/package.json
@@ -35,12 +35,17 @@
         "test:prettier": "jest -c node_modules/@solana/test-config/jest-prettier.config.ts --rootDir . --silent",
         "test:treeshakability:browser": "agadoo dist/index.browser.js",
         "test:treeshakability:node": "agadoo dist/index.node.js",
-        "test:typecheck": "tsc --noEmit"
+        "test:typecheck": "tsc --noEmit",
+        "test:unit:browser": "jest -c node_modules/@solana/test-config/jest-unit.config.browser.ts --rootDir . --silent --passWithNoTests",
+        "test:unit:node": "jest -c node_modules/@solana/test-config/jest-unit.config.node.ts --rootDir . --silent"
     },
     "browserslist": [
         "supports bigint and not dead",
         "maintained node versions"
     ],
+    "dependencies": {
+        "undici": "^6.2.2"
+    },
     "devDependencies": {
         "@solana/eslint-config-solana": "^1.0.2",
         "@solana/test-config": "workspace:*",
--- a/packages/fetch-impl/src/__tests__/fetch-test.node.ts
+++ b/packages/fetch-impl/src/__tests__/fetch-test.node.ts
@@ -1,0 +1,21 @@
+import { Dispatcher, fetch as fetchImpl } from 'undici';
+
+import fetch from '../index.node';
+
+jest.mock('undici');
+
+describe('fetch', () => {
+    it('should call the underlying `fetch` with the `dispatcher` supplied in `requestInit`', () => {
+        const explicitDispatcher = Symbol('explicitDispatcher') as unknown as Dispatcher;
+        fetch('http://solana.com', { dispatcher: explicitDispatcher });
+        expect(fetchImpl).toHaveBeenCalledWith('http://solana.com', {
+            dispatcher: explicitDispatcher,
+        });
+    });
+    it('should call the underlying `fetch` with an undefined `dispatcher` when an undefined is explicitly supplied in `requestInit`', () => {
+        fetch('http://solana.com', { dispatcher: undefined });
+        expect(fetchImpl).toHaveBeenCalledWith('http://solana.com', {
+            dispatcher: undefined,
+        });
+    });
+});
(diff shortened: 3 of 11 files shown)


steveluscher:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2178?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2179** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2179?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2178** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2178?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2177** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2177?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2176** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2176?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2178?utm_source=stack-comment).

packages/fetch-impl/src/dispatcher.ts
```diff
@@ -0,0 +1,16 @@
...
+export function getDefaultDispatcher(): Dispatcher {
+    if (!cachedDispatcher) {
+        cachedDispatcher = new Agent({
+            // FIXME: https://github.com/nodejs/undici/issues/2808
+            // allowH2: true,
+            connections: 16,
+            // One second fewer than the Solana RPC's keepalive timeout.
+            // Read more: https://github.com/solana-labs/solana/issues/27859#issuecomment-1340097889
+            keepAliveTimeout: 19000,
+        });
```
steveluscher, line 13: I made the decision to deliver a default dispatcher to everyone. This is what the current web3.js does. I _really_ hate this, because everyone's workload is different and requires tuning, but it's also true that having _no_ dispatching strategy has pretty awful performance.

lorisleiva:
Is there no way to bring this Dispatcher API to the browser?

steveluscher:
There is not. The browser doesn't _let_ you tune its networking stack.

packages/fetch-impl/src/dispatcher.ts
```diff
@@ -0,0 +1,18 @@
...
+
+let cachedDispatcher: Agent;
+export function getDefaultDispatcher(): Dispatcher {
+    if (!cachedDispatcher) {
+        cachedDispatcher = new Agent({
+            // FIXME: https://github.com/nodejs/undici/issues/2808
+            // allowH2: true,
+            // See https://github.com/solana-labs/solana-web3.js/pull/2179 for
+            // strategies on how to tune the number of connections through benchmarking.
+            connections: 32,
```
steveluscher, line 11: I'm starting to regret this in light of my latest benchmarks in #2179.

lorisleiva: If we're not creating a more performant Agent by providing these opinionated default values then that's good right? At least we don't need to be opinionated here.

lorisleiva (approved):

packages/fetch-impl/src/dispatcher.ts
steveluscher: Yeah, I think I might tear this default `Agent` out until we have evidence that `undici` with _no_ dispatcher is insufficient. The `getLatestBlockhash` benchmarks would indicate that it does a fine job of scheduling requests without one.

steveluscher:
### Merge activity

* **Feb 27, 10:58 AM PST**: @steveluscher started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2178).

PR #2181, opened 2024-02-24
refactor(experimental): improve codec field and variant types

### LINKED ISSUE #2159: Support for auto-inferring types for @solana/codecs.
## Motivation

@solana/codecs currently expect all types to be clearly defined when defining a `Codec`. This causes a lot of redundant code to be typed by developers when defining their own `Codec` instances. 

This, however, is unnecessary given that all codec construction methods already hold their expected encoded/decoded type in their return type signature.

This should work out of the box, but it doesn't:

```ts
export const tokenAmountCodecExpected = getDataEnumCodec([
  ["some", getStructCodec([["amount", getU64Codec()]])],
  ["all", getUnitCodec()],
]);
```

```
Type 'FixedSizeCodec<void & Omit<DataEnum, "__kind">, void & Omit<DataEnum, "__kind">>' is not assignable to type 'Codec<void> | Codec<Omit<DataEnum, "__kind">, Omit<DataEnum, "__kind">>'.
  Type 'FixedSizeCodec<void & Omit<DataEnum, "__kind">, void & Omit<DataEnum, "__kind">>' is not assignable to type 'FixedSizeCodec<Omit<DataEnum, "__kind">, Omit<DataEnum, "__kind">, number>'.
    Type 'FixedSizeCodec<void & Omit<DataEnum, "__kind">, void & Omit<DataEnum, "__kind">>' is not assignable to type 'BaseEncoder<Omit<DataEnum, "__kind">>'.
      Types of property 'encode' are incompatible.
        Type '(value: void & Omit<DataEnum, "__kind">) => Uint8Array' is not assignable to type '(value: Omit<DataEnum, "__kind">) => Uint8Array'.
          Types of parameters 'value' and 'value' are incompatible.
            Type 'Omit<DataEnum, "__kind">' is not assignable to type 'void & Omit<DataEnum, "__kind">'.
              Type 'Omit<DataEnum, "__kind">' is not assignable to type 'void'.ts(2322)
```

The only way to make this work is by redundantly typing each declared variant and field in `tokenAmount` twice:

```ts
export const tokenAmountCodec = getDataEnumCodec<
  { __kind: "some"; amount: bigint } | { __kind: "all" }
>([
  ["some", getStructCodec<{ amount: bigint }>([["amount", getU64Codec()]])],
  ["all", getUnitCodec()],
]);
```

lorisleiva:
This PR improves the type inferences of `getStructCodec` and `getDataEnumCodec` such that it becomes easier to create custom codecs using them.

Thank you @lithdew for raising this in #2159 and for providing a detailed explanation of the implementation.

Fixes #2159.

### DIFF
--- a/packages/codecs-data-structures/src/data-enum.ts
+++ b/packages/codecs-data-structures/src/data-enum.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 import {
     assertByteArrayIsNotEmptyForCodec,
     Codec,
@@ -56,42 +57,6 @@ export type GetDataEnumKindContent<T extends DataEnum, K extends T['__kind']> = 
     '__kind'
 >;
 
-/** Get the name and encoder of each variant in a data enum. */
-export type DataEnumToEncoderTuple<TFrom extends DataEnum> = Array<
-    TFrom extends never
-        ? never
-        : [
-              TFrom['__kind'],
-              keyof Omit<TFrom, '__kind'> extends never
-                  ? Encoder<Omit<TFrom, '__kind'>> | Encoder<void>
-                  : Encoder<Omit<TFrom, '__kind'>>,
-          ]
->;
-
-/** Get the name and decoder of each variant in a data enum. */
-export type DataEnumToDecoderTuple<TTo extends DataEnum> = Array<
-    TTo extends never
-        ? never
-        : [
-              TTo['__kind'],
-              keyof Omit<TTo, '__kind'> extends never
-                  ? Decoder<Omit<TTo, '__kind'>> | Decoder<void>
-                  : Decoder<Omit<TTo, '__kind'>>,
-          ]
->;
-
-/** Get the name and codec of each variant in a data enum. */
-export type DataEnumToCodecTuple<TFrom extends DataEnum, TTo extends TFrom = TFrom> = Array<
-    TFrom extends never
-        ? never
-        : [
-              TFrom['__kind'],
-              keyof Omit<TFrom, '__kind'> extends never
-                  ? Codec<Omit<TFrom, '__kind'>, Omit<TTo, '__kind'>> | Codec<void>
-                  : Codec<Omit<TFrom, '__kind'>, Omit<TTo, '__kind'>>,
-          ]
->;
-
 /** Defines the config for data enum codecs. */
 export type DataEnumCodecConfig<TDiscriminator = NumberCodec | NumberEncoder | NumberDecoder> = {
     /**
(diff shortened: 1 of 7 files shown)


lorisleiva:
* **#2181** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2181?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2181?utm_source=stack-comment).

lithdew (commented):
Read through all the code and it looks great. LGTM?

packages/transactions/src/serializers/message.ts
```diff
@@ -100,12 +91,20 @@ export function getCompiledMessageEncoder(): VariableSizeEncoder<CompiledMessage
...
 export function getCompiledMessageDecoder(): VariableSizeDecoder<CompiledMessage> {
-    return mapDecoder(getStructDecoder(getPreludeStructDecoderTuple()), ({ addressTableLookups, ...restOfMessage }) => {
-        if (restOfMessage.version === 'legacy' || !addressTableLookups?.length) {
-            return restOfMessage;
-        }
-        return { ...restOfMessage, addressTableLookups } as Exclude<CompiledMessage, { readonly version: 'legacy' }>;
-    });
+    return mapDecoder(
+        getStructDecoder(getPreludeStructDecoderTuple()) as VariableSizeDecoder<
+            CompiledMessage & { addressTableLookups?: ReturnType<typeof getCompiledAddressTableLookups> }
```
lithdew, line 96: The `ReturnType<..>` could be directly replaced with an `AddressTableLookup[]` here.

packages/transactions/src/serializers/message.ts
```diff
@@ -9,34 +9,27 @@ import {
...
+import type { getCompiledAddressTableLookups } from '../compile-address-table-lookups';
 import { CompiledMessage } from '../message';
 import { getAddressTableLookupDecoder, getAddressTableLookupEncoder } from './address-table-lookup';
 import { getMessageHeaderDecoder, getMessageHeaderEncoder } from './header';
 import { getInstructionDecoder, getInstructionEncoder } from './instruction';
 import { getTransactionVersionDecoder, getTransactionVersionEncoder } from './transaction-version';
 
 function getCompiledMessageLegacyEncoder(): VariableSizeEncoder<CompiledMessage> {
-    return getStructEncoder(getPreludeStructEncoderTuple());
+    return getStructEncoder(getPreludeStructEncoderTuple()) as VariableSizeEncoder<CompiledMessage>;
```
lithdew, line 24: Is the cast needed despite the return type already being specified?

lorisleiva: Yeah that's specific to the `CompiledMessage` type though. It's because `TransactionVersion` is of type `"legacy" | 0` whereas some nested type within the codec definition defines them as `number` so the cast is just for that.

We could probably improve some of the codec internals inside the transactions package but I wanted to make sure this PR didn't diverge into a big multi-package refactoring.

packages/transactions/src/serializers/message.ts
lorisleiva: I'm guessing it was designed that way because the return type is not exported.

buffalojoec (approved):
Nice, this looks slick!

slick

packages/codecs-data-structures/src/data-enum.ts
```diff
@@ -142,10 +127,11 @@ export function getDataEnumEncoder<TFrom extends DataEnum>(
  * @param variants - The variant decoders of the data enum.
  * @param config - A set of config for the decoder.
  */
-export function getDataEnumDecoder<T extends DataEnum>(
-    variants: DataEnumToDecoderTuple<T>,
+export function getDataEnumDecoder<const TVariants extends Variants<Decoder<any>>>(
```
lorisleiva, line 130: @steveluscher @buffalojoec Just wanted to point out that "Const Type Parameters" is a feature that was introduced in TypeScript 5.0. So that change officially says no to TypeScript 4 and below users.

I know we talked before about the fact that we're only planning to support TS5 but just wanted to make that explicit.

steveluscher: Oooh. More TypeScript docs for me to read. My favourite.

steveluscher: Uh, would this in any way have helped us to trim down the number of `getBlock()` overloads, @buffalojoec?

buffalojoec: Hmm this is an interesting feature. I don't think it helps the overloads specifically, but looking back at them right now, I think perhaps they can be written with better type functions to condense some of them.

buffalojoec: This could be really interesting for typed transactions though. You could get back a transaction wherein types are `const` therefore showing you things like actual addresses, etc.

PR #2185, opened 2024-02-24
refactor(experimental): errors: context: sort keys

buffalojoec:
Smh!

PR #2184, opened 2024-02-24
refactor(experimental): errors: remove cyclical dependency

buffalojoec:
In order to be able to add `SolanaError` to `@solana/addresses`, the
`devDependency` for `@solana/addresses` needs to be removed.

PR #2183, opened 2024-02-24
draft: account errors

buffalojoec:
DRAFT: NOT FOR MERGE: How's this looking?

I'm not sure if we want to stay consistent with any error enums from the SDK,
but here's a first brush at adding `SolanaError` codes to `@solana/accounts`.

If we are decoupling from the SDK, then providing pluralized errors might help
with this package.

PR #2182, opened 2024-02-24
refactor(experimental): errors: fix `README` link

buffalojoec:
Updates the links to the README in the `@solana/errors` package.

PR #2187, opened 2024-02-24
refactor(experimental): errors: `addresses` package

buffalojoec:
Adds custom `SolanaError` throws to the `@solana/addresses` package.


buffalojoec:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2187?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2187** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2187?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2186** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2186?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2185** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2185?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2184** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2184?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2182** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2182?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#2183](https://github.com/solana-labs/solana-web3.js/pull/2183) <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2183?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `master`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2187?utm_source=stack-comment).

packages/errors/src/context.ts
```diff
@@ -33,9 +35,15 @@ export type SolanaErrorContext = DefaultUnspecifiedErrorContextToUndefined<{
         addresses: string[];
         cause?: Error;
     };
+    [SOLANA_ERROR__INVALID_ADDRESS]: {
```
lorisleiva, line 38: I think we should also provide the invalid address in the context?

packages/errors/src/context.ts
```diff
@@ -33,9 +35,15 @@ export type SolanaErrorContext = DefaultUnspecifiedErrorContextToUndefined<{
...
+        reason?: string;
```
lorisleiva, line 39: I'd be interested to hear @steveluscher's opinion on error "sub-categories" which lead us to have this optional `reason` string here.

packages/errors/src/context.ts
```diff
@@ -33,9 +35,15 @@ export type SolanaErrorContext = DefaultUnspecifiedErrorContextToUndefined<{
...
+    };
     [SOLANA_ERROR__INVALID_KEYPAIR_BYTES]: {
         byteLength: number;
     };
+    [SOLANA_ERROR__INVALID_SEEDS]: {
```
lorisleiva, line 44: I think `SOLANA_ERROR__INVALID_PROGRAM_DERIVED_ADDRESS` would be more appropriate.

lorisleiva (changes requested):

packages/addresses/src/address.ts
```diff
@@ -55,18 +56,22 @@ export function assertIsAddress(putativeAddress: string): asserts putativeAddres
             // Highest address (32 bytes of 255)
             putativeAddress.length > 44
         ) {
-            throw new Error('Expected input string to decode to a byte array of length 32.');
+            throw new SolanaError(SOLANA_ERROR__INVALID_ADDRESS, {
```
steveluscher, line 59: ‘Invalid’ can mean _so_ many things where an address is concerned, but ‘malformed’ would be more to the point here.

packages/addresses/src/program-derived-address.ts
```diff
@@ -45,14 +46,14 @@ export function assertIsProgramDerivedAddress<TAddress extends string = string>(
...
+        throw new SolanaError(SOLANA_ERROR__INVALID_SEEDS, {
+            reason: `Expected given program derived address to have the following format: [Address, ProgramDerivedAddressBump].`,
+        });
     }
     if (value[1] < 0 || value[1] > 255) {
-        // TODO: Coded error.
-        throw new Error(`Expected program derived address bump to be in the range [0, 255], got: ${value[1]}.`);
+        throw new SolanaError(SOLANA_ERROR__INVALID_SEEDS, {
+            reason: `Expected program derived address bump to be in the range [0, 255], got: ${value[1]}.`,
+        });
```
steveluscher, line 56: This isn't true, right? This is rather that the `ProgramDerivedAddress` is malformed. We don't really know about the seeds – just that the `ProgramDerivedAddress` isn't of the form `[Address, Bump]`

packages/errors/src/context.ts
steveluscher: Let's not.

1. Reasons are not scrutable like error codes are, without matching on the _text_ of the `reason`
2. There's a chance someone will surface this message to the user (bad; English only)

packages/errors/src/context.ts
steveluscher: †`MALFORMED`

steveluscher (changes requested):
I'd rather have these all as separate error codes. You can always catch the `SolanaError` without caring what type it is.

```ts
try {
    const pda = await createProgramDerivedAddress(/* ... */);
} catch(e) {
    if (isSolanaError(e)) {
        // bad thing
    } else {
        throw e;
    }
}
```

steveluscher:
I see that this sub-error thing happens in later PRs too. Listen, one thing we _could_ do is to to make use of [`Error#cause`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause). We'd have to extend `SolanaError` somehow, maybe by adding `cause` to the context.

```ts
throw new SolanaError(SOLANA_ERROR__PDA_SUCKS, {
    cause: new SolanaError(
        SOLANA_ERROR__PDA_SPECIFICALLY_SUCKS_BECAUSE_OF_THIS,
        { /* context for cause */ },
    ),
    /* other context */
});
```

Then people can be like:

```ts
try {
    const pda = await createProgramDerivedAddress(/* ... */);
} catch (e) {
    if (isSolanaError(e, SOLANA_ERROR__PDA_SUCKS)) {
        const why = e.cause;
        // Do something with `cause`, or not.
    } else {
        throw e;
    }
}
```

I don't love it though. If we can get away without doing that, we probably should. If there are good use cases for sub-errors, then we should lean on `cause`.

Notes:

* You'll have to enable `ES2022.Error` in `tsconfig.json` to be able to use `cause`
* `super(message, { cause });`
* TypeScript typing will probably be super tough; you'll have to delete the `cause` property from `context` and write some type tests.

buffalojoec:
Alright I think I got the message, guys.

steveluscher (approved):
Added via Giphy

packages/errors/src/context.ts
```diff
@@ -38,9 +44,22 @@ export type SolanaErrorContext = DefaultUnspecifiedErrorContextToUndefined<{
...
+    };
     [SOLANA_ERROR__INVALID_KEYPAIR_BYTES]: {
         byteLength: number;
     };
+    [SOLANA_ERROR__MAX_NUMBER_OF_PDA_SEEDS_EXCEEDED]: {
+        maxSeeds: number;
+    };
+    [SOLANA_ERROR__MAX_PDA_SEED_LENGTH_EXCEEDED]: {
+        index: number;
+        maxSeedLength: number;
```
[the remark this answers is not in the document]
buffalojoec: > Also, it makes a value the maximum seed length that probably won't change in our lifetimes, at least not without a code change being necessary in web3.js.

Yeah, but I really didn't want to have to introduce two areas where we'd have to change that value, if we ever needed to update the code.

> On the other hand, inlining the seed into the error is maybe a bad idea (privacy?).

Probably safe to avoid this, yeah.

> I'm struggling to think of what this error context could contain that would actually help me fix the bug. Maybe the actual length of the violating seed is a start?

Index and actual length should do, no? I doubt we need more than that.

packages/addresses/src/address.ts
```diff
@@ -55,19 +61,21 @@ export function assertIsAddress(putativeAddress: string): asserts putativeAddres
             // Highest address (32 bytes of 255)
             putativeAddress.length > 44
         ) {
-            throw new Error('Expected input string to decode to a byte array of length 32.');
+            throw new SolanaError(SOLANA_ERROR__INCORRECT_BASE58_ADDRESS_LENGTH, {
+                actualLength: putativeAddress.length,
+            });
```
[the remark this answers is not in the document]
buffalojoec: I don't think this is a bad check, and the code is informative, but I can see how the two errors could be merged. I'm not really a huge fan of `actualLength: 'UNKNOWN'`, and I certainly don't think we should decode the string just to return the _real_ actual byte length, which could invoke another unrelated error.

packages/addresses/src/address.ts
```diff
@@ -55,19 +61,21 @@ export function assertIsAddress(putativeAddress: string): asserts putativeAddres
...
         }
         // Slow-path; actually attempt to decode the input string.
         const base58Encoder = getMemoizedBase58Encoder();
         const bytes = base58Encoder.encode(putativeAddress);
         const numBytes = bytes.byteLength;
         if (numBytes !== 32) {
-            throw new Error(`Expected input string to decode to a byte array of length 32. Actual length: ${numBytes}`);
+            throw new SolanaError(SOLANA_ERROR__INCORRECT_BASE58_ADDRESS_BYTE_LENGTH, {
```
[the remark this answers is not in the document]
buffalojoec: It really only makes sense to include the `_BASE58` to match the pattern of the error code above it, if it remains in the set.

packages/addresses/src/address.ts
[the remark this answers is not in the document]
steveluscher: `SOLANA_ERROR__WHERE_DID_YOU_EVEN_GET_THIS`

packages/errors/src/context.ts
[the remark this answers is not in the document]
buffalojoec: Btw, the `MAX_SEEDS` error gets thrown by `createProgramAddress` (also the case in the Rust library as well). This can be confusing to the downstream user, since if they provide 16 seeds (`MAX_SEEDS`), they'll actually be feeding `createProgramAddress` with 17 seeds, because the bump is included, and they'll get this error.

With the changes I have locally right now, you'd basically get a message "provided: 17" when you can visibly see you only provided 16. Is a descriptive message good enough to elaborate this fact? Or do we want to work out the "max seeds" concept to actually be 15?

steveluscher:
### Merge activity

* **Feb 29, 11:33 AM PST**: @steveluscher started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2187).

PR #2192, opened 2024-02-24
refactor(experimental): errors: `codecs-strings` package

buffalojoec:
Adds custom `SolanaError` throws to the `@solana/codecs-strings` package.


steveluscher (changes requested):
Added via Giphy

packages/errors/src/messages.ts
```diff
@@ -84,6 +85,7 @@ export const SolanaErrorMessages: Readonly<{
         'Invalid data enum variant. Expected one of [$variants], got $value',
     [SOLANA_ERROR__CODECS_INVALID_SCALAR_ENUM_VARIANT]:
         'Invalid scalar enum variant. Expected one of [$variants] or a number between $minRange and $maxRange, got $value',
+    [SOLANA_ERROR__CODECS_INVALID_STRING_FOR_BASE]: 'Expected a string of base $expectedBase, got $value.',
```
steveluscher, line 88: This is totally fine, but the root cause of the error is _actually_ more like ‘this string contains a character that's not in this alphabet.’ You could _theoretically_ have a ‘base 58’ number that's based on a different alphabet than `123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz`.

I think it's more to the point to be like ‘hey, this string has characters that aren't in this alphabet; allowable characters are \`${alphabet}\`’.

steveluscher (approved):
Added via Giphy

steveluscher:
### Merge activity

* **Feb 29, 11:33 AM PST**: @steveluscher started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2192).

PR #2212, opened 2024-02-28
Add functions to narrow instructions to `IInstructionWithAccounts` and `IInstructionWithData`

mcintyre94:
Our `IInstruction` type may not include accounts/data. When we serialize a transaction we only know that it has `instructions: IInstruction[]`

The generated Kinobi clients sensibly require data to identify eg a system instruction:

```ts
export function identifySplSystemInstruction(
  instruction: { data: Uint8Array } | Uint8Array
)
```

And accounts and data to parse one:

```ts
export function parseTransferSolInstruction<
  TProgram extends string,
  TAccountMetas extends readonly IAccountMeta[]
>(
  instruction: IInstruction<TProgram> &
    IInstructionWithAccounts<TAccountMetas> &
    IInstructionWithData<Uint8Array>
)
```

This PR adds functions to narrow the type of an `IInstruction` to `IInstruction & IInstructionWithAccounts` and `IInstruction & IInstructionWithData`

This will allow code using generated clients to do eg:

```ts
// instruction: IInstruction
if(isInstructionWithAccounts(instruction) && isInstructionWithData(instruction)) {
  const identified = identifySplSystemInstruction(instruction);
  const parsed = parseTransferSolInstruction(instruction)
}
```

Notes:

- I've added new coded errors, please LMK if I did that wrong!
- I tweaked the error message formatter to handle null/undefined as context values, since it's useful for a message identifying the instruction at fault to use these.


packages/errors/src/message-formatter.ts
```diff
@@ -16,6 +16,10 @@ function encodeValue(value: unknown): string {
         );
     } else if (typeof value === 'bigint') {
         return `${value}n`;
+    } else if (value === null) {
+        return 'null';
+    } else if (value === undefined) {
+        return 'undefined';
```
mcintyre94, line 22: These special cases are required because the `else` block calls `Object.getPrototypeOf(value)` which is a type error for these values:

```node
> Object.getPrototypeOf(null)
Uncaught TypeError: Cannot convert undefined or null to object
    at Function.getPrototypeOf (<anonymous>)
> Object.getPrototypeOf(undefined)
Uncaught TypeError: Cannot convert undefined or null to object
    at Function.getPrototypeOf (<anonymous>)
```

I was hitting this in my instructions test when it throws the error with undefined accounts/data.

packages/errors/src/messages.ts
```diff
@@ -23,6 +25,10 @@ export const SolanaErrorMessages: Readonly<{
 }> = {
     [SOLANA_ERROR__BLOCK_HEIGHT_EXCEEDED]:
         'The network has progressed past the last block for which this transaction could have been committed.',
+    [SOLANA_ERROR__EXPECTED_INSTRUCTION_TO_HAVE_ACCOUNTS]:
+        'The instruction with program address `$programAddress` and data `$data` does not have any accounts.',
+    [SOLANA_ERROR__EXPECTED_INSTRUCTION_TO_HAVE_DATA]:
+        'The instruction with program address `$programAddress` and accounts `$accountAddresses` does not have any data.',
```
lorisleiva, line 31: Since we don't know for sure that an account without data has accounts and vice versa, I wouldn't explicitly add them in the error message. I think it's fine to have in the context if someone wants to check their values but would be confusing in the error message for instruction that have neither.

packages/errors/src/codes.ts
```diff
@@ -13,6 +13,8 @@ export const SOLANA_ERROR__INVALID_KEYPAIR_BYTES = 4 as const;
 export const SOLANA_ERROR__BLOCK_HEIGHT_EXCEEDED = 5 as const;
 export const SOLANA_ERROR__NONCE_INVALID = 6 as const;
 export const SOLANA_ERROR__NONCE_ACCOUNT_NOT_FOUND = 7 as const;
+export const SOLANA_ERROR__EXPECTED_INSTRUCTION_TO_HAVE_ACCOUNTS = 70 as const;
+export const SOLANA_ERROR__EXPECTED_INSTRUCTION_TO_HAVE_DATA = 71 as const;
```
lorisleiva, line 17: Trying to avoid merge conflict wars are we? 😄

lorisleiva (approved):
A few nits but I think this is valuable.

packages/errors/src/messages.ts
mcintyre94: Ah that's what I was missing, I thought context was just about getting values into the error message but it's displayed separately too. Agreed, I'll remove the accounts/data from the messages and just include them in context.

packages/errors/src/codes.ts
mcintyre94: I saw your feedback PR had errors up to 51, and then Steve's latest jumped to 99901, so I figured I'd go somewhere in the middle 😄

steveluscher: I've just resigned myself to rebasing on top of everyone else's numbers. I don't mind.

packages/errors/src/message-formatter.ts
buffalojoec: I believe I have a fix for this in my stack with just `value != null`.

buffalojoec: [packages/errors/src/message-formatter.ts](https://github.com/solana-labs/solana-web3.js/blob/6cfd2d2d66972626c08f67e87c6ecfb8077c3111/packages/errors/src/message-formatter.ts#L22)

mcintyre94: Clever! TIL `undefined != null`. Any idea what else that catches?

buffalojoec: I originally had `value != null && value != undefined` but Steve pointed out that `!=` will actually evaluate both, as compared to `!==`. Seems it's because an `undefined` value with a single equals will evaluate the conditional as "falsy".

mcintyre94: Removed, Steve merged #2223 separately

packages/errors/src/__tests__/message-formatter-test.ts
```diff
@@ -94,5 +94,31 @@ describe('getErrorMessage', () => {
...
+                // @ts-expect-error Mock error config doesn't conform to exported config.
+                123: 'Here is a variable: $variable',
+            });
+            const message = getErrorMessage(
+                // @ts-expect-error Mock error context doesn't conform to exported context.
+                123,
+                { variable: undefined },
+            );
+            expect(message).toBe('Here is a variable: undefined');
+        });
```
steveluscher, line 122: borat-borat-very-nice

steveluscher (approved):
giphy

Issue #2118, opened 2024-02-15
[experimental] Convert all thrown errors to `SolanaError`


steveluscher:
1. Search everywhere except for `packages/library-legacy` for `throw new Error`
2. Convert the error to a coded `SolanaError`
    a. Create a code for it in `packages/errors/src/codes.ts`
    b. Optionally allow the caller to add typed context to the error by defining the type in `packages/errors/src/context.ts`
    c. Create a message for it in `packages/errors/src/messages.ts`
    d. Replace the callsite with `throw new SolanaError(SOLANA_ERROR__MY_NEW_ERROR, {/* optional context */})`
3. Make sure that the generated Kinobi clients throw `ProgramErrors` from on-chain programs in the same way (include the logs, message, program address, error code, etc in the error's context)

buffalojoec:
I want you guys to know I cried when I self-assigned this.

lorisleiva:
Thank you @buffalojoec 🫡

I can tackle the errors inside the `@solana/programs` package since the error resolution will have to change anyway such that we don't need to provide the array of programs as we do now.

lorisleiva:
@buffalojoec Actually you might also want to avoid tackling all RPC errors right now as I'm working on a massive refactoring stack.

buffalojoec:
> @buffalojoec Actually you might also want to avoid tackling all RPC errors right now as I'm working on a massive refactoring stack.

You got it, just throw a flag here whenever I'm good to go.

steveluscher:
We can split this up if you want! Want to work from A-Z and I'll work from Z-A @buffalojoec?

buffalojoec:
> We can split this up if you want! Want to work from A-Z and I'll work from Z-A @buffalojoec?

Sure! Are you just basing off of Loris's stack?

PR #2224, opened 2024-02-29
Convert errors in `@solana/signers` to coded exceptions

steveluscher:
Addresses #2118.

PR #2225, opened 2024-03-01
Refine `Address` error types

steveluscher:
#2187 buried these address sub errors in such a way that they'd never reach the caller. This PR allows them to bubble up.

Addresses #2118.


steveluscher:
* **#2224** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2224?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2225** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2225?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2225?utm_source=stack-comment).

packages/addresses/src/address.ts
```diff
@@ -53,29 +52,25 @@ export function isAddress(putativeAddress: string): putativeAddress is Address<t
...
+    }
+    // Slow-path; actually attempt to decode the input string.
+    const base58Encoder = getMemoizedBase58Encoder();
+    const bytes = base58Encoder.encode(putativeAddress);
+    const numBytes = bytes.byteLength;
+    if (numBytes !== 32) {
+        throw new SolanaError(SOLANA_ERROR__ADDRESS_BYTE_LENGTH_OUT_OF_RANGE, {
+            actualLength: numBytes,
+        });
     }
```
steveluscher, line 79: @buffalojoec, I missed this in #2187. This _used_ to aim to create an ‘all up’ error that the address is bad, with the inner error as the `cause`. Now that we have codes and separate errors for each cause, I don't think we need this.

Also, #2187 buried the sub errors so that they'd never reach the caller.

packages/errors/src/codes.ts
```diff
@@ -16,8 +16,8 @@ export const SOLANA_ERROR__MULTIPLE_ACCOUNTS_NOT_FOUND = 9 as const;
 export const SOLANA_ERROR__FAILED_TO_DECODE_ACCOUNT = 10 as const;
 export const SOLANA_ERROR__EXPECTED_DECODED_ACCOUNT = 11 as const;
 export const SOLANA_ERROR__NOT_ALL_ACCOUNTS_DECODED = 12 as const;
-export const SOLANA_ERROR__INCORRECT_BASE58_ADDRESS_LENGTH = 13 as const;
-export const SOLANA_ERROR__INCORRECT_BASE58_ADDRESS_BYTE_LENGTH = 14 as const;
+export const SOLANA_ERROR__ADDRESS_BYTE_LENGTH_OUT_OF_RANGE = 13 as const;
+export const SOLANA_ERROR__ADDRESS_STRING_LENGTH_OUT_OF_RANGE = 14 as const;
```
steveluscher, line 20: Massaged these names while I was here. tl;dr I'm converting `blockhash.ts` and it's exactly the same as `address.ts`.

packages/addresses/src/address.ts
buffalojoec: Ahh, I see. Nice catch! So can't `NOT_A_BASE_58_ADDRESS` come out then?

buffalojoec (approved):
very nice

packages/addresses/src/address.ts
steveluscher: That one would have bubbled up, yes, but the others would never have seen the light of day.

I think in other PRs we consolidated on throwing granular errors, rather than one error with a `cause`.

PR #2233, opened 2024-03-01
Don't check npm version of library-legacy before publishing

### LINKED ISSUE #2230: Legacy library publish is halted
This line tries to look up the about-to-be-published version on npm to decide whether the publish can be skipped or not.

https://github.com/solana-labs/solana-web3.js/blame/f054d90bb27d759c114bd53dee36de91551c9eba/packages/library-legacy/package.json#L47

What actually happens is this:

1. The version – governed by Semantic Release – isn't incremented in the `prepublishOnly` step
2. The lookup is unconditionally done for `@solana/web3.js@0.0.0-development`
3. That version _actually is_ found on npm, albeit as a deprecated version

Because the version will always be found, the legacy package will never publish.

mcintyre94:
The latest CI change stopped `library-legacy` from publishing, because we don't use `version-from-git` so the version being checked was incorrect.

This PR removes the `npm view` check for library-legacy. We rely entirely on `semantic-release` to determine whether and how to publish the version, as we did before any CI changes. This shouldn't cause any issues because there are no packages that depend on library-legacy, so if `semantic-release` errors in some cases then it won't affect publishing anything else - which is why we added this `npm view` check in the first place. 

Fixes #2230


steveluscher (approved):
Added via Giphy

PR #2234, opened 2024-03-01
Add a function to narrow an IInstruction to have a given program address

mcintyre94:
This is another function that is helpful for using generated clients.

The generated client includes a function to parse an `IInstruction`:
```ts
export function parseTransferSolInstruction<
  TProgram extends string,
  TAccountMetas extends readonly IAccountMeta[]
>(
  instruction: IInstruction<TProgram> & <snip>
): ParsedTransferSolInstruction<TProgram, TAccountMetas>
```

The problem is that if we pass an `IInstruction` from a deserialized transaction to this parse function, then it has `TProgram = string`, and we end up with `ParsedTransferSolInstruction<TProgram = string>`, ie we lose the known program address in the type. This can cause issues when we use this type and expect it to include a more specific type.

This PR adds a function to narrow an instruction type to include a known program address as `TProgram`. If we do this before calling eg this parse function, then we can maintain the program address in the type correctly.

Eg:
```ts
const instruction: IInstruction<string> // from eg a deserialized transaction
const address = '1234' as Address

if(instruction.programAddress === address) {
  const parsed = parseInstruction(instruction); // ParsedInstruction<string>
}

if(isInstructionWithKnownProgramAddress(instruction, address)) {
  const parsed = parseInstruction(instruction); // ParsedInstruction<'1234'>
}
 ```

### CODE REVIEW

File: packages/instructions/src/instruction.ts [hunk 1]
```diff
@@ -21,6 +22,28 @@ export interface IInstructionWithAccounts<TAccounts extends readonly (IAccountMe
     readonly accounts: TAccounts;
 }
 
+export function isInstructionWithKnownProgramAddress<TProgramAddress extends string, TInstruction extends IInstruction>(
+    instruction: TInstruction,
+    programAddress: Address<TProgramAddress>,
+): instruction is TInstruction & IInstruction<Address<TProgramAddress>> {
```
lorisleiva, line 28: I would narrow the intersected type here to:

```suggestion
): instruction is TInstruction & { programId: Address<TProgramAddress> } {
```

packages/errors/src/messages.ts
```diff
@@ -25,6 +25,7 @@ import {
     SOLANA_ERROR__EXPECTED_INSTRUCTION_TO_HAVE_ACCOUNTS,
     SOLANA_ERROR__EXPECTED_INSTRUCTION_TO_HAVE_DATA,
     SOLANA_ERROR__FAILED_TO_DECODE_ACCOUNT,
+    SOLANA_ERROR__INCORRECT_INSTRUCTION_PROGRAM_ID,
```
lorisleiva, line 28: nit: I'd like us to use consistent language when applicable. `INCORRECT_` isn't part of the [recommended guidelines](https://github.com/solana-labs/solana-web3.js/blob/ede7241966388f12ad312ea0215180a2a271d7fe/packages/errors/src/codes.ts#L9-L24) but `INVALID_` is. What do you think?

mcintyre94: I'm not sure about invalid here semantically because the unexpected program address isn't necessarily invalid, it's just not the expected one. What about `INSTRUCTION_PROGRAM_ID_MISMATCH`?

lorisleiva: I like that!

packages/instructions/src/__typetests__/instruction-typetest.ts
```diff
@@ -44,3 +48,20 @@ import {
         IInstructionWithAccounts<readonly (IAccountMeta | IAccountLookupMeta)[]> &
         IInstructionWithData<Uint8Array>;
 }
+
+// narrowing to known address
+{
+    const instruction = {} as unknown as IInstruction;
+    const myAddress = '1111' as Address<'1111'>;
+    type MyAddress = typeof myAddress;
```
steveluscher, line 56: I think this is meant to be `type MyAddress = '1111'` otherwise you end up with `IInstruction<MyAddress>` producing `Address<Address<'1111'>>`

steveluscher (approved):
I sort of prefer something like `isInstructionForProgram` rather than `isInstructionWithKnownProgramAddress`, but no big deal either way! I just think it's less clear what it means to be ‘with’ a program address.

packages/instructions/src/__typetests__/instruction-typetest.ts
mcintyre94: This was actually caused by a typo when I tweaked the function

With that fixed both work correctly:
```ts
instruction satisfies IInstruction<Address<'1111>>;
instruction satisfies IInstruction<'1111'>;
```

steveluscher (approved):

PR #2235, opened 2024-03-02
`SolanaError` now supports the `cause` property

steveluscher:
# Summary

The runtime has built-in support for ‘sub exceptions.’ You can read about it here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause

> It is used when catching and re-throwing an error with a more-specific or useful error message in order to still have access to the original error.

This PR brings `cause` to `SolanaError` so that we can use it to rethrow errors under one umbrella error. Developers will have access to the inner error via `SolanaError#cause`.

The first error that will make use of this is the websocket close error, in the next PR.

# Test Plan

```shell
```

### DIFF
--- a/packages/errors/tsconfig.json
+++ b/packages/errors/tsconfig.json
@@ -4,6 +4,7 @@
         "lib": [
             "DOM",
             "ES2015",
+            "ES2022.Error"
         ],
         "resolveJsonModule": true
     },
--- a/packages/errors/src/__tests__/error-test.ts
+++ b/packages/errors/src/__tests__/error-test.ts
@@ -4,22 +4,75 @@ import { getErrorMessage } from '../message-formatter';
 jest.mock('../message-formatter');
 
 describe('SolanaError', () => {
-    let error123: SolanaError;
-    beforeEach(() => {
-        error123 = new SolanaError(
-            // @ts-expect-error Mock error codes don't conform to `SolanaErrorCode`
-            123,
-            { foo: 'bar' },
-        );
+    describe('given an error with context', () => {
+        let errorWithContext: SolanaError;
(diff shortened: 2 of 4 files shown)


steveluscher:
* **#2236** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2236?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2235** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2235?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

lorisleiva (approved):
I think this makes sense and it's one step "closer to the platform".

packages/errors/tsconfig.json
```diff
@@ -4,6 +4,7 @@
         "lib": [
             "DOM",
             "ES2015",
+            "ES2022.Error"
```
buffalojoec, line 7: Does this mean people downstream would have to include this in their `tsconfig.json` as well?

buffalojoec (approved):

packages/errors/tsconfig.json
steveluscher: Yes, but if they don't have it in their tsconfig already then they're not missing it (ie. they don't have `cause` in their typespec for _any_ `Errors`).

steveluscher:
### Merge activity

* **Mar 3, 7:02 PM PST**: @steveluscher started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2235).

PR #2242, opened 2024-03-04
Create coded exceptions for invariant violations

steveluscher:
# Summary

Invariant violations are a kind of error that should never get hit ever, and if they do your program is broken.

In one sense we don't want error codes for them because they're not intended to be caught in downstream programs. On the other hand I've tried to come up with a reason _not_ to give them codes and I can't.

Making them a `@solana/error` gives us compression and error decoding for free.

Addresses #2118.


steveluscher:
* **#2242** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2242?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

steveluscher:
Add link to issue tracker

packages/rpc-types/src/commitment.ts
```diff
@@ -9,9 +11,9 @@ function getCommitmentScore(commitment: Commitment): number {
         case 'processed':
             return 0;
         default:
-            return ((_: never) => {
-                throw new Error(`Unrecognized commitment \`${commitment}\`.`);
-            })(commitment);
+            throw new SolanaError(SOLANA_ERROR__INVARIANT_VIOLATION_SWITCH_MUST_BE_EXHAUSTIVE, {
+                unexpectedValue: commitment satisfies never,
+            });
```
buffalojoec, line 16: This one smells a little bit. I guess the only way this gets thrown is if someone isn't using TypeScript (or ignores it) and sends a string that isn't a valid `Commitment`. Shouldn't the error say that, then? ie. `SOLANA_ERROR__INVALID_COMMITMENT_VALUE`?

steveluscher: Yeah, I was trying to make something that was generally usable for exhaustive switches. The error location will tell the developer where the unhandled value was thrown, which I figured was enough paired with the actual unexpected value.

steveluscher: https://dev.to/cefn/typescript-satisfies-never-exhaustiveness-checking-in-typescript-49-58fh

lorisleiva: I do actually like that we have utility errors for these sort of invariants.

lorisleiva (approved):

steveluscher:
### Merge activity

* **Mar 4, 3:29 PM PST**: @steveluscher started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2242).

PR #2258, opened 2024-03-05
[vote-program] adding instruction update validator identity

ochaloup:
I would like to use the `UpdateValidatorIdentity` instruction of VoteProgram in TS code and it is not available in the current web3.js library.
https://github.com/solana-labs/solana/blob/v1.18.4/sdk/program/src/vote/instruction.rs#L381

### DIFF
--- a/packages/library-legacy/src/programs/vote.ts
+++ b/packages/library-legacy/src/programs/vote.ts
@@ -551,7 +551,7 @@ export class VoteProgram {
   ): Transaction {
     if (params.lamports > currentVoteAccountBalance - rentExemptMinimum) {
       throw new Error(
-        'Withdraw will leave vote account with insufficient funds.',
+        'Withdraw will leave vote account with insuffcient funds.',
       );
     }
     return VoteProgram.withdraw(params);
--- a/packages/library-legacy/src/programs/vote.ts
+++ b/packages/library-legacy/src/programs/vote.ts
@@ -88,6 +88,15 @@ export type WithdrawFromVoteAccountParams = {
 };
 
 /**
+ * Update validator identity (node pubkey) vote account instruction params.
+ */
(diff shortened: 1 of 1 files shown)


packages/library-legacy/src/programs/vote.ts
```diff
@@ -535,9 +551,32 @@ export class VoteProgram {
   ): Transaction {
     if (params.lamports > currentVoteAccountBalance - rentExemptMinimum) {
       throw new Error(
-        'Withdraw will leave vote account with insuffcient funds.',
+        'Withdraw will leave vote account with insufficient funds.',
```
steveluscher, line 554: giphy

steveluscher (approved):
You're really going to like what we're working on next.

packages/library-legacy/src/programs/vote.ts
steveluscher, line 554:
```suggestion
        'Withdraw will leave vote account with insuffcient funds.',
```

steveluscher: https://github.com/solana-labs/solana-web3.js/commit/e587f8d08dd981afe4fbcfac731e14e95d275027

ochaloup:
> You're really going to like what we're working on next.

Thanks @steveluscher 

Could you tease me more details? Pretty interested :love_you_gesture: 
(btw. I was not able to get to finish my PR on vote accounts (https://github.com/solana-labs/solana-web3.js/pull/2013) and I trouble with stake accounts a bit as well. I would like potentially process through that while maybe not needed :-))

steveluscher:
> Could you tease me more details? Pretty interested 🤟

Autogenerated program clients compatible with the new web3.js are moments away. cc/ @lorisleiva

PR #2274, opened 2024-03-07
Publish the 2.0.0 line with `changesets` now

steveluscher:
# Summary

This PR aims to switch the experimental library over to Changesets for publishing.

Every commit to the experimental library will now create/update a changesets PR. When we want to release, we land that PR and Changesets does the following:

1. Increments the versions of everything in unison to `2.0.0-preview.{X+1}`
2. Commits the version numbers and the `CHANGELOG.md` files to the repo
3. Sets `hasChangesets` to `true` in the CI/CD GitHub Actions workflow, which causes our `publish-packages` script to publish the new versions.

# Test Plan

Have I tested this? Nope. The most I've done is to do this locally:

1. `pnpm changeset` – This is what the GitHub bot encourages you to do on PRs
2. `pnpm changeset version` – This is essentially what the Action in `cicd.yml` does

That looked good to me, but whether it will all work in production is anyone's guess.


steveluscher:
* **#2274** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2274?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

.changeset/config.json
```diff
@@ -0,0 +1,18 @@
...
+  "baseBranch": "master",
+  "bumpVersionsWithWorkspaceProtocolOnly": true,
+  "changelog": [
+    "@changesets/changelog-github",
+    {
+      "repo": "solana-labs/solana-web3.js"
+    }
+  ],
+  "commit": false,
+  "fixed": [["@solana/!({*-impl,build-scripts,crypto-impl,test-*,tsconfig,web3.js-legacy-sham})"]],
```
steveluscher, line 13: What we're saying here is ‘yeah, bump the version of everything, but leave the internal packages alone.’

.changeset/config.json
```diff
@@ -0,0 +1,18 @@
...
+  "ignore": ["@solana/web3.js"],
```
steveluscher, line 14: Look in `cicd.yml`. I left this thing publishing with Semantic Release.

.github/workflows/cicd.yml
```diff
@@ -47,3 +42,25 @@ jobs:
...
+
+      - name: Create Changesets Pull Request or Trigger an NPM Publish
+        id: changesets
+        uses: changesets/action@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Publish Experimental Library
+        if: steps.changesets.outputs.hasChangesets == 'false'
+        run: pnpm turbo publish-packages --concurrency=100% --filter=\!@solana/web3.js
```
steveluscher, line 66: Everything except for legacy.

.github/workflows/cicd.yml
```diff
@@ -47,3 +42,25 @@ jobs:
...
           publish_dir: ./packages/library-legacy/doc
+
+      - name: Configure NPM token
+        run: |
+          pnpm config set '//registry.npmjs.org/:_authToken' "${NPM_TOKEN}"
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Publish Legacy Library
+        run: pnpm turbo publish-packages --concurrency=100% --filter=@solana/web3.js
```
steveluscher, line 53: Only legacy.

.changeset/pre.json
```diff
@@ -0,0 +1,42 @@
...
+    "@solana/rpc-transport-http": "2.0.0-preview.1",
+    "@solana/rpc-types": "2.0.0-preview.1",
+    "@solana/rpc": "2.0.0-preview.1",
+    "@solana/signers": "2.0.0-preview.1",
+    "@solana/transaction-confirmation": "2.0.0-preview.1",
+    "@solana/transactions": "2.0.0-preview.1",
+    "@solana/web3.js": "0.0.0-development",
+    "@solana/web3.js-experimental": "2.0.0-preview.1",
+    "@solana/web3.js-legacy-sham": "1.0.0-preview.1",
+    "@solana/webcrypto-ed25519-polyfill": "2.0.0-preview.1"
```
steveluscher, line 39: All of these versions have been updated to `-preview.1`. The first publish we do will take it to `-preview.2` and we'll call that TP2. I _think_ what we'll do after TP2 is released is _not_ do any releases until TP3, and if we need a release then we'll do a [changeset snapshot](https://github.com/changesets/changesets/blob/main/docs/snapshot-releases.md).

.prettierignore
```diff
@@ -1,3 +1,4 @@
+.changeset/
```
steveluscher, line 1: There's some bug I think with Prettier and changesets not finding our local Prettier config. https://github.com/changesets/changesets/issues/1283#issuecomment-1981708259

packages/compat/package.json
```diff
@@ -76,7 +76,7 @@
         "@solana/addresses": "workspace:*",
         "@solana/keys": "workspace:*",
         "@solana/rpc-types": "workspace:*",
-        "@solana/web3.js": "workspace:*"
+        "@solana/web3.js": "^1"
```
steveluscher, line 79: The idea here is that we lock this to a public version so that Changesets leaves it alone. This OK, @mcintyre94?

.changeset/pre.json
steveluscher: Or maybe we could install this: https://github.com/seek-oss/changesets-snapshot

lorisleiva: Is it possible to have a pre-release number with two numbers? like `2.0.0-preview.2.42`.

lorisleiva (approved):
So excited about using changesets.

.changeset/pre.json
buffalojoec: I agree on continuous delivery. Surely there is a way to do what Loris is suggesting?

.changeset/config.json
steveluscher: > Is legacy-sham supposed to be internal?

Felt cute, might delete later. I'm no longer convinced it's going to be useful for any practical purpose.

steveluscher: > `crypto-impl` matches the first glob pattern

👀

.changeset/pre.json
steveluscher: Yep! That's snapshots. It'll make versions like `0.0.0-preview-20240307210610`. I'll install that in a follow up PR.

packages/compat/package.json
steveluscher: Because `@solana/compat` is and it's configured as `fixed`.

* https://github.com/changesets/changesets/pull/507
* https://github.com/changesets/changesets/issues/506
* https://github.com/changesets/changesets/blob/main/docs/fixed-packages.md#fixed-packages

steveluscher: I figured out a hack. If you specify the workspace as a relative path it gets ignored by `changeset version`.

.changeset/pre.json
steveluscher: For your amusement: https://github.com/solana-labs/solana-web3.js/commit/069e0006f493b83e842579fada1c983874f245b7#r139530996

PR #2213, opened 2024-02-28
Create coded errors to cover all `TransactionErrors` from the RPC

steveluscher:
# Summary

Wouldn't it be nice if you could `catch` _particular_ transaction errors in your application, like `BlockhashNotFound`, and choose the correct mitigation based on the type of transaction error?

In this PR, we introduce coded exceptions for each `TransactionError` returned from the RPC's `sendTransaction` method.

> [!NOTE]
> Because the RPC doesn't return structured errors or error codes, we had to break our own rules in this PR and hardcode a map between the error names and the code numbers. My [first crack](https://gist.github.com/steveluscher/aaa7cbbb5433b1197983908a40860c47#file-fml-ts-L12) at this employed a source code compression scheme that I later deemed too risky for the 250 gzipped bytes it saved. We might consider such a scheme in the future, especially since the next PR will add `InstructionError` to the mix.

# Test Plan

```shell
cd packages/errors
pnpm test:unit:browser
pnpm test:unit:node
```

Addresses #2118.

PR #2286, opened 2024-03-09
Replace `RpcError` with a coded exception

steveluscher:
# Summary

So wow. I completely missed these as part of #2118, and it turns out they're a really big deal and required a ton of changes.

There are:

* Errors that have enough data to format a message
* Errors that have no data, but also no context in the message
* Errors that have no data, but really should, because you can't format a message without it
* Preflight errors in which is nested a `TransactionError` (see #2213)

In this PR we create a helper that takes in the `RpcSimulateTransactionResult` from the RPC and reformats it as a coded `SolanaError`.

As always, everything you need to know is in the `packages/errors/src/__tests__/json-rpc-error-test.ts`.

# Test Plan

```
pnpm turbo test:unit:browser
pnpm turbo test:unit:node
```


steveluscher:
* **#2286** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2286?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/errors/src/__tests__/json-rpc-error.ts
```diff
@@ -0,0 +1,52 @@
...
+import { getSolanaErrorFromJsonRpcError } from '../json-rpc-error';
+import { getSolanaErrorFromTransactionError } from '../transaction-error';
+
+jest.mock('../transaction-error.ts');
+
+describe('getSolanaErrorFromJsonRpcError', () => {
+    it('produces a `SolanaError` given a JSON-RPC error with data', () => {
+        const expectedData = Symbol() as unknown as SolanaErrorContext[SolanaErrorCode];
+        const error = getSolanaErrorFromJsonRpcError({ code: 123, data: expectedData, message: 'o no' });
+        expect(error).toEqual(new SolanaError(123 as SolanaErrorCode, expectedData));
```
lorisleiva, line 13: I am sliiiiighly concerned about using RPC error codes as Solana error codes directly. Right now they are negative and cannot conflict with the rest of our codes but what if this change with RPC v2?

packages/errors/src/context.ts
```diff
@@ -120,8 +136,10 @@ import {
     SOLANA_ERROR__TRANSACTION_ERROR__UNKNOWN,
     SolanaErrorCode,
 } from './codes';
+import { RpcSimulateTransactionResult } from './json-rpc-error';
 
 type BasicInstructionErrorContext<T extends SolanaErrorCode> = Readonly<{ [P in T]: { index: number } }>;
+type BasicJsonRpcErrorContext<T extends SolanaErrorCode> = Readonly<{ [P in T]: { message: string } }>;
```
lorisleiva, line 142: Personally, I'd rather we parsed the message to grab a better context until we're able to get that information directly from the RPC. When that's fixed on the RPC layer, we can remove that parsing without breaking changes.

packages/errors/src/messages.ts
```diff
@@ -342,6 +363,30 @@ export const SolanaErrorMessages: Readonly<{
...
+    [SOLANA_ERROR__JSON_RPC__INVALID_REQUEST]: 'JSON-RPC error: The JSON sent is not a valid `Request` object',
+    [SOLANA_ERROR__JSON_RPC__METHOD_NOT_FOUND]: 'JSON-RPC error: The method does not exist / is not available',
+    [SOLANA_ERROR__JSON_RPC__PARSE_ERROR]:
+        'JSON-RPC error: An error occurred on the server while parsing the JSON text',
+    [SOLANA_ERROR__JSON_RPC__SCAN_ERROR]: '$message',
+    [SOLANA_ERROR__JSON_RPC__SERVER_ERROR_BLOCK_CLEANED_UP]: '$message',
+    [SOLANA_ERROR__JSON_RPC__SERVER_ERROR_BLOCK_NOT_AVAILABLE]: '$message',
+    [SOLANA_ERROR__JSON_RPC__SERVER_ERROR_BLOCK_STATUS_NOT_AVAILABLE_YET]: '$message',
+    [SOLANA_ERROR__JSON_RPC__SERVER_ERROR_KEY_EXCLUDED_FROM_SECONDARY_INDEX]: '$message',
+    [SOLANA_ERROR__JSON_RPC__SERVER_ERROR_LONG_TERM_STORAGE_SLOT_SKIPPED]: '$message',
```
lorisleiva, line 377: 😢

packages/errors/src/json-rpc-error.ts
```diff
@@ -0,0 +1,85 @@
...
+    unitsConsumed: number | null;
+}
+
+export function getSolanaErrorFromJsonRpcError({ code, data, message }: JsonRpcError): SolanaError {
+    if (code === -32002 /* JSON_RPC_SERVER_ERROR_SEND_TRANSACTION_PREFLIGHT_FAILURE */) {
+        const { err, ...preflightErrorContext } = (data ?? {}) as RpcSimulateTransactionResult;
+        return new SolanaError(SOLANA_ERROR__JSON_RPC__SERVER_ERROR_SEND_TRANSACTION_PREFLIGHT_FAILURE, {
+            ...(err ? { cause: getSolanaErrorFromTransactionError(err) } : null),
+            ...(preflightErrorContext as SolanaErrorContext[typeof SOLANA_ERROR__JSON_RPC__SERVER_ERROR_SEND_TRANSACTION_PREFLIGHT_FAILURE]),
+        });
```
lorisleiva, line 78: I'd love it if we could add some additional context data:

```ts
{
    failedInstruction: null | {
        index: number;
        programErrorCode: number;
        programAddress: number;
    }
}
```

Which we would parse ourselves here:

```ts
{
    // ...
   ...(isInstructionError(logs) ? parseInstructionError(logs) : null)
}
```

The logic here would most be the same as the one in `@solana/programs`.

packages/errors/src/__tests__/json-rpc-error.ts
steveluscher: They're ‘[reserved](https://www.jsonrpc.org/specification#error_object)’ by the JSON-RPC 2.0 specification, but I can always introduce an offset so that they're positive and unique within our `SolanaErrorCode` universe.

packages/errors/src/context.ts
steveluscher: We definitely can't do that.

* **Push-safety**: if I write a parser in `@solana/web3.js@X` and then the error message changes on the server, I won't be able to take back the version `X` clients that are already in the wild. They will just break.
* **Messages are non-standard**: Code numbers are core to the RPC specification. Error messages are not, and we don't want to elevate them to the level of spec.
  * Firedancer, Agave, and Sig are all likely to produce different error messages, and should be free to do so.
  * RPC implementers should be free to localize error messages in different languages.

Treating the messages themselves as non-opaque data will make the messages impossible to localize, to change, or to improve upon.

packages/errors/src/json-rpc-error.ts
steveluscher: I _think_ this already gives you that.

* Preflight fails. You get a preflight failed error
* The `cause` gets set to the result of `getSolanaErrorFromTransactionError()`
* Because the transaction error is _specifically_ an instruction error, it delegates to `getSolanaErrorFromInstructionError()`

So at the end you essentially get:

```ts
const e = new SolanaError(
    SOLANA_ERROR__JSON_RPC__SERVER_ERROR_SEND_TRANSACTION_PREFLIGHT_FAILURE,
    {
        cause: new SolanaError(
            SOLANA_ERROR__INSTRUCTION_ERROR__CUSTOM,
            { code: 123, index: 456 },
        ),
    },
);
```

You can decode the error using something like:

```ts
if (isSolanaError(e.cause, SOLANA_ERROR__INSTRUCTION_ERROR__CUSTOM)) {
    const { code, index } = e.cause.context;
    const { programAddress } = transaction.instructions[index];
    const customError = decodeThisShit(programAddress, code);
}
```

packages/errors/src/__tests__/json-rpc-error.ts
steveluscher: I made this change (ie. added an offset) and I think it made things way worse.

lorisleiva: No I think it's best without the offset. I was just saying we're merging two error code system together and right now it's okay because they don't overlap but are we worried that they might in the future?

packages/errors/src/json-rpc-error.ts
lorisleiva: Oh I see, that's great then!

packages/errors/src/context.ts
lorisleiva: Gotcha! Happy to disagree and commit here. 🫡

packages/errors/src/json-rpc-error.ts
```diff
@@ -0,0 +1,133 @@
...
+            case SOLANA_ERROR__JSON_RPC__SERVER_ERROR_KEY_EXCLUDED_FROM_SECONDARY_INDEX:
+            case SOLANA_ERROR__JSON_RPC__SERVER_ERROR_LONG_TERM_STORAGE_SLOT_SKIPPED:
+            case SOLANA_ERROR__JSON_RPC__SERVER_ERROR_SLOT_SKIPPED:
+            case SOLANA_ERROR__JSON_RPC__SERVER_ERROR_TRANSACTION_PRECOMPILE_VERIFICATION_FAILURE:
+            case SOLANA_ERROR__JSON_RPC__SERVER_ERROR_UNSUPPORTED_TRANSACTION_VERSION:
+                // The server supplies no structured data, but rather a pre-formatted message. Put
+                // the server message in `context` so as not to completely lose the data. The long
+                // term fix for this is to add data to the server responses and modify the
+                // messages in `@solana/errors` to be actual format strings.
+                errorContext = { __serverMessage: message };
```
steveluscher, line 120: Grumble.

packages/errors/src/json-rpc-error.ts
```diff
@@ -0,0 +1,133 @@
...
+                break;
+            default:
+                if (typeof data === 'object' && !Array.isArray(data)) {
+                    errorContext = data;
+                }
```
steveluscher, line 125: This protects against things like `data: 'abc'` turning into `{0: 'a', 1: 'b', 2: 'c'}`

packages/rpc-api/src/__tests__/send-transaction-test.ts
```diff
@@ -224,17 +246,53 @@ describe('sendTransaction', () => {
...
+                cause: new SolanaError(SOLANA_ERROR__TRANSACTION_ERROR__ACCOUNT_NOT_FOUND),
+                logs: [],
+                returnData: null,
+                unitsConsumed: 0,
+            }),
         );
-        await expect(resultPromise).rejects.toMatchObject({
-            code: -32002 satisfies (typeof SolanaRpcErrorCode)['JSON_RPC_SERVER_ERROR_SEND_TRANSACTION_PREFLIGHT_FAILURE'],
+    });
+    it('fatals when the fee payer has insufficient funds to pay rent', async () => {
```
steveluscher, line 259: Added this, because the _existing_ test wasn't actually testing anything about fees – only that the account _didn't exist_.

scripts/fixtures/send-transaction-fee-payer-insufficient-funds.json
```diff
@@ -0,0 +1,10 @@
+{
+    "pubkey": "6Zs91PMyhqyMgNVuT8EnM6f3YjVAVabvVWLjpFSC288q",
+    "account": {
+        "lamports": 1,
```
steveluscher, line 4: Added an account to test ‘not enough to pay …’ type tests.

lorisleiva (approved):
Thank you!

steveluscher:
### Merge activity

* **Mar 12, 10:58 AM PDT**: @steveluscher started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2286).

PR #2293, opened 2024-03-12
refactor(experimental): add resizeCodec helper to @solana/codecs-core

lorisleiva:
This PR adds a new `resizeCodec` helper function. See copy/pasted documentation below:

---

## Adjusting the size of codecs

The `resizeCodec` helper re-defines the size of a given codec by accepting a function that takes the current size of the codec and returns a new size. This works for both fixed-size and variable-size codecs.

```ts
// Fixed-size codec.
const getBiggerU32Codec = () => resizeCodec(getU32Codec(), size => size + 4);
getBiggerU32Codec().encode(42);
// 0x2a00000000000000
//   |       └-- Empty buffer space caused by the resizeCodec function.
//   └-- Our encoded u32 number.

// Variable-size codec.
const getBiggerStringCodec = () => resizeCodec(getStringCodec(), size => size + 4);
getBiggerStringCodec().encode('ABC');
// 0x0300000041424300000000
//   |             └-- Empty buffer space caused by the resizeCodec function.
//   └-- Our encoded string with a 4-byte size prefix.
```

Note that the `resizeCodec` function doesn't change any encoded or decoded bytes, it merely tells the `encode` and `decode` functions how big the `Uint8Array` should be before delegating to their respective `write` and `read` functions. In fact, this is completely bypassed when using the `write` and `read` functions directly. For instance:

```ts
const getBiggerU32Codec = () => resizeCodec(getU32Codec(), size => size + 4);

// Using the encode function.
getBiggerU32Codec().encode(42);
// 0x2a00000000000000

// Using the lower-level write function.
const myCustomBytes = new Uint8Array(4);
getBiggerU32Codec().write(42, myCustomBytes, 0);
// 0x2a000000
```

So when would it make sense to use the `resizeCodec` function? This function is particularly useful when combined with the `offsetCodec` function described below. Whilst the `offsetCodec` may help us push the offset forward — e.g. to skip some padding — it won't change the size of the encoded data which means the last bytes will be truncated by how much we pushed the offset forward. The `resizeCodec` function can be used to fix that. For instance, here's how we can use the `resizeCodec` and the `offsetCodec` functions together to create a struct codec that includes some padding.

```ts
const personCodec = getStructCodec([
    ['name', getStringCodec({ size: 8 })],
    // There is a 4-byte padding between name and age.
    [
        'age',
        offsetCodec(
            resizeCodec(getU32Codec(), size => size + 4),
            ({ preOffset }) => preOffset + 4,
        ),
    ],
]);

personCodec.encode({ name: 'Alice', age: 42 });
// 0x416c696365000000000000002a000000
//   |               |       └-- Our encoded u32 (42).
//   |               └-- The 4-bytes of padding we are skipping.
//   └-- Our 8-byte encoded string ("Alice").
```

As usual, the `resizeEncoder` and `resizeDecoder` functions can also be used to achieve that.

```ts
const getBiggerU32Encoder = () => resizeEncoder(getU32Codec(), size => size + 4);
const getBiggerU32Decoder = () => resizeDecoder(getU32Codec(), size => size + 4);
const getBiggerU32Codec = () => combineCodec(getBiggerU32Encoder(), getBiggerU32Decoder());
```


lorisleiva:
* **#2294** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2294?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2293** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2293?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

lithdew:
Hmm, some documentation on use cases for this codec would be good. I assume this would be useful for dealing with alignment/padding of i.e. struct/union fields?

If that is the case, I assume that every time we resize a codec, we would also want to do something to transform the input of the bytes fed into the codec as well (i.e. set the value of all extraneous bytes to some explicit padding constant to handle, for example, null-terminated strings).

lorisleiva:
@lithdew Good point on the documentation example I'll provide one.

We've actually already got a function that transforms the codec types by accepting mapping functions between these types. It's called [`mapCodec`](https://github.com/solana-labs/solana-web3.js/tree/master/packages/codecs-core#mapping-codecs). So if you wanted to resize and map the codec then you could use these two primitives together.

But actually the main reason I need `resizeCodec` right now is to cater for paddings inside structs. Say you're using bytemuck and you have 4 bytes of padding that represent nothing in your buffer for alignment. Then in your struct codec, you can use the `offsetCodec` helper to push the offset by 4 bytes but when encoding the data nothing tells the codec that the data should be 4 bytes longer. Now `resizeCodec` can help with that. Here's an example in the test:

https://github.com/solana-labs/solana-web3.js/blob/9b9a1336e60f31008ab6b2d7f16a57791d81d65f/packages/codecs-data-structures/src/__tests__/struct-test.ts#L62-L78

lithdew:
@lorisleiva Hmm, wouldn't `mapCodec` only support transforming the input value itself and not the encoded bytes of the input though? So i.e. how would I use `mapCodec` after expanding the number of bytes the codec takes in using `resizeCodec` to set the extra bytes values to a padding constant like `0xFF`?

lorisleiva:
Oh I see what you mean.

Yeah resizing doesn't affect any of the encoded or decoded bytes. In fact, it's only purpose is to know how big the `Uint8Array` should be when calling the `encode` and `decode` methods because, technically, anyone can use the `write` and `read` functions directly instead and provide their own `Uint8Array` with any size they want. Here's an example:

```ts
const codec = resizeCodec(getU32Codec(), size => size + 4);

codec.encode(42);
// 0x2a00000000000000

const myCustomBytes = new Uint8Array(4);
codec.write(42, myCustomBytes, 0);
// 0x2a000000
```

So I don't think `resizeCodec` should be responsible for changing any bytes or even have some concept of a padding constant, it's just here to expand or contract the size of a codec.

Now, say you wanted to describe a byte array that's composed of a `u32` followed by four `0xff` bytes, as you mentioned.
In this case, you wouldn't even need a `resizeCodec` call, you could just do the following:

```ts
const codec = mapCodec(
    getTupleCodec([getU32Codec(), getBytesCodec({ size: 4 })]),
    (value: number): [number, Uint8Array] => [value, new Uint8Array([255, 255, 255, 255])],
    ([value, bytes]: [number, Uint8Array]): number => value,
);
```

lithdew:
Ahh that makes sense - appreciate the code example 🙏.

lorisleiva:
I'm gonna update the docs accordingly because I can see why this can be confusing.

packages/codecs-core/src/__typetests__/resize-codec-typetest.ts
```diff
@@ -0,0 +1,77 @@
...
+    VariableSizeCodec,
+    VariableSizeDecoder,
+    VariableSizeEncoder,
+} from '../codec';
+import { resizeCodec, resizeDecoder, resizeEncoder } from '../resize-codec';
+
+type NumberToArray<N extends number, T extends unknown[] = []> = T['length'] extends N
+    ? T
+    : NumberToArray<N, [...T, unknown]>;
+type Increment<N extends number> = [...NumberToArray<N>, unknown]['length'];
```
steveluscher, line 17: Added via Giphy

packages/codecs-core/src/resize-codec.ts
```diff
@@ -0,0 +1,103 @@
...
+import { combineCodec } from './combine-codec';
+
+/**
+ * Updates the size of a given encoder.
+ */
+export function resizeEncoder<TFrom, TSize extends number, TNewSize extends number>(
+    encoder: FixedSizeEncoder<TFrom, TSize>,
+    resize: (size: TSize) => TNewSize,
+): FixedSizeEncoder<TFrom, TNewSize>;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
steveluscher, line 23: Damn.

I was able to get this to work without `any` by using `Encoder<never>`, `Decoder<unknown>` and `Codec<never, unknown>`, but then `Codec<never, unknown>` itself becomes a type error. Can't decide which is worse: this any suppression, or…

```ts
export function resizeCodec<
    TCodec extends Codec<
        never,
        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
        // @ts-expect-error
        unknown
    >,
>(codec: TCodec, resize: (size: number) => number): TCodec;
```

lorisleiva: Whilst `any` is a bad word in TypeScript, I do think it describes best the situation here — i.e. we accepts codecs for _any_ types and return you the exact same type.

buffalojoec (approved):
grow

lorisleiva:
### Merge activity

* **Mar 14, 6:08 AM EDT**: @lorisleiva started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2293).

PR #2294, opened 2024-03-12
refactor(experimental): add offsetCodec helper to @solana/codecs-core

lorisleiva:


lorisleiva:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2294?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2294** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2294?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2293** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2293?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

lithdew:
My initial thoughts are that it might be good to make the wrapping behavior of the pre/post offset optional, such that the codec will throw an error if the offset specified is greater or less than the size of the input.

Otherwise, everything else looks good.

lithdew:
One other comment is, what do you think about calling this codec `seekCodec` rather than `offsetCodec`?

lorisleiva:
@lithdew Thanks for the feedback! Tbh I went back and forth multiple times on whether to throw or modulo byte overflows.

The reason I went for the later is that it enables negative offsets so using `() => -4` will take you 4 bytes before the end of the buffer which can be super helpful.

An option could be nice though. So the API would look something like:

```ts
const u32InTheMiddleCodec = offsetCodec(biggerU32Codec, {
    preOffset: ({ preOffset }) => preOffset + 2,
    postOffset: ({ postOffset }) => postOffset + 2,
    wrapBytes: true, // false by default?
});
```

The only annoyance is it makes the simple `() => -4` example more verbose but maybe that's fine.

```ts
// Before.
offsetCodec(innerCodec, () => -4);

// After.
offsetCodec(innerCodec, {
    preOffset: () => -4,
    wrapBytes: true,
});
```

A conditional codec could be fun to implement as well yeah. I guess its logic would be fairly close to the `getOptionCodec` function.

I'm not sure I like `seekCodec` tbh because, for me, it sounds more like `findCodec` which doesn't really describe what's happening.

packages/codecs-core/src/__tests__/offset-codec-test.ts
```diff
@@ -0,0 +1,278 @@
...
+
+describe('offsetCodec', () => {
+    it.each(testCases)(
+        '$_it',
+        ({ preOffset, size, innerCodecSize, codec: codecFactory, expectedPostOffset, expectedPreOffset }) => {
+            const mockCodec = getMockCodec({ size });
+            mockCodec.write.mockImplementation((_value, _bytes, offset) => offset + innerCodecSize);
+            mockCodec.read.mockImplementation((bytes, offset) => [bytes, offset + innerCodecSize]);
+
+            const codec = codecFactory(mockCodec);
```
steveluscher, line 169: Inlining all of these test cases (as separate, hand-hewn artisanal `its()`) would reduce the debugging burden on the person who broke them. As it is, they're going to have to jump in and out of this code to figure out what the factory is, and what the expected values are.

packages/codecs-core/src/__tests__/offset-codec-test.ts
```diff
@@ -0,0 +1,278 @@
...
+            const hexBytes = '00'.repeat(size);
+
+            expect(codec.write(42, b(hexBytes), preOffset)).toBe(expectedPostOffset);
+            expect(mockCodec.write).toHaveBeenCalledWith(42, b(hexBytes), expectedPreOffset);
+
+            expect(codec.read(b(hexBytes), preOffset)[1]).toBe(expectedPostOffset);
+            expect(mockCodec.read).toHaveBeenCalledWith(b(hexBytes), expectedPreOffset);
```
steveluscher, line 176: This test exercises multiple expectations. Breaking something about the pre offset could break the test named “does not modulo the post-offset if it has reached the end of the buffer,” which is not a great situation to be in if you want to know what broke, why.

steveluscher (approved):
Added via Giphy

One day I'll convince you to write one expectation per test. And to proclaim your undying love for Jest. OK maybe just the first.

lorisleiva:
@steveluscher Following the conversation on this PR and talking to @joncinque about it, I'm gonna refactor the `offsetCodec` API slightly so it accepts a config object (see code in my previous message).

Instead of the `wrapBytes` boolean though I'm thinking of passing a `wrapBytes` function within the scope of both preOffset and postOffset functions.

So the API would look like this. What do you all think?

```ts
offsetCodec(innerCoder, {
    preOffset: ({ wrapBytes }) => wrapBytes(-4),
    postOffset: ({ preOffset }) => preOffset,
});
```

packages/codecs-core/src/__tests__/offset-codec-test.ts
lorisleiva: Thanks of the feedback. I've updated all the tests. Lmk what you think.

.eslintrc.js
```diff
@@ -1,4 +1,10 @@
 module.exports = {
     extends: ['turbo', '@solana/eslint-config-solana', '@solana/eslint-config-solana/jest'],
     root: true,
+    rules: {
+        'jest/expect-expect': [
+            'error',
+            { assertFunctionNames: ['expect', 'expectNewPreOffset', 'expectNewPostOffset'] },
+        ],
+    },
```
lorisleiva, line 9: @steveluscher Is this the best place to put this?

packages/codecs-core/src/offset-codec.ts
```diff
@@ -0,0 +1,83 @@
...
+            const wrapBytes = (offset: Offset) => modulo(offset, bytes.length);
+            const newPreOffset = config.preOffset ? config.preOffset({ bytes, preOffset, wrapBytes }) : preOffset;
+            assertByteArrayOffsetIsNotOutOfRange('offsetEncoder', newPreOffset, bytes.length);
+            const postOffset = encoder.write(value, bytes, newPreOffset);
+            const newPostOffset = config.postOffset
+                ? config.postOffset({ bytes, newPreOffset, postOffset, preOffset, wrapBytes })
+                : postOffset;
+            assertByteArrayOffsetIsNotOutOfRange('offsetEncoder', newPostOffset, bytes.length);
+            return newPostOffset;
+        },
```
buffalojoec, line 46: Small up-to-you nit: I think it's a little more clear if you plumb `offset` through this block, as defined by the `createEncoder` API. Same thing below with `offsetDecoder`. It makes it a bit clearer to see that you're using both `preOffset` and `postOffset` to slide the actual `offset` along.

buffalojoec (approved):
Lgtm! How does this bad boy work inside of an `ArrayDecoder`? Can I use an `OffsetDecoder` for each element in my array? In other words:

```ts
createArrayDecoder(
  createOffsetDecoder(/* .. */)
)
```

If so, not sure if you want to add some test coverage on arrays.

packages/codecs-core/src/offset-codec.ts
lorisleiva: Hmm tbh, as we're always introducing 4 new terms to describe offsets (`preOffset`, `newPreOffset`, `postOffset` and `newPostOffset`), I'd rather be extra-explicit here and avoid having a generic `offset` attribute.

lorisleiva:
@buffalojoec Yeah offsets will apply to each array items which can create some crazy use-cases haha. I'll add some array tests in sub-sequent PRs.

lorisleiva:
### Merge activity

* **Mar 14, 6:08 AM EDT**: @lorisleiva started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2294).

packages/codecs-core/src/offset-codec.ts
buffalojoec: A generic attribute isn't what I meant. But nbd, merged.

PR #2297, opened 2024-03-12
Use `@solana-program/system` in the Nonce utilties of `@solana/transactions`

### LINKED ISSUE #2127: [experimental] Replace implementation of `AdvanceNonce` instruction with that from the generated System program client
https://github.com/solana-labs/solana-web3.js/blob/b2c1b82d0da512ad485e9b81c2d9db12e3ce3588/packages/transactions/src/durable-nonce.ts#L73-L92

steveluscher:
This PR is not ready to land as is and probably won't make it into Technology Preview 2.

It suffers from a circular dependency between `@solana/transactions` and `@solana-program/system`, but on the basis of types only. The code compiles just fine, but TypeScript wont:

```shell
error TS5055: Cannot write file '/home/sol/src/solana-web3.js-git/packages/transactions/dist/types/accounts.d.ts' because it would overwrite input file.
error TS5055: Cannot write file '/home/sol/src/solana-web3.js-git/packages/transactions/dist/types/blockhash.d.ts' because it would overwrite input file.
error TS5055: Cannot write file '/home/sol/src/solana-web3.js-git/packages/transactions/dist/types/compilable-transaction.d.ts' because it would overwrite input file.
error TS5055: Cannot write file '/home/sol/src/solana-web3.js-git/packages/transactions/dist/types/compile-address-table-lookups.d.ts' because it would overwrite input file.
...
```

Additionally, there appear to be some problems with the conditional types in the raw instruction creator: https://github.com/solana-program/create-solana-program/issues/19

Addresses #2127.


steveluscher:
* **#2297** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2297?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

steveluscher:
This will never work because of the circular dependency it creates.

PR #2307, opened 2024-03-12
refactor(experimental): sysvars package: recent blockhashes

buffalojoec:
This commit introduces the `RecentBlockhashes` sysvar to the `@solana/sysvars` package.

### DIFF
--- a/packages/sysvars/src/recent-blockhashes.ts
+++ b/packages/sysvars/src/recent-blockhashes.ts
@@ -1,0 +1,75 @@
+import { assertAccountExists, decodeAccount, type FetchAccountConfig } from '@solana/accounts';
+import {
+    combineCodec,
+    getArrayDecoder,
+    getArrayEncoder,
+    getStructDecoder,
+    getStructEncoder,
+    type VariableSizeCodec,
+    type VariableSizeDecoder,
+    type VariableSizeEncoder,
+} from '@solana/codecs';
+import type { GetAccountInfoApi } from '@solana/rpc-api';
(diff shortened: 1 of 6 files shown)


packages/sysvars/src/recent-blockhashes.ts
```diff
@@ -0,0 +1,51 @@
...
+ * Information about recent blocks and their fee calculators.
+ */
+export type SysvarRecentBlockhashes = Entry[];
+
+export function getSysvarRecentBlockhashesCodec(): Codec<SysvarRecentBlockhashes> {
+    const lamportsCodec: Codec<LamportsUnsafeBeyond2Pow53Minus1> = mapCodec(
+        getU64Codec(),
+        val => val,
+        val => lamports(val),
+    );
```
lorisleiva, line 29: Since this is reused a few times, it might be good to extract it under the `@solana/rpc-types` package where the lamport type lives (with separate `Encoders` and `Decoders` ideally).

lorisleiva (approved):

packages/sysvars/src/recent-blockhashes.ts
buffalojoec: Yeah I was thinking the same thing, but this opens a box to adding a codec for every type in that library that uses `unique symbol`. If you think it should be in there I'll add it.

buffalojoec:
### Merge activity

* **Mar 14, 12:46 PM EDT**: @buffalojoec started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2307).

Issue #2289, opened 2024-03-12
[experimental] Validate that the public key generated from `createKeyPairFromBytes()` belongs to the private key


steveluscher:
## Motivation

Presently, when you pass 64 bytes of key material to `createKeyPairFromBytes()`, we don't validate that the public key part of the key pair actually belongs to the private key.

Signing a message using a public key _other_ than that which belongs to the private key can result in a [leak of the private key bytes](https://github.com/MystenLabs/ed25519-unsafe-libs?tab=readme-ov-file#what-is-the-issue) to someone.

This is generally not a problem with our APIs because they _only_ take the private key and message as an argument, and derive the public key internally; the worst thing that can happen is for verification of the resulting message to fail because you're holding a public key that doesn't correspond to the private key you signed the message with.

Someone who uses a mismatching key pair from `createKeyPairFromBytes()` with _other_ code though could end up in the scenario described above.

## Details

1. Throw unless the public key and private key produced by `createKeyPairFromBytes()` form a pair. Since there's no API in `SubtleCrypto` to derive an Ed25519 public key from a private key, the best we can do here is to sign a message with the private key and ensure that it verifies with the public key.
2. Offer a `createKeyPairFromBytes_DANGEROUSLY_SKIP_VALIDATION()` method that performs no validation.

PR #2329, opened 2024-03-16
[experimental] Validate that the public key generated from createKeyPairFromBytes() belongs to the private key 

luu-alex:
This PR addresses #2289 

updated `createKeyPairFromBytes`. Adds a validation check to verify the public and private key by signing data and verifying that the signed data matches the public key.

Added `createKeyPairFromBytes_DANGEROUSLY_SKIP_VALIDATION` that skips the validation above ^

Adds error code `SOLANA_ERROR__KEYS__VERIFY_SIGNATURE_FAILURE`

Included in this PR is testcases to test the possible errors within these functions and creating a succesful keypair from a 64-byte array.


packages/errors/src/codes.ts
```diff
@@ -96,6 +96,7 @@ export const SOLANA_ERROR__KEYS__INVALID_KEY_PAIR_BYTE_LENGTH = 3704000 as const
 export const SOLANA_ERROR__KEYS__INVALID_PRIVATE_KEY_BYTE_LENGTH = 3704001 as const;
 export const SOLANA_ERROR__KEYS__INVALID_SIGNATURE_BYTE_LENGTH = 3704002 as const;
 export const SOLANA_ERROR__KEYS__SIGNATURE_STRING_LENGTH_OUT_OF_RANGE = 3704003 as const;
+export const SOLANA_ERROR__KEYS__VERIFY_SIGNATURE_FAILURE = 3704004 as const;
```
lorisleiva, line 99: To be more consistent with the other errors this should be `SOLANA_ERROR__KEYS__FAILED_TO_VERIFY_SIGNATURE`.

However, I think we should instead call this error something like `SOLANA_ERROR__KEYS__PUBLIC_KEY_MUST_MATCH_PRIVATE_KEY` because what we're actually trying to test here is that the private and public keys match. The fact that we are verifying a signature is secondary.

packages/errors/src/messages.ts
```diff
@@ -402,6 +403,7 @@ export const SolanaErrorMessages: Readonly<{
         'Expected base58-encoded signature to decode to a byte array of length 64. Actual length: $actualLength.',
     [SOLANA_ERROR__KEYS__SIGNATURE_STRING_LENGTH_OUT_OF_RANGE]:
         'Expected base58-encoded signature string of length in the range [64, 88]. Actual length: $actualLength.',
+    [SOLANA_ERROR__KEYS__VERIFY_SIGNATURE_FAILURE]: 'The provided private key and public key are not a valid key pair',
```
lorisleiva, line 406: We should probably provide a context for that error as well, even if that only includes the public key.

Then this message can be something like:

```ts
'The provided private key does not match the provided public key at address [$address].'
```

packages/keys/src/key-pair.ts
```diff
@@ -21,5 +24,26 @@ export async function createKeyPairFromBytes(bytes: Uint8Array, extractable?: bo
...
+    if (bytes.byteLength !== 64) {
+        throw new SolanaError(SOLANA_ERROR__KEYS__INVALID_KEY_PAIR_BYTE_LENGTH, { byteLength: bytes.byteLength });
+    }
+    const [publicKey, privateKey] = await Promise.all([
+        crypto.subtle.importKey('raw', bytes.slice(32), 'Ed25519', /* extractable */ true, ['verify']),
+        createPrivateKeyFromBytes(bytes.slice(0, 32), extractable),
+    ]);
+
     return { privateKey, publicKey } as CryptoKeyPair;
 }
```
lorisleiva, line 49: I know the original issue suggested we expose such function but IMO we shouldn't.

If the user is looking to tradeoff that layer of security for performance, it's a very small function to copy/paste.

lorisleiva:
Thank you for your contribution! I just added a few comments. 🙂

luu-alex:
Thanks for the quick feedback! Your comments make sense, changes has been applied. lemme know if theres anything else

packages/keys/src/key-pair.ts
```diff
@@ -29,21 +29,8 @@ export async function createKeyPairFromBytes(bytes: Uint8Array, extractable?: bo
     const signedData = await signBytes(privateKey, TEST_DATA);
     const isValid = await verifySignature(publicKey, signedData, TEST_DATA);
     if (!isValid) {
-        throw new SolanaError(SOLANA_ERROR__KEYS__VERIFY_SIGNATURE_FAILURE);
+        throw new SolanaError(SOLANA_ERROR__KEYS__PUBLIC_KEY_MUST_MATCH_PRIVATE_KEY, { address: bytes.slice(32) });
```
lorisleiva, line 32: Thanks for the update! Just one more comment. It's probably more useful for the end-user if that address is base-58 encoded.

You can use string codecs to transform it like so:

```ts
import { getStringDecoder, getBase58Decoder } from "@solana/codecs-strings";

const addressDecoder = getStringDecoder({ encoding: getBase58Decoder(), size: 32 });
const address = addressDecoder.decode(bytes.slice(32));
```

Note that you could also use `getAddressDecoder` from `@solana/addresses` but that would add a new dependency to `@solana/keys` when it's not really needed.

lorisleiva:
Ah also, don't forget to fix the formatting. `pnpm style:fix` at the root should do the trick.

packages/keys/src/key-pair.ts
```diff
@@ -1,7 +1,10 @@
 import { assertKeyGenerationIsAvailable } from '@solana/assertions';
-import { SOLANA_ERROR__KEYS__INVALID_KEY_PAIR_BYTE_LENGTH, SolanaError } from '@solana/errors';
+import { SOLANA_ERROR__KEYS__INVALID_KEY_PAIR_BYTE_LENGTH, SolanaError, SOLANA_ERROR__KEYS__PUBLIC_KEY_MUST_MATCH_PRIVATE_KEY } from '@solana/errors';
 
 import { createPrivateKeyFromBytes } from './private-key';
+import { verifySignature, signBytes } from './signatures';
+
+const TEST_DATA = new Uint8Array([1, 2, 3, 4, 5]);
```
steveluscher, line 7: I'd like this to be randomly generated on each pass through the function.

packages/keys/src/key-pair.ts
```diff
@@ -21,5 +24,13 @@ export async function createKeyPairFromBytes(bytes: Uint8Array, extractable?: bo
         crypto.subtle.importKey('raw', bytes.slice(32), 'Ed25519', /* extractable */ true, ['verify']),
         createPrivateKeyFromBytes(bytes.slice(0, 32), extractable),
     ]);
+
+    // Verify the key pair
+    const signedData = await signBytes(privateKey, TEST_DATA);
+    const isValid = await verifySignature(publicKey, signedData, TEST_DATA);
```
steveluscher, line 30:
```suggestion
    const randomBytes = new Uint8Array(32);
    crypto.getRandomValues(randomBytes);
    const signedData = await signBytes(privateKey, randomBytes);
    const isValid = await verifySignature(publicKey, signedData, randomBytes);
```

packages/keys/src/key-pair.ts
```diff
@@ -1,7 +1,10 @@
...
-import { SOLANA_ERROR__KEYS__INVALID_KEY_PAIR_BYTE_LENGTH, SolanaError } from '@solana/errors';
+import { SOLANA_ERROR__KEYS__INVALID_KEY_PAIR_BYTE_LENGTH, SolanaError, SOLANA_ERROR__KEYS__PUBLIC_KEY_MUST_MATCH_PRIVATE_KEY } from '@solana/errors';
 
 import { createPrivateKeyFromBytes } from './private-key';
+import { verifySignature, signBytes } from './signatures';
+
+const TEST_DATA = new Uint8Array([1, 2, 3, 4, 5]);
 
 export async function generateKeyPair(): Promise<CryptoKeyPair> {
     await assertKeyGenerationIsAvailable();
```
steveluscher, line 10: To generate a test message randomly each time, we'll need to assert that the PRNG is available (that `crypto.getRandomValues` is a function`. Can you add that, and a related error code, to `@solana/assertions`.

```suggestion
    await Promise.all([
        assertKeyGenerationIsAvailable(),
        assertPRNGIsAvailable(), // <-- add this to `@solana/assertions` & `@solana/errors`
    ]);
```

packages/keys/src/key-pair.ts
steveluscher: I actually think we should probably not log the bytes here. Imagine that someone got the order wrong, and now their private bytes are logged to the console. Include no context.

packages/errors/src/messages.ts
```diff
@@ -402,6 +403,7 @@ export const SolanaErrorMessages: Readonly<{
         'Expected base58-encoded signature to decode to a byte array of length 64. Actual length: $actualLength.',
     [SOLANA_ERROR__KEYS__SIGNATURE_STRING_LENGTH_OUT_OF_RANGE]:
         'Expected base58-encoded signature string of length in the range [64, 88]. Actual length: $actualLength.',
+    [SOLANA_ERROR__KEYS__VERIFY_SIGNATURE_FAILURE]: 'The provided private key does not match the provided public key at address [$address].',
```
steveluscher, line 406: This is the wrong error code, right?
```suggestion
    [SOLANA_ERROR__KEYS__PUBLIC_KEY_MUST_MATCH_PRIVATE_KEY]: 'The provided private key does not match the provided public key at address [$address].',
```

steveluscher (changes requested):
Added via Giphy

steveluscher:
Thanks for doing this @luu-alex!

packages/errors/src/codes.ts
```diff
@@ -90,12 +90,17 @@ export const SOLANA_ERROR__SUBTLE_CRYPTO__GENERATE_FUNCTION_UNIMPLEMENTED = 3610
 export const SOLANA_ERROR__SUBTLE_CRYPTO__SIGN_FUNCTION_UNIMPLEMENTED = 3610005 as const;
 export const SOLANA_ERROR__SUBTLE_CRYPTO__VERIFY_FUNCTION_UNIMPLEMENTED = 3610006 as const;
 
+// Crypto-related errors.
```
luu-alex, line 93: Added new error code range, crypto.getRandomValues is not part of subtle :p

packages/keys/src/key-pair.ts
luu-alex: I think it would make more sense to put `assertPRNGIsAvailable()` in `createKeyPairFromBytes` than `generateKeyPair` right?

luu-alex:
Thanks for the feedback! 👍  let me know if theres anything else :)

packages/errors/src/messages.ts
luu-alex: Good catch, thanks!

packages/keys/src/key-pair.ts
steveluscher: Oh, of course. Oops; I left my comment in the wrong spot!

packages/keys/src/key-pair.ts
lorisleiva: That's a very good point.

packages/assertions/src/crypto.ts
```diff
@@ -0,0 +1,7 @@
+import { SOLANA_ERROR__CRYPTO__RANDOM_VALUES_FUNCTION_UNIMPLEMENTED, SolanaError } from '@solana/errors';
+
+export async function assertPRNGIsAvailable() {
```
steveluscher, line 3: Hmm… I wonder why some of these were made `async`. Seems like a mistake we shouldn't replicate here.

steveluscher (commented):
I don't have the rights to push to your `master` to update this PR, so here's the final boss level before merge!

```diff
diff --git a/packages/assertions/src/__tests__/crypto-test.ts b/packages/assertions/src/__tests__/crypto-test.ts
index a676917c4..5357cd790 100644
--- a/packages/assertions/src/__tests__/crypto-test.ts
+++ b/packages/assertions/src/__tests__/crypto-test.ts
@@ -3,9 +3,13 @@ import { SOLANA_ERROR__CRYPTO__RANDOM_VALUES_FUNCTION_UNIMPLEMENTED, SolanaError
 import { assertPRNGIsAvailable } from '../crypto';
 
 describe('assertPRNGIsAvailable()', () => {
-    it('resolves to `undefined` without throwing', async () => {
-        expect.assertions(1);
-        await expect(assertPRNGIsAvailable()).resolves.toBeUndefined();
+    describe('when getRandomValues is available', () => {
+        it('does not throw', () => {
+            expect(assertPRNGIsAvailable).not.toThrow();
+        });
+        it('returns `undefined`', () => {
+            expect(assertPRNGIsAvailable()).toBeUndefined();
+        });
     });
     describe('when getRandomValues is not available', () => {
         let oldCrypto: InstanceType<typeof Crypto>['getRandomValues'];
@@ -18,9 +22,8 @@ describe('assertPRNGIsAvailable()', () => {
         afterEach(() => {
             globalThis.crypto.getRandomValues = oldCrypto;
         });
-        it('rejects', async () => {
-            expect.assertions(1);
-            await expect(() => assertPRNGIsAvailable()).rejects.toThrow(
+        it('throws', () => {
+            expect(assertPRNGIsAvailable).toThrow(
                 new SolanaError(SOLANA_ERROR__CRYPTO__RANDOM_VALUES_FUNCTION_UNIMPLEMENTED),
             );
         });
diff --git a/packages/assertions/src/crypto.ts b/packages/assertions/src/crypto.ts
index bd62651c5..8752465e8 100644
--- a/packages/assertions/src/crypto.ts
+++ b/packages/assertions/src/crypto.ts
@@ -1,6 +1,6 @@
 import { SOLANA_ERROR__CRYPTO__RANDOM_VALUES_FUNCTION_UNIMPLEMENTED, SolanaError } from '@solana/errors';
 
-export async function assertPRNGIsAvailable() {
+export function assertPRNGIsAvailable() {
     if (typeof globalThis.crypto === 'undefined' || typeof globalThis.crypto.getRandomValues !== 'function') {
         throw new SolanaError(SOLANA_ERROR__CRYPTO__RANDOM_VALUES_FUNCTION_UNIMPLEMENTED);
     }
diff --git a/packages/assertions/src/index.ts b/packages/assertions/src/index.ts
index b3383c7a9..4d4f2dd31 100644
--- a/packages/assertions/src/index.ts
+++ b/packages/assertions/src/index.ts
@@ -1 +1,2 @@
+export * from './crypto';
 export * from './subtle-crypto';
diff --git a/packages/errors/src/codes.ts b/packages/errors/src/codes.ts
index feaac895e..5bac1bfea 100644
--- a/packages/errors/src/codes.ts
+++ b/packages/errors/src/codes.ts
@@ -92,7 +92,7 @@ export const SOLANA_ERROR__SUBTLE_CRYPTO__VERIFY_FUNCTION_UNIMPLEMENTED = 361000
 
 // Crypto-related errors.
 // Reserve error codes in the range [3611000-3611050].
-export const SOLANA_ERROR__CRYPTO__RANDOM_VALUES_FUNCTION_UNIMPLEMENTED = 3610007 as const;
+export const SOLANA_ERROR__CRYPTO__RANDOM_VALUES_FUNCTION_UNIMPLEMENTED = 3611000 as const;
 
 // Key-related errors.
 // Reserve error codes in the range [3704000-3704999].
diff --git a/packages/keys/src/key-pair.ts b/packages/keys/src/key-pair.ts
index 4adc36823..32c3264e8 100644
--- a/packages/keys/src/key-pair.ts
+++ b/packages/keys/src/key-pair.ts
@@ -1,4 +1,4 @@
-import { assertKeyGenerationIsAvailable } from '@solana/assertions';
+import { assertKeyGenerationIsAvailable, assertPRNGIsAvailable } from '@solana/assertions';
 import {
     SOLANA_ERROR__KEYS__INVALID_KEY_PAIR_BYTE_LENGTH,
     SOLANA_ERROR__KEYS__PUBLIC_KEY_MUST_MATCH_PRIVATE_KEY,
@@ -19,6 +19,8 @@ export async function generateKeyPair(): Promise<CryptoKeyPair> {
 }
 
 export async function createKeyPairFromBytes(bytes: Uint8Array, extractable?: boolean): Promise<CryptoKeyPair> {
+    assertPRNGIsAvailable();
+
     if (bytes.byteLength !== 64) {
         throw new SolanaError(SOLANA_ERROR__KEYS__INVALID_KEY_PAIR_BYTE_LENGTH, { byteLength: bytes.byteLength });
     }

```

luu-alex:
Cool. feedback addressed 👍

steveluscher:
Awesome! I'd do it myself, but I can't push to your `master`, so can you run a quick `pnpm style:fix` in the root to get CI to pass?

steveluscher (approved):
giphy

luu-alex:
boom. tests are passing, thanks

steveluscher (approved):
Thank you for doing this! Where did you come from? How did you get here?!

Added via Giphy

luu-alex:
Was learning more about solana  and saw solana-web3 to be interesting!

PR #2349, opened 2024-03-20
refactor: polyfill importKey to satisfy instance of CryptoKey

thepiwo:
libraries like [jose](https://github.com/panva/jose) do check if keys are `instanceOf CryptoKey` (e.g. https://github.com/panva/jose/blob/main/src/runtime/browser/webcrypto.ts#L3)

Using `Object.setPrototypeOf` the created keys from the ED25519 polyfill can also satisfy this check.

### DIFF
--- a/packages/webcrypto-ed25519-polyfill/src/__tests__/secrets-test.ts
+++ b/packages/webcrypto-ed25519-polyfill/src/__tests__/secrets-test.ts
@@ -160,6 +160,10 @@ describe('importKeyPolyfill', () => {
         it('allows importing valid public key bytes', () => {
             expect(() => importKeyPolyfill('raw', MOCK_PUBLIC_KEY_BYTES, false, ['verify'])).not.toThrow();
         });
+        it('imported public key to be instance of CryptoKey', () => {
+            const key = importKeyPolyfill('raw', MOCK_PUBLIC_KEY_BYTES, false, ['verify']);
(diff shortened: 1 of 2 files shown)


steveluscher (changes requested):
Oh wow. I didn't even think of that. Hero!

Can you hop over to `secrets-test.ts` and write tests for both of these?

thepiwo:
@steveluscher I'll do that!

thepiwo:
Actually, my fix doesn't seem to work in the test environment.

in `Unit Test (Node)` it works, but in `Unit Test (Browser)` I receive `ReferenceError: CryptoKey is not defined`, in my use-case in a different project I do use it successfully in a browser environment.

steveluscher:
Look in `setup-webcrypto` and glue the CryptoKey class to the global scope if it doesn't exist.

thepiwo:
@steveluscher thank you for the support, I tried that, but `prototype` isn't defined then, I didn't find a way to add this as well and currently don't have a good solution. Maybe "faking" the instance via `setPrototypeOf` in the first place isn't the best of ideas, even if it works in my use-case...

Would be open to try other suggestions if you have any.

I added this snippet in `setup-webcrypto.ts`: 
```
if (typeof globalThis.CryptoKey === 'undefined') {
     Object.defineProperty(globalThis, 'CryptoKey', {
        value: crypto.webcrypto.CryptoKey,
    });
}```

PR #2351, opened 2024-03-20
[WIP] Create subpackages for internal-ish code

steveluscher:
# Summary

The entire problem that this PR tries to solve is this.

1. Retain the simplicity of being able to `export * from '@solana/errors'` in `packages/library/`
2. Prevent things like `safeCaptureStackTrace()` from being exported from `packages/library/`

The steps taken to that end were: 

1. Change `moduleResolution` everywhere to `NodeNext`. This enables subpackage exports
2. Add `.js` extensions everywhere to satisfy that resolution regime
3. Create a subpackage export in `@solana/errors` for all of the stuff that we want to share internally but _not_ to export from `packages/library/`

Now any package in the `@solana` scope can use `@solana/errors/internal` but we can _still_ enjoy the simplicity of `export * from '@solana/errors'` in `packages/library/`


steveluscher:
* **#2351** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2351?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/build-scripts/package.json
```diff
@@ -3,7 +3,6 @@
     "version": "0.0.0",
     "private": true,
     "files": [
-        "add-js-extension-to-types.mjs",
```
steveluscher, line 6: Nuked this, in favour of _actually_ adding the .js extension everywhere.

packages/build-scripts/getBaseConfig.ts
```diff
@@ -25,7 +25,7 @@ export function getBaseConfig(platform: Platform, formats: Format[], _options: O
                               __REACTNATIVE__: `${platform === 'native'}`,
                               __VERSION__: `"${env.npm_package_version}"`,
                           },
-                          entry: [`./src/index.ts`],
+                          entry: ['./src/index.ts', './src/__internal.ts'],
```
steveluscher, line 28: This makes the build pick up this `__internal.ts` entrypoint if it exists.

packages/build-scripts/getBaseConfig.ts
```diff
@@ -70,6 +70,7 @@ export function getBaseConfig(platform: Platform, formats: Format[], _options: O
                           platform: platform === 'node' ? 'node' : 'browser',
                           pure: ['process'],
                           sourcemap: format !== 'iife' || isDebugBuild,
+                          splitting: false,
```
steveluscher, line 73: This prevents `tsup` from doing fancy bundle splitting between entrypoints.

packages/errors/src/__internal.ts
```diff
@@ -0,0 +1 @@
+export * from './stack-trace.js';
```
steveluscher, line 1: Here's that new ‘internal stuff’ entrypoint.

packages/rpc/src/rpc-integer-overflow-error.ts
```diff
@@ -1,4 +1,5 @@
-import { safeCaptureStackTrace, SOLANA_ERROR__RPC__INTEGER_OVERFLOW, SolanaError } from '@solana/errors';
+import { SOLANA_ERROR__RPC__INTEGER_OVERFLOW, SolanaError } from '@solana/errors';
+import { safeCaptureStackTrace } from '@solana/errors/internal';
```
steveluscher, line 2: Here's an example of making use of the internal subpackage.

packages/tsconfig/base.json
```diff
@@ -9,7 +9,8 @@
         "forceConsistentCasingInFileNames": true,
         "inlineSources": false,
         "isolatedModules": true,
-        "moduleResolution": "node",
+        "module": "NodeNext",
+        "moduleResolution": "NodeNext",
```
steveluscher, line 13: This is the fancy new resolution strategy that allows all of this to work.

packages/test-config/jest-unit.config.common.ts
```diff
@@ -2,6 +2,11 @@ import { Config } from '@jest/types';
 import path from 'path';
 
 const config: Partial<Config.InitialProjectOptions> = {
+    moduleNameMapper: {
+        // Strip `.js` from imports
+        // https://github.com/swc-project/jest/issues/64#issuecomment-1029753225
+        '^(\\.{1,2}/.*)\\.js$': '$1',
+    },
```
steveluscher, line 9: Of course Jest doesn't know any of this, so we have to re-strip the .js extensions.

lorisleiva:
Ugh it's so sad because I do love the end result but I cannot bear having to use `.js` extensions on `.ts` files. And the fact that we then have to strip them for Jest is the icing on the cake.

Is there no way we can do this as a build step instead?

steveluscher:
> Is there no way we can do this as a build step instead?

Do what; add the extensions? No, because once you change to `"moduleResolution": "NodeNext"` it's gg.

lorisleiva:
Well, adding `"moduleResolution": "NodeNext"` to our tsconfig changes how things are compiled, right?

What I'm saying is, instead of using the module resolution, can we not replicate the changes that this does at compile time to make subpackages work properly?

What exactly is `NodeNext` changing in our compiled code that makes subpackages possible?

steveluscher:
> What exactly is NodeNext changing in our compiled code that makes subpackages possible?

I think subpackage imports just straight up don't work unless `moduleResolution` is `node16` or higher.

With `moduleResolution: NodeNext`:



With `moduleResolution: Node`:

lorisleiva:
Oh I see, as a _consumer_ of our own API, we need `NodeNext` to import sub-packages. No matter how we export them. I'm happy to disagree and commit if everyone agrees that the benefits are worth it. `.js` extensions on `.ts` files do make me want to cry though.

steveluscher:
Meh.

PR #2352, opened 2024-03-20
Convert `SubtleCrypto` assertion methods to non-async

steveluscher:
# Summary

`assertKeyGenerationIsAvailable()` definitely needs to be async because it feature-detects `Ed25519`. These other assertions don't seem to need to be async though; they just test for the _presence_ of a function.

Unless there's a good reason to yield the loop here, we should probably just make these checks sync for performance.

# Test Plan

```
pnpm turbo test:typecheck test:unit:node test:unit:browser
```

### DIFF
--- a/.changeset/wild-lobsters-kick.md
+++ b/.changeset/wild-lobsters-kick.md
@@ -1,0 +1,7 @@
+---
+"@solana/addresses": patch
+"@solana/assertions": patch
+"@solana/keys": patch
+---
+
+`SubtleCrypto` assertion methods that can make their assertions synchronously are now synchronous, for performance.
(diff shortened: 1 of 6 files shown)


steveluscher:
* **#2352** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2352?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

steveluscher:
Please do check me on this though. Maybe there's something that I've forgotten? An environment where this check would need to be async?

lorisleiva (approved):

PR #2387, opened 2024-03-27
add a new ITransactionWithSignatures and IFullySignedTransaction type

### LINKED ISSUE #2362: [experimental] Decoding and then re-compiling a transaction invalidates existing signatures

mcintyre94:
This PR is the first in a stack to refactor transactions and properly separate `TransactionMessage` (unsigned transactions) from `ITransactionWithSignatures` (signed transactions that cannot be modified). See https://github.com/solana-labs/solana-web3.js/pull/2377#issuecomment-2021674511 for context/more details

It adds a `new-signatures` in the Transactions package that mirrors the existing `signatures` types + functions. These are not yet used by any callers, and have `new` prepended to their names to make them unique.

The rest of the stack will refactor uses of the old versions to use the new ones, and then eventually remove the old versions. New functionality will be added as required, this initial PR is quite minimal.

**Changes from the previous versions**

`signatures` is now an `OrderedMap`, with an entry for every expected signer of the transaction. Missing signatures are explicitly recorded in this map as null.

This means:

- `newPartiallySignTransaction` creates an `OrderedMap` for signatures with an entry for every expected signer, in order of the accounts in the compiled message. Signers that have not signed are stored as null.
- `newGetSignatureFromTransaction` returns the first transaction signature by order. This is assumed to be the fee payer. It does not decompile the message bytes to verify this assumption. It throws if this first signature entry is null.

Closes #2362.


mcintyre94:
* **#2387** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2387?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

steveluscher:
Haha. Oh, are _you_ going to smash this work out? Good thing I hadn't quite started :)

mcintyre94:
I was planning to, but happy for someone else to take it too! It sounded on the call earlier like everyone was very busy though, and I'm keen to unblock this

steveluscher (approved):
Added via Giphy

packages/transactions/src/new-signatures.ts
```diff
@@ -0,0 +1,96 @@
...
+    const compiledMessage = compileMessage(transaction);
+    const transactionSigners = compiledMessage.staticAccounts.slice(0, compiledMessage.header.numSignerAccounts);
+    const messageBytes = getCompiledMessageEncoder().encode(compiledMessage) as TransactionMessageBytes;
+
+    const publicKeySignaturePairs = await Promise.all(
+        keyPairs.map(keyPair =>
+            Promise.all([getAddressFromPublicKey(keyPair.publicKey), signBytes(keyPair.privateKey, messageBytes)]),
+        ),
+    );
+    const publickKeySignatureMap = new Map(publicKeySignaturePairs);
```
steveluscher, line 55:
```suggestion
    const publicKeySignatureMap = new Map(publicKeySignaturePairs);
```

packages/transactions/src/ordered-map.ts
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    firstValue(): V | undefined {
+        if (this.keysOrder.length === 0) {
+            return undefined;
+        }
+        const firstKey = this.keysOrder[0];
+        return this.map.get(firstKey);
+    }
+}
```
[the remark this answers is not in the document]
mcintyre94: Oh sweet, TIL!

packages/transactions/src/__tests__/new-signatures-test.ts
```diff
@@ -0,0 +1,359 @@
...
+    const MOCK_SIGNATURE_B = new Uint8Array(Array(64).fill(2));
+    const MOCK_SIGNATURE_C = new Uint8Array(Array(64).fill(3));
+    const mockKeyPairA = { privateKey: {} as CryptoKey, publicKey: {} as CryptoKey } as CryptoKeyPair;
+    const mockKeyPairB = { privateKey: {} as CryptoKey, publicKey: {} as CryptoKey } as CryptoKeyPair;
+    const mockKeyPairC = { privateKey: {} as CryptoKey, publicKey: {} as CryptoKey } as CryptoKeyPair;
+    const mockPublicKeyAddressA = 'AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA' as Address<'AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'>;
+    const mockPublicKeyAddressB = 'BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB' as Address<'BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB'>;
+    const mockPublicKeyAddressC = 'CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC' as Address<'CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC'>;
+    const mockCompiledMessageBytes = new Uint8Array(Array(100)).fill(4);
+    beforeEach(() => {
```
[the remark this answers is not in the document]
mcintyre94: These are the benefits of having signatures-test.ts to use as a starting point!

packages/transactions/src/new-signatures.ts
```diff
@@ -0,0 +1,96 @@
...
+    for (const signerAddress of transactionSigners) {
+        const signature = publickKeySignatureMap.get(signerAddress) ?? null;
+        signatures.set(signerAddress, signature);
+    }
+
+    const out: INewTransactionWithSignatures = {
+        messageBytes,
+        signatures,
+    };
+    Object.freeze(out);
```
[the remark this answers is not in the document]
mcintyre94: That's an annoying limitation, but I'm inclined to just leave it as-is. The `CryptoKey` approach feels a bit overkill for this, and I think storing this as anything but a `Uint8Array` is just going to add extra copies whenever someone wants to do anything with it. Ultimately it's probably going to get serialised and sent somewhere, there are limits to how much we can help you if you start randomly modifying byte arrays!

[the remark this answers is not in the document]
steveluscher: > Though I'm a bit confused why I'm not getting an error on the Object.freeze call…

It freezes shallowly. If we're not freezing every node in the tree elsewhere, we should be.

> …I think storing this as anything but a Uint8Array is just going to add extra copies whenever someone wants to do anything with it.

Copies are Bad™, but this very well may be literally the only place in `@solana/transactions` that you can't do `tx === otherTx` and know for sure that they're identical. Making `SignatureBytes` an `Array` would mean copying things back and forth, but we're also talking about fewer bytes than saying ‘Two coffees and a beer’ three times.

You're almost certainly correct, just wanted to give it another think.

packages/transactions/src/new-signatures.ts
```diff
@@ -0,0 +1,96 @@
...
+    // first signature is the fee payer
+    const signatureBytes = transaction.signatures.firstValue();
+    if (!signatureBytes) {
+        throw new SolanaError(SOLANA_ERROR__TRANSACTION__FEE_PAYER_SIGNATURE_MISSING);
+    }
+    const transactionSignature = base58Decoder.decode(signatureBytes);
+    return transactionSignature as Signature;
+}
+
+export async function newPartiallySignTransaction<TTransaction extends CompilableTransaction>(
```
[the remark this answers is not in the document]
mcintyre94: To clarify, this would be an overload with the type `(CryptoKeyPair[], ITransactionWithSignatures) -> ITransactionWithSignatures`, used to add additional signers to a partially signed tx by signing the same message bytes?

packages/transactions/src/new-signatures.ts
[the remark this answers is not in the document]
lorisleiva: We had this conversation a little while ago on a PR and decided at the time to put a pin on it. Since we're breaking things here, maybe now is the time to revisit this?

https://github.com/solana-labs/solana-web3.js/pull/1710#discussion_r1389234589

packages/transactions/src/new-signatures.ts
```diff
@@ -0,0 +1,96 @@
...
+import { compileMessage } from './message';
+import { OrderedMap } from './ordered-map';
+import { getCompiledMessageEncoder } from './serializers';
+
+export type TransactionMessageBytes = Uint8Array & { readonly __brand: unique symbol };
+
+export interface INewTransactionWithSignatures {
+    messageBytes: TransactionMessageBytes;
+    signatures: OrderedMap<Address, SignatureBytes | null>;
+}
```
lorisleiva, line 21: Since I'm assuming we're gonna call the other type `TransactionMessage`, why not take that opportunity to make _this_ type the main `Transaction` type?

packages/transactions/src/new-signatures.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export interface INewFullySignedTransaction extends INewTransactionWithSignatures {
+    readonly __brand: unique symbol;
+}
```
lorisleiva, line 25: How do we feel about dropping the `I` prefixes here?

I was thinking something like this for the overall API.

image

packages/transactions/src/new-signatures.ts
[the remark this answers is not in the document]
lorisleiva: My understanding here is that the new `partialSignTransaction` should either:

- Accept a `TransactionMessage` and return a new `Transaction`, or
- Accept a `TTransaction extends Transaction` and return `TTransaction`.

And the new `signTransaction` should either:

- Accept a `TransactionMessage` and return a new `Transaction & FullySignedTransaction`, or
- Accept a `TTransaction extends Transaction` and return `TTransaction & FullySignedTransaction`.

_Note that I'm naming types based on my previous comment._

Wdyt?

lorisleiva (approved):
Thank you for taking a stab at this! 🙏

packages/transactions/src/new-signatures.ts
mcintyre94: This API diagram looks like a good endpoint for this, I'll rename based on this.

packages/transactions/src/new-signatures.ts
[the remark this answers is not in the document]
mcintyre94: Alright cool this is using `ReadonlyUint8Array` now, added further up the stack

packages/transactions/src/new-signatures.ts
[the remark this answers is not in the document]
mcintyre94: I've refactored this:
- partial sign is `NewTransaction -> NewTransaction`
- sign is `NewTransaction -> NewFullySignedTransaction`

The logic to compile from `TransactionMessage -> NewTransaction` is further up the stack. I think it makes sense to just keep these steps separate because this ends up doing too much otherwise and using completely different types. It'll be trivial to write a helper to compile + sign if someone wants to.

I also didn't use `TTransaction extends Transaction` because I don't think anyone will need to extend `NewTransaction`. It's just message + signature, you really can't/shouldn't do anything else with it.

packages/transactions/src/new-signatures.ts
```diff
@@ -0,0 +1,78 @@
...
+    const transactionSignature = base58Decoder.decode(signatureBytes);
+    return transactionSignature as Signature;
+}
+
+export async function newPartiallySignTransaction(
+    keyPairs: CryptoKeyPair[],
+    transaction: NewTransaction,
+): Promise<NewTransaction> {
+    const addressPrivateKeyPairs = await Promise.all(
+        keyPairs.map(keyPair => Promise.all([getAddressFromPublicKey(keyPair.publicKey), keyPair.privateKey])),
```
lorisleiva, line 37: This `Promise.all()` is a bit weird since only the first item is a promise.

packages/transactions/src/new-signatures.ts
```diff
@@ -0,0 +1,78 @@
...
+        ([address, _privateKey]) => transaction.signatures[address] === null,
+    );
+    const newAddressSignaturePairs = await Promise.all(
+        newAddressPrivateKeyPairs.map(([address, privateKey]) =>
+            Promise.all([address, signBytes(privateKey, transaction.messageBytes)]),
+        ),
+    );
+    for (const [address, signature] of newAddressSignaturePairs) {
+        transaction.signatures[address] = signature;
+    }
```
lorisleiva, line 49: Here as well, we should probably create a new object (since `transaction.signatures` should be frozen) and freeze that new object as well.

packages/transactions/src/new-signatures.ts
```diff
@@ -0,0 +1,78 @@
...
+): Promise<NewTransaction> {
+    const addressPrivateKeyPairs = await Promise.all(
+        keyPairs.map(keyPair => Promise.all([getAddressFromPublicKey(keyPair.publicKey), keyPair.privateKey])),
+    );
+    const newAddressPrivateKeyPairs = addressPrivateKeyPairs.filter(
+        ([address, _privateKey]) => transaction.signatures[address] === null,
+    );
+    const newAddressSignaturePairs = await Promise.all(
+        newAddressPrivateKeyPairs.map(([address, privateKey]) =>
+            Promise.all([address, signBytes(privateKey, transaction.messageBytes)]),
```
lorisleiva, line 44: That `Promise.all()` is also weird as only the second item is a promise.

packages/transactions/src/new-signatures.ts
```diff
@@ -0,0 +1,78 @@
...
+        ),
+    );
+    for (const [address, signature] of newAddressSignaturePairs) {
+        transaction.signatures[address] = signature;
```
steveluscher, line 48: This function needs a test that asserts we _refuse_ to compute or add a signature that's not already in the list. Like, I shouldn't be able to add a signature to this thing unless there's a key in the `signatures` map for my keypair's public key.

steveluscher (approved):
Some nits about pipelining operations, and also I still would _really_ like to see the function return the identical transaction if nothing has changed (ie. if a partial sign operation resulted in no new signatures because the signature was _already there_ or the key pair supplied was not destined to be a signer.

packages/transactions/src/new-signatures.ts
mcintyre94: Nice, refactored to use this format. I was originally trying not to create a signature at all when there's already one, but this definitely makes it simpler/clearer. Added a test for the transaction being returned untouched too.

packages/transactions/src/new-signatures.ts
```diff
@@ -0,0 +1,98 @@
...
+): Promise<NewTransaction> {
+    let newSignatures: Record<Address, SignatureBytes> | undefined;
+
+    await Promise.all(
+        keyPairs.map(async keyPair => {
+            const address = await getAddressFromPublicKey(keyPair.publicKey);
+            const newSignature = await signBytes(keyPair.privateKey, transaction.messageBytes);
+            const existingSignature = transaction.signatures[address];
+
+            // what to do if existingSignature is undefined, ie not a valid signer?
```
steveluscher, line 48: I think you should drop the signer up there at the point where you've derived the public key. If `address` is not a key of `transaction.signatures` then either silently drop it or throw something like ‘hey this keypair isn't an eligible signer.’

steveluscher: Of note: we throw in the existing library.

https://github.com/solana-labs/solana-web3.js/blob/a98958b63f203613b84cd6b905765debab46ae51/packages/library-legacy/src/transaction/legacy.ts#L753-L755

mcintyre94: I was undecided whether to throw but I've added an error for this. It includes all the unexpected signers and also all the expected ones in the error context. I've hit that error a lot and that's what I tend to end up using to figure it out!

packages/transactions/src/new-signatures.ts
```diff
@@ -0,0 +1,115 @@
...
+
+            // Check if the address is expected to sign the transaction
+            if (existingSignature === undefined) {
+                // address is not an expected signer for this transaction
+                unexpectedSigners ||= new Set();
+                unexpectedSigners.add(address);
+                return;
+            }
+
+            const newSignature = await signBytes(keyPair.privateKey, transaction.messageBytes);
```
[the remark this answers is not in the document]
steveluscher: Like, not just for you, but for others in the `Promise.all`.

PR #2397, opened 2024-03-29
refactor(experimental): add prefixCodec to @solana/codecs-core

lorisleiva:
This PR adds a new `prefixCodec` primitive to `@solana/codecs-core`. This primitives accepts any codec and prefixes its size using a `NumberCodec` provided as a second argument.

For instance:

```ts
const codec = prefixCodec(getBase58Codec(), getU32Codec());

codec.encode('hello world');
// 0x0b00000068656c6c6f20776f726c64
//   |       └-- Our encoded base-58 string.
//   └-- Our encoded u32 size prefix.
```

This PR also updates the `getStringCodec` and `getBytesCodec` as they can now use the `prefixCodec` primitive to simplify their internal logic.

P.S.: Now that `getStringCodec` and `getBytesCodec` are simple aggregator of `fixCodec` and `prefixCodec`, we should consider removing them or adjusting them further. (Pros: no default values => better tree-shakeability, Cons: slightly less friendly API?).

```ts
// With getStringCodec.
getStringCodec();
getStringCodec({ encoding: getBase58Codec() });
getStringCodec({ size: 32 });
getStringCodec({ size: getU16Codec() });
getStringCodec({ size: "variable" });

// Without getStringCodec.
prefixCodec(getUtf8Codec(), getU32Codec());
prefixCodec(getBase58Codec(), getU32Codec());
fixCodec(getUtf8Codec(), 32);
prefixCodec(getUtf8Codec(), getU16Codec());
getUtf8Codec();
```

### DIFF
--- a/packages/codecs-core/README.md
+++ b/packages/codecs-core/README.md
@@ -362,6 +362,31 @@ const get32BytesBase58Decoder = () => fixDecoder(getBase58Decoder(), 32);
 const get32BytesBase58Codec = () => combineCodec(get32BytesBase58Encoder(), get32BytesBase58Codec());
 ```
 
+## Prefixing the size of codecs
+
+The `prefixCodec` function allows you to store the byte size of any codec as a number prefix. This allows you to contain variable-size codecs to their actual size.
+
+When encoding, the size of the encoded data is stored before the encoded data itself. When decoding, the size is read first to know how many bytes to read next.
+
+For example, say we want to represent a variable-size base-58 string using a `u32` size prefix — the equivalent of a Borsh `String` in Rust. Here’s how you can use the `prefixCodec` function to achieve that.
+
+```ts
+const getU32Base58Codec = () => prefixCodec(getBase58Codec(), getU32Codec());
+
+getU32Base58Codec().encode('hello world');
+// 0x0b00000068656c6c6f20776f726c64
+//   |       └-- Our encoded base-58 string.
+//   └-- Our encoded u32 size prefix.
+```
+
+You may also use the `prefixEncoder` and `prefixDecoder` functions to separate your codec logic like so:
+
+```ts
+const getU32Base58Encoder = () => prefixEncoder(getBase58Encoder(), getU32Encoder());
+const getU32Base58Decoder = () => prefixDecoder(getBase58Decoder(), getU32Decoder());
+const getU32Base58Codec = () => combineCodec(getU32Base58Encoder(), getU32Base58Decoder());
+```
+
 ## Adjusting the size of codecs
 
 The `resizeCodec` helper re-defines the size of a given codec by accepting a function that takes the current size of the codec and returns a new size. This works for both fixed-size and variable-size codecs.
--- a/packages/codecs-core/src/prefix-codec.ts
+++ b/packages/codecs-core/src/prefix-codec.ts
@@ -1,0 +1,118 @@
+import { assertByteArrayHasEnoughBytesForCodec } from './assertions';
+import {
+    Codec,
+    createDecoder,
+    createEncoder,
+    Decoder,
+    Encoder,
+    FixedSizeCodec,
+    FixedSizeDecoder,
+    FixedSizeEncoder,
+    getEncodedSize,
+    isFixedSize,
+    VariableSizeCodec,
+    VariableSizeDecoder,
+    VariableSizeEncoder,
+} from './codec';
+import { combineCodec } from './combine-codec';
+
+type NumberEncoder = Encoder<bigint | number> | Encoder<number>;
+type FixedSizeNumberEncoder<TSize extends number = number> =
+    | FixedSizeEncoder<bigint | number, TSize>
+    | FixedSizeEncoder<number, TSize>;
+type NumberDecoder = Decoder<bigint> | Decoder<number>;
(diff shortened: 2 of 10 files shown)


packages/codecs-core/README.md
```diff
@@ -362,6 +362,31 @@ const get32BytesBase58Decoder = () => fixDecoder(getBase58Decoder(), 32);
 const get32BytesBase58Codec = () => combineCodec(get32BytesBase58Encoder(), get32BytesBase58Codec());
 ```
 
+## Prefixing the size of codecs
+
+The `prefixCodec` function allows you to store the byte size of any codec as a number prefix. This allows you to contain variable-size codecs to their actual size.
```
steveluscher, line 367: I wonder if there's a name we can give this codec to make it more clear that the prefix it produces is _particularly_ the size of the field.

I wonder what other libraries call this.

* Length-delimited records ([protobuf](https://protobuf.dev/programming-guides/encoding/#length-types))
* Length-prefix framing

Doesn't super matter. I can't find strong evidence that there's even a consensus on what to call such a thing.

packages/codecs-core/README.md
```diff
@@ -362,6 +362,31 @@ const get32BytesBase58Decoder = () => fixDecoder(getBase58Decoder(), 32);
...
+//   |       └-- Our encoded base-58 string.
+//   └-- Our encoded u32 size prefix.
+```
+
+You may also use the `prefixEncoder` and `prefixDecoder` functions to separate your codec logic like so:
+
+```ts
+const getU32Base58Encoder = () => prefixEncoder(getBase58Encoder(), getU32Codec());
+const getU32Base58Decoder = () => prefixDecoder(getBase58Decoder(), getU32Codec());
+const getU32Base58Codec = () => combineCodec(getU32Base58Encoder(), getU32Base58Codec());
```
steveluscher, line 387: This looks like it needs another pass.

steveluscher (changes requested):
tl;dr unless I missed it, I _think_ you need something to prevent the actual size of the data from being larger than that which you can represent using the length prefix. We either need to throw when that's the case, or truncate the data to the actual length that will eventually be decoded.

packages/codecs-core/src/prefix-codec.ts
```diff
@@ -0,0 +1,118 @@
...
+    prefix: FixedSizeNumberEncoder,
+): FixedSizeEncoder<TFrom>;
+export function prefixEncoder<TFrom>(encoder: Encoder<TFrom>, prefix: NumberEncoder): VariableSizeEncoder<TFrom>;
+export function prefixEncoder<TFrom>(encoder: Encoder<TFrom>, prefix: NumberEncoder): Encoder<TFrom> {
+    const write = ((value, bytes, offset) => {
+        // Here we exceptionally use the `encode` function instead of the `write`
+        // function to contain the content of the encoder within its own bounds.
+        const encoderBytes = encoder.encode(value);
+        offset = prefix.write(encoderBytes.length, bytes, offset);
+        bytes.set(encoderBytes, offset);
```
steveluscher, line 46: We have to do _something_ to prevent this scenario, right?

```ts
const { decode, encode } = prefixCodec(getUtf8Encoder(), getU8Encoder());
const bytes = encode('string longer [...] than 256 characters');
bytes >= 257; // true
const str = decode(bytes); // string longer [...] than 256 ch
// And now the buffer after that point is corrupt
```

lorisleiva: I'll had a test to be explicit about it but if the u8 codec isn't able to store the string size, it will throw an error.

It's the responsibility of the size codec to determine if it can indeed encode/decode the provided size because you could technically provide more complicated number codecs such as the `shortU16`.

steveluscher: Ah, excellent.

steveluscher (approved):
Added via Giphy

packages/codecs-core/README.md
lorisleiva: Yeah naming this one has been tricky. The constraints for naming this were:

- Needs to start with a verb and finish with `Codec` since it's a codec primitive in the likes of `mapCodec` or `fixCodec`.
- Somewhat echo the `fixCodec` primitive which makes any codec a fixed-size one by providing the byte size explicitly. This is because technically, both of these codecs are responsible for "sizing" the codec they wrap.

`prefixCodec` seemed to do the job okay but I agree it's not ideal. My main issues with it are:

- Not clear that it's prefixing the size.
- A bit too close to the new `getHiddenPrefixCodec` helper at the end of this stack.

An alternative could be `prefixCodecSize` and we could rename `fixCodec` to `fixCodecSize` to bring the two closer together.

Wdyt?

lorisleiva: I made the suggested change in a subsequent PR: https://github.com/solana-labs/solana-web3.js/pull/2411.

packages/codecs-core/src/prefix-codec.ts
lorisleiva: I added a test for this.

lorisleiva:
### Merge activity

* **Apr 2, 6:48 PM EDT**: @lorisleiva started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2397).

PR #2398, opened 2024-04-01
refactor(experimental): add getUnionCodec to @solana/codecs-data-structures

lorisleiva:
This PR adds a new `getUnionCodec` helper that can be used to encode/decode any TypeScript union.

It accepts the following arguments:

-   An array of codecs, each defining a variant of the union.
-   A `getIndexFromValue` function which, given a value of the union, returns the index of the codec that should be used to encode that value.
-   A `getIndexFromBytes` function which, given the byte array to decode at a given offset, returns the index of the codec that should be used to decode the next bytes.

```ts
const codec: Codec<number | boolean> = getUnionCodec(
    [getU16Codec(), getBooleanCodec()],
    value => (typeof value === 'number' ? 0 : 1),
    (bytes, offset) => (bytes.slice(offset).length > 1 ? 0 : 1),
);

codec.encode(42); // 0x2a00
codec.encode(true); // 0x01
```


packages/codecs-data-structures/src/union.ts
```diff
@@ -0,0 +1,128 @@
...
+ * Creates a union encoder from the provided array of encoder.
+ *
+ * @param variants - The variant encoders of the union.
+ * @param getIndexFromValue - A function that returns the index of the variant from the provided value.
+ */
+export function getUnionEncoder<const TVariants extends readonly Encoder<any>[]>(
+    variants: TVariants,
+    getIndexFromValue: (value: GetEncoderTypeFromVariants<TVariants>) => number,
+): Encoder<GetEncoderTypeFromVariants<TVariants>> {
+    type TFrom = GetEncoderTypeFromVariants<TVariants>;
```
steveluscher, line 35: You… can… do this… inline?

steveluscher (approved):
Added via Giphy

lorisleiva:
### Merge activity

* **Apr 2, 6:48 PM EDT**: @lorisleiva started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2398).

PR #2416, opened 2024-04-02
Prevent the use of mismatching RPCs in factories that require a pair of RPCs (one-shot and subscriptions)

### LINKED ISSUE #2284: [experimental] Any method that takes in an `rpc` _and_ an `rpcSubscriptions` should enforce that they're for the same cluster
If it's possible to do this with TypeScript, lets.

steveluscher:
# Summary

If you try to, say, get an airdrop but you use a devnet RPC to get the airdrop, and a testnet RPC to confirm the signature, you're going to have a bad time.

This PR attempts to enforce, via TypeScript, that you supply _matching_ RPCs wherever a pair is required.

Closes #2284.

# Test Plan

```
pnpm turbo test:typecheck
```

### DIFF
--- a/packages/rpc-subscriptions/src/rpc-subscriptions-clusters.ts
+++ b/packages/rpc-subscriptions/src/rpc-subscriptions-clusters.ts
@@ -29,3 +29,13 @@ export type RpcSubscriptionsFromTransport<
       : TRpcSubscriptionsTransport extends RpcSubscriptionsTransportMainnet
         ? RpcSubscriptionsMainnet<TRpcMethods>
         : RpcSubscriptions<TRpcMethods>;
+export type RpcSubscriptionsFromCluster<
+    TRpcMethods,
+    TCluster extends 'devnet' | 'mainnet' | 'testnet' | undefined = undefined,
+> = TCluster extends 'devnet'
+    ? RpcSubscriptionsDevnet<TRpcMethods>
+    : TCluster extends 'testnet'
+      ? RpcSubscriptionsTestnet<TRpcMethods>
+      : TCluster extends 'mainnet'
+        ? RpcSubscriptionsMainnet<TRpcMethods>
+        : RpcSubscriptions<TRpcMethods>;
--- a/packages/rpc-subscriptions/src/rpc-subscriptions-clusters.ts
+++ b/packages/rpc-subscriptions/src/rpc-subscriptions-clusters.ts
@@ -29,13 +29,3 @@ export type RpcSubscriptionsFromTransport<
       : TRpcSubscriptionsTransport extends RpcSubscriptionsTransportMainnet
         ? RpcSubscriptionsMainnet<TRpcMethods>
         : RpcSubscriptions<TRpcMethods>;
-export type RpcSubscriptionsFromCluster<
-    TRpcMethods,
-    TCluster extends 'devnet' | 'mainnet' | 'testnet' | undefined = undefined,
-> = TCluster extends 'devnet'
-    ? RpcSubscriptionsDevnet<TRpcMethods>
-    : TCluster extends 'testnet'
-      ? RpcSubscriptionsTestnet<TRpcMethods>
-      : TCluster extends 'mainnet'
-        ? RpcSubscriptionsMainnet<TRpcMethods>
-        : RpcSubscriptions<TRpcMethods>;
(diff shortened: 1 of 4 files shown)


steveluscher:
* **#2416** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2416?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/rpc-subscriptions/src/rpc-subscriptions-clusters.ts
```diff
@@ -29,3 +29,13 @@ export type RpcSubscriptionsFromTransport<
...
+export type RpcSubscriptionsFromCluster<
+    TRpcMethods,
+    TCluster extends 'devnet' | 'mainnet' | 'testnet' | undefined = undefined,
+> = TCluster extends 'devnet'
+    ? RpcSubscriptionsDevnet<TRpcMethods>
+    : TCluster extends 'testnet'
+      ? RpcSubscriptionsTestnet<TRpcMethods>
+      : TCluster extends 'mainnet'
+        ? RpcSubscriptionsMainnet<TRpcMethods>
+        : RpcSubscriptions<TRpcMethods>;
```
buffalojoec, line 41:
```suggestion
export type RpcSubscriptionsFromCluster<
    TRpcMethods,
    TCluster extends ClusterUrl,
> = TCluster extends DevnetUrl
    ? RpcSubscriptionsDevnet<TRpcMethods>
    : TCluster extends TestnetUrl
      ? RpcSubscriptionsTestnet<TRpcMethods>
      : TCluster extends MainnetUrl
        ? RpcSubscriptionsMainnet<TRpcMethods>
        : RpcSubscriptions<TRpcMethods>;
```

packages/rpc/src/rpc-clusters.ts
```diff
@@ -24,6 +24,16 @@ export type RpcFromTransport<TRpcMethods, TRpcTransport extends RpcTransport> =
...
+export type RpcFromCluster<
+    TRpcMethods,
+    TCluster extends 'devnet' | 'mainnet' | 'testnet' | undefined = undefined,
+> = TCluster extends 'devnet'
+    ? RpcDevnet<TRpcMethods>
+    : TCluster extends 'testnet'
+      ? RpcTestnet<TRpcMethods>
+      : TCluster extends 'mainnet'
+        ? RpcMainnet<TRpcMethods>
+        : Rpc<TRpcMethods>;
```
buffalojoec, line 36:
```suggestion
export type RpcFromCluster<
    TRpcMethods,
    TCluster extends ClusterUrl,
> = TCluster extends DevnetUrl
    ? RpcDevnet<TRpcMethods>
    : TCluster extends TestnetUrl
      ? RpcTestnet<TRpcMethods>
      : TCluster extends MainnetUrl
        ? RpcMainnet<TRpcMethods>
        : Rpc<TRpcMethods>;
```

buffalojoec (commented):
Nice. Overall this makes a lot of sense, especially considering airdrop is only for test clusters.

I want to at least call out that I bet the enforcement of lower-level clusters, and explicit barring of generic cluster, will definitely frustrate some people.

That being said, why have cluster-specific RPCs and APIs in the first place if we're not going to use them?! It's not unreasonable to require someone explicitly define a test-cluster transport/RPC in order to request test lamports with our API.

do it

lorisleiva (changes requested):
IIRC, we agreed that the generic RPC should not fail in this case because it'll likely be used for local validators which do support airdrops.

buffalojoec:
> IIRC, we agreed that the generic RPC should not fail in this case because it'll likely be used for local validators which do support airdrops.

Bah, you're right. This made things much tricker! I've got a working implementation now for the type-checking, but the error messages are horrendous. Trying to iron that bit out, then I'll push another commit.

lorisleiva (approved):
This looks great!

steveluscher:
Nice. I intended to do this for all of the other places, other than airdrop, too. Search the code for `rpcSubscriptions: ` and you'll find the rest of them. Want to replicate what you've done here, there, @buffalojoec?

buffalojoec:
> Nice. I intended to do this for all of the other places, other than airdrop, too. Search the code for `rpcSubscriptions: ` and you'll find the rest of them. Want to replicate what you've done here, there, @buffalojoec?

#2534

PR #2481, opened 2024-04-12
attach a typed lifetime constraint field when we compile a transaction

mcintyre94:


mcintyre94:
* **#2481** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2481?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/transactions/src/new-compile-transaction.ts
```diff
@@ -21,7 +39,19 @@ export function compileTransaction(transactionMessage: CompilableTransactionMess
...
+    let lifetimeConstraint: TransactionWithLifetime['lifetimeConstraint'];
+    if ('blockhash' in transactionMessage.lifetimeConstraint) {
+        lifetimeConstraint = {
+            blockhash: transactionMessage.lifetimeConstraint.blockhash,
+        };
+    } else {
+        lifetimeConstraint = {
+            nonce: transactionMessage.lifetimeConstraint.nonce,
+        };
+    }
```
buffalojoec, line 51: Since you're already doing this check in two spots, and we could introduce more, wdyt about following our library's pattern like `assertIsDurableNonceTransaction` and rolling one of these?

```ts
assertIsBlockhashLifetime(lifetime: T): asserts lifetime is BlockhashLifetime;
```

packages/transactions/src/new-compile-transaction.ts
```diff
@@ -4,12 +4,30 @@ import { SignatureBytes } from '@solana/keys';
...
 } from '@solana/transaction-messages';
 
+import { TransactionBlockhashLifetime, TransactionDurableNonceLifetime, TransactionWithLifetime } from './lifetime';
 import { NewTransaction, OrderedMap, TransactionMessageBytes } from './transaction';
 
-export function compileTransaction(transactionMessage: CompilableTransactionMessage): NewTransaction {
+export function compileTransaction(
+    transactionMessage: ITransactionMessageWithBlockhashLifetime &
+        Omit<CompilableTransactionMessage, 'lifetimeConstraint'>,
+): NewTransaction & { lifetimeConstraint: TransactionBlockhashLifetime };
```
buffalojoec, line 18: Heh, this confused me for a minute, but I think I see what you're doing here. You do the specific lifetime-type then omit `lifetimeConstraint` from `CompilableTransactionMessage` to ensure it's the specific lifetime, right?

If so, might it be a bit cleaner to just do this?

```ts
export function compileTransaction(
    transactionMessage: CompilableTransactionMessage & { lifetimeConstraint: TransactionBlockhashLifetime }
): NewTransaction & { lifetimeConstraint: TransactionBlockhashLifetime };
```

buffalojoec (commented):
Looks like you might just be missing something little based on CI, but lgtm!

packages/transactions/src/new-compile-transaction.ts
mcintyre94: I've just rolled in our existing `isTransactionMessageWithBlockhashLifetime` instead of doing the ad-hoc check. I think that makes the most sense since we're dealing with `CompilableTransactionMessage` anyway.

packages/transactions/src/new-compile-transaction.ts
mcintyre94: Not sure why I thought I needed the `Omit`, removed it, thanks!

buffalojoec (approved):
nice

PR #2482, opened 2024-04-14
Optimize Cache-key function using fast-stable-stringify

Hrushi20:
## Aim-
Reduce the size of dependency to stringify various datatypes in project. 

Forked [fast-stable-stringify](https://www.npmjs.com/package/fast-stable-stringify) and added support to BigInt datatype.


packages/fast-stable-stringify/README.md
buffalojoec, line 1: Let's blow this README away and be sure to call out in the first line below the header that it's a fork and link to the original.

Then we can adapt a bit of the original documentation and morph it with some of ours. It doesn't need to say much.

packages/rpc-graphql/package.json
```diff
@@ -67,7 +67,7 @@
         "@solana/codecs-strings": "workspace:*",
         "dataloader": "^2.2.2",
         "graphql": "^16.8.0",
-        "json-stable-stringify": "^1.1.1"
+        "@solana/fast-stable-stringify": "workspace:*"
```
buffalojoec, line 70: Nice. Mind alpha-ordering these?

packages/fast-stable-stringify/src/__tests__/json_test.ts
```diff
@@ -0,0 +1,55 @@
+import stringify from '..';
+
+describe('test_fast_stable_stringify', function () {
+    it('big int', function () {
+        const obj = { age: 23, bigint: BigInt('200'), name: 'ABCD' };
+        expect(stringify(obj)).toMatch('{"age":23,"bigint":"200n","name":"ABCD"}');
+    });
```
buffalojoec, line 7: Nice!

packages/fast-stable-stringify/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "repository": {
+        "type": "git",
+        "url": "git+https://github.com/nickyout/fast-stable-stringify.git"
+    },
+    "author": "Hrushikesh Rao Naidnur",
+    "license": "MIT",
+    "bugs": {
+        "url": "https://github.com/nickyout/fast-stable-stringify/issues"
+    },
+    "homepage": "https://github.com/nickyout/fast-stable-stringify#readme"
```
buffalojoec, line 61: Let's make sure we get all of this updated as well.

buffalojoec (commented):
Awesome. Thanks for rolling this out! Just a few small requests.

packages/fast-stable-stringify/README.md
Hrushi20: What about the LICENSE?

buffalojoec: Ah, good call actually. You'll want to make sure you add the license as we do with the other packages in this repository, ie:
https://github.com/solana-labs/solana-web3.js/blob/master/packages/crypto-impl/LICENSE

Hrushi20:
Can you approve the workflow?

packages/fast-stable-stringify/README.md
Hrushi20: Updated the LICENSE :)

packages/fast-stable-stringify/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+
+## Running tests
+
+```
+npm run test:unit:browser
+npm run test:unit:node 
+```
+
+_Notice: The License of this repository has been changed from GPL-3.0 to MIT as of 2017-08-25. All following commits will fall under the MIT license._
```
buffalojoec, line 33: Drop this line, please.

packages/fast-stable-stringify/package.json
```diff
@@ -0,0 +1,73 @@
+{
+    "name": "@solana/fast-stable-stringify",
+    "version": "1.0.0",
```
buffalojoec, line 3: I know this is a brand-new package, but let's make this `2.0.0-preview.2` like the others!

.changeset/pre.json
```diff
@@ -12,6 +12,7 @@
     "@solana/codecs": "2.0.0-preview.1",
     "@solana/compat": "2.0.0-preview.1",
     "@solana/errors": "2.0.0-preview.1",
+    "@solana/fast-stable-stringify": "1.0.0",
```
buffalojoec, line 15: I don't think this needs to be added here in this PR. This comes after the first one is published, I believe.

buffalojoec (approved):
Thanks so much for doing this! Huge help!

rockstar

steveluscher:
This is awesome! I've been wanting to do this for a while.

We still have the problem that these two datastructures will produce the same hash:

```ts
stringify({foo: 10n});
stringify({foo: "10n"});
```

https://github.com/solana-labs/solana-web3.js/issues/2496

packages/fast-stable-stringify/src/index.ts
```diff
@@ -0,0 +1,77 @@
...
+                    str = '';
+                    i = 0;
+                    while (i < max) {
+                        key = keys[i];
+                        propVal = stringify(val[key], false);
+                        if (propVal !== undefined) {
+                            if (str) {
+                                str += ',';
+                            }
+                            str += JSON.stringify(key) + ':' + propVal;
```
steveluscher, line 51: Now that there is definitely _no_ expectation that the hash can be used with `JSON.parse`, I wonder if we can eliminate this `JSON.stringify` and just output `String(key).replace(/:/g, '\:')`?

steveluscher: Although, I just read this… https://github.com/nickyout/fast-stable-stringify?tab=readme-ov-file#note-about-string-escaping

steveluscher: To be clear: this would just be to reduce the size of the hash string.

Hrushi20: Can you give me the input and output of a sample json?

steveluscher: I'm actually a little hesitant to make this change now, since we no longer have _performance_ benchmarks here. I would not want to make it slower by making the output more compact.

I was talking about this:

```
s({"cool:beans": "bar"}) // {cool\:beans:"bar"}
```

PR #2484, opened 2024-04-15
Maintain transaction lifetime on transactions

mcintyre94:
This PR updates the usages of `NewTransaction` to maintain the transaction lifetime. I should have realised we'd have to extend it!

- The sign functions in the `transactions` package now take a `T extends NewTransaction` and maintain the type T. This means that if you input a transaction with a lifetime, the output transaction will have the same lifetime.

- The signers now return `NewTransaction & TransactionWithLifetime`. I haven't calculated the diff here because they take a transaction message as input and call `compileTransactionMessage`, so their output isn't based on the transaction message input.

PR #2485, opened 2024-04-15
Refactor transaction-confirmation to use new transaction lifetimes

mcintyre94:
This PR refactors transaction-confirmation to take `NewTransaction & {lifetimeConstraint: ...}` as input. This removes the dependency on the old transaction model.

- The new Transaction blockhash lifetime has `lastValidBlockHeight` added
- The new Transaction durable nonce lifetime has `nonceAccountAddress` added. Previously this was pulled from the transaction's instructions, but we no longer have access to decompiled instructions when we have a signed Transaction object.

Note: I'm leaning away from providing a helper to assert on the lifetime of a `NewTransaction` if you don't know it, at least for now. It would require decompiling almost all of the transaction (up to its first instruction), and potentially fetching address lookup tables (if they're used for any of the accounts in the advance nonce instruction). At that point I don't think it provides much value over our existing functions to decompile (using either known address table lookups or an rpc). We can always add this later if we find we need to, it doesn't block anything.


mcintyre94:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2485?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2485** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2485?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2484** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2484?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2481** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2481?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

mcintyre94:
I think it's only used for transaction confirmation, but I haven't exhausively worked through everything using transactions yet so I'm not 100% sure if there are any other places. I think any other use cases will be much smaller though.

It lives in `messageBytes` - sort of. It's a durable nonce transaction if the first instruction is a particular durable nonce instruction. Our check for that also includes a lookup for the accounts, so if you wanted to do it in full then you'd need to combine it with either known address lookup tables or an RPC to resolve them. The accounts in that instruction could be in address lookup tables.

So we could take a `NewTransaction` and add a lifetime to it when we need it. But I think probably one of  the most common use cases in web3js is that you create a transaction (at some point adding a lifetime), send and confirm it. We'd be requiring you to add this decompile (or use a helper that does it for you) on that path, even though you had the transaction message with a known lifetime. It's better performance and better DX.

I think in practice most people would probably just cast to whatever lifetime they were using instead of using a helper that requires address lookup tables or an RPC and not care that much, so the DX is the bigger point I think.

mcintyre94:
This PR has ended up pulling in most of https://github.com/solana-labs/solana-web3.js/pull/2487 because it broke `send-transaction`

lorisleiva:
Gotcha, it sounds like — whilst technically the information is there — the process required to get 100% of the compiled information is not simple enough to justify decompiling on-demand when we need access to that information. I'm happy with that.

I just wanted to make sure we didn't end up with another version of our "we need to keep our data in sync with the compiled bytes" problem — which is the primary reason for this entire refactoring haha.

lorisleiva (approved):

packages/library/src/send-transaction-internal.ts
```diff
@@ -23,7 +28,7 @@ interface SendAndConfirmDurableNonceTransactionConfig
             'getNonceInvalidationPromise' | 'getRecentSignatureConfirmationPromise'
         >,
     ) => Promise<void>;
-    transaction: IDurableNonceTransaction & SendableTransaction;
+    transaction: FullySignedTransaction & { lifetimeConstraint: TransactionDurableNonceLifetime };
```
steveluscher, line 31: Maybe `Readonly<{...}>` all of these?

steveluscher (approved):
Added via Giphy

packages/library/src/send-transaction-internal.ts
mcintyre94: Good idea! I'll do this in an update to https://github.com/solana-labs/solana-web3.js/pull/2486 to avoid rebase conflicts there

PR #2497, opened 2024-04-16
Restore original author to license file for `@solana/fast-stable-stringify`

steveluscher:
I'm pretty sure we are supposed to add a copyright notice for us, but retain the existing one.


buffalojoec (approved):
This would make sense! Any other license-related things we need to do? His is MIT.

steveluscher:
No idea, but I doubt this is less wrong than what's there right now.

PR #2505, opened 2024-04-17
Add glyphs to fix Hash collision with bigint and string

### LINKED ISSUE #2496: Hash collision with bigint and string in `@solana/fast-stable-stringify`
#2482 added native bigint support to the hasher, but now these two datastructures produce the same hash:

```ts
stringify({foo: 10n});
stringify({foo: "10n"});
```

Test case: https://codesandbox.io/p/devbox/goofy-shadow-jsnpv3

Hrushi20:
Added `‽` and `¶` which is a mix of question mark and exclamation. These glyphs are rarely used. Added two glyphs to reduce the possibility of collision.

Closes #2496.

### DIFF
--- a/packages/fast-stable-stringify/src/__tests__/json_test.ts
+++ b/packages/fast-stable-stringify/src/__tests__/json_test.ts
@@ -3,7 +3,7 @@ import stringify from '..';
 describe('test_fast_stable_stringify', function () {
     it('big int', function () {
         const obj = { age: 23, bigint: BigInt('200'), name: 'ABCD' };
-        expect(stringify(obj)).toMatch('{"age":23,"bigint":"200n‽","name":"ABCD"}');
+        expect(stringify(obj)).toMatch('{"age":23,"bigint":"200n‽¶","name":"ABCD"}');
     });
 
     it('only string', function () {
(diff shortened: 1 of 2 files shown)


packages/fast-stable-stringify/src/__tests__/json_test.ts
```diff
@@ -3,7 +3,7 @@ import stringify from '..';
 describe('test_fast_stable_stringify', function () {
     it('big int', function () {
         const obj = { age: 23, bigint: BigInt('200'), name: 'ABCD' };
-        expect(stringify(obj)).toMatch('{"age":23,"bigint":"200n","name":"ABCD"}');
+        expect(stringify(obj)).toMatch('{"age":23,"bigint":"200n‽¶","name":"ABCD"}');
```
steveluscher, line 6: If you put the glyph _outside_ the quotes, you can use a simpler character, and you will never get a collision.

Consider:

```ts
s(100n) // "100"n
s({foo: 100n}) // {"foo":"100"n}
```

steveluscher (changes requested):
This hash never has to parse as JSON, so we can do all kinds of illegal stuff.

steveluscher:
Be sure to rebase; I completely changed up your tests today.

PR #2511, opened 2024-04-17
Remove console.error

HBashanaE:
Remove console.error on 429, not to log error on client side even if the errors are handled.

### DIFF
--- a/packages/library-legacy/src/connection.ts
+++ b/packages/library-legacy/src/connection.ts
@@ -1635,9 +1635,6 @@ function createRpcClient(
         if (too_many_requests_retries === 0) {
           break;
         }
-        console.error(
-          `Server responded with ${res.status} ${res.statusText}.  Retrying after ${waitTime}ms delay...`,
-        );
         await sleep(waitTime);
         waitTime *= 2;
       }


rangatechnologies:
> Remove console.error on 429, not to log error on client side even if the errors are handled.

Yes i am also facing with same issue, this is fundamental. Please remove this.

buffalojoec (commented):
I think there's maybe a better way to do this. Some developers may actually like and/or depend on this message in their app somehow.

What about plumbing a config through `Connection` that can allow you to disable this log message (and likely others)? Something like `noLog: true`, maybe?

HBashanaE:
@buffalojoec Make sense. Let's do like that.

packages/library-legacy/src/connection.ts
```diff
@@ -1532,6 +1532,7 @@ function createRpcClient(
   fetchMiddleware?: FetchMiddleware,
   disableRetryOnRateLimit?: boolean,
   httpAgent?: NodeHttpAgent | NodeHttpsAgent | false,
+  log = false,
```
buffalojoec, line 1535:
```suggestion
  log = true,
```

buffalojoec (changes requested):
Cool! But you can't actually access that paremeter. This function is private and needs to be invoked by the `Connection` class.

buffalojoec (commented):
Great, thanks! I think if we're going to add this parameter, let's use it everywhere else there's a console output someone might want to disable. There's a handful more around the `Connection` class that can use this feature.

steveluscher (changes requested):
At this point, I think I'm unlikely to accept this change unless you can explain more about the problem.

> Yes i am also facing with same issue, this is fundamental.

Other than the log being annoying, is there any other issue? I'm looking for something that affects the operation of your app.

Console logs were never the right thing to do here, but we're here now, and I see only downsides to making a change now.

* This will break any developers who are spying on `console.error` and depend on the ability to scrutinize this log
* Adding config is always bad
  * every bit of config is cognitive overhead for the developer (log what? log to where? log, as opposed to what?)
  * when the config is switched _off_ you still bundle the _code_ – now dead code – controlled by that config. The right way to do config like this is by replacing constants, statically, at build time, so that the build system can reap the dead branches of code.

steveluscher:
I'm going to close this PR at this time; if you have time to revisit this and answer the question above, we can pick it up again!

PR #2512, opened 2024-04-17
Add glyphs to fix Hash collision with bigint and string 

### LINKED ISSUE #2496: Hash collision with bigint and string in `@solana/fast-stable-stringify`

Hrushi20:
Support BigInt and added more tests. Putting Glyph outside the quote avoid collisions. 

```
s(100n) // "100"n
s({foo: 100n}) // {"foo":"100"n}
```

Closes #2496.

### DIFF
--- a/packages/fast-stable-stringify/src/index.ts
+++ b/packages/fast-stable-stringify/src/index.ts
@@ -61,7 +61,7 @@ function stringify(val: unknown, isArrayProp: boolean) {
         case 'undefined':
             return isArrayProp ? null : undefined;
         case 'bigint':
-            return (JSON.stringify(val.toString()) + 'n');
+            return JSON.stringify(val.toString()) + 'n';
         case 'string':
             return JSON.stringify(val);
         default:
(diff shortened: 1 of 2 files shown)


packages/fast-stable-stringify/src/index.ts
```diff
@@ -61,7 +61,7 @@ function stringify(val: unknown, isArrayProp: boolean) {
         case 'undefined':
             return isArrayProp ? null : undefined;
         case 'bigint':
-            return JSON.stringify(val.toString() + 'n');
+            return JSON.stringify(val.toString()) + 'n';
```
steveluscher, line 64: Looks great! Can you think of a reason why we still need the `"` character though? Can we just do ```${val.toString()}n``` and have it work?

Hrushi20: Fixed it.

steveluscher (approved):
Added via Giphy

PR #2565, opened 2024-04-25
 refactor(experimental): graphql: relay: Add support for `id` field in transactions

tibi77:


packages/rpc-graphql/src/schema/transaction.ts
```diff
@@ -65,6 +65,7 @@ export const transactionTypeDefs = /* GraphQL */ `
     A Solana transaction
     """
     type Transaction {
+        id: Signature!
```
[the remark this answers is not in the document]
tibi77: thanks; pushed

packages/rpc-graphql/src/__tests__/transaction-tests.ts
```diff
@@ -68,10 +69,12 @@ describe('transaction', () => {
                 data: {
                     transaction: {
                         blockTime: expect.any(BigInt),
+                        id: expect.any(String),
```
buffalojoec, line 72: You can just put `signature` here.

packages/rpc-graphql/src/__tests__/transaction-tests.ts
```diff
@@ -59,6 +59,7 @@ describe('transaction', () => {
                 query testQuery($signature: Signature!) {
                     transaction(signature: $signature) {
                         blockTime
+                        id
```
buffalojoec, line 62: tiny nit: ID fields are always at the top!

buffalojoec (changes requested):
Cool, so this is the right first step, but our GraphQL client doesn't actually know how to populate that field. You'll need to hook that into the resolver here.

https://github.com/solana-labs/solana-web3.js/blob/35aed3c23bfb4d66585188a9633db3a9786a045b/packages/rpc-graphql/src/resolvers/transaction.ts#L105-L108

tibi77:
> Cool, so this is the right first step, but our GraphQL client doesn't actually know how to populate that field. You'll need to hook that into the resolver here.
> 
> https://github.com/solana-labs/solana-web3.js/blob/35aed3c23bfb4d66585188a9633db3a9786a045b/packages/rpc-graphql/src/resolvers/transaction.ts#L105-L108

should i be adding the ID inside the buildTransactionLoaderArgSetFromResolveInfo as well? https://github.com/solana-labs/solana-web3.js/blob/35aed3c23bfb4d66585188a9633db3a9786a045b/packages/rpc-graphql/src/resolvers/resolve-info/transaction.ts#L79-L88

buffalojoec:
> should i be adding the ID inside the buildTransactionLoaderArgSetFromResolveInfo as well?

_Technically,_ that's not necessary for cache matching - since we're using another field for that - but to follow the expected Relay pattern I guess we should also include the ID.

tibi77:
@buffalojoec I m trying to test locally the tests that are failing inside the pipeline, but i not getting same fails 

```
npm -v; pnpm -v
9.8.1
9.0.6
```
is it just because of the node versions? Or am i missing something else when i run the tests? 
```
node -v
v18.18.2
```

i do run 
```
```

buffalojoec:
> @buffalojoec I m trying to test locally the tests that are failing inside the pipeline, but i not getting same fails
> 
> ```
> npm -v; pnpm -v
> 9.8.1
> 9.0.6
> ```
> 
> is it just because of the node versions? Or am i missing something else when i run the tests?
> 
> ```
> node -v
> v18.18.2
> ```
> 
> i do run
> 
> ```
>  pnpm build --concurrency=100%
> ```
> 
> which seems to be the one that fails

I tend to get some flakiness from the main `pnpm turbo run build` command. But you definitely want to be using Node 20.

Try doing the following, and then running the tests only for the GraphQL library.

```
pnpm turbo clean
rm -rf node_modules
pnpm i
pnpm turbo compile:js compile:typedefs --force
cd packages/rpc-graphql
pnpm run test:unit:node
```

buffalojoec:
Hey @tibi77 I appreciate the contribution here, but I'm taking the Relay effort to a new branch `feature/graphql-relay`, and I've added the ID fields. Thanks!

Keep an eye on the issues if you want to keep hacking.

PR #2578, opened 2024-04-27
 refactor(experimental): graphql: token-2022 extensions: TransferCheckedWithFee

Hrushi20:
This PR adds support for Token-2022's TransferFeeExtension's TransferCheckedWithFee instruction
in the GraphQL schema.

### DIFF
--- a/packages/rpc-graphql/src/__tests__/transaction-tests.ts
+++ b/packages/rpc-graphql/src/__tests__/transaction-tests.ts
@@ -1662,7 +1662,7 @@ describe('transaction', () => {
                 });
             });
 
-            it('transfer-checked-with-fee', async () => {
+            it('transfer-checked-with-fee-with-authority', async () => {
                 expect.assertions(1);
                 const source = /* GraphQL */ `
                     query testQuery($signature: Signature!) {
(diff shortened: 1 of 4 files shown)


packages/rpc-graphql/src/__tests__/transaction-tests.ts
```diff
@@ -895,6 +895,168 @@ describe('transaction', () => {
...
+                                        },
+                                    },
+                                ]),
+                            },
+                        },
+                    },
+                });
+            });
+
+            it('transfer-checked-with-fee-with-multisigAuthority', async () => {
```
buffalojoec, line 979: I think you can probably combine this test and the one above it into the same test. It's exactly the same, just adding one field.

packages/rpc-graphql/src/schema/instruction.ts
```diff
@@ -553,6 +553,21 @@ export const instructionTypeDefs = /* GraphQL */ `
...
+    type SplTokenTransferCheckedWithFee implements TransactionInstruction {
+        programId: Address
+        authority: Account
+        destination: Account
+        feeAmount: TokenAmount
+        mint: Account
+        source: Account
+        tokenAmount: TokenAmount
+        multisigAuthority: Account
+        signers: [Address]
```
buffalojoec, line 568: Here we want to use `[Account]`, but we'll need to roll a new resolver to do multiple accounts. I _think_ it should be fairly trivial to do, but I can roll it soon if you wanted to keep chugging on the extensions. We can flip this to `[Account]` when I do. Unless you want to give it a whirl?

buffalojoec (commented):
Looking really good! Left a few comments.

packages/rpc-graphql/src/schema/instruction.ts
Hrushi20: Should I start by creating a new function `resolveAccounts` function with below signature 


```typescript

export const resolveAccounts = (fieldName?: string) => {
   return async (
       parent: { [x: string]: Array<Address> },
       args: { address?: Address; commitment?: Commitment; minContextSlot?: Slot },
       context: RpcGraphQLContext,
       info: GraphQLResolveInfo): Promise<AccountResult | null> => {

      const addresses = fieldName ? parent[fieldName] : null; // This will be an Array of address. 
      // Below handle logic for multiple address

}
}
```

Changing the existing function to return multiple address could cause things to break.

buffalojoec: Let me hack around on this for a minute. I think there's actually a simple way to do this with the way we've laid down the batch loader. I'll circle back here ASAP; working on it now.

buffalojoec: But yes, the interface you've posted is correct. I think we might be able to just roll that function signature as you've posted it, but then simply "map" the response from `resolveAccount`.

buffalojoec: I'm going to roll a new `resolveMultipleAccounts` resolver, as well as a `multipleAccounts` root query, as I mentioned [here](https://github.com/solana-labs/solana-web3.js/pull/2579#discussion_r1585502257). So go ahead and leave this as `[Address]` for now.

packages/rpc-graphql/src/__tests__/transaction-tests.ts
buffalojoec: Update this test to be combined as I mentioned [here](https://github.com/solana-labs/solana-web3.js/pull/2579#discussion_r1585501070), and then I think this PR is good to go!

buffalojoec (approved):
You _might_ have to rebase onto my latest merges, but looks great! Thanks!

Hrushi20:
This can be merged.

PR #2587, opened 2024-05-01
refactor(experimental): graphql: add `multipleAccounts` root query

buffalojoec:
#### Problem
The GraphQL resolver currently supports root queries for `account` and
`programAccounts`, but not for `multipleAccounts`. A user can craft multiple
`account` queries in one source, but why go through the trouble if all fields and
input parameters are the same?

Additionally, if the GraphQL resolver had a `resolveMultipleAccounts` resolver
function, it would make things much easier to add list-based account lookups
on schema types that would otherwise return `[Address]`. With this new resolver,
they can instead return `[Account]`, which would allow nested queries on these
lists.

#### Summary of Changes
Introduce the `mutipleAccounts` root query.

I've reimplemented the same tests for this new `multipleAccounts` query
as those already provided for the existing `accounts` query.

- `src/__tests__/account-test.ts`
- `src/loaders/__tests__/account-loader-test.ts`

ie:

- `src/__tests__/multiple-accounts-test.ts`
- `src/loaders/__tests__/multiple-accounts-loader-test.ts`

### DIFF
--- a/packages/rpc-graphql/src/schema/root.ts
+++ b/packages/rpc-graphql/src/schema/root.ts
@@ -2,6 +2,7 @@ export const rootTypeDefs = /* GraphQL */ `
     type Query {
         account(address: Address!, commitment: Commitment, minContextSlot: Slot): Account
         block(slot: Slot!, commitment: CommitmentWithoutProcessed): Block
+        multipleAccounts(addresses: [Address!]!, commitment: Commitment, minContextSlot: Slot): [Account]
         programAccounts(
             programAddress: Address!
             commitment: Commitment
--- a/packages/rpc-graphql/src/__tests__/multiple-accounts-test.ts
+++ b/packages/rpc-graphql/src/__tests__/multiple-accounts-test.ts
@@ -1,0 +1,1220 @@
+import {
+    GetAccountInfoApi,
+    GetBlockApi,
+    GetMultipleAccountsApi,
+    GetProgramAccountsApi,
+    GetTransactionApi,
+    Rpc,
+} from '@solana/rpc';
+
+import { createRpcGraphQL, RpcGraphQL } from '../index';
+import { createLocalhostSolanaRpc } from './__setup__';
+
+type GraphQLCompliantRpc = Rpc<
+    GetAccountInfoApi & GetBlockApi & GetMultipleAccountsApi & GetProgramAccountsApi & GetTransactionApi
+>;
+
+describe('multiple accounts', () => {
+    let rpc: GraphQLCompliantRpc;
+    let rpcGraphQL: RpcGraphQL;
+    beforeEach(() => {
+        rpc = createLocalhostSolanaRpc();
+        rpcGraphQL = createRpcGraphQL(rpc);
+    });
+
+    describe('basic queries', () => {
+        const variableValues = {
+            addresses: [
+                // See scripts/fixtures/spl-token-token-account.json
+                'AyGCwnwxQMCqaU4ixReHt8h5W4dwmxU7eM3BEQBdWVca',
+                // See scripts/fixtures/spl-token-mint-account.json
+                'Gh9ZwEmdLJ8DscKNTkTqPbNwLNNBjuSzaG9Vp2KGtKJr',
+            ],
+        };
+        it("can query multiple accounts' lamports balance", async () => {
+            expect.assertions(1);
+            const source = /* GraphQL */ `
+                query testQuery($addresses: [Address!]!) {
+                    multipleAccounts(addresses: $addresses) {
+                        lamports
+                    }
+                }
+            `;
+            const result = await rpcGraphQL.query(source, variableValues);
+            expect(result).toMatchObject({
+                data: {
+                    multipleAccounts: expect.arrayContaining([
+                        {
+                            lamports: 10290815n,
+                        },
+                    ]),
+                },
+            });
+        });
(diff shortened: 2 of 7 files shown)


buffalojoec:
* **#2587** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2587?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

 Join @buffalojoec and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

steveluscher (commented):
I think there are probably a number of good and perhaps surprising reasons not to make a plural field version of this, but I’ll have to spend some time explaining why. I’ll get to it tomorrow.

packages/rpc-graphql/src/schema/root.ts
```diff
@@ -2,6 +2,7 @@ export const rootTypeDefs = /* GraphQL */ `
     type Query {
         account(address: Address!, commitment: Commitment, minContextSlot: Slot): Account
         block(slot: Slot!, commitment: CommitmentWithoutProcessed): Block
+        multipleAccounts(addresses: [Address!]!, commitment: Commitment, minContextSlot: Slot): [Account]
```
steveluscher, line 5: Usually just follow pluralization rules.
```suggestion
        accounts(addresses: [Address!]!, commitment: Commitment, minContextSlot: Slot): [Account]
```

steveluscher, line 5: This field, as written, returns a nullable array. This will force the consumer to treat `null` and `[]` the same with two different checks. What you actually want is to unconditionally return an array, maybe empty, but in any case with `null` for every unfetchable account.
```suggestion
        multipleAccounts(addresses: [Address!]!, commitment: Commitment, minContextSlot: Slot): [Account]!
```

steveluscher (changes requested):
Alright, so let's talk about **plural fields**.

Let's break down the use cases for multiple account fetches.

# Use cases

## Case 1: You know the accounts statically

If your app straight up knows which accounts it needs, either literally statically (the addresses are constants in the code) or via arguments (there are always _n_ accounts it needs, and the addresses of those will be supplied as inputs to the application) then it should declare them each in the query.

```tsx
const data = useFragment(graphql`
    fragment Home_query on Query {
        userProfileAccount: account(address: $userProfileAccountAddress) {
            data
        }
        gameStateAccount: account(address: $gameStateAccountAddress) {
            data
        }
    }
`, props.query);
```

### Pros

* Each `account()` fetch can declare its own arguments (eg. `commitment`) independent of the others.
* Each account can be _refetched_ independently.

## Case 2: You need an unknown number of accounts, you don't have the addresses, but you know the root

In this case, the schema usually offers a field (consider `walletAccounts` in the example below) that resolves zero-or-more objects given some context/parent.

```ts
const data = useFragment(graphql`
    fragment UserProfile_user on User {
        walletAccounts(dataSlice: { length: 0, offset: 0 }) {
            ...Wallet_account
            address
        }
    }
`, props.user);
return (
    <ul>
        {data.walletAccounts.map(account => 
            <Wallet
                account={account}
                key={account.address}
            />
        }
    </ul>
);
```

### Pros

* You need not know the addresses of the entities up front to be able to fetch zero-or-more of them

### Cons

* The entire fetch for the list of accounts _must_ share the same arguments (eg. `dataSlice`).
* You get all of the entities over the wire, whether your app has space to render them or not. Plural fields offer no way to paginate.

## Case 3: You need a dynamic list of accounts for which you have the addresses

Consider an app that lets the user check the balance of multiple accounts, by literally supplying the accounts.

You could do this:

```ts
const [data, refetch] = useRefetchableFragment(graphql`
    fragment AccountBalances_query on Query
        @refetchable(queryName: "AccountBalancesRefetchQuery") {
        accounts(accounts: $accounts) {
            ...Balance_account
            address
        }
    }
`, props.query);
return (
    <>
        <textarea onChange={(e) => {
            const addresses = e.target.value.split('\n');
            refetch({ accounts:  addresses });
        }}></textarea>
        <ul>
            {data.accounts.map(account =>
                <Balance
                    account={account}
                    key={account.address}
                />
            }
        </ul>
    </>
);
```

### Cons

* Every time an address is added or removed, the _entire_ account list gets fetched again, even the ones you already have locally.
* The entire fetch for the list of accounts _must_ share the same arguments (eg. `commitment`).

# Discussion

1. The only use case that requires a plural _root_ field is case 3.
2. Case 3 can be implemented using a dynamic set of _entrypoints_ instead of a single page that queries a plural root field. Imagine each result UI being, itself, a mini-page based on the address as input. When the user adds or removes an address to their list, a new entrypoint is fetched and rendered, but the others don't change. Think of entrypoints as islands of one or more root queries that make up a larger app, and share data with the entire app.

buffalojoec:
@steveluscher Thanks for the detailed breakdown. It sounds to me like this is probably not The Way™. I'm mostly fixated on the fact that case 3 is the most valid use for such a query, and the cons I think are too much to bear.

Admittedly, I viewed Con 2 as a potential Pro, but now I'm starting to see how it could very easily become either one.

I think I may abandon this change.

buffalojoec:
@steveluscher Thanks again for the insight here. As I mentioned above, I'm going to revisit this idea once I've explored connections and pagination. In the meantime, I'll close this one.

PR #2606, opened 2024-05-03
refactor(experimental): use commonjs package type

lorisleiva:
The PR makes all packages of the new web3.js monorepo use the `"commonjs"` package type and, therefore, uses the explicit `.mjs` extension when building for ESM.

### Context

When exporting dual package libraries, we can either:
1. Use `"type": "module"` and export CJS builds using `.cjs`.
2. Use `"type": "commonjs"` and export ESM builds using `.mjs`.
3. Use whatever as a package type and use explicit extensions for both ESM and CJS.

Currently we are using strategy 1, this PR moves towards strategy 2 and 3 as we use explicit extensions for both exports. Thus, effectively, this PR continues to support both ESM and CJS, just using a different strategy.

### Problem

The problem with strategy 1 is that, having `"type": "module"` causes a lot of issues in how TypeScript resolves type declaration files, especially when using the `NodeNext` module resolution.

The work-around we have been using so far is to add a post-processing step to our build pipeline that adds `.js` or `/index.js` suffixes to relative imports inside type declaration files. However, this isn’t enough as the consumer packages must adjust their TypeScript config file and/or make their package a `"type": "module"` for TypeScript to resolve types properly.

For instance, a consumer package using the `NodeNext` resolution and  `"type": "commonjs"` — or not providing any explicit package type — will get TypeScript errors when trying to import our packages.

### Solutions

One way to fix this is to export two sets of TypeScript declarations. One for CJS and one for ESM that includes the post-processing step mentioned above. We then need to configure the `exports` property of our `package.json` to include both TypeScript declarations based on the `import` and `require` conditions.

OR

A much simpler solution — and one that popular JavaScript monorepos such as React and Stripe uses — is to use `"type": "commonjs"` on our packages. By doing this, we continue to support both ESM and CJS environments and TypeScript will understand how to resolve our type declarations without any post-processing step and without forcing consumer packagers to use one package type or the other. It just works.

This PR implements the latter.


lorisleiva:
* **#2606** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2606?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

mcintyre94 (approved):

steveluscher (approved):
Added via Giphy

PR #2611, opened 2024-05-03
refactor(experimental): graphql: token-2022 extensions: Reallocate Instruction

Hrushi20:
This PR adds support for Token-2022's Reallocate instruction in the GraphQL schema.


packages/rpc-graphql/src/__tests__/transaction-tests.ts
```diff
@@ -1480,6 +1480,89 @@ describe('transaction', () => {
...
+                                        account: {
+                                            address: expect.any(String),
+                                        },
+                                        extensionTypes: expect.any([expect.any(String)]),
+                                        multisigOwner: {
+                                            address: expect.any(String),
+                                        },
+                                        owner: {
+                                            address: expect.any(String),
+                                        },
```
buffalojoec, line 1550: `owner` should be null here.

packages/rpc-graphql/src/schema/instruction.ts
```diff
@@ -670,6 +670,20 @@ export const instructionTypeDefs = /* GraphQL */ `
         signers: [Address]
     }
 
+    """
+    SplToken-2022: Reallocate instruction
+    """
+    type SplTokenReallocate implements TransactionInstruction {
+        programId: Address
+        account: Account
+        extensionTypes: [String]
```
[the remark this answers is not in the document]
Hrushi20: Thanks. I'll add that as well.

packages/rpc-graphql/src/schema/instruction.ts
```diff
@@ -1093,6 +1093,22 @@ export const instructionTypeDefs = /* GraphQL */ `
         metadata: Account
         updateAuthority: Account
     }
+    """
+    SplToken-2022: Reallocate instruction
+    """
+    type SplTokenReallocate implements TransactionInstruction {
+        programId: Address
+        account: Account
+        extensionTypes: [String]
```
buffalojoec, line 1102: Just need to update this field and this one should be good to go.

packages/rpc-graphql/src/schema/types.ts
```diff
@@ -86,4 +86,32 @@ export const typeTypeDefs = /* GraphQL */ `
         BASE_58
         BASE_64
     }
+
+    enum ExtensionType {
```
buffalojoec, line 90: Actually I think you'll also need to define a resolver to go to and from string values here. You'll see the others under `resolvers/types.ts`

buffalojoec:
@Hrushi20 commandeered #2811.

PR #2642, opened 2024-05-06
refactor(experimental): graphql: group schema type resolvers

buffalojoec:
Grouping all schema type resolvers into a common subdirectory `type-resolvers`.

PR #2641, opened 2024-05-06
refactor(experimental): graphql: group schema typedefs

buffalojoec:
Grouping all schema type definitions into a common subdirectory `type-defs`.

PR #2643, opened 2024-05-06
refactor(experimental): graphql: slot-based nested queries

### LINKED ISSUE #1822: [GraphQL]: Add support for slot-based nested queries
## Motivation
Throughout the GraphQL schema, fields populated by an `Address` have been
given the type `Account`, which enabled nested queries on that address. Take,
for example, a lookup table account.

```gql
"""
A lookup table account
"""
type LookupTableAccount implements Account {
    address: Address
    data(encoding: AccountEncoding!, dataSlice: DataSlice): String
    executable: Boolean
    lamports: Lamports
    ownerProgram: Account
    space: BigInt
    rentEpoch: Epoch
    addresses: [Address]
    authority: Account
    deactivationSlot: Slot
    lastExtendedSlot: Slot
    lastExtendedSlotStartIndex: Int
}
```

You can see fields like `ownerProgram` and `authority` are queryable as
`Account` types, allowing full `account` queries to be performed on these nested
fields.

We can enable the same types of nested queries for `block` on any fields that are
populated with `Slot`.

## Details
This is actually quite trivial to implement, given the relationship of type definitions
and resolvers defined in this library. Here's an example.

```gql
"""
A lookup table account
"""
type LookupTableAccount implements Account {
    address: Address
    data(encoding: AccountEncoding!, dataSlice: DataSlice): String
    executable: Boolean
    lamports: Lamports
    ownerProgram: Account
    space: BigInt
    rentEpoch: Epoch
    addresses: [Address]
    authority: Account
    deactivationSlot: Block   # `Block` type, instead of `Slot`
    lastExtendedSlot: Block   # `Block` type, instead of `Slot`
    lastExtendedSlotStartIndex: Int
}
```

```ts
LookupTableAccount: {
    authority: resolveAccount('authority'),
    data: resolveAccountData(),
    ownerProgram: resolveAccount('ownerProgram'),
    // Block resolvers hooked up.
    deactivationSlot: resolveBlock('deactivationSlot'),
    lastExtendedSlot: resolveBlock('lastExtendedSlot'),
},
```

The main question here is "should we do this?" and unless we can produce
salient arguments against it, I believe we should.

buffalojoec:
As the linked issue describes, the GraphQL resolver sports full support for
nested `Account` queries wherever a field has type `Address`. Instead of
defining that field to simply be of type `Address`, it's defined with type `Account`,
which enables nested account queries on that field based on the address
returned.

In this PR, we're taking every field with type `Slot` and replacing it with `Block`,
to enable nested block queries in the same fashion.

Closes #1822.


buffalojoec:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2643?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2643** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2643?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2642** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2642?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2641** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2641?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

 Join @buffalojoec and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

packages/rpc-graphql/src/loaders/__tests__/transaction-loader-test.ts
```diff
@@ -55,7 +61,9 @@ describe('account loader', () => {
             const source = /* GraphQL */ `
                 query testQuery($signature: Signature!) {
                     transaction(signature: $signature) {
-                        slot
+                        slot {
```
steveluscher, line 64: Doesn't this need to be renamed to `block` then, to reflect the new return type?

packages/rpc-graphql/src/resolvers/block.ts
```diff
@@ -23,7 +23,7 @@ export const resolveBlock = (fieldName?: string) => {
     ) => {
         const slot = fieldName ? parent[fieldName] : args.slot;
 
-        if (slot) {
+        if (slot != undefined) {
             if (onlyFieldsRequested(['slot'], info)) {
                 return { slot };
             }
```
steveluscher, line 29: 👍

packages/rpc-graphql/src/schema/type-defs/account.ts
```diff
@@ -57,8 +57,8 @@ export const accountTypeDefs = /* GraphQL */ `
         rentEpoch: Epoch
         addresses: [Address]
         authority: Account
-        deactivationSlot: Slot
-        lastExtendedSlot: Slot
+        deactivationSlot: Block
+        lastExtendedSlot: Block
```
steveluscher, line 61: Still think these should be renamed to reflect the new return type.

packages/rpc-graphql/src/resolvers/block.ts
steveluscher: How does this behave if I do:

```graphql
{
    block: {
        __id
        __type
        slot
    }
}
```

Relay, for instance, basically adds `__id` and `__type` to every subselection automatically.

steveluscher (changes requested):
I think if we're going to drink deep from this cup, we should probably rename everything to reflect the return type.

packages/rpc-graphql/src/resolvers/block.ts
buffalojoec: Yeah GraphQL-JS adds those!

packages/rpc-graphql/src/schema/type-defs/account.ts
buffalojoec: I truly waffled on this a bit actually. I just wanted to be careful about straying too far from the defined data, such as an account whose fields are "something-slot" and now they have to call it "block".

I couldn't decide if this was acceptable, really.

buffalojoec:
> I think if we're going to drink deep from this cup, we should probably rename everything to reflect the return type.

I'm interested to know your thoughts on whether or not we _should_ drink from this cup some more.

packages/rpc-graphql/src/resolvers/block.ts
steveluscher: What I mean to ask is if this resolver will get deopted (ie. make the fetch) if the client asks for `__id` and `__type`.

packages/rpc-graphql/src/schema/type-defs/account.ts
steveluscher: If we leave it this way, it's going to lead to a lot of aggravating code like:

```ts
if (data.slot.slot < 123n) {
    // ...
}
```

packages/rpc-graphql/src/resolvers/block.ts
buffalojoec: Ohhh yeah it might actually. I'll add a PR for this. Good call.

buffalojoec:
I hate this PR. I think if people want this feature we can add it later, but it's kind of a beast.

PR #2652, opened 2024-05-06
refactor(experimental): graphql: token-2022 extensions: DepositConfidentialTransfer

nasjuice:
This PR adds support for Token-2022's DepositConfidentialTransfer  instruction in the GraphQL schema.

### DIFF
--- a/packages/rpc-graphql/src/schema/instruction.ts
+++ b/packages/rpc-graphql/src/schema/instruction.ts
@@ -873,7 +873,7 @@ export const instructionTypeDefs = /* GraphQL */ `
     type SplTokenDepositConfidentialTransfer implements TransactionInstruction {
         programId: Address
         amount: BigInt
-        decimals: Int
+        decimals: BigInt # FIXME:*
         destination: Account
         mint: Account
         multisigOwner: Account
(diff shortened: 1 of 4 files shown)


packages/rpc-graphql/src/schema/instruction.ts
```diff
@@ -866,6 +867,21 @@ export const instructionTypeDefs = /* GraphQL */ `
         signers: [Address]
     }
 
+    """
+    SplToken-2022: DepositConfidentialTransfer instruction
+    """
+    type SplTokenDepositConfidentialTransfer implements TransactionInstruction {
+        programId: Address
+        amount: BigInt
+        decimals: BigInt
```
buffalojoec, line 876: Decimals is a `u8` so this can just be `Int`.

buffalojoec (commented):
Looks great, just one nit.

nasjuice:
Sounds good, changed it 

I was a bit confused some places had these FIXME
image

Now it makes sense, I guess they are meant to be refactored to Int

buffalojoec (approved):
groovy

buffalojoec:
> I was a bit confused some places had these FIXME

Ahh wait, sorry, good call here. Web3.js 2.0 will turn those to `bigint`. Mind flipping it back, and adding the "FIXME" comment?

buffalojoec (changes requested):
Just marking this as non-approved until the decimals fix. Nice catch on that!

nasjuice:
Sounds good, fixed now

buffalojoec (approved):
Thanks for that, ship it!

PR #2687, opened 2024-05-08
refactor(experimental): graphql: token-2022 extensions: InitializeTokenGroup

nasjuice:
This PR adds support for Token-2022's InitializeTokenGroup instruction in the GraphQL schema.

### DIFF
--- a/packages/rpc-graphql/src/resolvers/instruction.ts
+++ b/packages/rpc-graphql/src/resolvers/instruction.ts
@@ -369,6 +369,12 @@ export const instructionResolvers = {
         delegate: resolveAccount('delegate'),
         mint: resolveAccount('mint'),
     },
+    SplTokenInitializeTokenGroup: {
+        group: resolveAccount('group'),
+        mint: resolveAccount('mint'),
+        mintAuthority: resolveAccount('mintAuthority'),
+        updateAuthority: resolveAccount('updateAuthority'),
+    },
     SplTokenInitializeTransferFeeConfig: {
         mint: resolveAccount('mint'),
         transferFeeConfigAuthority: resolveAccount('transferFeeConfigAuthority'),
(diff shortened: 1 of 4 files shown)


packages/rpc-graphql/src/schema/instruction.ts
```diff
@@ -1014,6 +1014,18 @@ export const instructionTypeDefs = /* GraphQL */ `
...
+    """
+    type SplTokenInitializeTokenGroup implements TransactionInstruction {
+        programId: Address
+        group: Account
+        maxSize: BigInt
+        mint: Account
+        mintAuthority: Account
+        updateAuthority: Account
+    }
+
```
buffalojoec, line 1028: So, this instruction set is actually an [interface](https://github.com/solana-labs/solana-program-library/blob/master/token-group/interface/src/instruction.rs), so is SPL Token Metadata, so let's actually define this separately from Token-2022.

```suggestion
    """
    Spl Token Group: InitializeGroup instruction
    """
    type SplTokenGroupInitializeGroup implements TransactionInstruction {
        programId: Address
        group: Account
        maxSize: BigInt
        mint: Account
        mintAuthority: Account
        updateAuthority: Account
    }
```

packages/rpc-graphql/src/resolvers/instruction.ts
```diff
@@ -882,6 +888,9 @@ export const instructionResolvers = {
                     if (jsonParsedConfigs.instructionType === 'initializeConfidentialTransferFeeConfig') {
                         return 'SplTokenInitializeConfidentialTransferFeeConfig';
                     }
+                    if (jsonParsedConfigs.instructionType === 'initializeTokenGroup') {
+                        return 'SplTokenInitializeTokenGroup';
+                    }
```
buffalojoec, line 893: This is still in the right spot, since for now the RPC is only going to parse this for Token-2022.

buffalojoec (commented):
Looks great! Just made a suggestion on interface naming conventions. Let's follow that pattern when we get to Token Metadata as well.

buffalojoec (approved):
Awesome! Thanks for continued great work.

PR #2699, opened 2024-05-09
refactor(experimental): graphql: token-2022 extensions account state: confidentialTransferAccount

Hrushi20:
This PR adds Token-2022 extension parsed account state support for ConfidentialTransferAccount.

Ref https://github.com/solana-labs/solana-web3.js/issues/2644.


buffalojoec (approved):
Looks great, thanks!

PR #2702, opened 2024-05-10
Update web3.js README for new TransactionMessage API

steveluscher:
I did not test these code snippets at all; I just eyeballed them.

PR #2701, opened 2024-05-10
Convert `tsx` codeblocks in README to `ts` because there's no React code in here

steveluscher:

PR #2703, opened 2024-05-10
A utility to estimate the compute unit consumption of a transaction message

steveluscher:
# Summary

Correctly budgeting a compute unit limit for your transaction message can increase the probabilty that your transaction will be accepted for processing. If you don't declare a compute unit limit on your transaction, validators will assume an upper limit of 200K compute units (CU) per instruction.

Since validators have an incentive to pack as many transactions into each block as possible, they may choose to include transactions that they know will fit into the remaining compute budget for the current block over transactions that might not. For this reason, you should set a compute unit limit on each of your transaction messages, whenever possible.

This is a utility that helps you to estimate the actual compute unit cost of a given transaction message using the simulator.

## Example

```ts
import { getSetComputeLimitInstruction } from "@solana-program/compute-budget";
import {
  createSolanaRpc,
  getComputeUnitEstimateForTransactionMessageFactory,
  pipe,
} from "@solana/web3.js";

// Create an estimator function.
const rpc = createSolanaRpc("http://127.0.0.1:8899");
const getComputeUnitEstimateForTransactionMessage =
  getComputeUnitEstimateForTransactionMessageFactory({
    rpc,
  });

// Create your transaction message.
const transactionMessage = pipe(
  createTransactionMessage({ version: "legacy" })
  /* ... */
);

// Request an estimate of the actual compute units this message will consume.
const computeUnitsEstimate = await getComputeUnitEstimateForTransactionMessage(
  transactionMessage
);

// Set the transaction message's compute unit budget.
const transactionMessageWithComputeUnitLimit =
  prependTransactionMessageInstruction(
    getSetComputeLimitInstruction({ units: computeUnitsEstimate }),
    transactionMessage
  );
```

## Notes

- The compute unit estimate is just that &ndash; an estimate. The compute unit consumption of the actual transaction might be higher or lower than what was observed in simulation. Unless you are confident that your particular transaction message will consume the same or fewer compute units as was estimated, you might like to augment the estimate by either a fixed number of CUs or a multiplier.
- If you are preparing an _unsigned_ transaction, destined to be signed and submitted to the network by a wallet, you might like to leave it up to the wallet to determine the compute unit limit. Consider that the wallet might have a more global view of how many compute units certain types of transactions consume, and might be able to make better estimates of an appropriate compute unit budget.

# Test Plan

```shell
cd packages/library
```


steveluscher:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2703?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2703** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2703?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2702** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2702?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2701** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2701?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

packages/library/README.md
```diff
@@ -878,6 +878,46 @@ const signedTransaction = await signTransaction([signer], transactionMessageWith
 // => "Property 'lifetimeConstraint' is missing in type"
 ```
 
+### Calibrating A Transaction Message's Compute Unit Budget
+
+Correctly budgeting a compute unit limit for your transaction message can increase the probabilty that your transaction will be accepted for processing. If you don't declare a compute unit limit on your transaction, validators will assume an upper limit of 200K compute units (CU) per instruction.
```
steveluscher, line 883: > …validators will assume an upper limit of 200K compute units (CU) per instruction.

I didn't look this up. Is this true?

packages/library/README.md
```diff
@@ -878,6 +878,46 @@ const signedTransaction = await signTransaction([signer], transactionMessageWith
...
+
+Since validators have an incentive to pack as many transactions into each block as possible, they may choose to include transactions that they know will fit into the remaining compute budget for the current block over transactions that might not. For this reason, you should set a compute unit limit on each of your transaction messages, whenever possible.
```
steveluscher, line 885: cc/ @joncinque to check over.

packages/library/README.md
```diff
@@ -878,6 +878,46 @@ const signedTransaction = await signTransaction([signer], transactionMessageWith
...
+    getSetComputeLimitInstruction({ units: computeUnitsEstimate }),
+    transactionMessage,
+);
+```
+
+> [!WARNING]
+> The compute unit estimate is just that &ndash; an estimate. The compute unit consumption of the actual transaction might be higher or lower than what was observed in simulation. Unless you are confident that your particular transaction message will consume the same or fewer compute units as was estimated, you might like to augment the estimate by either a fixed number of CUs or a multiplier.
+
+> [!NOTE]
+> If you are preparing an _unsigned_ transaction, destined to be signed and submitted to the network by a wallet, you might like to leave it up to the wallet to determine the compute unit limit. Consider that the wallet might have a more global view of how many compute units certain types of transactions consume, and might be able to make better estimates of an appropriate compute unit budget.
```
steveluscher, line 919: Thoughts on this, @jordaaash?

steveluscher: Feel free to tag wallet folks.

packages/library/src/compute-limit-internal.ts
```diff
@@ -0,0 +1,177 @@
...
+                ...simulateConfig,
+                encoding: 'base64',
+                replaceRecentBlockhash: !isDurableNonceTransactionMessage,
+                sigVerify: false,
+            })
+            .send({ abortSignal });
+        if (unitsConsumed == null) {
+            // This should never be hit, because all RPCs should support `unitsConsumed` by now.
+            throw new SolanaError(SOLANA_ERROR__TRANSACTION__FAILED_TO_ESTIMATE_COMPUTE_LIMIT);
+        }
```
steveluscher, line 167: cc/ @joncinque

packages/library/README.md
mcintyre94: We looked into this a bit when we were looking at priority fees a while back. It's accurate, 200k per instruction (excluding those to compute budget program) with a 1.4M cap (which is the max CU allowed per transaction): https://github.com/solana-labs/solana/blob/4293f11cf13fc1e83f1baa2ca3bb2f8ea8f9a000/program-runtime/src/compute_budget.rs#L13

packages/library/src/compute-limit-internal.ts
```diff
@@ -0,0 +1,187 @@
...
+            transactionMessage,
+        );
+    }
+    /**
+     * STEP 2: Ensure that the message has a `SetComputeLimit` instruction. The set compute limit
+     *         instruction itself consumes compute units, so it must be included in the simulation.
+     */
+    const existingSetComputeUnitLimitInstructionIndex =
+        transactionMessage.instructions.findIndex(isSetComputeLimitInstruction);
+    const maxComputeUnitLimitInstruction = createComputeUnitLimitInstruction(4294967295 /* U32::MAX */);
```
mcintyre94, line 144: Speculative: this might increase the estimate because it's higher than the actual max allowed, which is 1.4M - so there might be CUs consumed normalising it to 1.4M that won't be consumed when it's set to a realistic value.

packages/library/src/compute-limit-internal.ts
```diff
@@ -0,0 +1,187 @@
...
+     * STEP 3: Simulate the transaction to measure its compute unit consumption.
+     */
+    const compiledTransaction = compileTransaction(compilableTransactionMessage);
+    const wireTransactionBytes = getBase64EncodedWireTransaction(compiledTransaction);
+    try {
+        const {
+            value: { unitsConsumed },
+        } = await rpc
+            .simulateTransaction(wireTransactionBytes, {
+                ...simulateConfig,
```
mcintyre94, line 168: Since this is `library` and we can be opinionated, should we default the commitment to `Confirmed` instead of simulate's default of `Finalized`? It's most likely how they'll eventually send the transaction, and can still be explicitly set if needed.

packages/library/src/compute-limit-internal.ts
```diff
@@ -0,0 +1,187 @@
...
+            })
+            .send({ abortSignal });
+        if (unitsConsumed == null) {
+            // This should never be hit, because all RPCs should support `unitsConsumed` by now.
+            throw new SolanaError(SOLANA_ERROR__TRANSACTION__FAILED_TO_ESTIMATE_COMPUTE_LIMIT);
+        }
+        // FIXME: The simulation response returns compute units as a u64, but the `SetComputeLimit`
+        // instruction only accepts a u32. Until this changes, downcast it.
+        const downcastUnitsConsumed = unitsConsumed > 4294967295n ? 4294967295 : Number(unitsConsumed);
+        return downcastUnitsConsumed;
```
mcintyre94, line 181: I'm not sure how much knowledge we want to bake in here, but in practice if this is >1.4M it's an error

mcintyre94 (approved):

packages/library/src/compute-limit-internal.ts
steveluscher: Oh nice catch! I tried this.

|1.4M|u32::MAX|
|---|---|
|||

Same number of CUs. Looks like the `min()` gets called unconditionally, so takes the same amount?

https://github.com/anza-xyz/agave/blob/75cc70d3b6671362bf2605326b05e623b4a0e679/program-runtime/src/compute_budget_processor.rs#L128C9-L133

1. Bad: being wrong about this in the future
2. Also bad: hardcoding `MAX_COMPUTE_UNIT_LIMIT` into JS clients then not being able to take it back.

I think I'll leave it the way it is here.

packages/library/src/compute-limit-internal.ts
steveluscher: I think I just want to let the server tell me, and if it tells me a wrong thing then oh well.

packages/library/src/compute-limit-internal.ts
steveluscher: Actually the absence of an effect here probably has more to do with this.

https://github.com/anza-xyz/agave/blob/d87e23d8d91c32d5f2be2bb3557c730bee1e9434/programs/compute-budget/src/lib.rs#L3

packages/library/src/compute-limit-internal.ts
```diff
@@ -0,0 +1,188 @@
...
+
+function createComputeUnitLimitInstruction(units: number): IInstruction<typeof COMPUTE_BUDGET_PROGRAM_ADDRESS> {
+    const data = new Uint8Array(5);
+    data[0] = SET_COMPUTE_UNIT_LIMIT_INSTRUCTION_INDEX;
+    getU32Encoder().write(units, data, 1 /* offset */);
+    return Object.freeze({
+        data,
+        programAddress: COMPUTE_BUDGET_PROGRAM_ADDRESS,
+    });
+}
```
joncinque, line 49: Unless it's exposed somewhere else, it might be nice to expose this too, and add an example of adding the compute unit limit instruction after getting the units consumed.

steveluscher: It's actually here:

https://github.com/solana-program/compute-budget/blob/f425ee62857b3154c85dc0f551898d5cfef67fa8/clients/js/src/generated/instructions/setComputeUnitLimit.ts#L78-L95

I would have used that actual function here, but unfortunately it would cause a circular dependency.

joncinque: Gotcha, that works

PR #2704, opened 2024-05-10
improve: add support of compactUpdateVoteState instruction

EugeneButusov:


packages/library-legacy/src/programs/vote.ts
```diff
@@ -238,6 +250,41 @@ export class VoteInstruction {
...
+      VOTE_INSTRUCTION_LAYOUTS.CompactUpdateVoteState,
+      instruction.data,
+    );
+    let lockoutSlot = root;
+
+    return {
+      voteAccount: instruction.keys[0].pubkey,
+      voteAuthority: instruction.keys[1].pubkey,
+      voteStateUpdate: {
+        lockouts: lockoutOffsets.map(
```
EugeneButusov, line 272: this transformation is made to be compliant with the output provided by `api.mainnet-beta.solana.com`, since node returns lockouts, while instruction for "compact" contains offset values.

packages/library-legacy/src/programs/vote.ts
```diff
@@ -96,6 +97,17 @@ export type UpdateValidatorIdentityParams = {
   nodePubkey: PublicKey;
 };
 
+export type CompactUpdateVoteStateParams = {
```
buffalojoec, line 100: nit: can you document this type, like the other parameter types?

packages/library-legacy/src/programs/vote.ts
```diff
@@ -96,6 +97,17 @@ export type UpdateValidatorIdentityParams = {
...
+  voteAccount: PublicKey,
```
buffalojoec, line 101: nit: Just to follow the pattern of the others:

```suggestion
  votePubkey: PublicKey,
```

packages/library-legacy/src/programs/vote.ts
```diff
@@ -96,6 +97,17 @@ export type UpdateValidatorIdentityParams = {
...
+  voteAuthority: PublicKey,
+  voteStateUpdate: {
+    lockouts: Lockout[],
+    root: number,
+    hash: PublicKey,
```
buffalojoec, line 106: Unfortunately, the legacy library doesn't have an opaque type for a hash, but it looks like the rest of the package uses `string`. It definitely can't be `PublicKey` here, for a few reasons.

However, when it comes to decoding, you _can_ use the `publicKey` layout, since it will perform base58 decoding and assert the resulting length is 32 bytes.

packages/library-legacy/src/programs/vote.ts
```diff
@@ -342,6 +402,23 @@ const VOTE_INSTRUCTION_LAYOUTS = Object.freeze<{
       Layout.voteAuthorizeWithSeedArgs(),
     ]),
   },
+  CompactUpdateVoteState: {
+    index: 12,
+    layout: BufferLayout.struct<VoteInstructionInputData['CompactUpdateVoteState']>([
```
buffalojoec, line 407: You can add this whole layout to `layout.ts` and then just call `Layout.voteCompactUpdateVoteState()` here.

packages/library-legacy/src/programs/vote.ts
```diff
@@ -583,4 +660,43 @@ export class VoteProgram {
...
+        hash: toBuffer(hash.toBuffer()),
+        timestampOption: timestamp !== undefined ? 1 : 0,
+        timestamp: timestamp !== undefined ? timestamp : 0,
+      },
+    };
+    const data = encodeData(type, rawData);
+
+    const keys = [
+      {pubkey: voteAccount, isSigner: false, isWritable: true},
+      {pubkey: voteAuthority, isSigner: true, isWritable: true},
```
buffalojoec, line 693: Docs on this instruction only declare the authority as signer, not writable.

packages/library-legacy/src/programs/vote.ts
buffalojoec: I see what you're going for here. For anyone else reading, this is required due to the custom Serde serialization method used for `CompactUpdateVoteState`, ie. [`serde_compact_vote_state_update`](https://github.com/solana-labs/solana/blob/27eff8408b7223bb3c4ab70523f8a8dca3ca6645/sdk/program/src/vote/state/mod.rs#L811).

Since this is required at a serialization level, and you'll use this twice (here and in the transaction function), why not roll this into a custom serialization layout?

You can define a class `MyLayout extends Layout<T>` and implement the required methods that can transform from the instruction input type to the proper byte layout, and back.

We did this in SPL Token JS to mimic Rust's `COption::<Pubkey>` and you can use that as an example.
https://github.com/solana-labs/solana-program-library/blob/d1c03cc6164ebdeebef19190de28b7a71f4e4cb6/token/js/src/serialization.ts#L5-L39

buffalojoec (commented):
Thanks for adding! I left a few comments on this.

EugeneButusov:
@buffalojoec, thank you for the review, I'll handle the feedback in the closest days.

PR #2706, opened 2024-05-10
Update library readme to include extending RpcApiMethods

mcintyre94:
As written there's a type error because the `createRpcApi` expects the methods object to extend `RpcApiMethods`:
Screenshot 2024-05-10 at 16 54 27

This PR just updates the readme so the example `NftCollectionDetailsApi` extends `RpcApiMethods` which fixes the type error


buffalojoec (approved):
Lgtm!

https://github.com/solana-labs/solana-web3.js/blob/cbf8662d242b5ae9219c0c58aa2011c85ade8dab/packages/rpc-spec/src/__typetests__/rpc-api-typetest.ts#L15-L17

PR #2707, opened 2024-05-10
Allow creating crypto key/keypair from ReadonlyUint8Array

mcintyre94:
This PR updates `createKeyPairFromBytes`, ` createPrivateKeyFromBytes` and `createKeypairSignerFromBytes` so that they can take a `ReadonlyUint8Array` as input. They continue to also accept a `Uint8Array`.

This allows creating a keypair/key/signer from bytes produced by an encoder without casting.


steveluscher (approved):
It's as if you're actually building something with this stuff.

Added via Giphy

PR #2714, opened 2024-05-14
Make __DEV__ robust to environments without process.env

mcintyre94:
This PR updates the logic used to define `__DEV__` in the build scripts env-shim, so that if `process.env` doesn't exist  in the environment where we end up running doesn't exist, then it returns true instead of throwing a runtime error

The generated code in eg `packages/errors/dist/index.browser.cjs` is now:

```js
// ../build-scripts/env-shim.ts
function isDevelopment() {
  if (typeof process === "object" && process.env) {
    return process["env"].NODE_ENV === "development";
  }
  return true;
}
var __DEV__ = /* @__PURE__ */ isDevelopment();
```

Where previously it was:
```js
// ../build-scripts/env-shim.ts
var __DEV__ = /* @__PURE__ */ (() => process["env"].NODE_ENV === "development")();
```

### DIFF
--- a/packages/build-scripts/env-shim.ts
+++ b/packages/build-scripts/env-shim.ts
@@ -1,2 +1,10 @@
-// Clever obfuscation to prevent the build system from inlining the value of `NODE_ENV`
-export const __DEV__ = /* @__PURE__ */ (() => (process as any)['en' + 'v'].NODE_ENV === 'development')();
+function isDevelopment() {
+    if (typeof process === 'object' && process.env) {
+        // We use `'en' + 'v'` to prevent the build system from inlining the value of `NODE_ENV`
+        return (process as any)['en' + 'v'].NODE_ENV === 'development';
+    }
+    // if no process.env, use dev mode
+    return true;
+}
+
+export const __DEV__ = /* @__PURE__ */ isDevelopment();


lorisleiva:
Let me know if this is The Right Way as I'll need to update the generate program repos (and the generator).

steveluscher:
So confused why React can get away with just this.

steveluscher (changes requested):
Zooming out a bit:

1. This ‘call a function to get the value of this’ is dumb, and the only reason that I wrote it this way is because of how `inject` works in esbuild. Literally all I wanted to do was to [inject](https://esbuild.github.io/api/#inject), at the top level scope, the statement `const __DEV__ = process.env.NODE_ENV === 'development';`
2. See this conversation: https://github.com/evanw/esbuild/issues/2695
3. Have you tested that this actually results in code being reaped by a downstream bundler when `process.env.NODE_ENV` is `"production"`. I sort of suspect because of the added dynamic check of a global (`typeof process`) that this will never result in dead-code elimination.

mcintyre94:
I'm going to close this for now, I was able to use a Vite plugin to replace in the generated JS code. If this is something others end up needing then we can consider publishing a plugin to use without messing with the build output for everyone else.

PR #2715, opened 2024-05-14
refactor(experimental): consolidate Option-like codecs

lorisleiva:
This PR consolidates the `getNullableCodec` and `getOptionCodec` with their `Zeroable` counterparts and adds more configurations.

Namely, the `prefix` option can now be set to `null` and the `fixed` option was replaced with the `noneValue` option which can be set to `"zeroes"` for `Zeroable` codecs or a custom byte array for custom representations of none values. This means the `getZeroableNullableCodec` and `getZeroableOptionCodec` functions were removed in favor of the new options.

```ts
// Before.
getZeroableNullableCodec(getU16Codec());

// After.
getNullableCodec(getU16Codec(), { noneValue: 'zeroes', prefix: null });
```

As a result, this PR makes it possible to create nullable codecs that have no `prefix` nor `noneValue`. In this case, the existence of the nullable item is indicated by the presence of any remaining bytes left to decode.

```ts
const codec = getNullableCodec(getU16Codec(), { prefix: null });
codec.encode(42); // 0x2a00
codec.encode(null); // Encodes nothing.
codec.decode(new Uint8Array([42, 0])); // 42
codec.decode(new Uint8Array([])); // null
```

Also note that it is now possible for custom `noneValue` byte arrays to be of any length — previously, it had to match the fixed-size of the nullable item. This means the `SOLANA_ERROR__CODECS__EXPECTED_ZERO_VALUE_TO_MATCH_ITEM_FIXED_SIZE` error is no longer used. I did not remove it because it is illegal but thought I needed to mention that here.

Here is a recap of all supported scenarios, using a `u16` codec as an example:

| `encode(42)` / `encode(null)` | No `noneValue` (default) | `noneValue: "zeroes"`       | Custom `noneValue` (`0xff`) |
| ----------------------------- | ------------------------ | --------------------------- | --------------------------- |
| `u8` prefix (default)         | `0x012a00` / `0x00`      | `0x012a00` / `0x000000`     | `0x012a00` / `0x00ff`       |
| Custom `prefix` (`u16`)       | `0x01002a00` / `0x0000`  | `0x01002a00` / `0x00000000` | `0x01002a00` / `0x0000ff`   |
| No `prefix`                   | `0x2a00` / `0x`          | `0x2a00` / `0x0000`         | `0x2a00` / `0xff`           |

Reciprocal changes were made with `getOptionCodec`.

### Why?

As I needed the `getNullableCodec(x, { prefix: null })` variant, I could either add yet another Option-like codec pair such as `getRemainderNullableCodec` and `getRemainderOptionCodec` OR I could realise that actually, this is a simple 2x3 configuration matrix and refactor the `getNullableCodec` and `getOptionCodec` accordingly. I opted for the latter.


mcintyre94 (approved):
Nice, this looks like a cleaner API to me!

steveluscher (approved):
[didntread1](https://app.graphite.dev//settings/meme-library?org=solana-labs)

PR #2728, opened 2024-05-20
Simulate with max compute units, not u32::MAX

joncinque:
#### Problem

Looking through recent updates to the monorepo, this caught my eye: https://github.com/anza-xyz/agave/issues/885

The simulation utility sets `u32::MAX`, which will fail once that feature is enabled.

#### Solution

Set `MAX_COMPUTE_UNITS`, hardcoded to 1,400,000, instead of `u32::MAX`.


steveluscher (approved):
Added via Giphy

PR #2733, opened 2024-05-22
docs: add example on how to use connection.confirmTransaction in the non deprecated way.

spirobel:
Added this stackexchange answer by John as a tsdoc comment so people don't need to leave their editor to find the solution. 
https://solana.stackexchange.com/questions/2839/how-to-properly-call-confirmtransaction-so-that-it-is-not-deprecated

### DIFF
--- a/packages/library-legacy/src/connection.ts
+++ b/packages/library-legacy/src/connection.ts
@@ -3724,7 +3724,17 @@ export class Connection {
     commitment?: Commitment,
   ): Promise<RpcResponseAndContext<SignatureResult>>;
 
-  /** @deprecated Instead, call `confirmTransaction` and pass in {@link TransactionConfirmationStrategy} */
+  /** @deprecated Instead, call `confirmTransaction` and pass in {@link TransactionConfirmationStrategy}
+   *  @example
+   * ``` ts
+   * const latestBlockHash = await connection.getLatestBlockhash()
+   * await connection.confirmTransaction({
+   *   blockhash: latestBlockHash.blockhash,
+   *    lastValidBlockHeight: latestBlockHash.lastValidBlockHeight,
+   *    signature: airdropSignature,
+   *   })
+   *   ```
+   */
   // eslint-disable-next-line no-dupe-class-members
   confirmTransaction(
     strategy: TransactionSignature,
--- a/packages/library-legacy/src/connection.ts
+++ b/packages/library-legacy/src/connection.ts
@@ -3718,7 +3718,23 @@ export class Connection {
     }
     return res.result;
   }
-
+  /**
+   *
+   * @param strategy - {@link TransactionConfirmationStrategy} possibly blockhash or durable nonce based
+   * @param commitment - 'processed' | 'confirmed' | 'finalized' default: connection.commitment
+   * @example
+   * ``` ts
+   * const { blockhash, lastValidBlockheight } = await connection.getLatestBlockhash();
+   * const tx = new Transaction({ blockhash, lastValidBlockheight });
+   * const signature = await connection.sendTransaction(tx);
+   * const confirmationStrategy: BlockheightBasedTransactionConfirmationStrategy = {
+   *   blockhash,
+   *   lastValidBlockheight,
+   *   signature,
+   * };
+   * await connection.confirmTransaction(confirmationStrategy);
+   * ```
+   */
   confirmTransaction(
     strategy: TransactionConfirmationStrategy,
     commitment?: Commitment,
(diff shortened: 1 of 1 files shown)


packages/library-legacy/src/connection.ts
```diff
@@ -3724,7 +3724,17 @@ export class Connection {
     commitment?: Commitment,
   ): Promise<RpcResponseAndContext<SignatureResult>>;
 
-  /** @deprecated Instead, call `confirmTransaction` and pass in {@link TransactionConfirmationStrategy} */
+  /** @deprecated Instead, call `confirmTransaction` and pass in {@link TransactionConfirmationStrategy}
```
buffalojoec, line 3727: This should get moved upward to the non-deprecated overload.

packages/library-legacy/src/connection.ts
```diff
@@ -3724,7 +3724,17 @@ export class Connection {
...
+   *  @example
+   * ``` ts
+   * const latestBlockHash = await connection.getLatestBlockhash()
+   * await connection.confirmTransaction({
+   *   blockhash: latestBlockHash.blockhash,
+   *    lastValidBlockHeight: latestBlockHash.lastValidBlockHeight,
+   *    signature: airdropSignature,
+   *   })
+   *   ```
```
buffalojoec, line 3736: If we're going to include an example, I think it should be more explicit.

Such a usage description should be detailing exactly how to set up the confirmation strategy. We can just demonstrate the blockheight-based one for this doc.

The meat of the blockheight strategy is that the provided `blockhash` and `lastValidBlockheight` are used to constrain the range in which the transaction should be confirmed. Typically this is going to be the same values returned by `getLatestBlockhash` when the transaction is constructed, but for some use cases it may make sense to use an older blockhash.

For the sake of this example, I'd recommend something like this:

```ts
const { blockhash, lastValidBlockheight } = await connection.getLatestBlockhash();

const tx = new Transaction({ blockhash, lastValidBlockheight });
const signature = await connection.sendTransaction(tx);

const confirmationStrategy: BlockheightBasedTransactionConfirmationStrategy = {
  blockhash,
  lastValidBlockheight,
  signature,
};
await connection.confirmTransaction(confirmationStrategy);
```

But we should make sure to document that this is specifically for blockheight confirmation. Ideally we'd also want an accompanying nonce strategy example as well.

packages/library-legacy/src/connection.ts
spirobel: roger that. I added it there because I want to help people that come across old documentation.
The typical path is that `anchor test` fails because the airdrop transaction was not confirmed yet.
We could address the root cause by pointing people to this function:
``` ts
const newBalance = await airdropIfRequired(
  connection,
  keypair.publicKey,
  0.5 * LAMPORTS_PER_SOL,
  1 * LAMPORTS_PER_SOL,
);
```
found here: https://www.npmjs.com/package/@solana-developers/helpers

this package also contains the confirmTransaction function that was deprecated here.

packages/library-legacy/src/connection.ts
spirobel: I agree with this. Should we add it to this PR or make a separate one? (as mentioned in the other comment, the intention was to help people that want to use the deprecated function, that is still mentioned in guides, tutorials, SA answers)

packages/library-legacy/src/connection.ts
buffalojoec: This doesn't look like the deprecated version to me.

https://github.com/solana-developers/helpers/blob/f33977badf111033930663237de76552a3dd93c1/src/index.ts#L266-L271

packages/library-legacy/src/connection.ts
spirobel: Okay I added the example on the non deprecated `confirmTransaction` method. 
I also found a different spot where an example used the old signature. Updated that as well: https://github.com/solana-labs/solana-web3.js/pull/2733/commits/d672f34539b7d5f155268ec9f843a5d06ac4a645

Regarding the DurableNonce example: 
In most cases a blockhash based confirmation strategy will be used. I hinted at the possibility of a nonce based one next to the link  to the `TransactionConfirmationStrategy` type. It could be a good spot to put the example there. So anyone who is curious will click and everyone else can move forward.

spirobel:
This PR adds a ton of value for everyone who does not know immediately what to do when they see "TransactionConfirmationStrategy". @buffalojoec  you are looking at this code all day so you know what it is, but everyone else will not know. The comment above the airdrop function is straight up wrong. That needs to be fixed.

This is what people see when they follow your link to the docs of TransactionConfirmationStrategy
image
The sentence below just repeats what is already communicated in the word TransactionConfirmationStrategy. Completely useless.

PR #2736, opened 2024-05-24
Added transaction logs to send transaction functions

Woody4618:
- Added transaction logs to error case in sendAndConfirmRawTransaction
- Added transaction logs to error case in sendAndConfirmTransaction
- Added for skipPreflight and dont skipPreflight simulation log cases

### DIFF
--- a/packages/library-legacy/src/connection.ts
+++ b/packages/library-legacy/src/connection.ts
@@ -2927,7 +2927,12 @@ export type SignatureResult = {
 /**
  * Transaction error
  */
-export type TransactionError = {} | string;
+export type TransactionError = {
+  message: string;
+  data?: {
+    logs?: string[];
+  };
+};
 
 /**
  * Transaction confirmation status
@@ -5782,10 +5787,18 @@ export class Connection {
           console.error(res.error.message, logTrace);
         }
       }
-      throw new SendTransactionError(
-        'failed to simulate transaction: ' + res.error.message,
-        logs,
-      );
+      const transactionError: TransactionError = {
+        message: res.error.message,
+        data: {
+          logs: logs,
+        },
+      };
+
+      throw new SendTransactionError({
+        action: 'simulate',
+        signature: '',
+        transactionError: transactionError,
+      });
     }
     return res.result;
   }
@@ -5916,17 +5929,22 @@ export class Connection {
     const unsafeRes = await this._rpcRequest('sendTransaction', args);
     const res = create(unsafeRes, SendTransactionRpcResult);
     if ('error' in res) {
-      let logs;
+      let logs = undefined;
       if ('data' in res.error) {
         logs = res.error.data.logs;
       }
-      throw new SendTransactionError(
-        'failed to send transaction: ' +
-          res.error.message +
-          ' Logs: ' +
-          JSON.stringify(logs?.slice(-10), null, 2),
-        logs,
-      );
+      const transactionError: TransactionError = {
+        message: res.error.message,
+        data: {
+          logs: logs,
+        },
+      };
+
+      throw new SendTransactionError({
+        action: skipPreflight ? 'send' : 'simulate',
+        signature: '',
+        transactionError: transactionError,
+      });
     }
     return res.result;
   }
--- a/packages/library-legacy/src/connection.ts
+++ b/packages/library-legacy/src/connection.ts
@@ -5921,7 +5921,10 @@ export class Connection {
         logs = res.error.data.logs;
       }
       throw new SendTransactionError(
-        'failed to send transaction: ' + res.error.message,
+        'failed to send transaction: ' +
+          res.error.message +
+          ' Logs: ' +
+          JSON.stringify(logs, null, 2),
         logs,
       );
     }
--- a/packages/library-legacy/src/connection.ts
+++ b/packages/library-legacy/src/connection.ts
@@ -5924,7 +5924,7 @@ export class Connection {
         'failed to send transaction: ' +
           res.error.message +
           ' Logs: ' +
-          JSON.stringify(logs, null, 2),
+          JSON.stringify(logs?.slice(-10), null, 2),
         logs,
       );
     }
--- a/packages/library-legacy/src/utils/send-and-confirm-raw-transaction.ts
+++ b/packages/library-legacy/src/utils/send-and-confirm-raw-transaction.ts
@@ -8,6 +8,7 @@ import {
 } from '../connection';
 import type {TransactionSignature} from '../transaction';
 import type {ConfirmOptions} from '../connection';
+import {SendTransactionError} from '../errors';
 
 /**
  * Send and confirm a raw transaction
@@ -101,8 +102,9 @@ export async function sendAndConfirmRawTransaction(
         // If this does not work we'll just print the status
       }
       if (transaction?.meta?.err) {
-        throw new Error(
-          `Transaction ${signature} failed (${JSON.stringify(status)}): ${JSON.stringify(transaction.meta.logMessages?.slice(-10), null, 2)}`,
+        throw new SendTransactionError(
+          `Transaction ${signature} failed. Status: (${JSON.stringify(status)}) Logs: ${JSON.stringify(transaction.meta.logMessages?.slice(-10), null, 2)}`,
+          transaction?.meta?.logMessages as string[] | undefined,
         );
       }
     }
--- a/packages/library-legacy/src/utils/send-and-confirm-raw-transaction.ts
+++ b/packages/library-legacy/src/utils/send-and-confirm-raw-transaction.ts
@@ -7,7 +7,7 @@ import {
   TransactionConfirmationStrategy,
 } from '../connection';
 import type {TransactionSignature} from '../transaction';
-import type {ConfirmOptions} from '../connection';
+import type {ConfirmOptions, TransactionError} from '../connection';
 import {SendTransactionError} from '../errors';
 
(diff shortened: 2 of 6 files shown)


Woody4618:
The idea behind this pr is to improve developer experience by adding transaction logs which have more detailed informations for transaction errors. 

This PR will help turn these 0x0 errors: 

image

into errors with a full error logs like this: 

image

packages/library-legacy/src/utils/send-and-confirm-raw-transaction.ts
```diff
@@ -93,6 +93,14 @@ export async function sendAndConfirmRawTransaction(
...
 
   if (status.err) {
+    if (signature != null) {
+      const transaction = await connection.getTransaction(signature);
+      if (transaction?.meta?.err) {
+        throw new Error(
+          `Transaction ${signature} failed: ${JSON.stringify(transaction.meta.logMessages, null, 2)}`,
+        );
+      }
+    }
```
buffalojoec, line 103: This should probably be in a try-catch, in case the fetching of the transaction also fails. Also, we should still be logging the error code.

packages/library-legacy/src/utils/send-and-confirm-transaction.ts
```diff
@@ -89,6 +89,14 @@ export async function sendAndConfirmTransaction(
...
 
   if (status.err) {
+    if (signature != null) {
+      const transaction = await connection.getTransaction(signature);
+      if (transaction?.meta?.err) {
+        throw new Error(
+          `Transaction ${signature} failed: ${JSON.stringify(transaction.meta.logMessages, null, 2)}`,
+        );
+      }
+    }
```
buffalojoec, line 99: Same as previous comment.

buffalojoec (commented):
I'm a little wary about adding an unbounded log of transaction logs to these error logs.

Yes, there is a size limit to transaction logs, after which they become truncated, but it's still a large message to display in something like browser logs.

https://github.com/solana-labs/solana-web3.js/blob/71cba9915ce30b1bac11e4cac1afa0150e96536a/packages/rpc-graphql/src/__tests__/__setup__.ts#L755-L842

At a minimum, we should probably clip this to a subset of the logs - likely the last few lines, but I'm on the fence about whether or not to even do this.

cc @lorisleiva to weigh in.

packages/library-legacy/src/utils/send-and-confirm-raw-transaction.ts
Woody4618: Good point. Added

packages/library-legacy/src/utils/send-and-confirm-transaction.ts
Woody4618: Good point. Added

Woody4618:
> I'm a little wary about adding an unbounded log of transaction logs to these error logs.
> 
> Yes, there is a size limit to transaction logs, after which they become truncated, but it's still a large message to display in something like browser logs.
> 
> https://github.com/solana-labs/solana-web3.js/blob/71cba9915ce30b1bac11e4cac1afa0150e96536a/packages/rpc-graphql/src/__tests__/__setup__.ts#L755-L842
> 
> At a minimum, we should probably clip this to a subset of the logs - likely the last few lines, but I'm on the fence about whether or not to even do this.
> 
> cc @lorisleiva to weigh in.

True. I limited the logs to 10 lines 👍

lorisleiva:
I don't know that there is much value in concatenating the logs inside the error message as long as the error itself contains that information.

So I wouldn't update the `throw new SendTransactionError` to include the logs in the message as it already contains the logs and the end-user can decide to use it or not.

However, the other two updates are more interesting since we currently do not keep track of the information at all. I'd suggest replacing the `throw new Error` with a `throw new SendTransactionError` so that the logs are kept with the error (without affecting the error message). A new Error class may also be used if we fell it's necessary.

Woody4618:
The things is, when the logs are not concatenated onto the error message in most cases the users would need to add extra error handling for every case where the functions are used and usually the error message does not help: 
image

We could easily improve the default experience a lot by adding the last few lines of logs like so: 

image

What about a compromise and we use the SendTransactionError for all cases. So one can catch the error and get the whole logs, but also add the last few lines of logs to the error message to make it easy to figure out errors for new developers? 

I updated the PR with that suggestion. This is also a temporary fix probably until the new Web3Js is adopted everywhere, which has better error handling.

lorisleiva:
I understand your point but this seems more like a test framework issue to me.

The test frameworks I use display any extra attributes the error has, on top of its message. In this situation, you end up duplicating the logs on the console.

I also think that, because we need to cap the logs to a certain amount of lines, we're not even adding the full information to the message which means you're likely going to need to check the `logs` attribute anyway as the useful part tends to be at the end of the logs.

As I said though, it's great that we've added the `logs` attribute to some errors that didn't have it before. I'd just limit this PR to that.

Woody4618:
The logs are limited to the **last** few lines of the logs in this pr which contains the interesting error part not the first lines. So i think that would be helpful in all cases. 

Its not only a problem with the test framework. Its that every use of  ```web3.sendAndConfirmTransaction``` or ```web3.sendAndConfirmRawTransaction``` would need to be wrapped into a try catch: 

```
try {
          await sendAndConfirmTransaction(
            connection,
            transferSolTransaction,
            [keypair],
            confirmOptions,
          );
} catch (error) {
        if (error instanceof SendTransactionError && error.logs != null) {
          console.log('logs:', JSON.stringify(error.logs));
        }
}
``` 

Which is quite cumbersome. If its part of the error message it would immediately help me answer a lot questions less per day :D   

So to improve the devex I would need to update all examples and tutorials to add try catches. Like here for example: https://solana.com/de/docs/clients/javascript-reference

But I see your concerns. So if you still say to remove the logs from the message i'll do it and start teaching people to use try catch around their transaction sending once its merged. 

Ps: Also which test frame work are you using? I tried mocha, chai and jest and none of them showed more than the status message?  I might write a tutorial on that one as well then. Anchor is using mocha, so I would need to convince anchor somehow to switch to a test frame work that supports additional error fields, which will be hard.

lorisleiva:
Gotcha! I use Ava, Jest and Vitest. I thought Jest was showing these up but maybe we've got some custom configs.

I mean, if there's a clear value for the community to adding the last few lines of the logs to the message then it probably make sense to do it. 👌

Woody4618:
Cool, I think it would benefit the devs greatly. Any other tests I should add? Do you think this could break anything?

packages/library-legacy/src/connection.ts
```diff
@@ -5921,7 +5921,10 @@ export class Connection {
         logs = res.error.data.logs;
       }
       throw new SendTransactionError(
-        'failed to send transaction: ' + res.error.message,
+        'failed to send transaction: ' +
+          res.error.message +
+          ' Logs: ' +
+          JSON.stringify(logs?.slice(-10), null, 2),
```
steveluscher, line 5927: There are actually two `SendTransactionErrors` in this file.

steveluscher (changes requested):
So, we can't do this. Here's why.

This change modifies the latency characteristics of `sendAndConfirmTransaction()`. This method is latency sensitive because applications are _particularly_ interested in _failures_ to confirm transactions – the faster they can perform a remediation (show an error, retry, etc) the better the user experience. This change delays this response until a serial `getTransaction()` request completes, just to craft the `SendTransactionError` exception, which may or may not be of consequence regarding how the app decides to recover.

I think there's something we can do instead that gets you what you want.

1. Add all of the information necessary to fetch logs in the _constructor_ of `SendTransactionError`.
    ```ts
    export class SendTransactionError extends Error {
        #signature: TransactionSignature;
        #transactionError: TransactionError;
        constructor({
            action,
            signature,
            transactionError,
        }: {
            action: 'send' | 'simulate',
            signature: TransactionSignature,
            transactionError: TransactionError,
        }) {
            this.#signature = signature;
            this.#transactionError = transactionError;
        }
    }
    ```
2. Add a method to `SendTransactionError` that lets anyone get the logs _at their option_.
    ```ts
    export class SendTransactionError extends Error {
        #resolvedLogs: string[] | Promise<string[]> | undefined;
        async getLogs(connection: Connection): string[] {
            if (this.#resolvedLogs === undefined) {
                this.#resolvedLogs = new Promise((resolve, reject) => {
                    /* use the connection to fetch the logs */
                        .then(logs => {
                            this.#resolvedLogs = logs;
                            resolve(logs);
                        })
                        .catch(reject);
                });
            }
            return await this.#resolvedLogs;
        }
    }
    ```
3. `SendTransactionError` should not take in a message; it should take in the _raw_ material, and generate the message internally.
    ```ts
    export class SendTransactionError extends Error {
        constructor({action, signature, transactionError}) {
            const message =
                `Transaction \`${signature}\` failed to ${action}: ` +
                `${transactionError.message}. ` +
                'Call `getLogs()` on this error for full details.';
            super(message);
        }
    }
    ```

Woody4618:
Ohh, very true. I didn't think of that. Sorry. Your solution sounds perfect. I implemented it. Thank you for the review 🤗

Woody4618:
Thank you very much for the second review 🤗 , great tips!! I implemented them and added a feat: commit message.

steveluscher:
Dang, I think this broke typechecking on `library-legacy/` and I missed the CI failure. Can you take a look @Woody4618? Bunch of failures in `connection.ts` because of that change to `TransactionError`.

image

PR #2739, opened 2024-05-25
Turbocharge the `reverseCodec()`

steveluscher:
# Preamble

This is, what I hope to be, the first of many performance PRs. In general, we've prioritized correctness and readability in the codecs up until this point. The next step is to unroll that a bit to achieve great performance.

* Copy as little as possible.
  * Operate in-place wherever possible, like in this PR
  * Everywhere else, replace every possible call to `slice()` with `subarray()` which creates a new view over the same underlying data
* Do as little math as possible

# Summary

In this PR, we implement the reverse codec without using `reverse()` or cloning the source data. We instead write elements of the source into the target in reverse order.

# Results

* 7.5x speedup for decode
* 1.5x speedup for encode

## Before

```
> ./src/__benchmarks__/run.ts

┌─────────┬──────────────────┬─────────────┬───────────────────┬──────────┬─────────┐
│ (index) │ Task Name        │ ops/sec     │ Average Time (ns) │ Margin   │ Samples │
├─────────┼──────────────────┼─────────────┼───────────────────┼──────────┼─────────┤
│ 0       │ 'Decode reverse' │ '376,722'   │ 2654.4716237883   │ '±1.02%' │ 188362  │
│ 1       │ 'Encode reverse' │ '2,922,232' │ 342.2041376564442 │ '±0.10%' │ 1461117 │
└─────────┴──────────────────┴─────────────┴───────────────────┴──────────┴─────────┘
```

## After

```
> ./src/__benchmarks__/run.ts

┌─────────┬──────────────────┬─────────────┬────────────────────┬──────────┬─────────┐
│ (index) │ Task Name        │ ops/sec     │ Average Time (ns)  │ Margin   │ Samples │
├─────────┼──────────────────┼─────────────┼────────────────────┼──────────┼─────────┤
│ 0       │ 'Decode reverse' │ '2,961,665' │ 337.6479015528495  │ '±1.24%' │ 1480833 │
│ 1       │ 'Encode reverse' │ '4,234,486' │ 236.15613552370152 │ '±0.11%' │ 2117244 │
└─────────┴──────────────────┴─────────────┴────────────────────┴──────────┴─────────┘
```

### DIFF
--- a/packages/codecs-core/src/__tests__/reverse-codec-test.ts
+++ b/packages/codecs-core/src/__tests__/reverse-codec-test.ts
@@ -10,23 +10,29 @@ describe('reverseCodec', () => {
         const s = (size: number) => reverseCodec(fixCodecSize(base16, size));
 
         // Encode.
-        expect(s(1).encode('00')).toStrictEqual(b('00'));
-        expect(s(2).encode('00ff')).toStrictEqual(b('ff00'));
-        expect(s(2).encode('ff00')).toStrictEqual(b('00ff'));
-        expect(s(4).encode('00000001')).toStrictEqual(b('01000000'));
-        expect(s(4).encode('01000000')).toStrictEqual(b('00000001'));
-        expect(s(8).encode('0000000000000001')).toStrictEqual(b('0100000000000000'));
-        expect(s(8).encode('0100000000000000')).toStrictEqual(b('0000000000000001'));
-        expect(s(32).encode(`01${'00'.repeat(31)}`)).toStrictEqual(b(`${'00'.repeat(31)}01`));
-        expect(s(32).encode(`${'00'.repeat(31)}01`)).toStrictEqual(b(`01${'00'.repeat(31)}`));
+        expect(s(1).encode('99')).toStrictEqual(b('99'));
+        expect(s(2).encode('99ff')).toStrictEqual(b('ff99'));
+        expect(s(2).encode('ff9999')).toStrictEqual(b('99ff'));
+        expect(s(3).encode('9999ff')).toStrictEqual(b('ff9999'));
+        expect(s(3).encode('ff9999')).toStrictEqual(b('9999ff'));
+        expect(s(4).encode('999999ff')).toStrictEqual(b('ff999999'));
+        expect(s(4).encode('ff999999')).toStrictEqual(b('999999ff'));
+        expect(s(8).encode('99999999999999ff')).toStrictEqual(b('ff99999999999999'));
+        expect(s(8).encode('ff99999999999999')).toStrictEqual(b('99999999999999ff'));
+        expect(s(32).encode(`ff${'99'.repeat(31)}`)).toStrictEqual(b(`${'99'.repeat(31)}ff`));
+        expect(s(32).encode(`${'99'.repeat(31)}ff`)).toStrictEqual(b(`ff${'99'.repeat(31)}`));
 
         // Decode.
-        expect(s(2).decode(b('ff00'))).toBe('00ff');
-        expect(s(2).decode(b('00ff'))).toBe('ff00');
-        expect(s(4).decode(b('00000001'))).toBe('01000000');
-        expect(s(4).decode(b('01000000'))).toBe('00000001');
-        expect(s(4).read(b('aaaa01000000bbbb'), 2)).toStrictEqual(['00000001', 6]);
-        expect(s(4).read(b('aaaa00000001bbbb'), 2)).toStrictEqual(['01000000', 6]);
+        expect(s(2).decode(b('ff99'))).toBe('99ff');
+        expect(s(2).decode(b('99ff'))).toBe('ff99');
+        expect(s(3).decode(b('ff9999'))).toBe('9999ff');
+        expect(s(3).decode(b('9999ff'))).toBe('ff9999');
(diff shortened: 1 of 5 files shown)


steveluscher:
* **#2739** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2739?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/codecs-core/src/__tests__/reverse-codec-test.ts
```diff
@@ -10,23 +10,29 @@ describe('reverseCodec', () => {
...
-        expect(s(4).encode('01000000')).toStrictEqual(b('00000001'));
-        expect(s(8).encode('0000000000000001')).toStrictEqual(b('0100000000000000'));
-        expect(s(8).encode('0100000000000000')).toStrictEqual(b('0000000000000001'));
-        expect(s(32).encode(`01${'00'.repeat(31)}`)).toStrictEqual(b(`${'00'.repeat(31)}01`));
-        expect(s(32).encode(`${'00'.repeat(31)}01`)).toStrictEqual(b(`01${'00'.repeat(31)}`));
+        expect(s(1).encode('99')).toStrictEqual(b('99'));
+        expect(s(2).encode('99ff')).toStrictEqual(b('ff99'));
+        expect(s(2).encode('ff9999')).toStrictEqual(b('99ff'));
+        expect(s(3).encode('9999ff')).toStrictEqual(b('ff9999'));
+        expect(s(3).encode('ff9999')).toStrictEqual(b('9999ff'));
```
steveluscher, line 17: I added odd numbered arrays to exercise the ‘value in the middle’ edge case.

packages/codecs-core/src/__tests__/reverse-codec-test.ts
```diff
@@ -10,23 +10,29 @@ describe('reverseCodec', () => {
...
-        expect(s(1).encode('00')).toStrictEqual(b('00'));
-        expect(s(2).encode('00ff')).toStrictEqual(b('ff00'));
-        expect(s(2).encode('ff00')).toStrictEqual(b('00ff'));
-        expect(s(4).encode('00000001')).toStrictEqual(b('01000000'));
-        expect(s(4).encode('01000000')).toStrictEqual(b('00000001'));
-        expect(s(8).encode('0000000000000001')).toStrictEqual(b('0100000000000000'));
-        expect(s(8).encode('0100000000000000')).toStrictEqual(b('0000000000000001'));
-        expect(s(32).encode(`01${'00'.repeat(31)}`)).toStrictEqual(b(`${'00'.repeat(31)}01`));
-        expect(s(32).encode(`${'00'.repeat(31)}01`)).toStrictEqual(b(`01${'00'.repeat(31)}`));
+        expect(s(1).encode('99')).toStrictEqual(b('99'));
```
steveluscher, line 13: I replaced `00` in here with `99`

* visually distinct from `ff`
* catches bugs – forgetting to write in the middle value would still pass the test if the value in the middle happened to be `00`

packages/codecs-core/src/reverse-codec.ts
```diff
@@ -36,13 +59,16 @@ export function reverseDecoder<TTo, TSize extends number>(
...
+            const reversedBytes = new Uint8Array(decoder.fixedSize);
+            copySourceToTargetInReverse(
+                bytes /* source */,
+                reversedBytes /* target_WILL_MUTATE */,
+                offset /* sourceOffset */,
+                offset + decoder.fixedSize /* sourceLength */,
+                -offset /* targetOffset */,
+            );
+            const [decodedBytes, decodedOffset] = decoder.read(reversedBytes, 0);
+            return [decodedBytes, offset + decodedOffset];
```
lorisleiva, line 71: There is a slight issue here which is not obvious to see and tricky to test.

The issue is that the encoding logic is not completely symmetrical with the decoding logic.

Before, both the encoding and decoding were reversing bytes in situ. Meaning both the inner `encoder` and `decoder` would have access to the entirety of the byte array before/after the bytes have been reversed respectively.

After this PR, the encoding logic stays the same — i.e. the inner `encoder` has access to the full byte array — but **the decoding logic now creates a "box"** that only give access to the reversed bytes to the inner `decoder`.

This means that, if we had a `offsetCodec` inside of that inner codec, the results would be different when encoding and decoding.

CleanShot 2024-05-27 at 09 24 48@2x

Whatever we do, we need to ensure encoding and decoding with an `offsetCodec` inside yields the same result.

Here we can either box both functions or update the decoding logic so the inner `decoder` has access to the whole byte array such that part of it has been reversed. I am in favour of doing the latter since I think we should avoid creating boxes unless we really need to — e.g. `fixCodecSize`, `addCodecSizePrefix` and `addCodecSentinel` all need to be boxed.

We should probably add a test as well to ensure the codec is symmetrical with an `offsetCodec` inside the inner codec.

lorisleiva (commented):
This is amazing to see! There is just a little issue with the decoding logic but I'm hoping it shouldn't affect the performance gain too much.

packages/codecs-core/src/reverse-codec.ts
steveluscher: Amazing review. Thank you!

Fixed.

Also tested.

steveluscher: Also updated the benchmarks.

PR #2745, opened 2024-05-29
fix: Remove negative lookbehind from getHumanReadableErrorMessage regex

C-o-d-e-C-o-w-b-o-y:
Currently, the regex inside `getHumanReadableErrorMessage` in `@solana/errors` contains a negative lookbehind (`?<!`). When this regex is loaded into `iOS` browsers on `iOS < 16.4`, it breaks, as the lookbehind support in Javascript regex is relatively new. 

We discovered this when http://www.magiceden.io was broken for a number of iOS users, and patching the lookbehind regex was the fix. However it would be best if this could be patched at the source (here). 

In this PR we modify `getHumanReadableErrorMessage` to remove the negative lookbehind and replace it with a custom backslash check, and added a custom unit test for this case. 

(It is also worth noting that none of the error messages inside `SolanaErrorMessages` contain a backslash, I just wanted o keep the semantics the same)

More info about negative lookbehind iOS compatibility here:
* https://caniuse.com/js-regexp-lookbehind
* https://bugs.webkit.org/show_bug.cgi?id=174931

### DIFF
--- a/packages/errors/src/__tests__/message-formatter-test.ts
+++ b/packages/errors/src/__tests__/message-formatter-test.ts
@@ -94,6 +94,20 @@ describe('getErrorMessage', () => {
             );
             expect(message).toBe("Something awful happened: 'bar'. How awful!");
         });
+        it('does not interpolate escaped variables', () => {
+            const messagesSpy = jest.spyOn(MessagesModule, 'SolanaErrorMessages', 'get');
+            messagesSpy.mockReturnValue({
+                // @ts-expect-error Mock error config doesn't conform to exported config.
+                123: 'Here is a $variable and here is a \\$escapedVariable.',
+            });
+            const message = getErrorMessage(
+                // @ts-expect-error Mock error config doesn't conform to exported config.
+                123,
+                { escapedVariable: 'bar', variable: 'foo' },
+            );
+            // prettier-ignore
+            expect(message).toBe('Here is a foo and here is a \\$escapedVariable.');
+        });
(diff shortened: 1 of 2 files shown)


steveluscher (changes requested):
1. Ugh. Amazing find; thank you!
2. Neither of our implementations get it right! The point of this is to be able to print literal variable strings in the error. Replace your new test with an updated version of the one before it and make it pass?
    ```ts
    it('interpolates variables into a error message format string', () => {
        const messagesSpy = jest.spyOn(MessagesModule, 'SolanaErrorMessages', 'get');
        messagesSpy.mockReturnValue({
            // @ts-expect-error Mock error config doesn't conform to exported config.
            123: "Something $severity, $severity, $severity happened: '$foo'. How \\\\$severity\\\\ is the \\$severity!",
        });
        const message = getErrorMessage(
            // @ts-expect-error Mock error context doesn't conform to exported context.
            123,
            { foo: 'bar', severity: 'awful' },
        );
        expect(message).toBe("Something awful, awful, awful happened: 'bar'. How \\awful\\ is the $severity!");
    });
   ```

lukecaan:
This broke our UI at drift too .. super hard to fix because a bunch of dependencies now also have their own dependencies on it

lukecaan:
Tried to help with above but I can't even get `message-formatter-test.ts` running

lukecaan:
Hey guys, I've attempted to address the above changes requested and speed up a patch in this PR : https://github.com/solana-labs/solana-web3.js/pull/2784

Issue #2755, opened 2024-05-31
`innerInstructions` Missing From `simulateTransaction`


0xIchigo:
## Motivation
The `simulateTransaction` RPC method currently lacks support for the `innerInstructions` parameter. Adding this feature will allow users to include inner instructions in their simulations. This would be useful for developers who would like to simulate and debug complex transactions that involve multiple program invocations

## Example use case
If a developer is working on a program with nested calls without the `innerInstructions` parameter, they cannot debug issues related to inner program invocations. By enabling `innerInstructions`, the developer can get a detailed log of all nested instructions, making it easier to identify and fix problems.

Moreover, developers looking to optimize their compute usage by first simulating their transactions with inner instructions are currently unable to do so via web3.js and would have to make the HTTP call directly.

## Details
Behavior:
- The `simulateTransaction` method should accept an optional `innerInstructions` parameter of type `boolean`
- When the `innerInstructions` is set to `true`, the simulation response should include a detailed list of inner instructions executed during the transaction
- The response structure should match the structure defined in the [Solana documentation](https://solana.com/docs/rpc/json-structures#inner-instructions) for the inner instructions specifically, and the [RPC documentation for the response as a whole](https://solana.com/docs/rpc/http/simulatetransaction#result)

Implementation:
- Modify the `SimulateTransactionConfig` type to include the `innerInstructions` parameter
- Update the RPC call within the `simulateTransaction` method to pass in this parameter if provided

A PR implementing these changes will follow shortly! :)

PR #2756, opened 2024-05-31
feat(simulated_transaction): Add Inner Instructions

0xIchigo:
This PR aims to address the issue I've outlined in #2755 where the `innerInstructions` parameter is missing for transaction simulations

### DIFF
--- a/packages/library-legacy/src/connection.ts
+++ b/packages/library-legacy/src/connection.ts
@@ -1004,17 +1004,21 @@ const SimulatedTransactionResponseStruct = jsonRpcResultAndContext(
         }),
       ),
     ),
-    innerInstructions: optional(nullable(array(
-      pick({
-        index: number(),
-        instructions: array(
-          union([
-            ParsedInstructionStruct,
-            PartiallyDecodedInstructionStruct,
-          ]),
+    innerInstructions: optional(
+      nullable(
+        array(
+          pick({
+            index: number(),
+            instructions: array(
+              union([
+                ParsedInstructionStruct,
+                PartiallyDecodedInstructionStruct,
+              ]),
+            ),
+          }),
         ),
-      }),
-    ))),
+      ),
+    ),
   }),
 );
 
@@ -5738,7 +5742,11 @@ export class Connection {
         config.commitment = this.commitment;
       }
 
-      if (configOrSigners && typeof configOrSigners === 'object' && 'innerInstructions' in configOrSigners) {
+      if (
+        configOrSigners &&
+        typeof configOrSigners === 'object' &&
+        'innerInstructions' in configOrSigners
+      ) {
         config.innerInstructions = configOrSigners.innerInstructions;
       }
 
@@ -5832,7 +5840,11 @@ export class Connection {
       config.sigVerify = true;
     }
 
-    if (configOrSigners && typeof configOrSigners === 'object' && 'innerInstructions' in configOrSigners) {
+    if (
+      configOrSigners &&
+      typeof configOrSigners === 'object' &&
+      'innerInstructions' in configOrSigners
+    ) {
       config.innerInstructions = configOrSigners.innerInstructions;
     }
 
--- a/packages/library-legacy/src/connection.ts
+++ b/packages/library-legacy/src/connection.ts
@@ -956,6 +956,11 @@ export type SimulateTransactionConfig = {
   innerInstructions?: boolean;
 };
 
+export type ParsedInnerInstruction = {
+  index: number;
+  instructions: UiInstruction[];
+};
+
 export type UiCompiledInstruction = {
   programIdIndex: number;
   accounts: number[];
@@ -966,7 +971,11 @@ export type UiCompiledInstruction = {
 export type UiParsedInstruction = {
   program: string;
   programId: PublicKey;
-  parsed?: any;
+  parsed: {
+    info: any;
+    type: string;
+  };
+  stackHeight?: number;
 };
 
 export type UiPartiallyDecodedInstruction = {
@@ -977,18 +986,11 @@ export type UiPartiallyDecodedInstruction = {
 };
 
 export type UiInstruction =
-  | {type: 'Compiled'; value: UiCompiledInstruction}
-  | {
-      type: 'Parsed';
-      value: UiParsedInstruction | UiPartiallyDecodedInstruction;
-    };
+  | { type: 'Compiled'; value: UiCompiledInstruction }
+  | { type: 'Parsed'; value: UiParsedInstruction }
+  | { type: 'PartiallyDecoded'; value: UiPartiallyDecodedInstruction };
 
 export type UiInnerInstructions = {
-  index: number;
-  instructions: UiInstruction[];
-};
-
-export type ParsedNewInnerInstruction = {
   index: number;
   instructions: UiInstruction[];
 };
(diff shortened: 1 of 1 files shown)


mcintyre94 (commented):
Thanks for this! I haven't tested this as an RPC call yet, but comparing to the RPC docs there seem to be some differences here

I'm guessing that some of this is that `jsonParsed` encoding can produce decoded inner instructions, and the docs are missing that. But the field names also seem to differ from the docs a bit.

Do you by any chance have any example `simulateTransaction` calls you've been testing with that I could use to review this a bit better?

0xIchigo:
@mcintyre94 ah that's my bad! I was trying to reuse the [`ParsedInstruction`](https://github.com/solana-labs/solana-web3.js/blob/6340744e5cf0ea91ae677f381d5a187638a19597/packages/library-legacy/src/connection.ts#L1175-L1181) and [`PartiallyDecodedInstruction`](https://github.com/solana-labs/solana-web3.js/blob/6340744e5cf0ea91ae677f381d5a187638a19597/packages/library-legacy/src/connection.ts#L1149-L1156) types, but I think I'm better off creating an entirely new type. I'll refactor the PR to address those changes

packages/library-legacy/src/connection.ts
```diff
@@ -989,6 +1011,7 @@ const SimulatedTransactionResponseStruct = jsonRpcResultAndContext(
         }),
       ),
     ),
+    innerInstructions: optional(nullable(array(ParsedInnerInstructionStruct))),
```
mcintyre94, line 1014: Does this need to be optional + nullable? Will just optional suffice? Or will it be array, undefined or null in different situations?

steveluscher:
Do you have time to knock this out @0xIchigo, or would you like us to finish this up?

0xIchigo:
gm @steveluscher ! Totally up to you — I was waiting for us (i.e., Helius) to update our fleet to 1.18 so I could run some sample transactions to see the correct responses

mcintyre94:
Example of an unparsed inner instruction:

```
"innerInstructions": [
                {
                    "index": 0,
                    "instructions": [
                        {
                            "accounts": [
                                "9hUi6B4rqntMzvEz6HKTv9MAq2KJ7s3FvBxHBFf2WkS7",
                                "7Qpk9MmLevLeCKZqLQRfiucfR7P13XGtMhu7z5fbw3jN",
                                "8oxdgaw4EzgAVsisY1hhuZpTP3TLgkNYx7o1FrJwPfrs",
                                "Fo36995qyeK9n7sc7bxTTLgiPCQGukExGTvLdWRs5n68",
                                "J8x6y5G7GmTkuKTbbCAnfhn72vaUU2qsB6je9oKFigHM",
                                "CzK26LWpoU9UjSrZkVu97oZj63abJrNv1zp9Hy2zZdy5",
                                "AVNMK6wiGfppdQNg9WKfMRBXefDPGZFh2f3o1fRbgN8n"
                            ],
                            "data": "19TzxKPM1",
                            "programId": "zDEXqXEG7gAyxb1Kg9mK5fPnUdENCGKzWrM21RMdWRq",
                            "stackHeight": 2
                        }
                    ]
                }
            ],
```

Modified the simulate transaction call on https://explorer.solana.com/tx/5f47YKxsAmZV31QmFmVH8vHY9zdChAEJTKw2cc1nX8tdZ6XTTWdctevBLiCGncbTYqfL3LZZLHTwUBWCkHjyDCs5/inspect to pass `"innerInstructions": true` as a param.

mcintyre94:
@0xIchigo Sorry for all the back and forth here, I should have looked into this more closely before. We can make this PR much simpler, we don't need to define any new types.

The reason for the type errors you're getting is because the existing `ParsedTransactionMeta` has `innerInstructions?: ParsedInnerInstruction[] | null;`, and this is eventually using the `UiInnerInstruction` type that's been modified. The changes in this PR are incompatible.

But if you do `getTransaction` with the example transactions I posted above and `jsonParsed` encoding, you'll see that their `result.meta.innerInstructions` is identical to the inner instructions I posted above from simulating them. So we should just re-use those types exactly, ie probably just add `innerInstructions?: ParsedInnerInstruction[] | null;` to the simulate output.

The relevant types already defined here are:

```ts
export type ParsedInnerInstruction = {
  index: number;
  instructions: (ParsedInstruction | PartiallyDecodedInstruction)[];
};

export type ParsedInstruction = {
  /** Name of the program for this instruction */
  program: string;
  /** ID of the program for this instruction */
  programId: PublicKey;
  /** Parsed instruction info */
  parsed: any;
};

export type PartiallyDecodedInstruction = {
  /** Program id called by this instruction */
  programId: PublicKey;
  /** Public keys of accounts passed to this instruction */
  accounts: Array<PublicKey>;
  /** Raw base-58 instruction data */
  data: string;
};
```

This is exactly what we're seeing in practice from the `simulateTransaction` with innerInstructions on 1.18

Given the types are exactly the same, let's just re-use them. I think you can exactly re-use from `ParsedTransactionMeta` too:

```ts
  /** An array of cross program invoked parsed instructions */
  innerInstructions?: ParsedInnerInstruction[] | null;
```

0xIchigo:
Ahhh that's my bad for overcomplicating things — should be all good now!

packages/library-legacy/src/connection.ts
```diff
@@ -960,7 +962,20 @@ export type SimulatedTransactionResponse = {
...
   returnData?: TransactionReturnData | null;
+  innerInstructions?: ParsedInnerInstruction[] | null;
 };
+const ParsedInstructionStruct = pick({
+  program: string(),
+  programId: PublicKeyFromString,
+  parsed: any(),
+});
+
+const PartiallyDecodedInstructionStruct = pick({
```
mcintyre94, line 973: Are there existing structs that we can reuse here? No worries if not, just wondering if we have the equivalent for getTransaction too. Not exactly sure how the structs and types are related!

mcintyre94 (approved):

packages/library-legacy/src/connection.ts
0xIchigo: There is a [`ParsedInstructionResult` defined later on](https://github.com/solana-labs/solana-web3.js/blob/545c43992f555d54fe865788190e14eadd62264d/packages/library-legacy/src/connection.ts#L2225-L2229) that we could reuse — would just need to move it up, or we delete that and just use the new `ParsedInstructionStruct`

packages/library-legacy/src/connection.ts
```diff
@@ -1178,7 +1208,7 @@ export type ParsedInstruction = {
   /** ID of the program for this instruction */
   programId: PublicKey;
   /** Parsed instruction info */
-  parsed: any;
+  parsed?: any;
```
steveluscher, line 1211: I suppose this change is here for correctness – ie. this property can, in fact, be missing?

I don't think this will make a difference to consumers since this was `any` to begin with ([playground link](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAYg9nKBeKBvAUFLUwEMBOAzhACYBcUuAdiOgL7qiSwIByEA7smptnkaQD8FarQboAxnCqFgUAGYJuVAK4AbNZUJQVVANZU4HKlpZwA3JOmyFbTsvWbc23QaMnnZ9h0vpFcADp+YhIA2RV5eXMoAHoYqAB5AGk-Ow4gghCw4Aio2PiAZWAASw1EpKA)).

steveluscher: Actually, is this true? Is `parsed` not always defined, as per this? https://github.com/anza-xyz/agave/blob/5263c9d61f3af060ac995956120bef11c1bbf182/transaction-status/src/parse_instruction.rs#L83

steveluscher: https://docs.rs/serde_json/latest/serde_json/enum.Value.html

steveluscher: @0xlchigo, any thoughts on this? I really want to slam this in, but this is the last thing I'm stuck on.

0xIchigo: My apologies for missing your previous messages @steveluscher! The reason why I made it optional was because TypeScript throws some nasty type errors if it isn't set as optional. I agree that I don't think it'll make a difference to consumers — it was more so to make the compiler happy

steveluscher: Took a look. It's because `parsed` was wrong in `ParsedInstructionStruct`. It should be `unknown()`.

PR #2758, opened 2024-06-02
fix(library-legacy/deps): rpc-websockets@^7.11.0->^7.11.1

legobeat:
`7.11.0` is broken and deprecated. This upgrades to fixed `7.11.1`.

- [Context](https://github.com/elpheria/rpc-websockets/pull/161)


uafrontender:
@legobeat please check this issue https://github.com/solana-labs/solana-web3.js/issues/2762

In https://github.com/elpheria/rpc-websockets/pull/161 they renamed files and now imports in solana-web3 package work incorrect. We should or update imports or set fixed version of `"rpc-websockets": "7.11.0"`

legobeat:
@uafrontender Indeed. That's what this PR fixes.

uafrontender:
> @uafrontender Indeed. That's what this PR fixes.

Can you explain how does it fix? The problem arose because solana-web3 does not use a strict version of rpc-websockets, like 7.11.0. Instead, it uses ^7.11.0, which allows npm to install last compatible version, including 7.11.1. The 7.11.1 version of rpc-websockets was updated yesterday and is not compatible with the current code of solana-web3.

legobeat:
> > @uafrontender Indeed. That's what this PR fixes.
> 
> Can you explain how does it fix? The problem arose because solana-web3 does not use a strict version of rpc-websockets, like 7.11.0. Instead, it uses ^7.11.0, which allows npm to install last compatible version, including 7.11.1. The 7.11.1 version of rpc-websockets was updated yesterday and is not compatible with the current code of solana-web3.
> 
> @legobeat actually your PR changes nothing because now after I run `npm install @solana/web3.js` its dependencies already include `rpc-websockets: "^7.11.1"`

Sorry for the dismissiveness :facepalm: 

I missed to push the last of my changes and didn't look beyond the pending CI after that.

uafrontender:
> Pushed again with the actual fix.

Thanks!

steveluscher (approved):
Thanks for catching this problem that I caused, then fixed, then caused again, then you fixed.

image

softwarecurator:
pls fix

plustokens:
Please consider using a fixed version from rpc-websockets to prevent any upcoming breaking changes again.
Use rpc-websockets version "7.11.1" instead "^7.11.1"

Can we have the workaround in description?

- Fix rpc-websockets version to avoid using the latest one retrieved by solana web3js:
```
npm install rpc-websockets@7.11.0
```

nvm-delete:
@steveluscher This issue still persists in my Nextjs Pages router project, what should I do?

steveluscher:
Do you have a repro you can send me?

steveluscher:
Ironically, it’s not actually ESM only in 8.0.1 either. See my comment [here](https://github.com/elpheria/rpc-websockets/commit/6ff42dcd9f5db48deda6483e1d9e458e483ac39c#r142811170), and [this](https://github.com/elpheria/rpc-websockets/pull/164) follow up.

PR #2764, opened 2024-06-03
chore: Fixed SendTransactionError types

Woody4618:
Removed the logs from TransactionError and moved them in the SendTransactionError


steveluscher (approved):
image

PR #2765, opened 2024-06-03
chore: Fixed sendTransactionError types

Woody4618:
Reverted TransactionError types and added transactionMessage and logs directly into the SendTransactionError

### DIFF
--- a/packages/library-legacy/src/connection.ts
+++ b/packages/library-legacy/src/connection.ts
@@ -2927,12 +2927,7 @@ export type SignatureResult = {
 /**
  * Transaction error
  */
-export type TransactionError = {
-  message: string;
-  data?: {
-    logs?: string[];
-  };
-};
+export type TransactionError = {} | string;
 
 /**
  * Transaction confirmation status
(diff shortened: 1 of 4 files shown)


joncinque:
Was this meant to be a breaking change? There's some code in the account compression depending on the `logs` parameter as it used to be: https://github.com/solana-labs/solana-program-library/blob/b9d8fff34afe7ae852e13bae9063dd95f1af6501/account-compression/sdk/tests/utils.ts#L40

I'm happy to make the change on the library side, but I want to make sure this breakage won't cause other downstream issues.

steveluscher:
cc/ @Woody4618

Woody4618:
Hey, i can probably just fix that by renaming the type back to logs right?

steveluscher:
`git pull --rebase` your working copy and `pnpm i`; there was a problem with `rpc-websockets` that was fixed the same day too.

steveluscher:
I would probably satisfy backward compat by adding a getter for `logs` and deprecating it.
```ts
class SendTransactionError {
    /* @deprecated Use `await getLogs()` instead
    get logs() {
        return this.transactionError.logs;
    }
}
```

Woody4618:
I renamed it here: https://github.com/solana-labs/solana-web3.js/pull/2769 

Is that also okey?

PR #2769, opened 2024-06-04
chore: rename transactionLogs back to logs

Woody4618:
The rename to transactionLogs was breaking a downstream test and is not really necessary to be renamed.

### DIFF
--- a/packages/library-legacy/src/connection.ts
+++ b/packages/library-legacy/src/connection.ts
@@ -5787,7 +5787,7 @@ export class Connection {
         action: 'simulate',
         signature: '',
         transactionMessage: res.error.message,
-        transactionLogs: logs,
+        logs: logs,
       });
     }
     return res.result;
@@ -5928,7 +5928,7 @@ export class Connection {
         action: skipPreflight ? 'send' : 'simulate',
         signature: '',
         transactionMessage: res.error.message,
-        transactionLogs: logs,
+        logs: logs,
       });
     }
     return res.result;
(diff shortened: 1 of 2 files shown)


steveluscher (changes requested):
This creates a duality between `resolvedLogs` and the logs you provide in the constructor. Which is right? Which is up to date? Which is complete? It's null, does that mean there are no logs?

Perhaps we can pave over this by creating a getter that routes to `resolvedLogs` but _also_ tells you not to use it.

```ts
/* @deprecated Use `await getLogs()` */
get logs() {
    return this.resolvedLogs;
}
```

steveluscher:
Oh, I see; you renamed `resolvedLogs`.

steveluscher:
Do you think we could get away with doing the getter thing anyway, so that `logs` can't be overwritten from the outside? This is also a breaking change, but if someone is _writing_ logs _into_ a `SendTransactionError` then they're almost certainly doing something weird, right?

Woody4618:
Logs is private though. Can it still be overwritten from outside? Just calling it logs if probably cleaner. Wdyt?

steveluscher:
Yeah, unfortunately those typescript `private` markers don't do anything. They're just type guards.

https://www.typescriptlang.org/play/?#code/MYGwhgzhAEBiD29oG8BQ0PWPAdhALgE4Cuw+8hAFAA6ECWAbmPgKbS3zUBc0ARoiBZgcAShTQAvqimpsefNABm0ALzQcLAO5xElIsRYiA3KkUA6DtVVKwICCyPQA9E+gAVAJ7U2LQoQrQuCAejmBkxLbB0JoUANYQZqgu0AB6APyyuBDwgmYg8ADmlOaWxkA

PR #2771, opened 2024-06-04
chore: Add getter for logs to sendTransactionError

Woody4618:
- Add a getter for logs with the old types to be a non breaking change

### DIFF
--- a/packages/library-legacy/src/errors.ts
+++ b/packages/library-legacy/src/errors.ts
@@ -4,7 +4,7 @@ import {TransactionSignature} from './transaction';
 export class SendTransactionError extends Error {
   private signature: TransactionSignature;
   private transactionMessage: string;
-  private logs: string[] | Promise<string[]> | undefined;
+  private transactionLogs: string[] | Promise<string[]> | undefined;
 
   constructor({
     action,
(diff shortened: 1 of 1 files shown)


packages/library-legacy/src/errors.ts
```diff
@@ -44,25 +44,32 @@ export class SendTransactionError extends Error {
...
-      logs: Array.isArray(this.logs) ? this.logs : undefined,
+      logs: Array.isArray(this.transactionLogs)
+        ? this.transactionLogs
+        : undefined,
     };
   }
 
+  /* @deprecated Use `await getLogs()` instead */
+  get logs() {
+    return this.transactionLogs as string[] | undefined;
```
steveluscher, line 61: This may not be true, right? It could be a `Promise`? I think you have to just return `undefined` if `'then' in this.transactionLogs`

steveluscher (changes requested):

packages/library-legacy/src/errors.ts
Woody4618: Thank you. I added that check.

Woody4618: Hopefully now it will not be a breaking change anymore.

PR #2772, opened 2024-06-04
A series of Solana React hooks for Wallet Standard wallets

steveluscher:


steveluscher:
* **#2772** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2772?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/react/README.md
```diff
@@ -15,3 +15,226 @@
 # @solana/react
 
 This package contains React hooks for building Solana apps.
+
+## Hooks
```
steveluscher, line 19: I'm really tired; can you help me proofread this?

packages/react/README.md
```diff
@@ -15,3 +15,226 @@
...
+            <textarea name="message"></textarea>
+            {signature ? <span>Signature: {signature.toString()}</span> : null}
+            <button type="submit">Sign Message</button>
+        </form>
+    );
+}
+```
+
+> [!NOTE]
+> There exists a plural version of this hook &ndash; `useSignMessages` &ndash; that accepts multiple inputs and returns an array of outputs.
```
steveluscher, line 63: Anything else we want to say about this, @jordaaash?

packages/react/package.json
```diff
@@ -0,0 +1,93 @@
+{
+    "name": "@solana/react",
+    "version": "2.0.0-preview.2",
```
mcintyre94, line 3: Shall we align with `preview.3` on the other packages?

mcintyre94 (approved):
LGTM, and docs look good!

packages/react/src/__tests__/useSignMessage-test.ts
```diff
@@ -0,0 +1,99 @@
...
+            // eslint-disable-next-line jest/no-conditional-in-test
+            if (result.__type === 'error' || !result.current) {
+                throw result.current;
+            } else {
+                const signMessage = result.current;
+                signMessage({ message: new Uint8Array([1, 2, 3]) });
+                await jest.runAllTimersAsync();
+                signMessage({ message: new Uint8Array([1, 2, 3]) });
+                // eslint-disable-next-line jest/no-conditional-expect
+                expect(mockSignMessage).toHaveBeenCalledTimes(2);
```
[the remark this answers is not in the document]
steveluscher: This test asserts that calling `signMessage` results in a call to the underlying wallet feature's `signMessage` (in this test, `mockSignMessage`) so it's not actually important to wait on the result. I just want to know that it was called and with the params I expect.

packages/react/src/test-renderer.tsx
```diff
@@ -0,0 +1,70 @@
...
+    });
+    return {
+        rerenderHook(nextExecutor) {
+            act(() => {
+                testRenderer.update(<TestComponent executor={nextExecutor} resultRef={result} />);
+            });
+        },
+        result,
+    };
+}
```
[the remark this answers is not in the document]
steveluscher: It's the literally dumbest thing: if I put it in `__tests__` then ESLint complains that it has no tests in it.

I could make the file match pattern more strict and have it only match files that end in `-test.ts` or whatever, but I feel as soon as I did that we'd:

1. Miss some tests that are not thusly named, so
2. I'd have to go and do an audit to make sure that every test everywhere was exactly named `*-test.ts`, and then
3. Make some sort of counter-enforcement rule that says if you have an `it()` inside you you have to be named `*-test.ts`

…

packages/react/package.json
steveluscher: TY!

packages/react/src/useSignMessage.ts
```diff
@@ -0,0 +1,58 @@
...
+        SolanaSignMessage,
+    ) as SolanaSignMessageFeature[typeof SolanaSignMessage];
+    const account = getWalletAccountForUiWalletAccount_DO_NOT_USE_OR_YOU_WILL_BE_FIRED(uiWalletAccount);
+    return useCallback(
+        async (...inputs) => {
+            const inputsWithAccount = inputs.map(input => ({ ...input, account }));
+            const results = await signMessageFeature.signMessage(...inputsWithAccount);
+            const resultsWithoutSignatureType = results.map(
+                ({
+                    signatureType: _, // Solana signatures are always of type `ed25519` so drop this property.
```
buffalojoec, line 50: In the future, this would be where we'd support other means of encryption (non-Ed25519), right? It might not be necessary now, but we _could_ accept an optional parameter for this on `Input` which defaults to `ed25519`.

packages/react/src/chain.ts
```diff
@@ -0,0 +1,7 @@
+import { IdentifierArray } from '@wallet-standard/base';
+
+type AssertSolanaChain<T> = T extends `solana:${string}` ? T : never;
+
+export type OnlySolanaChains<T extends IdentifierArray> = T extends IdentifierArray
+    ? AssertSolanaChain<T[number]>
+    : never;
```
buffalojoec, line 7: What does Pythnet use, for example? And what about "the rise of SVM rollups"? 😅

packages/react/src/useSignTransaction.ts
```diff
@@ -0,0 +1,73 @@
...
+            chain,
+            featureName: SolanaSignAndSendTransaction,
+            supportedChains: [...uiWalletAccount.chains],
+            supportedFeatures: [...uiWalletAccount.features],
+        });
+    }
+    const signTransactionFeature = getWalletAccountFeature(
+        uiWalletAccount,
+        SolanaSignTransaction,
+    ) as SolanaSignTransactionFeature[typeof SolanaSignTransaction];
```
buffalojoec, line 59: Just for my understanding: The docs say that these inputs dictate how a wallet should simulate a transaction (min context slot, etc.). Is this where wallets plug in that simulation?

packages/react/src/useSignAndSendTransaction.ts
```diff
@@ -0,0 +1,72 @@
...
+        async input => {
+            const [result] = await signAndSendTransactions(input);
+            return result;
+        },
+        [signAndSendTransactions],
+    );
+}
+
+/**
+ * The plural version of `useSignAndSendTransactions`. Call this function with more than one input
```
buffalojoec, line 37:
```suggestion
 * The plural version of `useSignAndSendTransaction`. Call this function with more than one input
```

buffalojoec (approved):
I'm missing a bit of context on the changes made to wallet standard, so I was reading through that repo a bit as I reviewed.

Overall I think this React API looks great!

packages/react/README.md
```diff
@@ -15,3 +15,155 @@
...
+A config object with the following properties:
+
+-   `message`: A `Uint8Array` of bytes to sign
+
+#### Returns
+
+An object with the following properties:
+
+-   `signature`: The 64-byte Ed25519 signature as a `Uint8Array`
+-   `signedMessage`: The `Uint8Array` of bytes signed by the wallet. These bytes might differ from the input bytes if the wallet modified the message
```
jordaaash, line 36: We added a `signatureType` field to this to be able to support non-Ed25519 signatures but no one uses it yet, and I think it's probably the wrong way to even try. A more appropriate way to do it would be a separate feature like `signMessageWithSecp256r1` or something.

packages/react/README.md
```diff
@@ -15,3 +15,155 @@
...
+Given a `UiWalletAccount` and a chain that begins with `solana:`, this hook returns a function you can call to sign a serialized transaction.
+
+#### Arguments
+
+A config object with the following properties:
+
+-   `options`: An object with the following properties that govern how the wallet should simulate the transaction for the benefit of the user
+    -   `minContextSlot`: The minimum slot at which to simulate the transaction
+    -   `preflightCommitment`: The commitment at which to simulate the transaction
+-   `transaction`: A `Uint8Array` of bytes that conforms to the [Solana transaction schema](https://solana.com/docs/core/transactions#transaction)
```
jordaaash, line 78: We should eliminate `preflightCommitment` entirely, as it's not used by wallets (and shouldn't be). They handle their own simulation, and they aren't sending it anyway.

packages/react/README.md
```diff
@@ -15,3 +15,155 @@
...
+
+A config object with the following properties:
+
+-   `options`: An object with the following properties that govern how the wallet should simulate the transaction for the benefit of the user
+    -   `commitment`: The commitment at which to confirm the transaction
+    -   `maxRetries`: The number of retries the wallet's RPC should make when forwarding the transaction to the leader.
+    -   `minContextSlot`: The minimum slot at which to simulate the transaction
+    -   `mode`: One of `parallel` or `serial`. It's up to the wallet to interpret how to present this to the user.
+    -   `preflightCommitment`: The commitment at which to simulate the transaction
+    -   `skipPreflight`: When `true` asks the wallet to disable transaction verification when sending the transaction to an RPC.
```
jordaaash, line 129: We should deprecate and eliminate `commitment`, `maxRetries`, `preflightCommitment`, `skipPreflight`. Wallets should not let apps influence how these are simulated or sent (that's up to them) or confirmed (should always use `commitment` `confirmed` anyway, which is what they do).

`mode` can also go (it's not used, and was the wrong way to handle `signAndSendAllTransactions`), which needs its own feature (and no wallet has it yet).

packages/react/README.md
```diff
@@ -15,3 +15,155 @@
...
+            }}
+        >
+            Sign and Send Transaction
+        </button>
+    );
+}
+```
+
+> [!NOTE]
+> There exists a plural version of this hook &ndash; `useSignAndSendTransactions` &ndash; that accepts multiple inputs and returns an array of outputs.
```
jordaaash, line 169: We need a new feature for this, I'll work on these.

packages/react/src/chain.ts
jordaaash: I doubt Pythnet is even aware of this, since it's purely in the client. I imagine e.g. Eclipse should use either `solana:eclipse-mainnet` or `eclipse:mainnet`

packages/react/README.md
steveluscher: Yeah, I stripped out the `signatureType` because this is the `@solana/react` packages and I took a guess that we'd never have any other type of signatures, ever.

If this were the hook for `experimental:signMessage` then it would be in the returns.

packages/react/README.md
steveluscher: I mean if they don't support it, they can throw exceptions, I guess.

One reason that we might have to expose this plural API is that Loris' `@solana/signers` API presumes that everything is plural (ie. `TransactionSendingSigner` et al take in an array of transactions/messages).

packages/react/README.md
steveluscher: OK, I'll nuke the whole `options` object. If `preflightCommitment` isn't relevant, neither is `minContextSlot`.

packages/react/src/chain.ts
steveluscher: They can make `@pyth/react`, wrap this API, and override the types, maybe?

packages/react/src/useSignMessage.ts
steveluscher: Encryption would be a different feature/hook, so we could _not_ strip this argument out there. Will `solana:signMessage` ever produce signatures other than `ed25519`? ¯\\\_(°ペ)\_/¯

packages/react/src/useSignTransaction.ts
steveluscher: I've stripped this out of the docs and the inputs. Wallets decide; apps have no say.

packages/react/README.md
jordaaash: `minContextSlot` is probably still relevant -- I think it can be used to locate the valid blockhash (assuming the wallet uses it). When the wallet receives a blockhash in a tx from the app, it doesn't know if it's valid, and there's no good API for determining this. But it could theoretically fetch the latest slot from the RPC and know if the `minContextSlot` will be satisfied?

packages/react/README.md
jordaaash: Generally looks fine to me at a first pass!

packages/react/README.md
steveluscher: So IIUC the threat here is that the application happened to fetch the very newest blockhash at slot X, but the wallet's RPC is somehow behind the application at slot X-1 and doesn't know about that blockhash. Because of this the wallet shows the ‘you waited too long to send this transaction’ warning.

If providing `minContextSlot` is important in this context, then isn't `preflightCommitment` also important? Both come into play when deciding if a blockhash is valid.

Also, I literally don't understand why any of this matters in the context of `signTransaction`.

* If I'm a wallet, and
* I'm not being asked to send a transaction, just sign it, and
* It's not a nonce transaction

Why do I care if the blockhash contained within is valid? I should just simulate it with a valid blockhash, sign the thing, and give it back, right?

jordaaash: The blockhash is important for determining what chain you're on since we don't have this otherwise in the tx, so if it's provided by the app, the wallet wants to know as much as it can about how it got there. Realistically most apps aren't passing this right now and wallets aren't using it, but I think we have an opportunity to make that the default in The New Way. Come to think of it, where's the optional chain ID we expect?

steveluscher: Oh I see; this is that feature where if your wallet is ‘on devnet’ and the app sends you a transaction with a blockhash that's valid on mainnet, then the wallet can be like ‘hey hey whoa, this blockhash is valid on mainnet – if you sign this then someone could land it on main.’

steveluscher: > Come to think of it, where's the optional chain ID we expect?

This API asks for the chain in render.

```ts
const doTheThing = useTheThing(account, chain);
```

That way, if the account doesn't support that chain, you find that out in render, rather than later on when you try to sign something.

packages/react/src/useSignAndSendTransaction.ts
```diff
@@ -0,0 +1,90 @@
...
+import { getWalletAccountForUiWalletAccount_DO_NOT_USE_OR_YOU_WILL_BE_FIRED } from '@wallet-standard/ui-registry';
+import { useCallback } from 'react';
+
+import { OnlySolanaChains } from './chain';
+
+type Input = Readonly<
+    Omit<SolanaSignAndSendTransactionInput, 'account' | 'chain' | 'options'> & {
+        options?: Readonly<{
+            minContextSlot?: bigint;
+        }>;
```
steveluscher, line 21: Wait, this makes no sense. These are per-transaction-send level options, not wallet-account-feature level options.

steveluscher:
### Merge activity

* **Jun 12, 6:46 PM PDT**: @steveluscher started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2772).

PR #2784, opened 2024-06-06
[2] fix: Remove negative lookbehind from getHumanReadableErrorMessage regex

lukecaan:
Trying to fix the issue mentioned here : https://github.com/solana-labs/solana-web3.js/pull/2745

This is me attempting to speed up getting a patch through because this is resulting in some bugs in production which are hard to resolve due to the nested dependencies on the currently broken version of solana-web3.js

I tried to incorporate the changes from the above PR, while addressing the changes requested in [this comment](https://github.com/solana-labs/solana-web3.js/pull/2745#pullrequestreview-2089069306). I don't know enough about the context of the code to be 100% sure that this provides the exact expected functionality, but it passes the given unit tests.

cc: @steveluscher

### DIFF
--- a/packages/errors/src/__tests__/message-formatter-test.ts
+++ b/packages/errors/src/__tests__/message-formatter-test.ts
@@ -85,14 +85,14 @@ describe('getErrorMessage', () => {
             const messagesSpy = jest.spyOn(MessagesModule, 'SolanaErrorMessages', 'get');
             messagesSpy.mockReturnValue({
                 // @ts-expect-error Mock error config doesn't conform to exported config.
-                123: "Something $severity happened: '$foo'. How $severity!",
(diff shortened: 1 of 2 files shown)


mcintyre94 (approved):
Hey Luke, thanks for this! The logic looks good to me, but I'll let Steve double check it covers the use cases he's expecting.

Would you mind splitting the test change into a new test? So keep `'interpolates variables into a error message format string'` unchanged, and put the new test string into a new test named something like `'interpolates escaped variables into a error message format string'`. Just so that if we break this in future we know which bit of the logic we broke.

lukecaan:
^ Addressed the above feedback 🙏 .. out of interest if this is good is there an ETA for how quickly a patch could go out that we could consume in our UI?

steveluscher:
Give me a few minutes.

PR #2826, opened 2024-06-18
feat: add isPublicKey with test

aspnxdd:
This PR aims to add a new functionality to validate if a `PublicKeyInitData `can actually construct a `PublicKey`

### DIFF
--- a/packages/library-legacy/src/publickey.ts
+++ b/packages/library-legacy/src/publickey.ts
@@ -251,6 +251,17 @@ export class PublicKey extends Struct {
     const pubkey = new PublicKey(pubkeyData);
     return isOnCurve(pubkey.toBytes());
   }
+
+  /**
+   * Check that a pubkey is valid.
+   */
+  static isPublicKey(value: PublicKeyInitData): boolean {
+    try{
+      return new PublicKey(value).toBase58().length === 44;
+    } catch (err) {
+      return false;
+    }
+  }
 }
 
 SOLANA_SCHEMA.set(PublicKey, {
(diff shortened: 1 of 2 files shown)


aspnxdd:
> Thanks for the submission!
> 
> Can you talk a little bit about what this new static method adds over this?
> 
> ```ts
> try {
>   // Construct a new public key from some `initData`
>   new PublicKey(initData);
> } catch (e) {
>   // Run this code when construction fails.
> }
> ```
> 
> Your new static method calls the constructor, and the `constructor()` function already:
> 
> * decodes base58 strings, throwing if they are invalid
> * asserts that they are 32 bytes, decoded
> 
> Your new method won't actually work for some public keys, because base58 encoded public keys can be anywhere [between 32 and 44 characters](https://github.com/solana-labs/solana-web3.js/blob/master/packages/addresses/src/address.ts#L57-L67).

Thanks for your reply. It is just a handy method, I personally usually build this function in every project.But I think it can actually be simplified to your suggestion, or I could add that the following instead: 
```ts
const { length } = new PublicKey(value).toBase58();
return length >= 32 && length <= 44;
```
I just find it handy to have it as a static method in this class, in many projects I have to build this myself as a util, and I think it can help a bit. For example in JS you have the `Array.isArray()` or in etherjs you have `Signer.isSigner()` ([source](https://docs.ethers.org/v5/api/signer/#Signer-isSigner))

steveluscher:
The difference between `Array.isArray()` is that you get it for free – it's built into the JS runtime. `PublicKey` already has too many methods on it that _your application may not use_. Adding another punishes every application that doesn't need `isPublicKey`.

We're changing this architecture with the new 2.0 line of web3.js. I suggest that you take a look, and kick the tires if you have time! It won't be long before we recommend it for general use.

https://github.com/solana-labs/solana-web3.js/releases/tag/tp3

PR #2817, opened 2024-06-14
An example React app that exercises the new `@solana/web3.js` and the new Wallet Standard wallet adapter in `@solana/react`

steveluscher:
This example app brings together the work on `@solana/web3.js@2`, `@wallet-standard/react`, and `@solana/react`. In it you will find an example of how to build an app that:

* Connects to Wallet Standard wallets
* Uses RPC requests and subscriptions to display a wallet's balance
* Signs messages with a wallet account
* Signs and sends a (token transfer) transaction between two connected wallet accounts

PR #2840, opened 2024-06-22
Replace the `__DEV__` flag with literally `process.env.NODE_ENV` in the built source

steveluscher:
# Summary

Previously, we were using the hair-brained `inject` feature of ESBuild to try to replace all instances of `__DEV__` with something resembling `process.env.NODE_ENV === 'development'`. This just didn't work.

In this PR we replace it with the _literal_ `process.env.NODE_ENV === 'development'`, pre-sourcemaps, and only do the replacement in iife bundles.

# Test Plan

Consider the application in #2817. `pnpm build` it.

# Before

TBD

# After

TBD


steveluscher:
* **#2840** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2840?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

steveluscher:
Definitely! It basically doesn't work with the `env-shim` method. This method results in actual DCE.

lorisleiva:
How would you feel about `process.env.NODE_ENV === 'development'` being generated directly by Kinobi instead of using a `__DEV__` variable that needs to be replaced at build time? The issue with the latter is that anyone the generates code using Kinobi now needs to implement the same pre-processing build step which is a [concern we've already received](https://github.com/kinobi-so/kinobi/pull/49) from the community.

lorisleiva:
@steveluscher Just bumping that in case the GitHub bot made you mark this as done. 😇

steveluscher:
> How would you feel about process.env.NODE_ENV === 'development' being generated directly by Kinobi instead of using a __DEV__ variable that needs to be replaced at build time?

Much better. This is what web3.js does now, essentially. Nobody except for _us_ developers ever see `__DEV__`; it's just a shorthand for us that gets replaced with { development expression du jour }.

PR #2852, opened 2024-06-25
refactor(experimental): assert single sending signer inside signAndSend function

lorisleiva:
Fixes https://github.com/solana-labs/solana-web3.js/issues/2818

This PR calls the `assertIsTransactionMessageWithSingleSendingSigner` inside the `signAndSendTransactionMessageWithSigners` so the end-user doesn't need to do it explicitly.

This is partly to improve the developer experience and partly because TS doesn't fail properly when the transaction message hasn't been asserted on prior to calling the sign and send function.

Note that I did not remove the `assertIsTransactionMessageWithSingleSendingSigner` and the `isTransactionMessageWithSingleSendingSigner` from the package's exports since they may still be useful for custom use-cases such as the one described in the documentation:

```ts
let transactionSignature: SignatureBytes;
if (isTransactionMessageWithSingleSendingSigner(transaction)) {
    transactionSignature = await signAndSendTransactionMessageWithSigners(transaction);
} else {
    const signedTransaction = await signTransactionMessageWithSigners(transaction);
    const encodedTransaction = getBase64EncodedWireTransaction(signedTransaction);
    transactionSignature = await rpc.sendTransaction(encodedTransaction).send();
}
```

### DIFF
--- a/packages/signers/README.md
+++ b/packages/signers/README.md
@@ -549,9 +549,9 @@ Here as well, composite transaction signers are treated such that at least one s
 -   `TransactionModifyingSigner`, if no other `TransactionModifyingSigner` exists.
 -   `TransactionPartialSigner`, otherwise.
 
-The provided transaction must be of type `ITransactionWithSingleSendingSigner` meaning that it must contain exactly one `TransactionSendingSigner` inside its account metas. If more than one composite signers implement the `TransactionSendingSigner` interface, one of them will be selected as the sending signer.
+The provided transaction must contain exactly one `TransactionSendingSigner` inside its account metas. If more than one composite signers implement the `TransactionSendingSigner` interface, one of them will be selected as the sending signer. Otherwise, if multiple `TransactionSendingSigners` must be selected, the function will throw an error.
 
-Therefore, you may use the `assertIsTransactionWithSingleSendingSigner()` function to ensure the transaction is of the expected type.
+If you'd like to assert that a transaction makes use of exactly one `TransactionSendingSigner` _before_ calling this function, you may use the `assertIsTransactionWithSingleSendingSigner` function.
 
 ```ts
 assertIsTransactionWithSingleSendingSigner(myTransaction);
--- a/.changeset/short-jokes-begin.md
+++ b/.changeset/short-jokes-begin.md
@@ -1,0 +1,5 @@
+---
+'@solana/signers': patch
+---
+
+The `signAndSendTransactionMessageWithSigners` function now automatically asserts that the provided transaction message contains a single sending signer and fails otherwise.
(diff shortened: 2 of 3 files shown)


lorisleiva:
* **#2852** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2852?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/signers/README.md
```diff
@@ -549,9 +549,9 @@ Here as well, composite transaction signers are treated such that at least one s
 -   `TransactionModifyingSigner`, if no other `TransactionModifyingSigner` exists.
 -   `TransactionPartialSigner`, otherwise.
 
-The provided transaction must be of type `ITransactionWithSingleSendingSigner` meaning that it must contain exactly one `TransactionSendingSigner` inside its account metas. If more than one composite signers implement the `TransactionSendingSigner` interface, one of them will be selected as the sending signer.
+The provided transaction must contain exactly one `TransactionSendingSigner` inside its account metas. If more than one composite signers implement the `TransactionSendingSigner` interface, one of them will be selected as the sending signer. Otherwise, if multiple `TransactionSendingSigners` must be selected, the function will throw an error.
 
-Therefore, you may use the `assertIsTransactionWithSingleSendingSigner()` function to ensure the transaction is of the expected type.
+If you'd like to a single `TransactionSendingSigner` can be selected from the provided transaction _before_ calling this function, you may use the `assertIsTransactionWithSingleSendingSigner` function.
```
steveluscher, line 554:
```suggestion
If you'd like assert that a transaction makes use of exactly one `TransactionSendingSigner` _before_ calling this function, you may use the `assertIsTransactionWithSingleSendingSigner` function.
```

steveluscher (approved):
You don't want to delete `ITransactionMessageWithSingleSendingSigner` hey? It's now in use nowhere.

lorisleiva:
I think that since there's some value in exporting the helper functions for single sending signer transaction messages, it may be cleaner to keep the type as well so their implementation is consistent to some of the other assertions we provided that use an opaque type.

I don't feel too strongly about it though, so I can kill the type if you feel that's a better move.

steveluscher:
Totally up to you. It's just literally usable for nothing, until someone marks an input to a _new_ function as requiring `ITransactionMessageWithSingleSendingSigner`, right?

lorisleiva:
I'm leaning towards leaving it in. Otherwise the `assertIs*` and `is*` helper methods no longer narrow down the type which makes them less valuable — even though we don't make use of that type internally.

lorisleiva:
### Merge activity

* **Jun 26, 5:05 AM EDT**: @lorisleiva started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2852).

PR #2858, opened 2024-06-26
Add `minContextSlot` as an option to `Transaction{Sending}SignerConfig`

### LINKED ISSUE #2846: Add `minContextSlot` into the options of `TransactionSendingSignerConfig`
See discussion here: https://github.com/solana-labs/solana-web3.js/pull/2795#discussion_r1645752484

steveluscher:
# Summary

Instead of the cumbersome `getOptions()` solution that requires callers of the hook to track transaction / context slot pairs and return a getter for them… just add `minContextSlot` to the options.

# Test Plan

```
```

Closes #2846

### DIFF
--- a/.changeset/wicked-cougars-matter.md
+++ b/.changeset/wicked-cougars-matter.md
@@ -1,0 +1,5 @@
+---
+'@solana/signers': minor
+---
+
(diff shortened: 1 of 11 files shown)


lorisleiva (approved):
Perfection!

PR #2865, opened 2024-06-27
Correct the return type of `simulateTransaction` when accounts are not loaded

steveluscher:
This was returning `accounts: null` at the top level instead of inside `value`

See https://github.com/solana-labs/solana-web3.js/commit/7192f3d37e6cf65e144c8cd920fa20a71b2b1fe0

### DIFF
--- a/packages/rpc-api/src/simulateTransaction.ts
+++ b/packages/rpc-api/src/simulateTransaction.ts
@@ -127,7 +127,7 @@ export interface SimulateTransactionApi extends RpcApiMethods {
     simulateTransaction(
         base58EncodedWireTransaction: Base58EncodedBytes,
         config?: SigVerifyAndReplaceRecentBlockhashConfig & SimulateTransactionConfigBase,
-    ): SimulateTransactionApiResponseBase & { accounts: null };
+    ): SimulateTransactionApiResponseBase & SolanaRpcResponse<{ readonly accounts: null }>;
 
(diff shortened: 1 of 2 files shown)


steveluscher:
* **#2865** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2865?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/rpc-api/src/simulateTransaction.ts
```diff
@@ -127,7 +127,7 @@ export interface SimulateTransactionApi extends RpcApiMethods {
     simulateTransaction(
         base58EncodedWireTransaction: Base58EncodedBytes,
         config?: SigVerifyAndReplaceRecentBlockhashConfig & SimulateTransactionConfigBase,
-    ): SimulateTransactionApiResponseBase & { accounts: null };
+    ): SimulateTransactionApiResponseBase & SolanaRpcResponse<{ readonly accounts: null }>;
```
buffalojoec, line 130: I think this points out a broader issue.

In lots of places, we're including the `RpcResponse` wrapper in the "base response", but in others, we do something like:

```
RpcResponse<Base & { ... }>
```

I think it should always be the latter. I'll make a note to comb for this in my pass over RPC stuff.

buffalojoec (approved):

Issue #2859, opened 2024-06-26
Eliminate deprecated RPC methods


steveluscher:
The following deprecated methods are [being removed in Agave 2.0](https://github.com/anza-xyz/agave/pull/1886):

* confirmTransaction
* getSignatureStatus
* getSignatureConfirmation
* getTotalSupply
* getConfirmedSignaturesForAddress
* getConfirmedBlock
* getConfirmedBlocks
* getConfirmedBlocksWithLimit
* getConfirmedTransaction
* getConfirmedSignaturesForAddress2
* getRecentBlockhash
* getFees
* getFeeCalculatorForBlockhash
* getFeeRateGovernor
* getSnapshotSlot
* getStakeActivation

Clean all references and code for these from web3.js.

steveluscher:
I've started to have a few second thoughts here.

1. Responsibly removing these methods would require a major version bump, which would take legacy web3.js to 2.0.
2. The new library that we _actually_ intend on releasing as web3.js 2.0 already has these methods gone.

I think maybe we just cut our losses for legacy, and close #2859 by releasing the new library. Two birds, one stone.

sekaieth:
> I've started to have a few second thoughts here.
> 
> 1. Responsibly removing these methods would require a major version bump, which would take legacy web3.js to 2.0.
> 2. The new library that we _actually_ intend on releasing as web3.js 2.0 already has these methods gone.
> 
> I think maybe we just cut our losses for legacy, and close #2859 by releasing the new library. Two birds, one stone.

Hi Steve,

Is there an estimate on when the new library will be released?  Or, if it is already released, can you help me with a link?

steveluscher:
Hey @sekaieth; you can follow along with the Release Candidates [here](https://github.com/solana-labs/solana-web3.js/releases/tag/v2.0.0-rc.1). It's looking like the final will come shortly after Solana Breakpoint this year.

sekaieth:
Thanks!  

I'm on a wallet team.  Will the Agave 2.0 fork break my application if we don't upgrade this sdk immediately?

steveluscher:
The versions aren't related. `@solana/web3.js` 2.0 breaks all APIs from the application perspective, not the network perspective. There are no RPC API changes between the network at 1.18 or 2.0, _except_ that some long-deprecated methods have been permanently removed.

sekaieth:
Thanks!

steveluscher:
[...]

PR #2896, opened 2024-06-29
reafactor(Agave 2.0): Remove All Instances of `getFees` and `getFeeCalculatorForBlockhash`

0xIchigo:
This PR aims to start addressing #2859 by eliminating all instances of `getFees` and `getFeeCalculatorForBlockhash` as they'll be removed in Agave 2.0


steveluscher:
Thanks for this!

Before you go too far down this rabbit hole, I've started to have a few second thoughts here.

1. Responsibly removing these methods would require a major version bump, which would take legacy web3.js to 2.0.
2. The new library that we _actually_ intend on releasing as web3.js 2.0 already has these methods gone.

I think maybe we just cut our losses for legacy, and close #2859 by releasing the new library. Two birds, one stone.

0xIchigo:
No worries @steveluscher! I figured I'd start with this PR and go from there. Should I close this PR then or leave it open for the time being?

PR #2902, opened 2024-07-01
Change the webcrypto polyfill to require calling an `install` function

### LINKED ISSUE #2898: You should be able to time when to polyfill Ed25519 crypto methods
## Motivation

Presently, for environments that don't support Ed25519 natively (React Native, old browsers) we need a polyfill. We would benefit from being able to time when that polyfill installs.

## Example use case

Our current implementation of that is to do a raw require of this:

```ts
import '@solana/webcrypto-ed25519-polyfill';
```

There comes a problem when you're building a React Native application. RN applications themselves need a polyfill of the entire `SubtleCrypto` API because it doesn't exist to begin with.

One example of this is https://github.com/margelo/react-native-quick-crypto. This doesn't work with our solution because doing this:

```ts
import { install } from 'react-native-quick-crypto';
install();
import '@solana/webcrypto-ed25519-polyfill';
```

Results in this sequence of events:

1. Our polyfill's module factory runs first and creates `globalThis.crypto.subtle`
2. The `install()` method of `react-native-quick-crypto` runs second and overwrites `globalThis.crypto`

And you're left without the Ed25519 polyfill. Womp womp.

## Details

Either

1. Stop polyfilling in the module factory and instead export an `install()` method that people have to call. Update all the docs and example apps, or
2. Export _two_ versions of `@solana/webcrypto-ed25519-polyfill` (eg. export a `@solana/webcrypto-ed25519-polyfill/installer`) so that you can either bare-import the automatic-install version or import an `install()` method that you can call at the right time.

The second would be elegant, but [subpackage exports](https://webpack.js.org/guides/package-exports/) might be a huge pain in the ass so don't kill yourself trying to make it work.

mcintyre94:
This PR refactors the webcrypto polyfill so that all its functionality is in an `install` function. This must be called to enable the polyfill and it is no longer automatically installed by just importing. This gives apps more control over when the polyfill is called.

Example:

```ts
import { install } from '@solana/webcrypto-ed25519-polyfill';

async function main() {
    try {
        await crypto.subtle.generateKey('Ed25519', false, ['sign']);
    } catch (e) {
        console.log('unable to create keypair first time');
    }
    install();
    const keyPair = await crypto.subtle.generateKey('Ed25519', false, ['sign']); 
    console.log(keyPair);
}

main().then(() => console.log('done!'));
```

This will first log "unable to create keypair first time", as the polyfill is not yet installed. It will then log an Ed25519 keypair object created after installing the polyfill. 

Fixes #2898


lorisleiva:
Nice! Since that's all we're exporting, how would feel about exporting the `install` function as the default export?

mcintyre94:
Actually that's probably a nice idea because in the case of Steve's example in the issue, another library has a named `install` export. Letting callers name ours themselves probably makes it a bit nicer to use.

steveluscher (approved):
This is great. We can always add (back) an automatic option if we want to figure out submodule exports (ie. `import '@solana/webcrypto-ed25519-polyfill/auto';` or something).

Added via Giphy

PR #2907, opened 2024-07-03
`__DEV__` logic is now ‘NODE_ENV is not production’

steveluscher:
# Summary

A few folks have been caught out _not_ having _any_ setting for `NODE_ENV`. In cases like this, the `__DEV__` flag will be `false` because `NODE_ENV` is not `'development'`.

It would be preferable for development to be the default when there is no `NODE_ENV`. This means it will be harder to get ‘into’ production mode (you have to set an env variable) but it will mean that people fall into the pit of success when debug functionality is what they're looking for.

# Test Plan

```
pnpm i
pnpm turbo compile:js
```

Observe that all of the `__DEV__` constants have been replaced with `process.env.NODE_ENV !== 'production'`

### DIFF
--- a/packages/rpc-subscriptions-spec/src/__tests__/rpc-subscription-test.ts
+++ b/packages/rpc-subscriptions-spec/src/__tests__/rpc-subscription-test.ts
@@ -2,7 +2,6 @@ import {
     SOLANA_ERROR__RPC_SUBSCRIPTIONS__CANNOT_CREATE_SUBSCRIPTION_REQUEST,
     SOLANA_ERROR__RPC_SUBSCRIPTIONS__EXPECTED_SERVER_SUBSCRIPTION_ID,
     SolanaError,
-    SolanaErrorCode,
 } from '@solana/errors';
 import { createRpcMessage } from '@solana/rpc-spec-types';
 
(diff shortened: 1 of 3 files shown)


steveluscher:
* **#2907** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2907?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

steveluscher:
cc/ @jac0xb

steveluscher:
This mirrors, for instance, React's behaviour. Their library uses `process.env.NODE_ENV === 'production'` to kick it into production mode.

packages/rpc-subscriptions-spec/src/__tests__/rpc-subscription-test.ts
```diff
@@ -263,12 +266,14 @@ describe('JSON-RPC 2.0 Subscriptions', () => {
...
-                error: { code: 123, message: 'o no' },
+                error: { code: SOLANA_ERROR__SUBTLE_CRYPTO__DIGEST_UNIMPLEMENTED, message: 'o no block height' },
                 id: 0,
             });
         });
         const subscribePromise = rpc.thingNotifications().subscribe({ abortSignal: new AbortController().signal });
-        await expect(subscribePromise).rejects.toThrow(new SolanaError(123 as SolanaErrorCode, undefined));
+        await expect(subscribePromise).rejects.toThrow(
+            new SolanaError(SOLANA_ERROR__SUBTLE_CRYPTO__DIGEST_UNIMPLEMENTED),
+        );
```
mcintyre94, line 276: CI error seems to be related to these changes - are they intended to be in this PR at all?

mcintyre94 (approved):
LGTM!

lorisleiva (approved):
Thanks, I'll make sure I replicate on Kinobi when I'm out of this documentation hole.

packages/rpc-subscriptions-spec/src/__tests__/rpc-subscription-test.ts
steveluscher: Hmm. This change was supposed to fix a test failure owing to the fact that error 123 doesn’t exist.

steveluscher: Oh, it’s lint I left behind. 

Maybe I’ll try to do this properly with mocks this time.

PR #2910, opened 2024-07-04
The RPC request coalescer now throws exceptions whether you configure it with an `AbortSignal` or not

Jac0xb:
# Summary

When an `AbortSignal` was supplied to the RPC request coalescer, it would fail to attach a rejection handler to the shim promise returned to the app.

# Test Plan

```
cd packages/rpc/
```

### DIFF
--- a/.changeset/chilled-walls-scream.md
+++ b/.changeset/chilled-walls-scream.md
@@ -1,0 +1,5 @@
+---
+'@solana/rpc': patch
+---
+
+Fixed a bug where the RPC would fail to throw errors in the event that you configured it with an `AbortSignal`
--- a/packages/rpc/src/__tests__/rpc-request-coalescer-test.ts
+++ b/packages/rpc/src/__tests__/rpc-request-coalescer-test.ts
@@ -218,7 +218,8 @@ describe('RPC request coalescer', () => {
             });
         });
     });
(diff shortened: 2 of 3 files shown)


Jac0xb:
Open to other solutions just want to surface this

steveluscher (changes requested):
It's almost certainly this:

```ts
            return await new Promise<TResponse>((resolve, reject) => {
                const handleAbort = (e: AbortSignalEventMap['abort']) => {
                    signal.removeEventListener('abort', handleAbort);
                    coalescedRequest.numConsumers -= 1;
                    Promise.resolve().then(() => {
                        if (coalescedRequest.numConsumers === 0) {
                            const abortController = coalescedRequest.abortController;
                            abortController.abort((EXPLICIT_ABORT_TOKEN ||= createExplicitAbortToken()));
                        }
                    });
                    reject((e.target as AbortSignal).reason);
                };
                signal.addEventListener('abort', handleAbort);
                responsePromise.then(resolve).finally(() => {
                    signal.removeEventListener('abort', handleAbort);
                }); // <==== this is missing a `catch(reject)`
            });
```

When there's an `abortSignal` we drop into that shim, and `responsePromise` is not folded into the shim promise, so it just explodes all over the event loop.

I'm happy to fix this, but if you want to fix it, it's gonna need a realllllly good failing test!

PR #2911, opened 2024-07-04
Bust Turborepo cache when the test validator version changes

### LINKED ISSUE #2718: Add test validator to Turborepo dependencies so that cache gets bust when validator version changes
https://github.com/solana-labs/solana-web3.js/pull/2717 passed CI in the pull request but not after land. This is because the PR CI used the Turborepo cache, which didn't actually run the tests against the new validator anew.

The reason is because `turbo.json` is not configured to bust the cache when the validator version changes.

Find a way to add `.agave/` or `.agave/active_release/version.yml` to the Turborepo dependencies.

steveluscher:
Closes #2718.

# Test Plan

```
pnpm turbo test:unit:node --filter=@solana/rpc-spec
pnpm turbo test:unit:node --filter=@solana/rpc-spec
nano .agave/active_release/version.yml  # Change version
pnpm turbo test:unit:node --filter=@solana/rpc-spec
```

Observe that first call is uncached, second is cached, and third is uncached.

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -81,11 +81,11 @@
         },
         "test:unit:browser": {
             "dependsOn": ["^compile:js"],
-            "inputs": ["$TURBO_DEFAULT$", "src/**"]
+            "inputs": ["$TURBO_DEFAULT$", "../../.agave/**/version.yml", "src/**"]
         },
         "test:unit:node": {
             "dependsOn": ["^compile:js"],
-            "inputs": ["$TURBO_DEFAULT$", "src/**"]
+            "inputs": ["$TURBO_DEFAULT$", "../../.agave/**/version.yml", "src/**"]
         },
         "test:treeshakability:browser": {
             "dependsOn": ["compile:js"]
@@ -146,7 +146,7 @@
         },
         "@solana/web3.js#test:live-with-test-validator": {
             "dependsOn": ["compile:js"],
-            "inputs": ["$TURBO_DEFAULT$", "src/**", "test/**"]
+            "inputs": ["$TURBO_DEFAULT$", "../../.agave/**/version.yml", "src/**", "test/**"]
         },
         "@solana/web3.js#test:typecheck": {
             "dependsOn": ["compile:typedefs"],


steveluscher:
* **#2911** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2911?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

turbo.json
```diff
@@ -81,11 +81,11 @@
         },
         "test:unit:browser": {
             "dependsOn": ["^compile:js"],
-            "inputs": ["$TURBO_DEFAULT$", "src/**"]
+            "inputs": ["$TURBO_DEFAULT$", ".agave/active_release/version.yml", "src/**"]
```
steveluscher, line 84: Surely likely to break one day, but close enough for jazz?

steveluscher:
Oh crap, this path needs to be relative to the workspace root.

turbo.json
```diff
@@ -81,11 +81,11 @@
         },
         "test:unit:browser": {
             "dependsOn": ["^compile:js"],
-            "inputs": ["$TURBO_DEFAULT$", "src/**"]
+            "inputs": ["$TURBO_DEFAULT$", "../../.agave/**/version.yml", "src/**"]
```
steveluscher, line 84: K don't even get me started on tooling _not_ following symlinks (I couldn't write `active_release/` here; had to write `**/`).

buffalojoec: But wait, does this work? Won't two asterisks match `version.yml` across multiple versions? Why doesn't Turbo recognize the symlink?

buffalojoec: It will only work if we can reliably assume our CI never installs more than one version. Which is true now, but will surely break one day.

steveluscher: You're right! In any case, it's likely that a change in the set of `versions.yml` will imply that you installed a _newer_ version and are using that now.

buffalojoec (approved):
Wfm!

PR #2934, opened 2024-07-11
Prevent the balance component's Suspense fallback from, itself, suspending

steveluscher:
# Summary

Turns out, this mistake was the entire reason that I needed a `startTransition()` here where one didn't make any sense. The `fallback` here could, itself, suspend.

The only reason I stuck `<Balance />` in there was so that it would hold open the same height and not cause a layout shift. Bad move. Just draw the box the size it needs to be.

# Test Plan

Select a bunch of wallets and change the cluster. Observe the balance spinner appear and disappear without causing layout shift.


steveluscher:
* **#2934** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2934?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

lorisleiva (approved):

PR #2950, opened 2024-07-17
RPC: Remove JSON RPC specific behaviour from rpc-spec, move into default transformer

### LINKED ISSUE #2946: Experimental: Custom transports are required to wrap their response in a `{result: ...}` property
## Overview

I've been testing writing a custom transport for the new web3js, to handle cases like Helius where some endpoints are JSON-RPC while others are simple fetch requests to an API server. 

A transport for simple fetch requests could look something like:

```ts
function createFetchTransport(heliusApiUrl: HeliusApiUrl): RpcTransport {
    return async <TResponse>(...args: Parameters<RpcTransport>): Promise<TResponse> => {
        // const url = ...
        const response = await fetch(url, { signal: args[0].signal });
        return (await response.json()) as TResponse;
    };
}
```

But I've found that wrapping such a transport in an RPC returns undefined when it's called.

This is because the web3js code expects to read a `result` property, ie it treats all transports like a JSON-RPC transport. 

## Steps to reproduce

```ts
import { createRpc, createRpcApi, RpcApiMethods, RpcTransport } from "@solana/web3.js";

interface MyCustomApi extends RpcApiMethods {
  myFunction(): number[];
}

function createCustomTransport(): RpcTransport {
  return async <TResponse>(): Promise<TResponse> => {
    return [1, 2, 3, 4, 5] as TResponse;
  };
}

const api = createRpcApi<MyCustomApi>();

async function main() {
  const transport = createCustomTransport();
  const transportResponse = await transport({
    payload: { method: "myFunction", params: {} },
  });
  // [1, 2, 3, 4, 5]
  console.log({ transportResponse });

  const rpc = createRpc({ api, transport });
  const noResultRpcResponse = await rpc.myFunction().send();
  // undefined
  console.log({ noResultRpcResponse });
}

main();
```

Replit: https://replit.com/@mcintyre94/SolanaTransportTest?v=1

## Description of bug

I think we should either:
- Allow transports to return arbitrary data
- Add a result mapper to RPC, which modifies the data returned by the transport before returning it. For our JSON-RPC implementation this would return the `result` property. But by default for a custom transport it'd just return it as-is. 

Or document (and enforce with types) that transports must return their data in a `result` property.

mcintyre94:
This PR does the following:

- `rpc-spec` no longer returns the `result` field of the transport. This means that a custom transport does not need to use this JSON RPC concept. 
- Instead the default transformer now operates on the `result` field, since its input is a JSON RPC transport. 
- `rpc-spec` no longer detects an `error` field and converts it to a `SolanaError`. This means that custom transports can have arbitrary error handling. It now throws any error from the transport as-is. 
- Instead the default transformer now detects an `error` field and converts it to a `SolanaError` using the same logic, so that our default RPC has the same error handling as before.

This removes the need for transports to implement parts of the JSON RPC spec. They are free to return any arbitrary data, and the RPC will return it as-is. They can also implement any error handling and are not limited to JSON RPC style errors, nor our `SolanaError` wrapper.  

Note that this PR includes only the minimum necessary change to `rpc-subscriptions-spec`, which no longer unpacks the `result` field before calling its response processor, since it shares the same default one. I think error handling would need to be a bit different if we removed it from `rpc-subscriptions-spec`, and that might require using a different default response processor. 

Fixes #2946

### DIFF
--- a/packages/rpc-spec/package.json
+++ b/packages/rpc-spec/package.json
@@ -63,7 +63,6 @@
         "maintained node versions"
     ],
     "dependencies": {
-        "@solana/errors": "workspace:*",
         "@solana/rpc-spec-types": "workspace:*"
     },
     "peerDependencies": {
--- a/packages/rpc-spec/src/__tests__/rpc-test.ts
+++ b/packages/rpc-spec/src/__tests__/rpc-test.ts
@@ -1,4 +1,3 @@
-import { SOLANA_ERROR__JSON_RPC__PARSE_ERROR, SolanaError } from '@solana/errors';
 import { createRpcMessage } from '@solana/rpc-spec-types';
 
 import { createRpc, Rpc } from '../rpc';
@@ -35,19 +34,16 @@ describe('JSON-RPC 2.0', () => {
     });
     it('returns results from the transport', async () => {
         expect.assertions(1);
-        (makeHttpRequest as jest.Mock).mockResolvedValueOnce({ result: 123 });
+        (makeHttpRequest as jest.Mock).mockResolvedValueOnce(123);
         const result = await rpc.someMethod().send();
         expect(result).toBe(123);
     });
     it('throws errors from the transport', async () => {
         expect.assertions(1);
-        (makeHttpRequest as jest.Mock).mockResolvedValueOnce({
-            error: { code: SOLANA_ERROR__JSON_RPC__PARSE_ERROR, message: 'o no' },
-        });
+        const transportError = new Error('o no');
+        (makeHttpRequest as jest.Mock).mockRejectedValueOnce(transportError);
         const sendPromise = rpc.someMethod().send();
-        await expect(sendPromise).rejects.toThrow(
-            new SolanaError(SOLANA_ERROR__JSON_RPC__PARSE_ERROR, { __serverMessage: 'o no' }),
-        );
+        await expect(sendPromise).rejects.toThrow(transportError);
     });
     describe('when calling a method having a concrete implementation', () => {
         let rpc: Rpc<TestRpcMethods>;
@@ -94,13 +90,13 @@ describe('JSON-RPC 2.0', () => {
         });
         it('calls the response transformer with the response from the JSON-RPC 2.0 endpoint', async () => {
             expect.assertions(1);
-            (makeHttpRequest as jest.Mock).mockResolvedValueOnce({ result: 123 });
+            (makeHttpRequest as jest.Mock).mockResolvedValueOnce(123);
             await rpc.someMethod().send();
             expect(responseTransformer).toHaveBeenCalledWith(123, 'someMethod');
         });
         it('returns the processed response', async () => {
             expect.assertions(1);
-            (makeHttpRequest as jest.Mock).mockResolvedValueOnce({ result: 123 });
+            (makeHttpRequest as jest.Mock).mockResolvedValueOnce(123);
             const result = await rpc.someMethod().send();
             expect(result).toBe('123 processed response');
         });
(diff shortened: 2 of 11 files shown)


packages/rpc-spec/package.json
```diff
@@ -63,7 +63,6 @@
         "maintained node versions"
     ],
     "dependencies": {
-        "@solana/errors": "workspace:*",
```
steveluscher, line 66: Added via Giphy

packages/rpc-api/src/index.ts
```diff
@@ -188,7 +188,7 @@ export function createSolanaRpcApi<
         parametersTransformer: getDefaultParamsTransformerForSolanaRpc(config) as (params: unknown[]) => unknown[],
         responseTransformer: getDefaultResponseTransformerForSolanaRpc({
             allowedNumericKeyPaths: getAllowedNumericKeypaths(),
-        }),
+        }) as RpcApiConfig['responseTransformer'],
```
steveluscher, line 191: This is probably because `packages/rpc-transformers/src/response-transformer.ts` is missing a return type annotation. Does the need to do this go away if you add one?

packages/rpc-spec/src/__tests__/rpc-test.ts
```diff
@@ -35,19 +34,16 @@ describe('JSON-RPC 2.0', () => {
     });
     it('returns results from the transport', async () => {
         expect.assertions(1);
-        (makeHttpRequest as jest.Mock).mockResolvedValueOnce({ result: 123 });
```
steveluscher, line 38: Haha. Smoking gun, right here.

packages/rpc-subscriptions-api/src/index.ts
```diff
@@ -53,7 +54,7 @@ function createSolanaRpcSubscriptionsApi_INTERNAL<TApi extends RpcSubscriptionsA
         parametersTransformer: getDefaultParamsTransformerForSolanaRpc(config) as (params: unknown[]) => unknown[],
         responseTransformer: getDefaultResponseTransformerForSolanaRpc({
             allowedNumericKeyPaths: getAllowedNumericKeypaths(),
-        }),
+        }) as RpcSubscriptionsApiConfig['responseTransformer'],
```
steveluscher, line 57: Same question about `packages/rpc-transformers/src/response-transformer.ts`.

packages/rpc-transformers/src/__tests__/response-transformer-test.ts
```diff
@@ -1,12 +1,15 @@
+import { SOLANA_ERROR__JSON_RPC__PARSE_ERROR, SolanaError } from '@solana/errors';
+
 import { getDefaultResponseTransformerForSolanaRpc } from '../response-transformer';
 import { KEYPATH_WILDCARD } from '../tree-traversal';
 
 describe('getDefaultResponseTransformerForSolanaRpc', () => {
     describe('given an array as input', () => {
         const input = [10, 10n, '10', ['10', [10n, 10], 10]] as const;
+        const result = { result: input };
```
steveluscher, line 9: Maybe…

```suggestion
        const response = { result: input };
```

packages/rpc-transformers/src/response-transformer.ts
```diff
@@ -6,14 +8,19 @@ export type ResponseTransformerConfig<TApi> = Readonly<{
     allowedNumericKeyPaths?: AllowedNumericKeypaths<TApi>;
 }>;
 
+type JsonRpcResponse = { error: Parameters<typeof getSolanaErrorFromJsonRpcError>[0] } | { result: unknown };
+
 export function getDefaultResponseTransformerForSolanaRpc<TApi>(config?: ResponseTransformerConfig<TApi>) {
```
steveluscher, line 13: Yeah, this one. Maybe needs a return type annotation?

steveluscher (approved):
Added via Giphy

packages/rpc-transformers/src/response-transformer.ts
mcintyre94: Added the return type as `NonNullable<RpcApiConfig['responseTransformer']>`. I also needed to tweak the `rpc-spec` to make `methodName` optional on the response transformer, matching this implementation, which I think makes sense anyway.

PR #2969, opened 2024-07-22
Add a function to replace accounts in a transaction message using lookup tables

mcintyre94:
This PR adds a new function `useAddressLookupTablesForTransactionMessage` to `@solana/transaction-messages`.

It takes as input a `TransactionMessage` and a mapping of lookup table address to its list of addresses. For each account that is not a signer of the input transaction message, and is not already a lookup, if it appears in one of the lookup tables then it is replaced by a lookup meta for that lookup table.

This means that a developer can build a transaction without worrying about lookup tables, and then call this with the lookup tables they want to use. This is similar to the legacy `.compileToV0Message([lookupTable])` API.

This is not the same operation as is done when we compile a transaction message, this is simply replacing the accounts so that when the transaction message is compiled it will include the lookup tables.

Note: this is not as sophisticated as it could be, in terms of minimising the final transaction size. If you pass in multiple lookup tables, the output transaction message may use more of them than is optimal. Eg if we passed `{lookup1: [abc], lookup2: [def], lookup3: [abc, def]}` then it may use lookup1 and lookup2 instead of only using lookup3.

Also note that this doesn't guarantee that replaced accounts will use these lookup tables if the message is modified further. Eg if the message is modified such that an account becomes a signer after this is called, then `compileTransactionMessage` will change it back to a static account.


packages/transaction-messages/src/use-address-lookup-tables.ts
```diff
@@ -0,0 +1,86 @@
...
+    // All signers will need to be static keys, so we shouldn't replace them if they're used elsewhere in the transaction,
+    // even if they're also in a lookup table
+    const signers = new Set(
+        transactionMessage.instructions
+            .flatMap(i => (i.accounts ?? []).filter(a => isSignerRole(a.role)))
+            .map(a => a.address),
+    );
+    if ('feePayer' in transactionMessage) {
+        signers.add((transactionMessage as ITransactionMessageWithFeePayer).feePayer);
+    }
```
steveluscher, line 49: So, I think you should only ever bail on a _particular_ account entry if it's a signer, instead of looking across _all_ accounts to make sure the address has never been used as a signer.

1. After optimization, someone _could_ change the fee payer, delete an instruction, or change signers. Skipping this whole entry could result in a sub-optimal account lookup set after such a modification.
2. The message compiler already [throws out LUT entries that have to sign](https://github.com/solana-labs/solana-web3.js/blob/9239e6ec972b4de9f0d15b197fbef1d2871759d9/packages/transaction-messages/src/compile/accounts.ts#L120-L124).

So if you have two instructions, each with a reference to address `ABC`, one a signer and one not, I think you should leave the signer entry alone, and replace the non-signer one with an `IAccountLookupMeta` _just_ on the off chance that the transaction is later modified or split into two or whatever in such a way that LUT entry _can_ all of a sudden be used.

Also makes this code simpler.

packages/transaction-messages/src/use-address-lookup-tables.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    const lookupTableAddresses = new Set(Object.values(addressesByLookupTableAddress).flatMap(a => a));
+
+    const newInstructions: IInstruction[] = [];
```
steveluscher, line 53: You want ‘if nothing changed, return the exact same object’ behaviour here. Like [this](https://github.com/solana-labs/solana-web3.js/blob/9239e6ec972b4de9f0d15b197fbef1d2871759d9/packages/transactions/src/signatures.ts#L82-L84), tested [here](https://github.com/solana-labs/solana-web3.js/blob/9239e6ec972b4de9f0d15b197fbef1d2871759d9/packages/transactions/src/__tests__/signatures-test.ts#L257-L266).

packages/transaction-messages/src/use-address-lookup-tables.ts
```diff
@@ -0,0 +1,86 @@
...
+            // We already checked it's in one of the lookup tables
+            const lookupMetaAccount = findAddressInLookupTables(
+                account.address,
+                account.role,
+                addressesByLookupTableAddress,
+            )!;
+            newAccounts.push(lookupMetaAccount);
+        }
+
+        newInstructions.push({ ...instruction, accounts: newAccounts });
```
steveluscher, line 82: I think there's a bunch of `Object.freeze` that has to happen in here, and get tested like [this](https://github.com/solana-labs/solana-web3.js/blob/9239e6ec972b4de9f0d15b197fbef1d2871759d9/packages/signers/src/__tests__/add-signers-test.ts#L126-L141).

steveluscher (approved):
I like it! I'd like to change the name though.

Address lookup tables are an optimization that lets you compress a transaction over the wire. We should use one of those words.

* `compressTransactionMessageUsingAddressLookupTables(message, luts)`
* `optimizeTransactionMessageUsingAddressLookupTables(message, luts)`


I was initially going to say we should call it ‘optimize’ but ‘compress’ is alluring because of how specific and revelatory it is. An optimization could be anything, but what's _really_ going on here is that you're replacing `IAccountMeta` with `IAccountLookupMeta` that result in smaller compiled messages.

> If you pass in multiple lookup tables, the output transaction message may use more of them than is optimal

Ooh, interesting. We should maybe move this functionality into the `compileTransactionMessage`. There should be an algorithm in there that minimizes the number of lookup tables used.

packages/transaction-messages/src/use-address-lookup-tables.ts
mcintyre94: Got it, that makes sense, thanks!

mcintyre94:
Thanks @steveluscher! I like `compressTransactionMessageUsingAddressLookupTables` too, will rename to that.

lorisleiva (approved):
Perfect! 👌

packages/transaction-messages/src/compress-transaction-message.ts
```diff
@@ -0,0 +1,81 @@
...
+}
+
+type TransactionMessageNotLegacy = Exclude<TransactionMessage, { version: 'legacy' }>;
+
+export function compressTransactionMessageUsingAddressLookupTables<
+    TTransactionMessage extends TransactionMessageNotLegacy = TransactionMessageNotLegacy,
+>(
+    transactionMessage: TTransactionMessage,
+    addressesByLookupTableAddress: AddressesByLookupTableAddress,
+): TTransactionMessage {
```
steveluscher, line 38: This isn't exactly true, I guess. This function probably actually widens every account from whatever it is, back out to `IAccountMeta<TAddress> | IAccountLookupMeta<TAddress>`.

What would be the implications of mapping over this puppy and making that so?

```ts
// Pseudo TypeScript...
): 
  | TTransactionMessage  // Unchanged, or
  | TTransactionMessage & TTransactionMessage.instructions.map(i => ({
    ...i,
    accounts: i.accounts.map(a => IAccountMeta<typeof a.address> | IAccountLookupMeta<typeof a.address>)
  }))
{
```

lorisleiva: Haha I thought about this but the content of `AddressesByLookupTableAddress` would need to be fully static for this to even work which will probably never be the case.

mcintyre94: Fair point! I did this just to keep `ITransactionMessageWithFeePayer` etc intact on the output message, but it isn't strictly accurate if you've got the `accounts` field typed more specifically.

mcintyre94: Not 100% on this type but I think this works? https://github.com/solana-labs/solana-web3.js/pull/2969/commits/b643109329928112912a3bea5b6340f1135352a4

PR #2974, opened 2024-07-23
An example of how to use the new `@solana/web3.js` to deserialize and parse a transaction

mcintyre94:
This PR adds an example of how to use web3js to decode/inspect/parse transactions. It starts by creating a transaction, partially signing it, and encoding it as base64. It then demonstrates:

- Decoding it to a `Transaction`
- Inspecting the signatures of a `Transaction`
- Verifying a signature on a `Transaction`
- Decoding the message bytes to a `CompiledTransactionMessage`
- Decoding to a `TransactionMessage`, resolving the lookup tables
- Parsing the data of a transfer SOL instruction
- Parsing the data of a memo instruction

Example output:

```
INFO (Deserialize): [setup] Setting lookup table address
    address: "2JPQuT3dHtPjrdcbUQyrrT4XYRYaWpWfmAJ54SUapg6n"
INFO (Deserialize): [setup] Got a blockhash
    blockhash: "28zB6pcMaWZtUmpMPTGHYUy6T1AawusvogCrWA2emhK7"
    lastValidBlockHeight: 152903
(node:31022) ExperimentalWarning: The Ed25519 Web Crypto API algorithm is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
INFO (Deserialize): [setup] Created the transaction message
INFO (Deserialize): [setup] Compressed the transaction message using lookup table 2JPQuT3dHtPjrdcbUQyrrT4XYRYaWpWfmAJ54SUapg6n
INFO (Deserialize): [setup] Partially signed the transaction with a signature from ED1WqT2hWJLSZtj4TtTdoovmpMrr7zpkUdbfxmcJR1Fq
INFO (Deserialize): [setup] Encoded the transaction as base64: AgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA...
INFO (Deserialize): [step 1] Decoded the transaction
INFO (Deserialize): [step 1] 9xaf9RQvmr47tcKZ2y8KdpcSn6KyymGU6PZAFC9AKjPd is required to sign the transaction but hasn't yet
INFO (Deserialize): [step 1] ED1WqT2hWJLSZtj4TtTdoovmpMrr7zpkUdbfxmcJR1Fq has signed the transaction
INFO (Deserialize): [step 1] The signature for ED1WqT2hWJLSZtj4TtTdoovmpMrr7zpkUdbfxmcJR1Fq is valid for the transaction
INFO (Deserialize): [step 2] The transaction is version 0
INFO (Deserialize): [step 2] We can see the transaction lifetime token, but we don't know if it's a blockhash or durable nonce: 28zB6pcMaWZtUmpMPTGHYUy6T1AawusvogCrWA2emhK7
INFO (Deserialize): [step 2] Static accounts of the transaction
    0: "9xaf9RQvmr47tcKZ2y8KdpcSn6KyymGU6PZAFC9AKjPd"
    1: "ED1WqT2hWJLSZtj4TtTdoovmpMrr7zpkUdbfxmcJR1Fq"
    2: "11111111111111111111111111111111"
    3: "MemoSq4gqABAXKb96qnH8TysNcWxMyWCqXgDLGmfcHr"
INFO (Deserialize): [step 2] Address lookup tables for the transaction
    0: {
      "lookupTableAddress": "2JPQuT3dHtPjrdcbUQyrrT4XYRYaWpWfmAJ54SUapg6n",
      "writableIndices": [
        0
      ],
      "readableIndices": []
    }
INFO (Deserialize): [step 2] The first instruction of the compiled transaction message
    programAddressIndex: 2
    accountIndices: [
      1,
      4
    ]
    data: "(removed for brevity)"
INFO (Deserialize): [step 3] The transaction fee payer is 9xaf9RQvmr47tcKZ2y8KdpcSn6KyymGU6PZAFC9AKjPd
INFO (Deserialize): [step 3] The transaction lifetime constraint
    blockhash: "28zB6pcMaWZtUmpMPTGHYUy6T1AawusvogCrWA2emhK7"
    lastValidBlockHeight: 18446744073709552000
INFO (Deserialize): [step 3] The transaction blockhash is 28zB6pcMaWZtUmpMPTGHYUy6T1AawusvogCrWA2emhK7
INFO (Deserialize): [step 3] The first decoded instruction
    programAddress: "11111111111111111111111111111111"
    accounts: [
      {
        "address": "ED1WqT2hWJLSZtj4TtTdoovmpMrr7zpkUdbfxmcJR1Fq",
        "role": 3
      },
      {
        "address": "F1Vc6AGoxXLwGB7QV8f4So3C5d8SXEk3KKGHxKGEJ8qn",
        "addressIndex": 0,
        "lookupTableAddress": "2JPQuT3dHtPjrdcbUQyrrT4XYRYaWpWfmAJ54SUapg6n",
        "role": 1
      }
    ]
    data: "(removed for brevity)"
INFO (Deserialize): [step 3] The data bytes of the first instruction
    0: 2
    1: 0
    2: 0
    3: 0
    4: 78
    5: 97
    6: 188
    7: 0
    8: 0
    9: 0
    10: 0
    11: 0
INFO (Deserialize): [step 4] The first instruction calls the TransferSol instruction of the system program
INFO (Deserialize): [step 4] The parsed Transfer SOL instruction
    programAddress: "11111111111111111111111111111111"
    accounts: {
      "source": {
        "address": "ED1WqT2hWJLSZtj4TtTdoovmpMrr7zpkUdbfxmcJR1Fq",
        "role": 3
      },
      "destination": {
        "address": "F1Vc6AGoxXLwGB7QV8f4So3C5d8SXEk3KKGHxKGEJ8qn",
        "addressIndex": 0,
        "lookupTableAddress": "2JPQuT3dHtPjrdcbUQyrrT4XYRYaWpWfmAJ54SUapg6n",
        "role": 1
      }
    }
    data: {
      "amount": 12345678,
      "discriminator": 2
    }
INFO (Deserialize): [step 4] In the first instruction, ED1WqT2hWJLSZtj4TtTdoovmpMrr7zpkUdbfxmcJR1Fq transfers 12,345,678 lamports to F1Vc6AGoxXLwGB7QV8f4So3C5d8SXEk3KKGHxKGEJ8qn
INFO (Deserialize): [step 4] The second instruction calls the MemoSq4gqABAXKb96qnH8TysNcWxMyWCqXgDLGmfcHr program
INFO (Deserialize): [step 4] The second instruction calls the memo program
INFO (Deserialize): [step 4] The parsed Add Memo instruction
    programAddress: "MemoSq4gqABAXKb96qnH8TysNcWxMyWCqXgDLGmfcHr"
    data: {
      "memo": "hello from @solana/web3.js"
    }
INFO (Deserialize): [step 4] The second instruction adds a memo with message "hello from @solana/web3.js"
```

### DIFF
--- a/examples/deserialize-transaction/src/example.ts
+++ b/examples/deserialize-transaction/src/example.ts
@@ -1,0 +1,408 @@
+/**
+ * EXAMPLE
+ * Deserialize and inspect a transaction with @solana/web3.js.
+ *
+ * Before running any of the examples in this monorepo, make sure to set up a test validator by
+ * running `pnpm test:live-with-test-validator:setup` in the root directory.
+ *
+ * To run this example, execute `pnpm start` in this directory.
+ */
+import createLogger from '@solana/example-utils/createLogger.js';
+import {
+    Address,
+    appendTransactionMessageInstructions,
+    assertAccountDecoded,
+    assertAccountExists,
+    assertIsInstructionWithAccounts,
+    assertIsInstructionWithData,
+    compressTransactionMessageUsingAddressLookupTables,
+    createKeyPairSignerFromBytes,
+    createSolanaRpc,
+    createTransactionMessage,
+    decodeTransactionMessage,
+    fetchJsonParsedAccount,
+    getAddressEncoder,
+    getBase64EncodedWireTransaction,
+    getBase64Encoder,
+    getCompiledTransactionMessageDecoder,
+    getTransactionDecoder,
+    partiallySignTransactionMessageWithSigners,
+    pipe,
+    setTransactionMessageFeePayer,
+    setTransactionMessageLifetimeUsingBlockhash,
+    verifySignature,
+} from '@solana/web3.js';
+import { getAddMemoInstruction, MEMO_PROGRAM_ADDRESS, parseAddMemoInstruction } from '@solana-program/memo';
+import {
+    getTransferSolInstruction,
+    identifySystemInstruction,
+    parseTransferSolInstruction,
+    SystemInstruction,
+} from '@solana-program/system';
+
+const log = createLogger('Deserialize');
+
+/**
+ * SETUP: LOOKUP TABLE ADDRESS
+ * Our fixtures include a lookup table, which we're going to use for this transaction
+ * This is the address of the lookup table account from `scripts/fixtures/address-lookup-table-account.json`
+ */
+const LOOKUP_TABLE_ADDRESS = '2JPQuT3dHtPjrdcbUQyrrT4XYRYaWpWfmAJ54SUapg6n' as Address;
+log.info({ address: LOOKUP_TABLE_ADDRESS }, '[setup] Setting lookup table address');
+
+/**
+ * SETUP: RPC CONNECTION
+ * While in this example we won't send the transaction, we will use the remote procedure call (RPC) server to:
+ * - fetch the blockhash for the transaction lifetime
+ * - fetch the address lookup table used in the transaction
+ *
+ * This example uses your local test validator which must be running before you run this script.
+ */
+const rpc = createSolanaRpc('http://127.0.0.1:8899');
+
+/**
+ * SETUP: TRANSACTION LIFETIME
+ * We will fetch the latest blockhash, which we will use as the transaction's lifetime
+ * See `examples/transfer-lamports` for more detail on transaction lifetime
+ */
+const { value: latestBlockhash } = await rpc.getLatestBlockhash().send();
+log.info(latestBlockhash, '[setup] Got a blockhash');
+
+/**
+ * SETUP: CREATE A TRANSACTION MESSAGE
+ * We create a transaction that sends lamports from a source account to a destination account,
+ * where the destination account is in the lookup table.
+ * We will then serialize this transaction, and then explore how to deserialize and inspect it.
+ * See the `transfer-lamports` example for more detailed documentation on creating, signing and sending a transaction.
+ */
+const SOURCE_ACCOUNT_SIGNER = await createKeyPairSignerFromBytes(
+    /**
+     * These are the bytes that we saved at the time this account's key pair was originally
+     * generated. Here, they are inlined into the source code, but you can also imagine them being
+     * loaded from disk or, better yet, read from an environment variable.
+     */
+    new Uint8Array(
+        // prettier-ignore
+        [2, 194, 94, 194, 31, 15, 34, 248, 159, 9, 59, 156, 194, 152, 79, 148, 81, 17, 63, 53, 245, 175, 37, 0, 134, 90, 111, 236, 245, 160, 3, 50, 196, 59, 123, 60, 59, 151, 65, 255, 27, 247, 241, 230, 52, 54, 143, 136, 108, 160, 7, 128, 4, 14, 232, 119, 234, 61, 47, 158, 9, 241, 48, 140],
+    ), // Address: ED1WqT2hWJLSZtj4TtTdoovmpMrr7zpkUdbfxmcJR1Fq
+);
+log.info({ address: SOURCE_ACCOUNT_SIGNER.address }, '[setup] Loaded key pair for source account');
+
+// This is the first address in the lookup table
+const DESTINATION_ADDRESS = 'F1Vc6AGoxXLwGB7QV8f4So3C5d8SXEk3KKGHxKGEJ8qn' as Address;
+log.info({ address: DESTINATION_ADDRESS }, '[setup] Setting destination address');
+
+/**
+ * This is an arbitrary address that we set as the fee payer
+ * In this example we won't send the transaction so it doesn't matter if this account is funded
+ */
+const FEE_PAYER_ADDRESS = '9xaf9RQvmr47tcKZ2y8KdpcSn6KyymGU6PZAFC9AKjPd' as Address;
+log.info({ address: FEE_PAYER_ADDRESS }, '[setup] Setting fee payer address');
+
+const transactionMessage = pipe(
+    createTransactionMessage({ version: 0 }),
+    tx => setTransactionMessageFeePayer(FEE_PAYER_ADDRESS, tx),
+    tx => setTransactionMessageLifetimeUsingBlockhash(latestBlockhash, tx),
+    tx =>
+        appendTransactionMessageInstructions(
+            [
+                getTransferSolInstruction({
+                    amount: 12345678,
+                    destination: DESTINATION_ADDRESS,
+                    source: SOURCE_ACCOUNT_SIGNER,
+                }),
+                getAddMemoInstruction({
+                    memo: 'hello from @solana/web3.js',
+                }),
+            ],
+            tx,
+        ),
+);
+log.info('[setup] Created the transaction message');
+
+/**
+ * SETUP: COMPRESS WITH LOOKUP TABLE
+ * As the destination of our transfer SOL instruction is in our lookup table, we can use the
+ * lookup table to compress the transaction.
+ *
+ * This will modify the accounts referred to in the transaction, so that where possible they
+ * refer to an index of an address lookup table instead of an address directly. This makes the
+ * transaction smaller when it is compiled, particularly if many addresses can use a lookup table.
+ */
+
+// The only data from the lookup table we need is its addresses
+type LookupTableData = {
+    addresses: Address[];
+};
+
+// We fetch the JSON parsed representation of the lookup table from the RPC
+const lookupTableAccount = await fetchJsonParsedAccount<LookupTableData>(rpc, LOOKUP_TABLE_ADDRESS);
+assertAccountDecoded(lookupTableAccount);
+assertAccountExists(lookupTableAccount);
+
+const transactionMessageWithLookupTables = compressTransactionMessageUsingAddressLookupTables(transactionMessage, {
+    [LOOKUP_TABLE_ADDRESS]: lookupTableAccount.data.addresses,
+});
+log.info(`[setup] Compressed the transaction message using lookup table ${LOOKUP_TABLE_ADDRESS}`);
+
+/**
+ * SETUP: PARTIALLY SIGN THE TRANSACTION
+ * A Solana transaction can have multiple signers that must sign before it is sent to the network.
+ * In our case the fee payer and the source account would both need to sign the transaction
+ * We used a signer for the source account, so it will sign it when we call `signTransactionMessageWithSigners`
+ * But we only used an address for the fee payer, and we don't have its private key to sign the transaction
+ * So we will use `partiallySignTransactionMessageWithSigners` to get a transaction that has been signed by
+ * the source account, but not by the fee payer.
+ */
+const signedTransaction = await partiallySignTransactionMessageWithSigners(transactionMessageWithLookupTables);
+log.info(`[setup] Partially signed the transaction with a signature from ${SOURCE_ACCOUNT_SIGNER.address}`);
+
+/**
+ * SETUP: ENCODE THE TRANSACTION AS BASE64
+ * This is a common format to send a transaction between different systems
+ * If you wanted to convert it to a byte array instead, you could do this:
+ *
+ * ```ts
+ * const transactionEncoder = getTransactionEncoder();
+ * const transactionBytes = transactionEncoder.encode(signedTransaction);
+ * ```
+ */
+const base64EncodedTransaction = getBase64EncodedWireTransaction(signedTransaction);
+log.info(`[setup] Encoded the transaction as base64: ${base64EncodedTransaction.slice(0, 64)}...`);
+
+/**
+ * SETUP COMPLETE
+ * At this point, imagine we've received `base64EncodedTransaction` as input and don't know anything about it
+ * Let's use web3js to decode and inspect it
+ */
+
+/**
+ * STEP 1: DECODE TO TRANSACTION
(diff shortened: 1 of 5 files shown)


examples/deserialize-transaction/src/example.ts
```diff
@@ -0,0 +1,403 @@
...
+const transactionMessageWithLookupTables = compressTransactionMessageUsingAddressLookupTables(transactionMessage, {
+    [LOOKUP_TABLE_ADDRESS]: lookupTableAccount.data.addresses,
+});
+log.info(`[setup] Compressed the transaction message using lookup table ${LOOKUP_TABLE_ADDRESS}`);
+
+/**
+ * SETUP: PARTIALLY SIGN THE TRANSACTION
+ * A Solana transaction can have multiple signers that must sign before it is sent to the network.
+ * In our case the fee payer and the source account would both need to sign the transaction
+ * We used a signer for the source account, so it will sign it when we call `signTransactionMessageWithSigners`
```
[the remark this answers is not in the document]
buffalojoec: Looks like you maybe did this intentionally to articulate the other signing method `signTransactionMessageWithSigners`. If so, that's fine, just maybe add a few words like "instead we're going to use the partial version" or whatever.

examples/deserialize-transaction/src/example.ts
```diff
@@ -0,0 +1,403 @@
...
+ * const transactionBytes = transactionEncoder.encode(signedTransaction);
+ * ```
+ */
+const base64EncodedTransaction = getBase64EncodedWireTransaction(signedTransaction);
+log.info(`[setup] Encoded the transaction as base64: ${base64EncodedTransaction.slice(0, 64)}...`);
+
+/**
+ * SETUP COMPLETE
+ * At this point, imagine we've received `base64EncodedTransaction` as input and don't know anything about it
+ * Let's use web3js to decode and inspect it
```
buffalojoec, line 172: Hell yeah. 🔥

examples/deserialize-transaction/src/example.ts
```diff
@@ -0,0 +1,403 @@
...
+    }
+}
+
+// Now we know that we have a signature for `ED1WqT2hWJLSZtj4TtTdoovmpMrr7zpkUdbfxmcJR1Fq`
+// Let's verify that it's correct
+const signedByAddress = 'ED1WqT2hWJLSZtj4TtTdoovmpMrr7zpkUdbfxmcJR1Fq' as Address;
+const signedBySignature = decodedTransaction.signatures[signedByAddress]!;
+// We encode the source address to bytes
+const sourceAddressBytes = getAddressEncoder().encode(signedByAddress);
+// Then we create a public Ed25519 key with those bytes
```
buffalojoec, line 214: Wouldn't hurt to briefly plug in that this is a SubtleCrypto `CryptoKey` object, and you've designated that you want this key to have role `verify`.

examples/deserialize-transaction/src/example.ts
```diff
@@ -0,0 +1,403 @@
...
+ * representing the compiled version of the `TransactionMessage`. If we decode this we can see some information
+ * about the transaction. Let's do that next.
+ */
+
+// Again we have a decoder to convert from bytes to a Solana data structure, in this case the `CompiledTransactionMessage`
+const compiledTransactionMessageDecoder = getCompiledTransactionMessageDecoder();
+const compiledTransactionMessage = compiledTransactionMessageDecoder.decode(decodedTransaction.messageBytes);
+
+// This gives us the data stucture `CompiledTransactionMessage`. This is the format that transactions are
+// compiled to before being sent to the Solana network.
```
buffalojoec, line 234:
```suggestion
// compiled to before the entire transaction is encoded to base64 to be sent to the Solana network.
```

examples/deserialize-transaction/src/example.ts
```diff
@@ -0,0 +1,403 @@
...
+}
+
+// Now we can view address lookup tables:
+log.info(compiledTransactionMessage.addressTableLookups, '[step 2] Address lookup tables for the transaction');
+
+// We see that the transaction uses our address lookup table.
+
+// Let's look at a compiled instruction
+log.info(
+    { ...compiledTransactionMessage.instructions[0], data: '(removed for brevity)' },
```
buffalojoec, line 263: nit: Since this is a transfer, the instruction data should only be 8 bytes, right?

examples/deserialize-transaction/src/example.ts
```diff
@@ -0,0 +1,403 @@
...
+ * We know from the program address (11111111111111111111111111111111) that the first instruction
+ * is to the system program
+ * You can generate such a client for any Solana program using Kinobi
+ * See https://github.com/kinobi-so/kinobi for more information on Kinobi
+ */
+
+const firstInstruction = decodedTransactionMessage.instructions[0];
+// Narrow the type, an instruction must have data to be identified
+assertIsInstructionWithData(firstInstruction);
+const identifiedInstruction = identifySystemInstruction(firstInstruction);
```
buffalojoec, line 367: Instructions can have no data, though. There are plenty of instructions with no data that only require a list of accounts.

buffalojoec (commented):
Fantastic!! I left some small suggestions, but this is an awesome example. Nice job!!!

examples/deserialize-transaction/src/example.ts
```diff
@@ -0,0 +1,403 @@
...
+ * - fetch the address lookup table used in the transaction
+ *
+ * This example uses your local test validator which must be running before you run this script.
+ */
+const rpc = createSolanaRpc('http://127.0.0.1:8899');
+
+/**
+ * SETUP: TRANSACTION LIFETIME
+ * We will fetch the latest blockhash, which we will use as the transaction's lifetime
+ * See `examples/transfer-lamports` for more detail on transaction lifetime
```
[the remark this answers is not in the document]
lorisleiva: +1. These examples actually end up doing rather a lot so referring to one can be a bit confusing as the reader would then need to peruse the example to find the section that mentions transaction lifetimes.

It might be worth considering smaller examples such as "set blockhash transaction lifetime" and "set durable nonce transaction lifetime" that can be more easily referred by other examples in the docblocs (the code would still be duplicated so the example file is self-sufficient).

packages/transaction-messages/src/compress-transaction-message.ts
```diff
@@ -43,9 +43,15 @@ type WidenInstructionAccounts<TInstruction extends IInstruction> =
...
+type ExtractAdditionalProps<T, U> = Omit<T, keyof U>;
+
 type WidenTransactionMessageInstructions<TTransactionMessage extends TransactionMessage> =
     TTransactionMessage extends BaseTransactionMessage<infer TVersion, infer TInstruction>
-        ? BaseTransactionMessage<TVersion, WidenInstructionAccounts<TInstruction>>
+        ? BaseTransactionMessage<TVersion, WidenInstructionAccounts<TInstruction>> &
+              ExtractAdditionalProps<
+                  TTransactionMessage,
+                  BaseTransactionMessage<TVersion, WidenInstructionAccounts<TInstruction>>
+              >
```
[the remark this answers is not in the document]
mcintyre94: Yep - I missed this when I modified the type to adjust accounts. If you just return `BaseTransactionMessage<TVersion, WidenInstructionAccounts<TInstruction>>` then you lose eg `ITransactionMessageWithFeePayer` and then the functions that require compiling the message rightly error

This change means that everything additional to `BaseTransactionMessage` is retained, and it's only the accounts in the instructions in the `BaseTransactionMessage` that get manipulated.

examples/deserialize-transaction/src/example.ts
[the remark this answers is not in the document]
lorisleiva: Why not just use `setTransactionMessageFeePayerSigner` and then use the full power of signers? I think it's confusing to use both methods in the example.

examples/deserialize-transaction/src/example.ts
```diff
@@ -0,0 +1,403 @@
...
+
+const log = createLogger('Deserialize');
+
+/**
+ * SETUP: LOOKUP TABLE ADDRESS
+ * Our fixtures include a lookup table, which we're going to use for this transaction
+ * This is the address of the lookup table account from `scripts/fixtures/address-lookup-table-account.json`
+ */
+const LOOKUP_TABLE_ADDRESS = '2JPQuT3dHtPjrdcbUQyrrT4XYRYaWpWfmAJ54SUapg6n' as Address;
+log.info({ address: LOOKUP_TABLE_ADDRESS }, '[setup] Setting lookup table address');
```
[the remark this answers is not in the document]
mcintyre94: I copied the convention used in the `transfer-lamports` example, which prefixes everything with setup or a step number

https://github.com/solana-labs/solana-web3.js/blob/996234c6c69186bbe96240c734d9ed03c56009c8/examples/transfer-lamports/src/example.ts#L63-L64

examples/deserialize-transaction/src/example.ts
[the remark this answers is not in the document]
mcintyre94: Yep agreed, it'd be better to have a smaller example to refer to here

The main reason I didn't want to just copy the whole doc block is because some of the detail isn't necessarily relevant, since this example doesn't actually send/confirm the transaction

lorisleiva (approved):
Great job! My only concern is how lengthy these examples are becoming and maybe we should consider reducing the granularity of the next ones if possible so they are more digestible.

examples/deserialize-transaction/src/example.ts
```diff
@@ -0,0 +1,403 @@
...
+
+// We fetch the JSON parsed representation of the lookup table from the RPC
+const lookupTableAccount = await fetchJsonParsedAccount<LookupTableData>(rpc, LOOKUP_TABLE_ADDRESS);
+assertAccountDecoded(lookupTableAccount);
+assertAccountExists(lookupTableAccount);
+
+const transactionMessageWithLookupTables = compressTransactionMessageUsingAddressLookupTables(transactionMessage, {
+    [LOOKUP_TABLE_ADDRESS]: lookupTableAccount.data.addresses,
+});
+log.info(`[setup] Compressed the transaction message using lookup table ${LOOKUP_TABLE_ADDRESS}`);
```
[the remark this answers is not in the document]
mcintyre94: I don't think that should be in the logs, but I'll add a bit to the `SETUP: COMPRESS WITH LOOKUP TABLE` doc block on what's happening to the transaction message

examples/deserialize-transaction/src/example.ts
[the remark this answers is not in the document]
mcintyre94: I think I switched from sign to partially sign and didn't update this, it's probably less confusing to mention the `partiallySignTransactionMessageWithSigners` I'm actually using

examples/deserialize-transaction/src/example.ts
mcintyre94: I'm getting 12 bytes, I guess it's using a 4 byte discriminator

But the reason I did this is that the logger we use unrolls the array to one element per line, like this:

```
INFO (Deserialize): [step 2] The first instruction of the compiled transaction message
    programAddressIndex: 2
    accountIndices: [
      1,
      4
    ]
    data: {
      "0": 2,
      "1": 0,
      "2": 0,
      "3": 0,
      "4": 78,
      "5": 97,
      "6": 188,
      "7": 0,
      "8": 0,
      "9": 0,
      "10": 0,
      "11": 0
    }
```

Which I don't think is useful in this case until we actually want to look at the data later on

examples/deserialize-transaction/src/example.ts
mcintyre94: Are there programs that have multiple different instructions that are distinguished by the types of accounts passed in? Or would such a program (similarly to the memo program) just be treated as having one instruction type? 

My assumption was that a program that just takes accounts would (like the memo program) not have an `identify` function, and would just have one instruction that you can parse.

mcintyre94:
@lorisleiva Thanks! And yep agreed, I was thinking after I finished this that it might be better in a blog post format TBH!

examples/deserialize-transaction/src/example.ts
buffalojoec: That's a good point. I guess it's probably extremely rare, but you _could_ do different things based on the number of accounts provided. This would be an anti-pattern though.

buffalojoec: Re-reading this line in the example, I think I was just confused by what it was describing. You're saying an instruction must have data to be identified _by the assert function_, not in general. You can disregard my comment here.

examples/deserialize-transaction/src/example.ts
```diff
@@ -0,0 +1,408 @@
...
+} from '@solana-program/system';
+
+const log = createLogger('Deserialize');
+
+/**
+ * SETUP: LOOKUP TABLE ADDRESS
+ * Our fixtures include a lookup table, which we're going to use for this transaction
+ * This is the address of the lookup table account from `scripts/fixtures/address-lookup-table-account.json`
+ */
+const LOOKUP_TABLE_ADDRESS = '2JPQuT3dHtPjrdcbUQyrrT4XYRYaWpWfmAJ54SUapg6n' as Address;
```
steveluscher, line 50: We should create a new one of these instead of reusing an account that's also used by tests. I should be able to delete the test and the fixture, or delete the example and the fixture without breaking either.

Consider cloning this fixture and naming it `example-deserialize-transaction-address-lookup-table.json`?

PR #2985, opened 2024-07-25
An example that demonstrates how to add a custom JSON RPC API method to the base ones

steveluscher:
# Summary

Some JSON RPC servers implement additional methods, above and beyond the ones offered by the Solana JSON RPC API. We want developers to be able to mix these into their RPC instance commensurate with their RPC providers' capabilities.

In this PR we provide an example of how one might add a custom JSON RPC API method based on the `getAsset` method available on the public Solana Mainnet RPC server.

# Test Plan

```
cd examples/rpc-custom-api/
pnpm start

> @solana/example-rpc-custom-api@ start /home/sol/src/solana-web3.js-git/examples/rpc-custom-api
> tsx src/example.ts

INFO (Custom JSON RPC API): [step 1] Solana RPC methods like `getLatestBlockhash` still work
    blockhash: "C8Tqp653ZcyRcnPjnrwjRLfZBpwWPjSTFa5Fr8jA2iSL"
    lastValidBlockHeight: 258841174
INFO (Custom JSON RPC API): [step 2] The custom `getAssetMetadata` that we implemented also works
    metadata: {
      "name": "USD Coin",
      "symbol": "USDC"
    }
```

### DIFF
--- a/examples/rpc-custom-api/src/example.ts
+++ b/examples/rpc-custom-api/src/example.ts
@@ -1,0 +1,116 @@
+/**
+ * EXAMPLE
+ * Add a custom JSON RPC method to the base Solana JSON RPC API using @solana/web3.js.
+ *
+ * To run this example, execute `pnpm start` in this directory.
+ */
+import { createLogger } from '@solana/example-utils/createLogger.js';
+import {
+    Address,
+    address,
+    createDefaultRpcTransport,
+    createRpc,
+    createSolanaRpcApi,
+    DEFAULT_RPC_CONFIG,
+    mainnet,
+    RpcApi,
+    RpcApiMethods,
+    SolanaRpcApiMainnet,
+} from '@solana/web3.js';
+
+const log = createLogger('Custom JSON RPC API');
+
+/**
+ * STEP 1: CUSTOM JSON RPC API CALL SIGNATURE
+ * Define the call signature of the custom API. For this example we will use the Triton One
+ * `getAsset` API, available on the public mainnet RPC server.
+ * https://docs.triton.one/digital-assets-api/metaplex-digital-assets-api/get-asset
+ */
+type AssetMetadata = Readonly<{
+    description: string;
+    name: string;
+    symbol: string;
+}>;
+interface TritonGetAssetApi extends RpcApiMethods {
+    /**
+     * Define the ideal developer-facing API as a TypeScript type. Doing so will enable typechecking
+     * and autocompletion for it on the RPC instance.
+     */
+    getAssetMetadata(address: Address): AssetMetadata;
+}
+
+/**
+ * STEP 2: CUSTOM JSON RPC API IMPLEMENTATION
+ * Create an instance of the default JSON RPC API, then create a wrapper around it to intercept
+ * calls for custom API methods. The wrapper should format the inputs to satisfy the API of the JSON
+ * RPC server, and post-process the server response to satisfy the call signature defined above.
+ */
+const solanaRpcApi = createSolanaRpcApi<SolanaRpcApiMainnet>(DEFAULT_RPC_CONFIG);
+/**
+ * Create a proxy that wraps the Solana RPC API and adds extra functionality.
+ */
+const customizedRpcApi = new Proxy(solanaRpcApi, {
+    defineProperty() {
+        return false;
+    },
+    deleteProperty() {
+        return false;
+    },
+    get(target, p, receiver) {
+        const methodName = p.toString();
+        if (methodName === 'getAssetMetadata') {
+            /**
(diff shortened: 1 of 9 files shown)


steveluscher:
* **#2985** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2985?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/rpc-subscriptions/src/rpc-subscriptions.ts
```diff
@@ -9,7 +9,7 @@ import {
 } from '@solana/rpc-subscriptions-spec';
 import { ClusterUrl } from '@solana/rpc-types';
 
-import { DEFAULT_RPC_CONFIG } from './rpc-default-config';
+import { DEFAULT_RPC_SUBSCRIPTIONS_CONFIG } from './rpc-default-config';
```
steveluscher, line 12: I had to make these default configs public so that people could build atop a regular Solana RPC API. Any thoughts on:

1. Whether these belong here?
2. If they should be public or if we should figure something else out?
3. The fact that they're identical copies of each other for both RPC and RPC Subscriptions?

steveluscher: @lorisleiva, you last moved these around, right? Any thoughts on if this belongs here, or where it should go?

lorisleiva: Hey Steven, I'm not seeing anything on the example that justifies making this public. Could you provide a code snippet that shows why we're having to expose this? Regarding the fact that the configs are the same for RPC Subscriptions, I'd still leave them separate as we're bound to have more configs on them in the future that's likely to be specific to RPC or RPC Subscriptions.

examples/rpc-custom-api/src/example.ts
```diff
@@ -0,0 +1,116 @@
...
+ * SETUP: CUSTOM JSON RPC API IMPLEMENTATION
+ * Create an instance of the default JSON RPC API, then create a wrapper around it to intercept
+ * calls for custom API methods. The wrapper should format the inputs to satisfy the API of the JSON
+ * RPC server, and post-process the server response to satisfy the call signature defined above.
+ */
+const solanaRpcApi = createSolanaRpcApi<SolanaRpcApiMainnet>(DEFAULT_RPC_CONFIG);
+/**
+ * Create a proxy that wraps the Solana RPC API and adds extra functionality.
+ */
+const customizedRpcApi = new Proxy(solanaRpcApi, {
```
lorisleiva, line 52: Don't we have a `createRpcApi` helper that creates a proxy for you?

mcintyre94: We do, but it doesn't let you fiddle with method name the way this example is 

It does let you do per-method `paramsTransformer` and `responseTransformer` though, so I wonder if we could just extend it to also do a `methodNameTransformer`? I think that's the only thing this example does that we can't do with the existing `createRpcApi`.

steveluscher: Yeah, the point here was to show how you can layer _new_ methods onto the default RPC, as you would if you were using a server that supports a superset of the Solana RPC API (Triton, Quicknode, Helius, basically every one).

packages/rpc-subscriptions/src/rpc-subscriptions.ts
steveluscher: I need it to be able to create a default Solana RPC API to build new methods on top of.

```ts
createSolanaRpcApi<SolanaRpcApiMainnet>(DEFAULT_RPC_CONFIG);
```

I made the subscriptions one public for the sake of symmetry.

examples/rpc-custom-api/src/example.ts
steveluscher: The reason this is preferable to creating a new RPC instance is that they target the same server, so you want to share the connection `undici.Dispatcher`.

steveluscher: Just found another place where we claim this to be exported.

https://github.com/solana-labs/solana-web3.js/blob/master/packages/library/README.md?plain=1#L391

lorisleiva: RE the proxy vs the `createRpcApi` helper I think that's fine. IIRC we have other examples in READMEs that use the `createRpcApi` helper to add simpler methods.

packages/rpc-subscriptions/src/rpc-subscriptions.ts
lorisleiva: Makes sense. 👌

lorisleiva (approved):

PR #2991, opened 2024-07-26
Refactor number codecs + allow offloading lamport codec to an inner codec

mcintyre94:
This is a prerequisite for https://github.com/kinobi-so/kinobi/issues/141, as the `solAmountTypeNode` can be defined with any number type:

```json
"type": {
  "kind": "solAmountTypeNode",
  "number": {
    "kind": "numberTypeNode",
    "format": "u64",
    "endian": "be"
  }
}
```

In order to render this using `getLamportsEncoder` etc without restricting the wrapped number type, we need to allow passing an arbitrary number encoder/decoder as a lamports encoder.

This is enabled by generalising number encoders to all encode a `bigint | number`, instead of the smaller ones being restricted to only encoding numbers. They will still decode to only `number`. An existing range check restricts the values to the appropriate range. 

This means that any number encoder can receive a `bigint` value, ie a Lamports value, so any number encoder can be used to encode lamports. When decoding we additionally convert any `Number` to a `BigInt`, and pass to our `lamports` function to get a Lamports value.

### DIFF
--- a/packages/codecs-numbers/src/__typetests__/codecs.ts
+++ b/packages/codecs-numbers/src/__typetests__/codecs.ts
@@ -1,0 +1,169 @@
+import type {
+    FixedSizeNumberCodec,
+    FixedSizeNumberDecoder,
+    FixedSizeNumberEncoder,
+    NumberCodec,
+    NumberDecoder,
+    NumberEncoder,
+} from '../common';
+import { getF32Codec, getF32Decoder, getF32Encoder } from '../f32';
+import { getF64Codec, getF64Decoder, getF64Encoder } from '../f64';
+import { getI8Codec, getI8Decoder, getI8Encoder } from '../i8';
+import { getI16Codec, getI16Decoder, getI16Encoder } from '../i16';
+import { getI32Codec, getI32Decoder, getI32Encoder } from '../i32';
+import { getI64Codec, getI64Decoder, getI64Encoder } from '../i64';
+import { getI128Codec, getI128Decoder, getI128Encoder } from '../i128';
+import { getShortU16Codec, getShortU16Decoder, getShortU16Encoder } from '../short-u16';
+import { getU8Codec, getU8Decoder, getU8Encoder } from '../u8';
+import { getU16Codec, getU16Decoder, getU16Encoder } from '../u16';
+import { getU32Codec, getU32Decoder, getU32Encoder } from '../u32';
+import { getU64Codec, getU64Decoder, getU64Encoder } from '../u64';
+import { getU128Codec, getU128Decoder, getU128Encoder } from '../u128';
+
+getF32Encoder() satisfies NumberEncoder;
+getF32Encoder() satisfies FixedSizeNumberEncoder;
+getF32Decoder() satisfies NumberDecoder;
+getF32Decoder() satisfies FixedSizeNumberDecoder;
+getF32Codec() satisfies NumberCodec;
+getF32Codec() satisfies FixedSizeNumberCodec;
+getF32Codec() satisfies NumberEncoder;
+getF32Codec() satisfies FixedSizeNumberEncoder;
+getF32Codec() satisfies NumberDecoder;
+getF32Codec() satisfies FixedSizeNumberDecoder;
+
+getF64Encoder() satisfies NumberEncoder;
+getF64Encoder() satisfies FixedSizeNumberEncoder;
+getF64Decoder() satisfies NumberDecoder;
+getF64Decoder() satisfies FixedSizeNumberDecoder;
+getF64Codec() satisfies NumberCodec;
+getF64Codec() satisfies FixedSizeNumberCodec;
+getF64Codec() satisfies NumberEncoder;
+getF64Codec() satisfies FixedSizeNumberEncoder;
+getF64Codec() satisfies NumberDecoder;
+getF64Codec() satisfies FixedSizeNumberDecoder;
+
+getI8Encoder() satisfies NumberEncoder;
+getI8Encoder() satisfies FixedSizeNumberEncoder;
+getI8Decoder() satisfies NumberDecoder;
+getI8Decoder() satisfies FixedSizeNumberDecoder;
+getI8Codec() satisfies NumberCodec;
+getI8Codec() satisfies FixedSizeNumberCodec;
+getI8Codec() satisfies NumberEncoder;
+getI8Codec() satisfies FixedSizeNumberEncoder;
+getI8Codec() satisfies NumberDecoder;
+getI8Codec() satisfies FixedSizeNumberDecoder;
+
+getI16Encoder() satisfies NumberEncoder;
+getI16Encoder() satisfies FixedSizeNumberEncoder;
+getI16Decoder() satisfies NumberDecoder;
+getI16Decoder() satisfies FixedSizeNumberDecoder;
+getI16Codec() satisfies NumberCodec;
+getI16Codec() satisfies FixedSizeNumberCodec;
+getI16Codec() satisfies NumberEncoder;
+getI16Codec() satisfies FixedSizeNumberEncoder;
+getI16Codec() satisfies NumberDecoder;
+getI16Codec() satisfies FixedSizeNumberDecoder;
+
+getI32Encoder() satisfies NumberEncoder;
+getI32Encoder() satisfies FixedSizeNumberEncoder;
+getI32Decoder() satisfies NumberDecoder;
+getI32Decoder() satisfies FixedSizeNumberDecoder;
+getI32Codec() satisfies NumberCodec;
+getI32Codec() satisfies FixedSizeNumberCodec;
+getI32Codec() satisfies NumberEncoder;
+getI32Codec() satisfies FixedSizeNumberEncoder;
+getI32Codec() satisfies NumberDecoder;
+getI32Codec() satisfies FixedSizeNumberDecoder;
+
+getI64Encoder() satisfies NumberEncoder;
+getI64Encoder() satisfies FixedSizeNumberEncoder;
+getI64Decoder() satisfies NumberDecoder;
+getI64Decoder() satisfies FixedSizeNumberDecoder;
+getI64Codec() satisfies NumberCodec;
+getI64Codec() satisfies FixedSizeNumberCodec;
+getI64Codec() satisfies NumberEncoder;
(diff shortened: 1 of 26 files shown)


packages/codecs-numbers/src/__typetests__/codecs.ts
```diff
@@ -0,0 +1,169 @@
...
+import { getI64Codec, getI64Decoder, getI64Encoder } from '../i64';
+import { getI128Codec, getI128Decoder, getI128Encoder } from '../i128';
+import { getShortU16Codec, getShortU16Decoder, getShortU16Encoder } from '../short-u16';
+import { getU8Codec, getU8Decoder, getU8Encoder } from '../u8';
+import { getU16Codec, getU16Decoder, getU16Encoder } from '../u16';
+import { getU32Codec, getU32Decoder, getU32Encoder } from '../u32';
+import { getU64Codec, getU64Decoder, getU64Encoder } from '../u64';
+import { getU128Codec, getU128Decoder, getU128Encoder } from '../u128';
+
+getF32Encoder() satisfies NumberEncoder;
```
lorisleiva, line 23: Thanks for adding a typetest file! ❤️

packages/codecs-numbers/src/common.ts
```diff
@@ -1,12 +1,10 @@
 import { Codec, Decoder, Encoder, FixedSizeCodec, FixedSizeDecoder, FixedSizeEncoder } from '@solana/codecs-core';
 
 /** Defines a encoder for numbers and bigints. */
-export type NumberEncoder = Encoder<bigint | number> | Encoder<number>;
+export type NumberEncoder = Encoder<bigint | number>;
```
lorisleiva, line 4: So much cleaner. 🤤

packages/codecs-numbers/src/i16.ts
```diff
@@ -3,12 +3,12 @@ import { combineCodec, FixedSizeCodec, FixedSizeDecoder, FixedSizeEncoder } from
...
 import { numberDecoderFactory, numberEncoderFactory } from './utils';
 
-export const getI16Encoder = (config: NumberCodecConfig = {}): FixedSizeEncoder<number, 2> =>
+export const getI16Encoder = (config: NumberCodecConfig = {}): FixedSizeEncoder<bigint | number, 2> =>
     numberEncoderFactory({
         config,
         name: 'i16',
         range: [-Number('0x7fff') - 1, Number('0x7fff')],
-        set: (view, value, le) => view.setInt16(0, value, le),
+        set: (view, value, le) => view.setInt16(0, typeof value === 'bigint' ? Number(value) : value, le),
```
lorisleiva, line 11: nit: I suppose you could just use `Number(value)` here, right? Since I don't think there is an overhead to casting a number to a number.

packages/rpc-types/src/lamports.ts
```diff
@@ -46,18 +48,34 @@ export function lamports(putativeLamports: bigint): LamportsUnsafeBeyond2Pow53Mi
...
+    if (innerEncoder) {
+        return fixEncoderSize(innerEncoder, 8);
+    }
     return getMemoizedU64Encoder();
 }
 
-export function getLamportsDecoder(): FixedSizeDecoder<LamportsUnsafeBeyond2Pow53Minus1, 8> {
+export function getLamportsDecoder(
+    innerDecoder?: NumberDecoder,
+): FixedSizeDecoder<LamportsUnsafeBeyond2Pow53Minus1, 8> {
```
lorisleiva, line 62: I'm not sure why we are fixing the size of this codec? If I want to describe lamports using a u32, this will still use 8 bytes which I think is wrong. Unless I'm missing something?

packages/rpc-types/src/lamports.ts
```diff
@@ -46,18 +48,34 @@ export function lamports(putativeLamports: bigint): LamportsUnsafeBeyond2Pow53Mi
...
+    if (innerDecoder) {
```
lorisleiva, line 63: It might be best to offer two helper methods. One that accepts an explicit number codec to wrap and one that uses the memoized u64 codec as the number codec (which can delegate to the first helper function). That way the latter is flexible and is as "fixed" or "variable" as the provided number codec. Whereas the second one is a fixed codec of 8 bytes, always (and this can be used in most places within this library).

packages/rpc-types/src/lamports.ts
mcintyre94: Okay cool, I'll loosen the output type here to `NumberDecoder` and we can remove the fix. I thought the `FixedSizeDecoder<LamportsUnsafeBeyond2Pow53Minus1, 8>` was important in this signature and didn't want to break it!

packages/rpc-types/src/lamports.ts
mcintyre94: Is that to preserve the `FixedSizeDecoder<LamportsUnsafeBeyond2Pow53Minus1, 8>` output type when this is called without an inner de/encoder?

packages/rpc-types/src/lamports.ts
lorisleiva: Ideally this should keep the size of the provided number codec but map the number type to a Lamport type.

mcintyre94: Reworked this to use new functions like this:

```
export function getLamportsEncoderFromEncoder<TEncoder extends NumberEncoder>(
    innerEncoder: TEncoder,
): Encoder<LamportsUnsafeBeyond2Pow53Minus1> & ExtractAdditionalProps<TEncoder, NumberEncoder> {
    return innerEncoder;
}
```

Like you said, the version without a wrapped encoder is used all over the place, so best to keep it small and tree-shake this stuff away

And then when there is a wrapped encoder, it no longer gets its size changed, and the `size` remains on the lamport encoder type too.

packages/rpc-types/src/lamports.ts
```diff
@@ -46,18 +49,45 @@ export function lamports(putativeLamports: bigint): LamportsUnsafeBeyond2Pow53Mi
     return putativeLamports;
 }
 
+type ExtractAdditionalProps<T, U> = Omit<T, keyof U>;
+
 export function getLamportsEncoder(): FixedSizeEncoder<LamportsUnsafeBeyond2Pow53Minus1, 8> {
     return getMemoizedU64Encoder();
 }
 
+export function getLamportsEncoderFromEncoder<TEncoder extends NumberEncoder>(
```
lorisleiva, line 58: nit: How do you feel about making this the main `getLamportsEncoder` function, and call this other one something like `getU64LamportsEncoder` instead?

packages/rpc-types/src/lamports.ts
```diff
@@ -46,18 +49,45 @@ export function lamports(putativeLamports: bigint): LamportsUnsafeBeyond2Pow53Mi
...
+    innerEncoder: TEncoder,
+): Encoder<LamportsUnsafeBeyond2Pow53Minus1> & ExtractAdditionalProps<TEncoder, NumberEncoder> {
```
lorisleiva, line 60: Are we sure that this return type keeps the fixed-size or variable-size properties of the inner encoder? Should add a few typetests just to be safe?

packages/rpc-types/src/lamports.ts
```diff
@@ -46,18 +49,45 @@ export function lamports(putativeLamports: bigint): LamportsUnsafeBeyond2Pow53Mi
...
+    return innerEncoder;
+}
+
 export function getLamportsDecoder(): FixedSizeDecoder<LamportsUnsafeBeyond2Pow53Minus1, 8> {
     return transformDecoder(getMemoizedU64Decoder(), lamports);
```
lorisleiva, line 65: Here I would delegate to the other helper function that accepts an inner encoder/decoder. That way if we need to change its behaviour in the future, we have a single point of failure.

packages/rpc-types/src/lamports.ts
mcintyre94: I don't have a strong view either way, but I'd prefer to do it in a separate PR just to reduce churn and files here. I probably already should have split this one a bit!

packages/rpc-types/src/lamports.ts
mcintyre94: Added typetests: https://github.com/solana-labs/solana-web3.js/blob/7ec7cfcec42b3321201bf79e71df85875d0f993f/packages/rpc-types/src/__typetests__/lamports-typetests.ts

packages/rpc-types/src/lamports.ts
mcintyre94: Good idea, updated :)

packages/rpc-types/src/lamports.ts
lorisleiva: Sounds good!

packages/rpc-types/src/__typetests__/lamports-typetests.ts
```diff
@@ -0,0 +1,107 @@
...
+{
+    const encoder = getLamportsEncoder();
+    encoder satisfies FixedSizeEncoder<LamportsUnsafeBeyond2Pow53Minus1>;
+    encoder.fixedSize satisfies number;
+    encoder.encode(1n as LamportsUnsafeBeyond2Pow53Minus1) satisfies ReadonlyUint8Array;
+}
+
+// Fixed size inner encoder
+{
+    const innerEncoder = {} as FixedSizeEncoder<bigint | number>;
```
lorisleiva, line 31: Could we just add an explicit fixed size here to make sure it's also not lost?

```suggestion
    const innerEncoder = {} as FixedSizeEncoder<bigint | number, 42>;
```

packages/rpc-types/src/lamports.ts
```diff
@@ -46,12 +49,28 @@ export function lamports(putativeLamports: bigint): LamportsUnsafeBeyond2Pow53Mi
     return putativeLamports;
 }
 
+type ExtractAdditionalProps<T, U> = Omit<T, keyof U>;
+
 export function getLamportsEncoder(): FixedSizeEncoder<LamportsUnsafeBeyond2Pow53Minus1, 8> {
     return getMemoizedU64Encoder();
```
lorisleiva, line 55: super nit: maybe also wrap that one in the helper function that accepts an inner encoder for consistency?

lorisleiva (approved):
Amazing, thank you!

PR #2993, opened 2024-07-26
Update the README for `@solana/rpc`

steveluscher:
Addresses #2983.

### DIFF
--- a/packages/rpc/README.md
+++ b/packages/rpc/README.md
@@ -14,4 +14,110 @@
 
 # @solana/rpc
 
-TODO
+This package contains utilities for creating objects that you can use to communicate with a Solana JSON RPC server. It can be used standalone, but it is also exported as part of the Solana JavaScript SDK [`@solana/web3.js@experimental`](https://github.com/solana-labs/solana-web3.js/tree/master/packages/library).
+
+Unless you plan to create a custom RPC interface, you can use the [`createSolanaRpc(clusterUrl)`](#createsolanarpcclusterurl-config) function to obtain a default implementation of the [Solana JSON RPC API](https://solana.com/docs/rpc/http).
+
+## Types
+
+### `RpcTransport{Devnet|Testnet|Mainnet}`
+
+These types refine the base `RpcTransport` type. Each describes a transport that is specific in some way to a particular Solana cluster.
+
+For instance, a `RpcTransportDevnet` is understood to communicate with a RPC server related to devnet, and as such might only be accepted for use as the transport of a `RpcDevnet`.
+
+This is useful in cases where you need to make assertions about what capabilities a RPC offers. For example, RPC methods like `requestAirdrop` are not available on mainnet. You can use the ability to assert on the type of RPC transport at compile time to prevent calling unimplemented methods or presuming the existence of unavailable capabilities.
+
+### `RpcTransportFromClusterUrl<TClusterUrl extends ClusterUrl>`
+
+Given a `ClusterUrl`, this utility type will resolve to as specific a `RpcTransport` as possible.
+
+```ts
+function createCustomTransport<TClusterUrl extends ClusterUrl>(
+    clusterUrl: TClusterUrl,
+): RpcTransportFromClusterUrl<TClusterUrl> {
+    /* ... */
+}
+const transport = createCustomTransport(testnet('http://api.testnet.solana.com'));
+transport satisfies RpcTransportTestnet; // OK
+```
+
+### `Rpc{Devnet|Testnet|Mainnet}<TRpcMethods>`
+
+These types refine the base `Rpc` type. Each describes a RPC that is specific in some way to a particular Solana cluster and a corpus of RPC methods.
+
+This is useful in cases where you need to make assertions about the suitablilty of a RPC for a given purpose. For example, you might like to make it a type error to combine certain types with RPC belonging to certain clusters, at compile time.
+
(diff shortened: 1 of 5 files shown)


steveluscher:
* **#2993** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/2993?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/rpc/src/rpc-clusters.ts
```diff
@@ -5,7 +5,6 @@ import type { ClusterUrl, DevnetUrl, MainnetUrl, TestnetUrl } from '@solana/rpc-
 export type RpcTransportDevnet = RpcTransport & { '~cluster': 'devnet' };
 export type RpcTransportTestnet = RpcTransport & { '~cluster': 'testnet' };
 export type RpcTransportMainnet = RpcTransport & { '~cluster': 'mainnet' };
-export type RpcTransportWithCluster = RpcTransportDevnet | RpcTransportMainnet | RpcTransportTestnet;
```
steveluscher, line 8: I couldn't find or think of a use for this. In particular, the `Rpc*` types below have no such thing.

packages/rpc/src/rpc-default-config.ts
```diff
@@ -2,9 +2,9 @@ import type { createSolanaRpcApi } from '@solana/rpc-api';
...
 import { createSolanaJsonRpcIntegerOverflowError } from './rpc-integer-overflow-error';
 
-export const DEFAULT_RPC_CONFIG: Partial<Parameters<typeof createSolanaRpcApi>[0]> = {
+export const DEFAULT_RPC_CONFIG: Partial<NonNullable<Parameters<typeof createSolanaRpcApi>[0]>> = {
     defaultCommitment: 'confirmed',
     onIntegerOverflow(methodName, keyPath, value) {
         throw createSolanaJsonRpcIntegerOverflowError(methodName, keyPath, value);
     },
-};
+} as const;
```
steveluscher, line 10: This prevents us from adding properties here that are _not_ part of the config type.

packages/rpc/src/rpc-transport.ts
```diff
@@ -7,7 +7,7 @@ import { getRpcTransportWithRequestCoalescing } from './rpc-request-coalescer';
 import { getSolanaRpcPayloadDeduplicationKey } from './rpc-request-deduplication';
 
 type RpcTransportConfig = Parameters<typeof createHttpTransport>[0];
-export interface DefaultRpcTransportConfig<TClusterUrl extends ClusterUrl> extends RpcTransportConfig {
+interface DefaultRpcTransportConfig<TClusterUrl extends ClusterUrl> extends RpcTransportConfig {
```
steveluscher, line 10: This is in use nowhere, so I made it private.

packages/rpc/README.md
```diff
@@ -14,4 +14,108 @@
...
+    rpc: RpcMainnet,
+): Promise<SpecialAccountInfo>;
+async function getSpecialAccountInfo(
+    address: Address<'TeST1111111111111111111111111111'>,
+    rpc: RpcDevnet | RpcTestnet,
+): Promise<SpecialAccountInfo>;
+async function getSpecialAccountInfo(address: Address, rpc: Rpc): Promise<SpecialAccountInfo> {
+    /* ... */
+}
+const rpc = getSolanaRpc({ url: devnet('https://api.devnet.solana.com') });
```
lorisleiva, line 61: I think this should be:

```suggestion
const rpc = createSolanaRpc(devnet('https://api.devnet.solana.com'));
```

packages/rpc/README.md
```diff
@@ -14,4 +14,108 @@
...
+await getSpecialAccountInfo(address('TeST1111111111111111111111111111'), rpc); // ERROR
```
lorisleiva, line 62: According to your type signatures above this is valid. Maybe you meant to use the "real" address here which would cause a type failure.

packages/rpc/README.md
```diff
@@ -14,4 +14,108 @@
...
+
+#### Arguments
+
+A config object with the following properties:
+
+-   `dispatcher_NODE_ONLY`: An optional `Undici.Dispatcher` instance that governs how the networking stack should behave. This option is only relevant in Node applications. Consult the documentation for the various subclasses of `Undici.Dispatcher`, such as `Agent`, `Client`, and `Pool`, at https://undici.nodejs.org/#/docs/api/Client.
+-   `headers`: An optional object where the keys are HTTP header names and the values are HTTP header values. This parameter is typed to disallow certain headers from being overwritten.
+-   `url`: A `ClusterUrl` at which the RPC server can be contacted.
+
+### `createSolanaRpc(clusterUrl, config)`
```
lorisleiva, line 115: This is possibly the most important function of this whole package and perhaps we should find a way to bump its visibility on the readme. I know we're following a "Types" then "Functions" documentation pattern but maybe we can make an exemption here? Or just mention it in the intro with a link to it further down the page.

lorisleiva (approved):
Thanks for that! I'll gradually make my way to your other documentation/example PRs.

packages/rpc/README.md
steveluscher: I did! Oops.

lorisleiva (approved):
Thanks for the adjustments!

PR #3003, opened 2024-07-29
Update the README for Release Candidate

steveluscher:


steveluscher:
* **#3003** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3003?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

lorisleiva (approved):
Looks good! I had a quick look at the codec/Kinobi section and I don't think anything else needs updating.

packages/library/README.md
```diff
@@ -205,9 +180,9 @@ const rpc = createSolanaRpc('http://127.0.0.1:8899');
 const slot = await rpc.getSlot().send();
 ```
 
-### Using Custom RPC Transports
+### Custom RPC Transports
 
-The `createSolanaRpc` function communicates with the RPC server using a default HTTP transport that should satisfy most use cases. However, you may provide your own transport or decorate existing ones to communicate with RPC servers in any way you see fit. In the example below, we explicitly create a transport and use it to create a new RPC client via the `createSolanaRpcFromTransport` function.
+The `createSolanaRpc` function communicates with the RPC server using a default HTTP transport that should satisfy most use cases. You can provide your own transport or decorate an existing one to communicate with RPC servers in any way you see fit. In the example below, we explicitly create a transport and use it to create a new RPC client via the `createSolanaRpcFromTransport` function.
```
mcintyre94, line 185: Nitpick: I'm not sure quite what "decorate an existing one" means, given we don't have decorators. I wonder if "wrap an existing one" would be a bit clearer.

packages/library/README.md
```diff
@@ -401,7 +372,7 @@ const rpc = createRpc({ api, transport });
 
 Note that the `createSolanaRpcApi` function is a wrapper on top of the `createRpcApi` function which adds some Solana-specific transformers such as setting a default commitment on all methods or throwing an error when an integer overflow is detected.
 
-#### Creating your Own API Methods
+#### Creating Your Own API Methods
 
 The new library’s RPC specification supports an _infinite_ number of JSON-RPC methods with **zero increase** in bundle size.
```
mcintyre94, line 377: Nitpick: The unchanged bit below this refers to a custom Quicknode method for Ethereum. Their [Solana docs](https://www.quicknode.com/docs/solana) don't have any equivalent custom methods AFAICT. Might be worth switching out for a Helius link, eg https://docs.helius.dev/compression-and-das-api/digital-asset-standard-das-api/get-asset?

mcintyre94 (approved):
LGTM! A few small nitpicky suggestions but this looks really great IMO.

PR #3008, opened 2024-07-30
Update compat readme to only mention `fromVersionedTransaction`

mcintyre94:
This was missed as part of the transaction refactor. Compat now only has a `fromVersionedTransaction` function which gives a `Transaction`, and you can convert that further to a transaction message separately if required. Therefore it makes no distinction between a blockhash and durable nonce transaction, which are concerns of transaction message only.


lorisleiva (approved):
Good catch, thanks!

PR #3049, opened 2024-08-02
Add `getPublicKeyFromPrivateKey` helper

lorisleiva:
This PR adds a `getPublicKeyFromPrivateKey` helper method that can be used to retrieve the `CryptoKey` public key from an extractable `CryptoKey` private key.

```ts
import { createPrivateKeyFromBytes, getPublicKeyFromPrivateKey } from '@solana/keys';

const privateKey = await createPrivateKeyFromBytes(new Uint8Array([...]), true);

const publicKey = await getPublicKeyFromPrivateKey(privateKey);
const extractablePublicKey = await getPublicKeyFromPrivateKey(privateKey, true);
```


lorisleiva:
* **#3049** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3049?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/keys/src/__tests__/public-key-test.ts
```diff
@@ -0,0 +1,37 @@
...
+            const publicKeyBytes = new Uint8Array(await crypto.subtle.exportKey('raw', publicKey));
+            expect(publicKeyBytes).toEqual(EXPECTED_MOCK_PUBLIC_KEY_BYTES);
+        });
+    });
+    describe('given a non-extractable private key', () => {
+        let privateKey: CryptoKey;
+        beforeEach(async () => {
+            privateKey = await createPrivateKeyFromBytes(MOCK_PRIVATE_KEY_BYTES, false);
+        });
+        it('cannot get the associated public key', async () => {
```
mcintyre94, line 32: Is it possible for us to check if it's extractable? I'm wondering if we should add a Solana error for this case because I'm skeptical that subtle crypto will give a nice error message

mcintyre94 (approved):

packages/keys/src/public-key.ts
```diff
@@ -0,0 +1,19 @@
...
+    // Remove private data from JWK.
+    delete jwk.d;
+    delete jwk.dp;
+    delete jwk.dq;
+    delete jwk.q;
+    delete jwk.qi;
+    jwk.key_ops = ['verify'];
+
+    // Import public key.
+    return await crypto.subtle.importKey('jwk', jwk, 'Ed25519', extractable ?? false, ['verify']);
```
steveluscher, line 18: I think it's preferable to build this up according to what the importer needs, rather than to delete things. What if, in the future, the export API starts returning extra crap that we don't have a `delete` statement for?
```suggestion
    // Import public key.
    return await crypto.subtle.importKey(
        'jwk',
        {
            crv /* curve */: 'Ed25519',
            ext /* extractable */: extractable ?? false,
            key_ops /* key operations */: ['verify'],
            kty /* key type */: 'OKP' /* octet key pair */,
            x /* public key x-coordinate */: jwk.x,
        },
        'Ed25519',
        extractable ?? false,
        ['verify'],
    );
```

packages/keys/src/public-key.ts
```diff
@@ -0,0 +1,19 @@
...
+    // Export private key.
+    const jwk = await crypto.subtle.exportKey('jwk', privateKey);
+
+    // Remove private data from JWK.
+    delete jwk.d;
+    delete jwk.dp;
+    delete jwk.dq;
+    delete jwk.q;
+    delete jwk.qi;
+    jwk.key_ops = ['verify'];
```
steveluscher, line 15: Missed setting the `ext` parameter of the jwk according to the value of `extractable`. Might want to write a test to cover mismatches of that nature. My example below handles that in the implementation.

steveluscher (changes requested):
Wow! I didn't think this was possible! Amazing.

This can't go in without this PR _also_ including an update to the polyfill. The only reason the tests pass here is that the tests run in a haz-ed25519 environment. This will puke in a browser that doesn't.

To be clear, you need to add a polyfill that implements at least `jwk` export for private keys and import for public keys. Hint: the `x` and `d` coordinates are base64URL encoded.

lorisleiva:
Thanks @steveluscher, I'll apply your suggestions and prepend PRs to the stack for the polyfill.

packages/keys/src/public-key.ts
lorisleiva: I've added some tests.

packages/keys/src/__tests__/public-key-test.ts
lorisleiva: Good shout! I've added a new `SOLANA_ERROR__SUBTLE_CRYPTO__CANNOT_EXPORT_NON_EXTRACTABLE_KEY` error code.

steveluscher (approved):
Added via Giphy

lorisleiva:
### Merge activity

* **Aug 7, 4:34 AM EDT**: @lorisleiva started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3049).

PR #3080, opened 2024-08-07
Build with a concurrency of 1 on CI

steveluscher:
# Summary

GitHub Actions runners are anemic, and a new CPU-intensive test has made it impossible for CI to pass: https://github.com/solana-labs/solana-web3.js/actions/runs/10288741934/job/28474891140

I suspect this is because concurrent build tasks result in the CPU being starved for cycles. Let's try to put it back to 1 for CI only.


steveluscher:
### Merge activity

* **Aug 7, 12:09 PM PDT**: @steveluscher started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3080).

PR #3098, opened 2024-08-13
refactor(experimental): GraphQL: Refactor program accounts filters

buffalojoec:
#### Problem
The filters input parameter for the GraphQL resolver is malformed. This PR brings it to parity with the Solana JSON RPC API as well as the `@solana/rpc-api` package's definition for program accounts filters.

#### Summary of Changes
First add some test coverage to `@solana/rpc-api` for program accounts filters to ensure everything is working properly.

Then, refactor the GraphQL API for the `programAccounts` query to correct the program accounts filters input. Finally, add some tests for GraphQL.

Since GraphQL doesn't allow unions as input types, I had to split the two types of filters into two separate filters for the GraphQL API, then coalesce them into the same `filters` argument for the RPC.

### DIFF
--- a/packages/rpc-graphql/src/loaders/program-accounts.ts
+++ b/packages/rpc-graphql/src/loaders/program-accounts.ts
@@ -19,14 +19,7 @@ async function loadProgramAccounts(
     rpc: Rpc<GetProgramAccountsApi>,
     { programAddress, ...config }: ProgramAccountsLoaderArgs,
 ): Promise<ProgramAccountsLoaderValue> {
-    // @ts-expect-error FIX ME: https://github.com/microsoft/TypeScript/issues/43187
-    return await rpc
-        .getProgramAccounts(
-            programAddress,
-            // @ts-expect-error FIX ME: https://github.com/microsoft/TypeScript/issues/43187
-            config,
-        )
-        .send();
+    return await rpc.getProgramAccounts(programAddress, config).send();
 }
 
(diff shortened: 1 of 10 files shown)


packages/rpc-graphql/src/loaders/program-accounts.ts
```diff
@@ -19,14 +19,7 @@ async function loadProgramAccounts(
...
 ): Promise<ProgramAccountsLoaderValue> {
-    // @ts-expect-error FIX ME: https://github.com/microsoft/TypeScript/issues/43187
-    return await rpc
-        .getProgramAccounts(
-            programAddress,
-            // @ts-expect-error FIX ME: https://github.com/microsoft/TypeScript/issues/43187
-            config,
-        )
-        .send();
+    return await rpc.getProgramAccounts(programAddress, config).send();
```
mcintyre94, line 22: Awesome! Do you know why this is fixed? The TS issue it was referencing is still open and it doesn't look like we've changed anything here

mcintyre94 (approved):
Nice, LGTM!

packages/rpc-graphql/src/loaders/program-accounts.ts
buffalojoec: I must have thought it was a different problem, but since I wasn't defining the right type, of course it was throwing a type error (ha).

lorisleiva (approved):

PR #3119, opened 2024-08-20
fix: Compare public key values not references

### LINKED ISSUE #3118: createAccountWithSeed duplicates keys unnecessarily
In `createAccountWithSeed` there is a check whether the `fromPubkey` and the `basePubkey` are the same key to determine whether we need to add the `basePubKey` to the keys array.
In this method, the key objects are being compared but since they are objects the thing that is compared is their reference, meaning it will only resolve to `false` when they are exactly the same object.
This causes issues if the instruction is being decoded and later recompiled to a new transaction, as the public keys are the same but the objects are not, resolving this check incorrectly.

Solution suggested: compare the actual value of the keys, and not the objects

https://github.com/solana-labs/solana-web3.js/blob/4124b49fa74214dd46e0e60b6d2ecff3349283ba/src/programs/system.ts#L863

buffalojoec:
#### Problem

As mentioned in #3118, using `==` or `!=` to compare two `PublicKey` objects will compare their references. This means the conditional will only evaluate to be a match if they are the same object. Additionally, if two build targets or versions of `@solana/web3.js` are being used, these will always be different objects, even if the keys are the same value.

We've encountered the latter issue before, and this appears to be the last place in the legacy library where we still do this conditional without using the `PublicKey.equals` method.

#### Summary of Changes

Run a search through the codebase for any public key comparisons using `==` or `!=` and update them to use `PublicKey.equals`. There's only one.

Closes #3118

### DIFF
--- a/src/programs/system.ts
+++ b/src/programs/system.ts
@@ -860,7 +860,7 @@ export class SystemProgram {
       {pubkey: params.fromPubkey, isSigner: true, isWritable: true},
       {pubkey: params.newAccountPubkey, isSigner: false, isWritable: true},
     ];
-    if (params.basePubkey != params.fromPubkey) {
+    if (!params.basePubkey.equals(params.fromPubkey)) {
       keys.push({
         pubkey: params.basePubkey,
         isSigner: true,


steveluscher (approved):
image

PR #3123, opened 2024-08-20
chore: fix publish CI

buffalojoec:
As described in https://github.com/semantic-release/release-notes-generator/issues/657, add an override for `conventional-changelog-conventionalcommits` to fix the date parsing issue in semantic release.

The PR to update the dependency within `@commitlint/config-conventional` is still open.


buffalojoec:
YOLO

PR #3133, opened 2024-08-21
Implements a JSON.parse that transforms unsafe numbers to bigints

lorisleiva:
This PR implements and uses a custom `JSON.parse` implementation that supports large unsafe JavaScript integers by wrapping them in `bigints`.

This means we can start accepting RPC values as `bigint` right now without loss of precision.

It does so by first transforming the JSON string such that any integer outside of quoted strings will be wrapped in a `BitIngValueObject` if and only if `Number.isSafeInteger` returns `false` for that number.

It then uses a `JSON.parse` reviver to identify that special object wrapper and casts that to a `bigint` without loss of information.

There are still a lot more tests I'd like to add to this PR but wanted to gather early feedback before continuing my work.


lorisleiva:
* **#3133** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3133?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/rpc-transport-http/src/json-parse-with-bigint.ts
```diff
@@ -0,0 +1,76 @@
...
+        // If the first group is matched, it means we are at the
+        // beginning of the JSON string and we have an unquoted segment.
+        if (firstGroup) return transform(firstGroup);
+
+        // Otherwise, we have a double quoted string followed by an unquoted segment.
+        return `${secondGroup}${transform(thirdGroup)}`;
+    });
+}
+
+type WrappedBigInt = { __brand: 'bigint'; value: string };
```
steveluscher, line 57: The properties and discriminator of this value object should probably be ‘compressed’ from the start so that they take up as little memory as possible once they're on the client.

```suggestion
type BigIntValueObject = {
    // $ implies 'this is a value object'; `n` implies 'interpret the value as a bigint'
    $n: string,
};
```

Thoughts?

steveluscher:
```ts
JSON.parse(
    "{\"hi\":9223372036854775807,\"bye\":\"content\"}",
    (_, value, { source }) =>
        source?.match(/^\d+$/)
            ? BigInt(source)
            : value,
);
```

Sorry. :)

packages/rpc-transport-http/src/json-parse-with-bigint.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    return json.replaceAll(unquotedSegmentsRegex, (_, firstGroup, secondGroup, thirdGroup) => {
+        // If the first group is matched, it means we are at the
+        // beginning of the JSON string and we have an unquoted segment.
+        if (firstGroup) return transform(firstGroup);
+
+        // Otherwise, we have a double quoted string followed by an unquoted segment.
+        return `${secondGroup}${transform(thirdGroup)}`;
+    });
+}
```
steveluscher, line 55: I would be really tempted to implement this as a parser instead of a regex. Walking every character:

1. Toggle `inQuote` on and off when you encounter a non-escaped `"`.
2. If `inQuote` is `false` and you encounter a contiguous run of numeric digits without a `.` in the middle, then wrap 'em.

```ts
const str = '{"hi":9223372036854775807,"there":123.4,"bye":"content is \\"1337\\""}';
let out = [];
let inQuote = false;
for (let ii = 0; ii < str.length; ii++) {
    let isEscaped = false;
    if (str[ii] === '\\') {
        out.push(str[ii++]);
        isEscaped = !isEscaped;

    }
    if (ii >= str.length) {
        break;
    }
    if (str[ii] === '"') {
        out.push(str[ii++]);
        if (!isEscaped) {
            inQuote = !inQuote;
        }
    }
    if (ii >= str.length) {
        break;
    }
    if (!inQuote) {
        let consumedNumber = '';
        while (str[ii].match(/[\d\.]/)) {
            consumedNumber += str[ii++];
        }
        if (consumedNumber.length) {
            if (consumedNumber.includes('.')) {
                out.push(consumedNumber);
            } else {
                out.push(`{"$n":"${consumedNumber}"}`)
            }
        }
    }
    out.push(str[ii]);
}
console.log(out.join(''));
```

packages/rpc-transport-http/src/json-parse-with-bigint.ts
lorisleiva: Yeah I'm happy with that. I knew the type was gonna need optimising which is why I extracted helpers for it. 😊

packages/rpc-transport-http/src/json-parse-with-bigint.ts
```diff
@@ -0,0 +1,76 @@
...
+     *
+     * For instance, in the string `{"age":42,"name":"Alice \"The\" 2nd"}`, it would the
+     * following parts: `{`, `:42,`, `:`, `}`. Notice the whole "Alice \"The\" 2nd" string
+     * is not matched as it is wrapped in double quotes and contains escaped double quotes.
+     *
+     * The regex is composed of two parts:
+     *
+     *   1. The first part `^([^"]+)` matches any character until we reach the first double quote.
+     *   2. The second part `("(?:\\"|[^"])+")([^"]+)` matches any double quoted string that may
+     *      and any unquoted segment that follows it. To match a double quoted string, we use the
```
buffalojoec, line 41: nit: are you missing a word or line here? after "string that may".

lorisleiva:
Damn that's a shame about the `source` attribute. It would have been perfect.

packages/rpc-transport-http/src/json-parse-with-bigint.ts
lorisleiva: I don't disagree. Regexes always feel a bit hacky but at the same time I was struggling to come up with examples that would not work with this implementation so I though it might be the least intrusive way to implement this. That being said, we still need a shit loads more test to make sure we're not messing up with the integrity of the response data.

I'm up for giving the parser a go though. I don't know if you've seen [this library](https://github.com/sidorares/json-bigint/blob/master/lib/parse.js) but it could be useful to get some inspiration from. Tbh, I started looking at this parsing logic first and though: hmm maybe a parser is more complicated than it looks like and moved to regexes instead haha. But I think this library does a bit too much for our need anyway.

Anyway, to summary my chaotic thoughts, we can either:
1. Continue to explore regexes.
2. Move towards a parsing solution that only transforms the string (what you're suggesting).
3. Move towards a full parsing solution that outputs parsed data directly by forking/editing [this library](https://github.com/sidorares/json-bigint/blob/master/lib/parse.js).

In any case, we need lots more tests.

I lean towards solution 2, it seems like a clean and robust approach.

buffalojoec (commented):
This is awesome @lorisleiva  - nice work! I'm stoked to get rid of all the `UnsafeBeyond2Pow53Minus1` now. 🚀

I think we should strongly consider benchmarking this parser, though, before cutting a release with it in the response transformer. Something like taking some _huge_ JSON payloads that could come back from the RPC (think `getProgramAccounts`) and comparing native `JSON.parse` (or `.json()`) to this modded version with the reviver. Maybe @steveluscher knows some slick ways to do these benches?

packages/rpc-transport-http/src/json-parse-with-bigint.ts
lorisleiva: Oh yeah I must have deleted a comment line by mistake like an idiot. I've forgotten what I initially wrote now but I've updated the docs so it's a least a proper sentence. 😄

packages/rpc-transport-http/src/json-parse-with-bigint.ts
steveluscher: String-to-string parsing will be way more resilient, I think. It's already resilient to annoying shit like newlines, spaces, and infinite `\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\` escape sequences, without added complexity.

lorisleiva: I've implemented solution 2 in two different ways:
- One that uses a simple regex to consume numbers.
- One that uses plain loops to consume numbers.

The `while (str[ii]?.match(/[\d\.Ee]/))` wasn't enough to correctly match numbers — e.g. it was matching `true` because of the `e` and returning `{"result":tru{"$n":"e"}}`.

I've also left the old regex-only implementation to compare in the benchmarks but I'm happy to remove it because the parsing is much more robust.

PR #3137, opened 2024-08-22
Add edge-light export using Node output for Vercel edge runtime

mcintyre94:
- Vercel edge runtime uses the `browser` export by default. This doesn't work well for us, because our crypto libs depend on a secure context to function in a browser environment, which the edge runtime is not. 
- This PR adds an `edge-light` export, which is the export version that can be used to target the Vercel edge runtime. We just duplicate our Node export here, as this matches the behaviour we want in an edge API route.
- In Next, the `edge-light` export must come before the `browser` one to be used, so I've made it the first export. See https://github.com/vercel/next.js/issues/69175

### DIFF
--- a/packages/ws-impl/package.json
+++ b/packages/ws-impl/package.json
@@ -8,6 +8,11 @@
             "require": "./dist/index.node.cjs",
             "types": "./dist/types/index.browser.d.ts"
         },
+        "workerd": {
+            "import": "./dist/index.node.mjs",
+            "require": "./dist/index.node.cjs",
+            "types": "./dist/types/index.browser.d.ts"
+        },
         "browser": {
             "import": "./dist/index.browser.mjs",
             "require": "./dist/index.browser.cjs",
--- a/packages/ws-impl/package.json
+++ b/packages/ws-impl/package.json
@@ -3,6 +3,11 @@
     "version": "0.0.0",
     "private": true,
     "exports": {
+        "edge-light": {
+            "import": "./dist/index.node.mjs",
+            "require": "./dist/index.node.cjs",
+            "types": "./dist/types/index.browser.d.ts"
+        },
         "browser": {
             "import": "./dist/index.browser.mjs",
             "require": "./dist/index.browser.cjs",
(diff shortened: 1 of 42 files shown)


mcintyre94:
React's [server export](https://github.com/vercel/next.js/blob/3ab6d1790998f0984b97c4abf8c96ffe1eaad865/packages/next/src/compiled/react-dom/package.json#L13-L23) is:

```
      "react-server": "./server.react-server.js",
      "workerd": "./server.edge.js",
      "bun": "./server.bun.js",
      "deno": "./server.browser.js",
      "worker": "./server.browser.js",
      "node": "./server.node.js",
      "edge-light": "./server.edge.js",
      "browser": "./server.browser.js",
      "default": "./server.node.js"
```

Should do a quick check to see which of these are worth adding. Probably just `workerd` for Cloudflare workers which should also point at Node.

packages/ws-impl/package.json
```diff
@@ -3,6 +3,11 @@
     "version": "0.0.0",
     "private": true,
     "exports": {
+        "edge-light": {
+            "import": "./dist/index.node.mjs",
+            "require": "./dist/index.node.cjs",
+            "types": "./dist/types/index.browser.d.ts"
```
buffalojoec, line 9: Is this the only package we need to export the types for?

buffalojoec (approved):
Lgtm

steveluscher (approved):
As we mentioned in person, might as well throw in `workerd` for Cloudflare and any others if it turns out they need them too.

packages/ws-impl/package.json
mcintyre94: I think there was one other too - I copied the `node` export, which included `types` only in these cases. I'm not sure why that is but figured I should just do the exact same for `edge-light`.

mcintyre94:
Added `workerd` because Workers seems to behave the same as Vercel Edge - falls back to browser + has to come before browser. Bun and Deno seem to fall back on Node by default which makes sense so I've left them for now.

PR #3168, opened 2024-08-27
Add `toPayload` and `toText` optional functions to `RpcRequest`

lorisleiva:
This PR adds the following optional functions to the `RpcRequest` type:

```diff
  export type RpcRequest<TParams = unknown> = {
      readonly methodName: string;
      readonly params: TParams;
+     readonly toPayload?: (methodName: string, params: unknown) => unknown;
+     readonly toText?: (payload: unknown) => string;
  };
```

This enables `RpcRequestTransformers` to inject and even augment functions that define:
- How the payload should be constructed from a method name and a set of parameters.
- How the raw text body should be constructed from a payload.

If provided, the `toPayload` function is used by the `Rpc` object to create a payload before sending it to the `RpcTransformer`.

If provided, the `toText` function is passed as-is to the `RpcTransformer` which will use it to convert the payload to a string instead of `JSON.stringify`.

This is important because, `RpcRequestTransformers` can now affect how the JSON string will be send to the RPC server which makes it possible for an RPC API to provide unsafe integers in the JSON string as long as their RPC allows it (which is the case for Solana RPCs and `u64`/`i64` integers).

Currently, it is only possible to achieve that when we get the response back from the RPC Transformer. This PR allows the other side — the request — to have the same ability.

---

Here are some diagrams illustrating the RPC changes made by this PR.

### Current state

Current Rpc Refactoring@2x

### Proposed state

Proposed Rpc Refactoring@2x

### DIFF
--- a/.changeset/twelve-toys-guess.md
+++ b/.changeset/twelve-toys-guess.md
@@ -1,0 +1,6 @@
+---
+'@solana/rpc-transport-http': patch
+'@solana/rpc-spec': patch
+---
+
+Add `toPayload` and `toText` optional functions to `RpcRequest`
--- a/packages/rpc-spec/README.md
+++ b/packages/rpc-spec/README.md
@@ -45,8 +45,10 @@ An object that exposes all of the functions described by `TRpcMethods`, and fulf
 
 An object that describes the elements of a JSON RPC request. It consists of the following properties:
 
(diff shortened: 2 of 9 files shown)


lorisleiva:
We've decided to take another approach on this.

PR #3213, opened 2024-09-06
Clean up SolanaRpcApi: no longer extend RpcApiMethods

mcintyre94:
Extending `RpcApiMethods` removes the ability to access typed keys on `SolanaRpcApi` etc, because it's typed as `[method: string]: RpcApiMethod`. 

In this PR we change from:
```ts
interface GetAssetApi extends RpcApiMethods {
```

To:
```ts
type GetAssetApi = {
```

(Thanks @lorisleiva!)

This allows `createRpcApi` to maintain its typing, meaning that it still constrains any API methods we attempt to build an RPC for. But we now have typesafe `keyof SolanaRpcApi` etc. 

I've also removed the export of `RpcApiMethods`, since this should now be used only as a constraint internally and not a type externally.

I've added a typetest with a bit more detail, but this basically makes this work:
```ts
'getAccountInfo' satisfies keyof SolanaRpcApi;
// @ts-expect-error RPC API does not have this method
'someMadeUpMethod' satisfies keyof SolanaRpcApi;
```

Previously `someMadeUpMethod` would satisfy it, because `keyof SolanaRpcApi` was just string.

### DIFF
--- a/.changeset/green-eels-applaud.md
+++ b/.changeset/green-eels-applaud.md
@@ -1,0 +1,8 @@
+---
+'@solana/rpc-graphql': patch
+'@solana/accounts': patch
+'@solana/rpc-spec': patch
+'@solana/rpc-api': patch
+---
+
+Clean up SolanaRpcApi: no longer extend RpcApiMethods + remove export
--- a/README.md
+++ b/README.md
@@ -389,8 +389,8 @@ type GetAssetApiResponse = Readonly<{
     /* ...etc... */
 }>;
 
-// Set up an interface for the request method.
-interface GetAssetApi extends RpcApiMethods {
+// Set up a type spec for the request method.
+type GetAssetApi = {
     // Define the method's name, parameters and response type
     getAsset(args: { id: Address }): GetAssetApiResponse;
 }
--- a/examples/rpc-custom-api/src/example.ts
+++ b/examples/rpc-custom-api/src/example.ts
@@ -14,7 +14,6 @@ import {
     DEFAULT_RPC_CONFIG,
     mainnet,
     RpcApi,
-    RpcApiMethods,
     SolanaRpcApiMainnet,
 } from '@solana/web3.js';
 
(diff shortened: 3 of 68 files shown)


lorisleiva (approved):
Much better 👌

mcintyre94:
* **#3212** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3212?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3213** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3213?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

buffalojoec (approved):
Looks great! I dug up the original reason why I added this.

[The change](https://github.com/solana-labs/solana-web3.js/pull/1843) to the API methods went in here, and it was really only to tee up [this change](https://github.com/solana-labs/solana-web3.js/pull/1781) which added the `createJsonRpcApi` generic helper.

It seems like somewhere along the way the helper was updated and it will no longer throw a type error, cool! Seems like it might be the change from this:

```ts
const params = config?.parametersTransformer
    ? config?.parametersTransformer(rawParams, methodName)
    : rawParams;
const responseTransformer = config?.responseTransformer
    ? config?.responseTransformer<ReturnType<TRpcMethods[TMethodName]>>
    : (rawResponse: unknown) => rawResponse as ReturnType<TRpcMethods[TMethodName]>;
return {
    methodName,
    params,
    responseTransformer,
};
```

... to this:

```ts
const rawRequest = { methodName, params: rawParams };
const request = config?.requestTransformer
    ? config?.requestTransformer(Object.freeze(rawRequest))
    : rawRequest;
return Object.freeze({
    ...request,
    ...(config?.responseTransformer
        ? {
                responseTransformer: config.responseTransformer as RpcResponseTransformer<
                    ReturnType<TRpcMethods[TMethodName]>
                >,
            }
        : {}),
});
```

Note: This is mirrored in `rpc-subscriptions-api`, so that API should also get the same treatment.

mcintyre94:
> Note: This is mirrored in rpc-subscriptions-api, so that API should also get the same treatment.

Thanks! Added a new PR to the stack: #3218

mcintyre94:
Yep, still seems ~instant for me!

mcintyre94:
### Merge activity

* **Sep 9, 1:11 PM EDT**: @mcintyre94 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3213).

PR #3227, opened 2024-09-10
Remove request arg in API plan's response transformer

lorisleiva:
This PR removes the unnecessary `request` argument from the API plan's `responseTransformer` function. This is because the plan already knows the `request` so it can inject it to the `responseTransformer` provided in its config.

PR #3229, opened 2024-09-10
Make API plans return payloads

lorisleiva:
This PR alters the boundaries of the RPC API layer by making the `RpcApiRequestPlan` return a `payload` of type `unknown` instead of `methodName` and `params`. This means, it is now the responsibility of the RPC API to crunch the request into a payload. The outer RPC layer now simply need to forward that `payload` to the RPC transport.

### DIFF
--- a/packages/rpc-spec/src/rpc.ts
+++ b/packages/rpc-spec/src/rpc.ts
@@ -60,7 +60,7 @@ function makeProxy<TRpcMethods, TRpcTransport extends RpcTransport>(
                 const createRpcRequest = Reflect.get(target, methodName, receiver);
                 const newRequest = createRpcRequest
                     ? createRpcRequest(...rawParams)
-                    : { methodName, params: rawParams };
+                    : { payload: createRpcMessage(methodName, rawParams) };
                 return createPendingRpcRequest(rpcConfig, newRequest);
             };
         },
@@ -73,11 +73,8 @@ function createPendingRpcRequest<TRpcMethods, TRpcTransport extends RpcTransport
 ): PendingRpcRequest<TResponse> {
     return {
         async send(options?: RpcSendOptions): Promise<TResponse> {
-            const { methodName, params, responseTransformer } = apiPlan;
-            const rawResponse = await rpcConfig.transport<TResponse>({
-                payload: createRpcMessage(methodName, params),
-                signal: options?.abortSignal,
-            });
+            const { payload, responseTransformer } = apiPlan;
+            const rawResponse = await rpcConfig.transport<TResponse>({ payload, signal: options?.abortSignal });
             return responseTransformer ? responseTransformer(rawResponse) : rawResponse;
         },
     };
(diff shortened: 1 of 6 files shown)


lorisleiva:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3229?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3229** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3229?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#3227** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3227?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

packages/rpc-spec/src/rpc.ts
```diff
@@ -60,7 +60,7 @@ function makeProxy<TRpcMethods, TRpcTransport extends RpcTransport>(
                 const createRpcRequest = Reflect.get(target, methodName, receiver);
                 const newRequest = createRpcRequest
                     ? createRpcRequest(...rawParams)
-                    : { methodName, params: rawParams };
+                    : { payload: createRpcMessage(methodName, rawParams) };
```
lorisleiva, line 63: There's a little caveat here.

Since it is possible that `Reflect.get` will return undefined when trying to access the request plan, the outer RPC layer needs to defined a "fallback plan" which is why, in some situations, it may end up using `createRpcMessage`.

An alternative would be to throw an error if no plan can be constructed from the given API.

mcintyre94: I think I'd prefer to have a solana error thrown if `createRpcRequest` is undefined after the `Reflect.get` 

I'm guessing this would happen if you eg called `rpc.methodThatDoesntExist()`?

mcintyre94 (approved):

packages/rpc-spec/src/rpc.ts
lorisleiva: This would only happen if you provided an RPC API that explicitly defined all the supported methods but none of the others. Our RPC API does not do this, it uses a proxy object to catch any method call. This means, with our default RPC API, this "fallback" scenario is never used. Which makes me think it's probably a good thing to throw if a plan could not be created.

steveluscher: I actually did exactly this in the subscriptions API that I'm refactoring, because it makes zero sense to have a ‘fallback’ strategy for making a subscription. Like, what would that even be – subscription protocols are all so specific.

lorisleiva: Thanks guys! I've made the change in [a follow-up PR](https://github.com/solana-labs/solana-web3.js/pull/3245).

steveluscher (approved):
Added via Giphy

lorisleiva:
### Merge activity

* **Sep 24, 11:36 AM EDT**: @lorisleiva started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3229).

PR #3251, opened 2024-09-13
Update `isAddress` to not throw

ryoid:
When supplied with a non-base58 encoded string the function would unexpectedly throw. 

This PR updates it to return `false` instead and adds similar tests as `assertIsAddress`

### DIFF
--- a/packages/addresses/src/address.ts
+++ b/packages/addresses/src/address.ts
@@ -46,14 +46,10 @@ export function isAddress(putativeAddress: string): putativeAddress is Address<t
     // Slow-path; actually attempt to decode the input string.
     const base58Encoder = getMemoizedBase58Encoder();
     try {
-        const bytes = base58Encoder.encode(putativeAddress);
-        const numBytes = bytes.byteLength;
-        if (numBytes !== 32) {
-            return false;
-        }
-        return true;
-    } catch (e) {
-        return false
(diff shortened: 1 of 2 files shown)


packages/addresses/src/address.ts
```diff
@@ -45,12 +45,16 @@ export function isAddress(putativeAddress: string): putativeAddress is Address<t
...
-        return false;
+    try {
+        const bytes = base58Encoder.encode(putativeAddress);
+        const numBytes = bytes.byteLength;
+        if (numBytes !== 32) {
+            return false;
+        }
+        return true;
+    } catch (e) {
+        return false
```
buffalojoec, line 56: Heh, we already had this pretty verbose, but what do you think about simplifying?

```suggestion
    try {
        return base58Encoder.encode(putativeAddress).byteLength === 32;
    } catch {
        return false;
    }
```

ryoid: Sure looks fine helps to simplify.

buffalojoec (approved):
Nice work, and thanks for the useful tests!

buffalojoec:
O no, just a prettier issue. Then lgtm.

ryoid:
> O no, just a prettier issue. Then lgtm.

Should be resolved now

PR #3288, opened 2024-09-26
Change return type of getFeeForMessage to Lamports

mcintyre94:
Looking at server code this is definitely measured in Lamports, seems more helpful to use that type here


steveluscher (approved):
Added via Giphy

buffalojoec (approved):

PR #3290, opened 2024-09-27
Throw if the transaction fails when simulated to estimate CUs

mcintyre94:
When we simulate a transaction and it fails in simulation, we get a `unitsConsumed` field but it will be based on how many CUs the transaction took to error. Supposing the transaction is successful later, the CU limit returned by this simulation is probably not helpful.

Currently if you use the function from `getComputeUnitEstimateForTransactionMessageFactory`, and the transaction fails in simulation, you'll just be returned the value of `unitsConsumed` with no indication that the transaction failed. A developer has no way to know if the CU limit returned is based on a successful simulation or not.

This PR modifies the return type from just `number` to `{computeUnitEstimate: number; transactionError: TransactionError | null}`

The `err` field of the simulate transaction result is passed to `transactionError`. This enables a developer to identify that the CU estimate returned is based on a failed transaction, and decide how to deal with that in their code.


steveluscher (changes requested):
Oooh, this is a good catch, but we should probably actually just _throw_ a new error, include the `TransactionError` as the `cause`, and include the `unitsConsumed` in the `SolanaError` `context` for good measure, right?

This way, if people aren't aware of the presence/implication of `transactionError` being non-null, they will get the message loud and clear.

packages/errors/src/codes.ts
```diff
@@ -205,6 +205,7 @@ export const SOLANA_ERROR__TRANSACTION__ADDRESSES_CANNOT_SIGN_TRANSACTION = 5663
 export const SOLANA_ERROR__TRANSACTION__CANNOT_ENCODE_WITH_EMPTY_SIGNATURES = 5663016 as const;
 export const SOLANA_ERROR__TRANSACTION__MESSAGE_SIGNATURES_MISMATCH = 5663017 as const;
 export const SOLANA_ERROR__TRANSACTION__FAILED_TO_ESTIMATE_COMPUTE_LIMIT = 5663018 as const;
+export const SOLANA_ERROR__TRANSACTION__FAILED_WHEN_SIMULATED_TO_ESTIMATE_COMPUTE_LIMIT = 5663019 as const;
```
steveluscher, line 208: Failed when simulating to…? ¯\\_(°ペ)\_/¯

steveluscher (approved):
Added via Giphy

PR #3172, opened 2024-08-28
A utility that makes an `AsyncIterator` interruptible

steveluscher:
# Summary

Some async iterators do work when you poll, and that work is expected to finish in a finite amount of time.

```ts
const stakeAccountKey = 0;
const stakeAccountAddresses = {
  async next() {
    const nextStakeAccountAddress = await deriveStakeAccountAddress(
      owner,
      `stake:${stakeAccountKey++}`
    );
    return { done: false, value: nextStakeAccountAddress };
  },
};
```

Async iterators that wrap push APIs like `EventEmitter` however can get into a situation where a consumer has polled (ie. called `next()`) before the emitter has produced the next value.

```ts
// Ignore all the obvious bugs.
let resolveNextNotification!;
const notifications = {
  async next() {
    const nextNotification = await new Promise((resolve) => {
      resolveNextNotification = resolve;
    });
    return { done: false, value: nextNotification };
  },
  return() {
    emitter.off("notificaton");
  },
};
emitter.on("notification", resolveNextNotification);
```

This implies that – unless the developer is extremely careful – an async iterator that wraps a _push_ API like this can easily find itself deadlocked indefinitely if the pusher never emits another event.

This PR introduces something called an interruptible iterator that races every call to an inner iterator's methods with a ‘termination promise’ that gets triggered immediately on calls to the outer iterator's `return` or `throw` methods.

See the README and the tests for more.

PR #3301, opened 2024-10-02
A package that makes it easier to work with subscribable objects

steveluscher:
# Summary

At the root of the subscriptions infra is `WebSocket` which conforms to the `EventEmitter` interface. In this new package are utilties for making them easier to work with.

1. `getDataPublisherFromEventEmitter` lets you convert an emitter into something with an `on()` method that vends unsubscribe functions. Unsubscribe functions are much easier to work with because you can pass them around as values to unsubscribe from an event _without_ having to hold on to a reference to the original listener function.
2. `createAsyncIterableFromDataPublisher` lets you create an `AsyncIterable` from a `DataPublisher`. This iterable handles cancellation, queueing when there is backpressure, and error propagation.

Read the `README` and the tests for more detail.


steveluscher:
* **#3301** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3301?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/subscribable/src/__typetests__/data-publisher-typetest.ts
```diff
@@ -0,0 +1,60 @@
...
+// [DESCRIBE] getDataPublisherFromEventEmitter
+{
+    // It materializes listener signatures based on the events of the emitter
+    {
+        const eventEmitter = null as unknown as TypedEventEmitter<{ foo: CustomEvent<'bar'> }>;
+        const publisher = getDataPublisherFromEventEmitter(eventEmitter);
+        publisher.on('foo', data => {
+            data satisfies 'bar';
+        });
+    }
```
lorisleiva, line 22: Could we add a typetest for non-custom events? I'm guessing, based on the type definition below, that `null` will be provided as the first argument.

Something like this:

```ts
const eventEmitter = null as unknown as TypedEventEmitter<{ foo: SomeEvent }>;
const publisher = getDataPublisherFromEventEmitter(eventEmitter);
publisher.on('foo', (...args) => {
    args satisfies readonly [null]; // Is this right?
});
```

lorisleiva (approved):
This is great! I love that we define our own `DataPublisher` type abstraction instead of working directly with event emitters. And I love that the transformation to an async iterable is the very last step, essentially making it an "API decorator".

packages/subscribable/src/__tests__/async-iterable-test.ts
```diff
@@ -0,0 +1,220 @@
...
+                done: false,
+                value: 'queued message 1',
+            });
+            await expect(iteratorB.next()).resolves.toStrictEqual({
+                done: false,
+                value: 'queued message 1',
+            });
+            await expect(iteratorA.next()).rejects.toThrow(new Error('o no'));
+            await expect(iteratorB.next()).rejects.toThrow(new Error('o no'));
+        });
```
steveluscher, line 196: One key difference between this implementation and the last is that the last one would nuke the iterators immediately upon receiving an error. This might have caused code that was behind processing notifications to miss notifications that had already arrived on the client.

This implementation continues to vend any messages in the queue before vending the error. Errors are now a queue member like anything else.

packages/subscribable/src/async-iterable.ts
```diff
@@ -0,0 +1,172 @@
...
+                        __type: PublishType.DATA,
+                        data: data as TData,
+                    });
+                }
+            });
+        },
+        options,
+    );
+    return {
+        async *[Symbol.asyncIterator]() {
```
steveluscher, line 117: This implementation still uses generators instead of the manual approach cancellation (via `return()`) in #3172. This means that the only way to interrupt this generator is to call the abort signal, which aborts the whole subscription – you can't just call `.return()` on the iterator. This is _probably_ sufficient for most use cases today, but we should consider rolling in some of the ideas and tests from #3172 anyway.

packages/subscribable/src/__typetests__/data-publisher-typetest.ts
steveluscher: Done!

steveluscher:
### Merge activity

* **Oct 4, 10:55 AM PDT**: @steveluscher started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3301).

PR #3340, opened 2024-10-06
The default RPC transport now requests wire compression in Node

steveluscher:
# Summary

The `fetch` API in browsers automatically requests the wire compression encoding that the browser supports, in the order it wants it. Node's `fetch` API does not.

In this PR we add all of the default algorithms supported by Node LTS at the time of writing.

Closes #3338.

# Test Plan

1. Install Node LTS (`n lts`)
2. Test that it actually supports these

```
> await (await fetch('https://httpbin.org/deflate', {headers:{'Accept-Encoding': 'deflate', 'Content-Type': 'application/json'}})).json()
{
  deflated: true,
  headers: {
    Accept: '*/*',
    'Accept-Encoding': 'deflate',
    'Accept-Language': '*',
    'Content-Type': 'application/json',
    Host: 'httpbin.org',
    'Sec-Fetch-Mode': 'cors',
    'User-Agent': 'node',
    'X-Amzn-Trace-Id': 'Root=1-67020914-22d36c347b3b8b0f356d9a98'
  },
  method: 'GET',
  origin: '145.40.88.235'
}
> await (await fetch('https://httpbin.org/gzip', {headers:{'Accept-Encoding': 'gzip', 'Content-Type': 'application/json'}})).json()
{
  gzipped: true,
  headers: {
    Accept: '*/*',
    'Accept-Encoding': 'gzip',
    'Accept-Language': '*',
    'Content-Type': 'application/json',
    Host: 'httpbin.org',
    'Sec-Fetch-Mode': 'cors',
    'User-Agent': 'node',
    'X-Amzn-Trace-Id': 'Root=1-67020916-28ed8e6b0b8c612f3be0f76f'
  },
  method: 'GET',
  origin: '145.40.88.235'
}
> await (await fetch('https://httpbin.org/brotli', {headers:{'Accept-Encoding': 'br', 'Content-Type': 'application/json'}})).json()
{
  brotli: true,
  headers: {
    Accept: '*/*',
    'Accept-Encoding': 'br',
    'Accept-Language': '*',
    'Content-Type': 'application/json',
    Host: 'httpbin.org',
    'Sec-Fetch-Mode': 'cors',
    'User-Agent': 'node',
    'X-Amzn-Trace-Id': 'Root=1-6702091a-1a5970103512353b1eafdfcf'
  },
  method: 'GET',
  origin: '145.40.88.235'
}
```

PR #3347, opened 2024-10-07
Repair benchmarks

steveluscher:
# Summary

Because we moved a build-time contstant to the top level in #3340, we can no longer run these scripts without replacing the build time constants. That means that running this with `tsx` will yield a `__NODEJS__ is not defined` error.

In this PR, we change the benchmarks to run on the built source, and teach Turborepo to build it beforehand. You now run benchmarks with `pnpm turbo benchmark`.

### DIFF
--- a/packages/codecs-core/src/__benchmarks__/run.ts
+++ b/packages/codecs-core/src/__benchmarks__/run.ts
@@ -1,4 +1,4 @@
-#!/usr/bin/env -S pnpm dlx tsx
+#!/usr/bin/env -S pnpm dlx tsx -r ../build-scripts/register-node-globals.cjs
 
 import { webcrypto as crypto } from 'node:crypto';
 
--- a/packages/build-scripts/package.json
+++ b/packages/build-scripts/package.json
@@ -3,6 +3,7 @@
     "version": "0.0.0",
     "private": true,
     "files": [
+        "register-node-globals.mjs",
         "tsup.config.library.ts",
         "tsup.config.package.ts"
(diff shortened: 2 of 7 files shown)


steveluscher:
* **#3347** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3347?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

steveluscher (commented):
Unless anyone knows of a better way to run a script with `tsx` _and_ replace build-time constants on the fly?

packages/codecs-core/src/__benchmarks__/run.ts
```diff
@@ -4,9 +4,9 @@ import { webcrypto as crypto } from 'node:crypto';
 
 import { Bench } from 'tinybench';
 
-import { createCodec, FixedSizeCodec } from '../codec';
-import { ReadonlyUint8Array } from '../readonly-uint8array';
-import { reverseCodec } from '../reverse-codec';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
```
steveluscher, line 8: These ignore directives imply that these benchmarks will break frequently without warning, but I don't know of any other way to:

1. Refer to the built source
2. Before it exists
3. In a way that will typecheck

steveluscher: I figured it out.

PR #3374, opened 2024-10-11
Replace abuse of promises with the `queueMicrotask` API in the RPC request coalescer

steveluscher:
# Summary

The `queueMicrotask` API is well-supported enough now to use for this use case. https://caniuse.com/mdn-api_queuemicrotask

### DIFF
--- a/packages/rpc/src/__tests__/rpc-request-coalescer-test.ts
+++ b/packages/rpc/src/__tests__/rpc-request-coalescer-test.ts
@@ -21,10 +21,10 @@ describe('RPC request coalescer', () => {
             coalescedTransport({ payload: null }).catch(() => {});
             expect(mockTransport).toHaveBeenCalledTimes(1);
         });
-        it('multiple requests in different ticks each produce their own transport request', async () => {
+        it('multiple requests in different ticks each produce their own transport request', () => {
             expect.assertions(1);
             coalescedTransport({ payload: null }).catch(() => {});
-            await jest.runOnlyPendingTimersAsync();
+            jest.runAllTicks();
             coalescedTransport({ payload: null }).catch(() => {});
             expect(mockTransport).toHaveBeenCalledTimes(2);
         });
@@ -46,7 +46,7 @@ describe('RPC request coalescer', () => {
             mockTransport.mockResolvedValueOnce(mockResponseA);
             mockTransport.mockResolvedValueOnce(mockResponseB);
             const responsePromiseA = coalescedTransport({ payload: null });
-            await jest.runOnlyPendingTimersAsync();
+            jest.runAllTicks();
             const responsePromiseB = coalescedTransport({ payload: null });
             await Promise.all([
                 expect(responsePromiseA).resolves.toBe(mockResponseA),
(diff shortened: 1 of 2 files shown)


steveluscher:
* **#3374** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3374?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/rpc/src/__tests__/rpc-request-coalescer-test.ts
```diff
@@ -21,10 +21,10 @@ describe('RPC request coalescer', () => {
             coalescedTransport({ payload: null }).catch(() => {});
             expect(mockTransport).toHaveBeenCalledTimes(1);
         });
-        it('multiple requests in different ticks each produce their own transport request', async () => {
+        it('multiple requests in different ticks each produce their own transport request', () => {
             expect.assertions(1);
             coalescedTransport({ payload: null }).catch(() => {});
-            await jest.runOnlyPendingTimersAsync();
+            jest.runAllTicks();
```
steveluscher, line 27: This is a simplification of these tests, but it should be noted that the tests in their pre-PR state also pass with the change to `queueMicrotask`.

mcintyre94: I'm not really familiar with these jest timing functions, but would there be any value in having tests that use `await jest.runOnlyPendingTimersAsync();` too? Or are these doing the same thing in Jest?

mcintyre94 (approved):

packages/rpc/src/__tests__/rpc-request-coalescer-test.ts
steveluscher: The [docs](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/58aabc0cfd2baf08f5f71a2712ae7baa6cb2a3ce/types/jest/index.d.ts#L299-L329) are actually pretty great for those functions.

```ts
/**
 * Exhausts the micro-task queue (i.e., tasks in Node.js scheduled with `process.nextTick()`).
 */
function runAllTicks(): void;
/**
 * Executes only the macro-tasks that are currently pending (i.e., only the tasks that
 * have been queued by `setTimeout()`, `setInterval()` and `setImmediate()` up to this point).
 */
function runOnlyPendingTimers(): void;
/**
 * Asynchronous equivalent of `jest.runOnlyPendingTimers()`. It also yields to the event loop,
 * allowing any scheduled promise callbacks to execute _before_ running the timers.
 *
 * @remarks
 * Not available when using legacy fake timers implementation.
 */
function runOnlyPendingTimersAsync(): Promise<void>;
```

The metapoint here is that you want to run tests in such a way that they test async operations _without actually waiting for them_. If you can _simulate_ `await` without actually doing it, your tests run a little faster and with less fuss.

This change replaces the `await` with something that just pumps the microtask queue, to trigger the `queueMicrotask` in this PR to run.

steveluscher:
### Merge activity

* **Oct 15, 10:14 AM PDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3374).

PR #3392, opened 2024-10-17
fix: cover logs error for ws message

barbarosh:
Some message from websocket contain error 
```json
{
  "id": 59,
  "jsonrpc": "2.0",
  "error": {
    "code": -32601,
    "message": "Subscriptions unsupported for this network"
  }
}
```
Which not present any where, just add console log for represent error for developer

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -6251,13 +6251,15 @@ export class Connection {
                   serverSubscriptionId
                 ] = subscription.callbacks;
                 await this._updateSubscriptions();
-              } catch (e) {
+              } catch (e: any) {
                 if (e instanceof Error) {
                   console.error(
                     `${method} error for argument`,
                     args,
                     e.message,
                   );
+                } else if (typeof e === 'object' && e.code) {
+                  console.log(`${method} error for argument`, args, e.message);
                 }
                 if (!isCurrentConnectionStillActive()) {
                   return;
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -6251,16 +6251,14 @@ export class Connection {
                   serverSubscriptionId
                 ] = subscription.callbacks;
                 await this._updateSubscriptions();
-              } catch (e: any) {
-                if (e instanceof Error) {
-                  console.error(
-                    `${method} error for argument`,
+              } catch (e) {
+                console.error(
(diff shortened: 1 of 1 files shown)


steveluscher:
This is a good change. This does _not_ of course solve the problem of the subscription system endlessly looping when it encounters an error like this. Consider this:

```ts
import { Connection } from '@solana/web3.js';
const connection = new Connection('https://api.devnet.solana.com');
connection.onAccountChange({ toBase58() { return '1'.repeat(31); }}, console.log);
```

This invalid pubkey will generate a JSON-RPC error:

```
Received JSON-RPC error calling `accountSubscribe` {
  args: [
    '1111111111111111111111111111111',
    { encoding: 'base64', commitment: 'finalized' }
  ],
  error: { code: -32602, message: 'Invalid Request: Invalid pubkey provided' }
}
```

But then the thing will loop forever, retrying the subscription endlessly.

The problem is that this API has no practical way to communicate subscription failure to the client. This has been fixed in `@solana/web3.js@2.0.0` and will not be addressed in the 1.x line.

barbarosh:
> This is a good change. This does _not_ of course solve the problem of the subscription system endlessly looping when it encounters an error like this. Consider this:
> 
> ```ts
> import { Connection } from '@solana/web3.js';
> const connection = new Connection('https://api.devnet.solana.com');
> connection.onAccountChange({ toBase58() { return '1'.repeat(31); }}, console.log);
> ```
> 
> This invalid pubkey will generate a JSON-RPC error:
> 
> ```
> Received JSON-RPC error calling `accountSubscribe` {
>   args: [
>     '1111111111111111111111111111111',
>     { encoding: 'base64', commitment: 'finalized' }
>   ],
>   error: { code: -32602, message: 'Invalid Request: Invalid pubkey provided' }
> }
> ```
> 
> But then the thing will loop forever, retrying the subscription endlessly.
> 
> The problem is that this API has no practical way to communicate subscription failure to the client. This has been fixed in `@solana/web3.js@2.0.0` and will not be addressed in the 1.x line.

Thanks, for explanation, of course this changes just for notify developer what something wrong with subscribe, and I like about 2 version which fixes this behavior, will wait for stable release

PR #3418, opened 2024-10-23
feat: Agave v2 RPC: replace `getConfirmedTransaction` with `getTransaction`

buffalojoec:
#### Problem
With the upcoming upgrade from 1.18 to 2.0 on Solana mainnet-beta, deprecated RPC methods have been removed, therefore they will no longer be available through Web3.js client requests.

The [Agave 2.0 Migration Guide](https://github.com/anza-xyz/agave/wiki/Agave-v2.0-Transition-Guide) lists semi-equivalent RPC method counterparts for each of the removed methods.

#### Summary of Changes
Replace `getConfirmedTransaction` with `getTransaction`. This method was already requiring a confirmation level of `confirmed` or `finalized`, so the behavior should be unchanged.

PR #3419, opened 2024-10-23
feat: Agave v2 RPC: replace `getRecentBlockhash` with `getLatestBlockhash`

buffalojoec:
#### Problem
With the upcoming upgrade from 1.18 to 2.0 on Solana mainnet-beta, deprecated RPC methods have been removed, therefore they will no longer be available through Web3.js client requests.

The [Agave 2.0 Migration Guide](https://github.com/anza-xyz/agave/wiki/Agave-v2.0-Transition-Guide) lists semi-equivalent RPC method counterparts for each of the removed methods.

#### Summary of Changes
Replace `getRecentBlockhash` with `getLatestBlockhash`. This method still vends a blockhash, but the field `lamportsPerSignature` cannot be served, so accessing it throws an error. However, if apps use the deprecated `getRecentBlockhash` method and only access the `blockhash` field, they will not see an error, and the request is served by `getLatestBlockhash`.

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -2600,20 +2600,6 @@ const GetParsedTransactionRpcResult = jsonRpcResult(
 );
 
 /**
- * Expected JSON RPC response for the "getRecentBlockhash" message
- *
- * @deprecated Deprecated since RPC v1.8.0. Please use {@link GetLatestBlockhashRpcResult} instead.
- */
-const GetRecentBlockhashAndContextRpcResult = jsonRpcResultAndContext(
-  pick({
-    blockhash: string(),
-    feeCalculator: pick({
-      lamportsPerSignature: number(),
-    }),
-  }),
-);
-
-/**
  * Expected JSON RPC response for the "getLatestBlockhash" message
  */
 const GetLatestBlockhashRpcResult = jsonRpcResultAndContext(
@@ -4567,13 +4553,25 @@ export class Connection {
       feeCalculator: FeeCalculator;
     }>
   > {
-    const args = this._buildArgs([], commitment);
-    const unsafeRes = await this._rpcRequest('getRecentBlockhash', args);
-    const res = create(unsafeRes, GetRecentBlockhashAndContextRpcResult);
-    if ('error' in res) {
-      throw new SolanaJSONRPCError(res.error, 'failed to get recent blockhash');
-    }
-    return res.result;
+    const {
+      context,
+      value: {blockhash},
+    } = await this.getLatestBlockhashAndContext(commitment);
+    return {
+      context,
+      value: {
+        blockhash,
+        feeCalculator: {
+          get lamportsPerSignature(): number {
+            throw new Error(
+              'The capability to fetch `lamportsPerSignature` using the `getRecentBlockhash` API is ' +
+                'no longer offered by the network. Use the `getFeeForMessage` API to obtain the fee ' +
+                'for a given message.',
+            );
+          },
+        },
+      },
+    };
   }
 
   /**
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -5353,7 +5353,7 @@ export class Connection {
     commitment?: Finality,
   ): Promise<ConfirmedTransaction | null> {
     const args = this._buildArgsAtLeastConfirmed([signature], commitment);
-    const unsafeRes = await this._rpcRequest('getConfirmedTransaction', args);
+    const unsafeRes = await this._rpcRequest('getTransaction', args);
     const res = create(unsafeRes, GetTransactionRpcResult);
     if ('error' in res) {
       throw new SolanaJSONRPCError(res.error, 'failed to get transaction');
@@ -5384,7 +5384,7 @@ export class Connection {
       commitment,
       'jsonParsed',
     );
-    const unsafeRes = await this._rpcRequest('getConfirmedTransaction', args);
+    const unsafeRes = await this._rpcRequest('getTransaction', args);
     const res = create(unsafeRes, GetParsedTransactionRpcResult);
     if ('error' in res) {
       throw new SolanaJSONRPCError(
@@ -5411,7 +5411,7 @@ export class Connection {
         'jsonParsed',
       );
       return {
-        methodName: 'getConfirmedTransaction',
+        methodName: 'getTransaction',
         args,
       };
     });
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -5225,7 +5225,7 @@ export class Connection {
     commitment?: Finality,
   ): Promise<ConfirmedBlock> {
     const args = this._buildArgsAtLeastConfirmed([slot], commitment);
-    const unsafeRes = await this._rpcRequest('getConfirmedBlock', args);
+    const unsafeRes = await this._rpcRequest('getBlock', args);
     const res = create(unsafeRes, GetConfirmedBlockRpcResult);
 
     if ('error' in res) {
@@ -5331,7 +5331,7 @@ export class Connection {
         rewards: false,
       },
     );
-    const unsafeRes = await this._rpcRequest('getConfirmedBlock', args);
+    const unsafeRes = await this._rpcRequest('getBlock', args);
     const res = create(unsafeRes, GetBlockSignaturesRpcResult);
     if ('error' in res) {
       throw new SolanaJSONRPCError(res.error, 'failed to get confirmed block');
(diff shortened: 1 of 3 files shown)


buffalojoec:
* **#3591** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3591?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3419** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3419?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3419?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3418** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3418?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3417** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3417?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `maintenance/v1.x`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

mcintyre94:
What's our semver deal here, are we committed to backward compatibility across 1.x? I'd like to just remove the deprecated method TBH, but maybe we can't do that in a 1.x release. Failing that I'd prefer to just hardcode 5000, but not too worried either way.

buffalojoec:
> What's our semver deal here, are we committed to backward compatibility across 1.x? I'd like to just remove the deprecated method TBH, but maybe we can't do that in a 1.x release. Failing that I'd prefer to just hardcode 5000, but not too worried either way.

We can't just remove them, even if they are deprecated. I can change this to be hard-coded, but we'd lose out on the fact that fees are supposed to change under varying network load.
https://github.com/anza-xyz/agave/blob/96249691b4b7c873220b27376f271ead38392541/sdk/fee-calculator/src/lib.rs#L109

This might not be the prettiest solution, but it does capture fee rate changes with accuracy.

It's worth noting that the probability of someone using this field - on a long-deprecated method no less - is quite low.

@steveluscher what do you think here?

steveluscher:
OK, hear me out. How absolutely nuts would it be to:

1. Deploy a program to mainnet/testnet/devnet that gave you the fee-per-signature as `returnData`
2. Create a message that calls that program, having _no_ blockhash
3. Use `simulateTransaction` with that message, configured with `replaceRecentBlockhash` set to `true`
4. Pull out the [`replacementBlockhash`](https://github.com/anza-xyz/agave/blob/master/rpc-client-api/src/response.rs#L411) and `returnData` from that simulation

That way you could get this all done in a single call to the server.

buffalojoec:
> 1. Deploy a program to mainnet/testnet/devnet that gave you the fee-per-signature as `returnData`

How does a program get this info?

steveluscher:
> How does a program get this info?

I was hoping you'd know better than me. [This](https://github.com/anza-xyz/agave/blob/ef5c7a3859e8ca0d8538b52ccb422fd0f8216f36/sdk/program/src/syscalls/definitions.rs#L51)?

buffalojoec:
> > How does a program get this info?
> 
> I was hoping you'd know better than me. [This](https://github.com/anza-xyz/agave/blob/ef5c7a3859e8ca0d8538b52ccb422fd0f8216f36/sdk/program/src/syscalls/definitions.rs#L51)?

Actually I was tempted to use this and just load it from account data, but it's deprecated. So we can officially kick the can down the road and just load from a deprecated sysvar for now, and probably avoid the requirement for an on-chain program altogether.

https://github.com/anza-xyz/agave/blob/32b677bf4f2b4a718f62683e1daec87a3a83b83d/sdk/program/src/sysvar/fees.rs#L42
https://github.com/anza-xyz/agave/blob/32b677bf4f2b4a718f62683e1daec87a3a83b83d/sdk/fee-calculator/src/lib.rs#L23

This sysvar is still updated per-slot by Bank, so it would totally work (for now).

steveluscher:
> It would still be two calls…

Well, the difference with that approach is that you could _parallelize_ the calls, so that's much, much better than the implementation with serial calls. Go for it.

steveluscher:
Oof. I'm not sure who we're building this for anymore.

* The person who can't just switch to `getFeeForMessage()`
* …but who _is_ willing to upgrade all the way to the latest 1.x version

Surely that's nobody, right?

At this point I'm tempted to just:

```ts
return {
  context,
  value: {
    blockhash,
    feeCalculator: {
      get lamportsPerSignature() {
        throw new Error(
          'The capability to fetch `lamportsPerSignature` using the `getRecentBlockhash` API is ' +
          'no longer offered by the network. Use the `getFeeForMessage` API to obtain the fee ' +
          'for a given message.',
        );
      }
    },
  },
}
```

buffalojoec:
> Oof. I'm not sure who we're building this for anymore.
> 
> * The person who can't just switch to `getFeeForMessage()`
> * …but who _is_ willing to upgrade all the way to the latest 1.x version
> 
> Surely that's nobody, right?
> 
> At this point I'm tempted to just:
> 
> ```ts
> return {
>   context,
>   value: {
>     blockhash,
>     feeCalculator: {
>       get lamportsPerSignature() {
>         throw new Error(
>           'The capability to fetch `lamportsPerSignature` using the `getRecentBlockhash` API is ' +
>           'no longer offered by the network. Use the `getFeeForMessage` API to obtain the fee ' +
>           'for a given message.',
>         );
>       }
>     },
>   },
> }
> ```

Honestly this works for me. With this approach, we aren't technically breaking our API. However, the network simply doesn't support this field anymore, so they're going to get an error anyway. At least here we can vend a response for the endpoint, instead of the app getting an error outright (which it should).

src/connection.ts
```diff
@@ -4567,13 +4553,25 @@ export class Connection {
...
+      value: {
+        blockhash,
+        feeCalculator: {
+          get lamportsPerSignature(): number {
+            throw new Error(
+              'The capability to fetch `lamportsPerSignature` using the `getRecentBlockhash` API is ' +
+                'no longer offered by the network. Use the `getFeeForMessage` API to obtain the fee ' +
+                'for a given message.',
+            );
+          },
```
steveluscher, line 4571: You know what's annoying? This works just fine with APIs like `console.log`, but if you try to pass this to `JSON.stringify()` it will trigger a read and blow up, which could probably really screw up some error logging products.

```ts
const a = {stuff: 'things', get foo() { throw new Error('o no'); }}
console.log(a); // OK
JSON.stringify(a); // BOOM
```

Add this:

```ts
  get lamportsPerSignature(): number { throw new Error(/* ... */); },
  toJSON() {
    const safe = {};
    Object.keys(this)
      .filter((k) => k !== "lamportsPerSignature")
      .forEach((k) => {
        safe[k] = this[k];
      });
    return safe;
  },
```

…and then maybe a test for that.

steveluscher (approved):
Added via Giphy

src/connection.ts
buffalojoec: Sure, but we are already doing this in a few areas in `Connection` already. Do you want all of those to get this treatment?

steveluscher: Doing what? Throwing? Every response that would nuke JSON.stringify should probably get protected against doing so, yeah.

buffalojoec: Ohh I misunderstood the exact problem. This is the only spot where we do this on a field.

I can add this, but TypeScript will barf on it unless I also add `toJSON?: object` to the `FeeCalculator` interface. Additionally, since `lamportsPerSignature` is the only field, we can just make this return `{}` can't we?

steveluscher: Add it with `Object.defineProperty()`.

buffalojoec: Haha actually you can trick TypeScript by just declaring it before returning it. 🫠

steveluscher: “Structural typing.”

steveluscher:
~Ship it, except you can't because you need to sign your commits now~.

steveluscher (dismissed):

buffalojoec:
### Merge activity

* **Dec 15, 11:14 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3419).
* **Dec 15, 11:18 PM EST**: [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3419) couldn't merge this pull request because a downstack PR #3418 failed to merge.

PR #3430, opened 2024-10-23
Add `execute` function to `RpcPlan`

lorisleiva:
Now we're getting in the fun bit.

This PR adds an `execute` function to the `RpcPlan` type whose entire purpose is to dictate how the request should be forwarded to the transport and how its response should be transformed.

Note that the following PR will remove the `payload` and `responseTransformer` attributes from the plan as they are no longer necessary.

Also note that a follow-up PR will add a changeset for these changes.

PR #3435, opened 2024-10-24
Run examples in CI checks

lorisleiva:
This PR adds the `start` script in the `build` pipeline such that CI ends up not just linting and type-checking examples but also making sure they run without failing.

To do so, I've had to tweak the `pressAnyKeyToContinue` helper since otherwise CI just force-closes the session after a timeout. For that, we can use the `process.env.CI` variable provided in GitHub Actions environments.

Also note that CI purposefully fails here (for the correct reason) since we currently have a failing example (due to one of my previous refactoring). The following PR fixes the example to show that fixes the example also makes CI pass.

**EDIT**: Additionally, I've had to change the RPC URL in CI, otherwise we will keep getting 429 when trying to hit mainnet.

### DIFF
--- a/examples/rpc-custom-api/src/example.ts
+++ b/examples/rpc-custom-api/src/example.ts
@@ -12,12 +12,27 @@ import {
     createRpc,
     createSolanaRpcApi,
     DEFAULT_RPC_CONFIG,
+    isJsonRpcPayload,
     mainnet,
     RpcApi,
+    RpcTransport,
+    RpcTransportMainnet,
     SolanaRpcApiMainnet,
 } from '@solana/web3.js';
 
 const log = createLogger('Custom JSON RPC API');
+
+// This is a workaround that avoids the need to send RPC calls to mainnet during CI.
+function mockRpcTransportForCI(): RpcTransportMainnet {
+    const transport = createDefaultRpcTransport({ url: 'http://127.0.0.1:8899' });
+    return (async <TResponse>({ payload, signal }: Parameters<RpcTransport>[0]) => {
+        if (isJsonRpcPayload(payload) && payload.method === 'getAsset') {
+            const metadata = { name: 'USD Coin', symbol: 'USDC' };
+            return { result: { content: { metadata } } } as TResponse;
+        }
+        return await transport({ payload, signal });
+    }) as RpcTransportMainnet;
+}
 
 /**
  * STEP 1: CUSTOM JSON RPC API CALL SIGNATURE
@@ -97,7 +112,9 @@ const customizedRpcApi = new Proxy(solanaRpcApi, {
  */
 const customizedRpc = createRpc({
     api: customizedRpcApi,
-    transport: createDefaultRpcTransport({ url: mainnet('https://api.mainnet-beta.solana.com') }),
+    transport: process.env.CI
+        ? mockRpcTransportForCI()
+        : createDefaultRpcTransport({ url: mainnet('https://api.mainnet-beta.solana.com') }),
 });
 
 /**
--- a/examples/rpc-transport-throttled/src/example.ts
+++ b/examples/rpc-transport-throttled/src/example.ts
@@ -107,9 +107,8 @@ function getThrottledTransport<TClusterUrl extends ClusterUrl>(
  * Create a default RPC transport, wrap it in a throttled transport, then create a Solana RPC
  * instance from the resulting transport.
  */
-const defaultTransport = createDefaultRpcTransport({
-    url: mainnet('https://api.mainnet-beta.solana.com'),
-});
+const url = process.env.CI ? 'http://127.0.0.1:8899' : mainnet('https://api.mainnet-beta.solana.com');
+const defaultTransport = createDefaultRpcTransport({ url });
 const throttledTransport = getThrottledTransport(defaultTransport);
 const throttledRpc = createSolanaRpcFromTransport(throttledTransport);
 
--- a/examples/utils/pressAnyKeyPrompt.ts
+++ b/examples/utils/pressAnyKeyPrompt.ts
@@ -3,8 +3,12 @@ import colors from 'yoctocolors';
 
 export default createPrompt<void, string>((message = 'Press any key to exit', done) => {
     const prefix = usePrefix({});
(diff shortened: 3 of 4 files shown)


lorisleiva:
* **#3435** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3435?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

examples/rpc-custom-api/src/example.ts
```diff
@@ -12,13 +12,28 @@ import {
...
 const log = createLogger('Custom JSON RPC API');
 
+// This is a workaround that avoids the need to send RPC calls to mainnet during CI.
+function mockRpcTransportForCI(): RpcTransportMainnet {
+    const transport = createDefaultRpcTransport({ url: 'http://127.0.0.1:8899' });
+    return (async <TResponse>({ payload, signal }: Parameters<RpcTransport>[0]) => {
+        if (isJsonRpcPayload(payload) && payload.method === 'getAsset') {
+            const metadata = { name: 'USD Coin', symbol: 'USDC' };
+            return { result: { content: { metadata } } } as TResponse;
+        }
```
mcintyre94, line 32: I think it'd be worth verifying `params` here too, just to make sure we don't break anything in how they're passed through

I'd also throw an error or at least log the payload in the else case to help with debugging

mcintyre94 (approved):
Nice! These are useful tests to have running in CI

mcintyre94:
Just to point out, this doesn't run the `react` example, which I was wondering how that would run/what that would do in CI

That example doesn't have a `start` job, which makes sense. I'm just wondering if `start` is clear enough though.

turbo.json
```diff
@@ -45,6 +47,10 @@
             ],
             "passThroughEnv": ["GH_TOKEN", "NPM_TOKEN"]
         },
+        "start": {
+            "dependsOn": ["^compile:js", "^compile:typedefs"],
+            "inputs": ["$TURBO_DEFAULT$", "src/**", "../utils/**"]
```
steveluscher, line 52:
```suggestion
            "inputs": ["$TURBO_DEFAULT$", "src/**", "../utils/**"],
            "outputs": []
```

turbo.json
```diff
@@ -1,5 +1,6 @@
 {
     "$schema": "https://turbo.build/schema.json",
+    "globalEnv": ["CI"],
```
steveluscher, line 3: This will disable remote caching on PRs, forever. Locally run builds will never match ones run on PRs.

turbo.json
```diff
@@ -45,6 +47,10 @@
             ],
             "passThroughEnv": ["GH_TOKEN", "NPM_TOKEN"]
         },
+        "start": {
+            "dependsOn": ["^compile:js", "^compile:typedefs"],
```
steveluscher, line 51: Running the examples doesn't actually depend on typedefs.
```suggestion
            "dependsOn": ["^compile:js"],
```

examples/utils/pressAnyKeyPrompt.ts
```diff
@@ -3,8 +3,12 @@ import colors from 'yoctocolors';
...
     const prefix = usePrefix({});
-    useKeypress(() => {
+    if (process.env.CI) {
         done();
-    });
+    } else {
+        useKeypress(() => {
+            done();
+        });
+    }
```
steveluscher, line 12: Added via Giphy

examples/rpc-transport-throttled/src/example.ts
```diff
@@ -107,9 +107,8 @@ function getThrottledTransport<TClusterUrl extends ClusterUrl>(
  * Create a default RPC transport, wrap it in a throttled transport, then create a Solana RPC
  * instance from the resulting transport.
  */
-const defaultTransport = createDefaultRpcTransport({
-    url: mainnet('https://api.mainnet-beta.solana.com'),
-});
+const url = process.env.CI ? 'http://127.0.0.1:8899' : mainnet('https://api.mainnet-beta.solana.com');
```
steveluscher, line 110: Added via Giphy

examples/rpc-custom-api/src/example.ts
```diff
@@ -12,13 +12,28 @@ import {
...
     RpcApi,
+    RpcTransport,
+    RpcTransportMainnet,
     SolanaRpcApiMainnet,
 } from '@solana/web3.js';
 
 const log = createLogger('Custom JSON RPC API');
 
+// This is a workaround that avoids the need to send RPC calls to mainnet during CI.
+function mockRpcTransportForCI(): RpcTransportMainnet {
```
steveluscher, line 26: Added via Giphy

turbo.json
```diff
@@ -15,7 +16,8 @@
                 "test:treeshakability:node",
                 "test:typecheck",
                 "test:unit:browser",
-                "test:unit:node"
+                "test:unit:node",
+                "start"
```
steveluscher, line 20: I feel like the _semantics_ of what you're trying to do here is ‘test the examples.’ Instead of adding a new dimension (`start`) to the build, what would you think of adding a `test:unit:node` _task_ to each example (ie. that just ran `run:example`).

steveluscher:
> Just to point out, this doesn't run the `react` example

But it is already caught by `compile:js`.

steveluscher (commented):
Two things are true.

1. The change in this PR would have caught the regression from #3430 in CI.
2. If the proxy's `get()` method had a typehint, that would have also been sufficient to catch the regression in CI.

```diff
-    get(target, p, receiver) {
+    get(target, p, receiver): (address: Address) => RpcPlan<AssetMetadata> {
         const methodName = p.toString();
         if (methodName === 'getAssetMetadata') {
             /**
              * When the `getAssetMetadata` method is called on the RPC, return a custom definition.
              */
-            return (address: Address) => {
+            return (address) => {
```

I'm of the opinion that testing the examples, particularly with all the special casing required, is too large a hammer, especially with the terminal damage it does to our ability to cache steps on PR runs.

lorisleiva:
@steveluscher Fair enough! Let's close this then and make sure we add more types to our examples.

PR #3453, opened 2024-10-25
Rename decodeTransactionMessage to decompileTransactionMessageFetchingLookupTables

mcintyre94:
This function had a confusing name, the new name better describes its functionality. It uses the passed in RPC to fetch lookup tables, and then passes them to `decompileTransactionMessage`.


lorisleiva (approved):
Thanks!

steveluscher (approved):
Added via Giphy

PR #3496, opened 2024-10-30
Change references from Kinobi to Codama

mcintyre94:
Let's point the web3js docs at our fork to avoid confusion.


lorisleiva (approved):
Oh thanks, I missed that one.

PR #3506, opened 2024-10-31
Add a failing test that verifies we can subscribe to an RPC subscription method

mcintyre94:
In rc2 there's a bug where `await rpcSubscriptions.anyNotifications().subscribe()` hangs. Unlike `rpc-api` [all tests in `rpc-subscriptions-api` are currently commented out.  ](https://github.com/solana-labs/solana-web3.js/blob/master/packages/rpc-subscriptions-api/src/__tests__/account-notifications-test.ts)

This PR adds a (currently failing) test to subscribe to notifications against the localhost test validator.

Notes:

- I've put this test in `rpc-subscriptions` rather than `rpc-subscriptions-api` because it relies on transport functions from `rpc-subscriptions`. If we wanted to build a full suite of `rpc-subscriptions-api` tests later then we should probably move these transport functions to a different package where they could be a dependency of `rpc-subscriptions-api`. We currently have `rpc-transport-http` for this on the `rpc-api` side. 

- I used `it.skip` instead of `it.failing` because the test times out rather than failing, and Jest doesn't treat that as a pass for `it.failing`. Anyway I don't plan to leave it failing long!

### DIFF
--- a/packages/rpc-subscriptions/src/__tests__/rpc-subscriptions-functional-test.ts
+++ b/packages/rpc-subscriptions/src/__tests__/rpc-subscriptions-functional-test.ts
@@ -1,0 +1,40 @@
+import { Address } from '@solana/addresses';
+import { AccountNotificationsApi, SolanaRpcSubscriptionsApi } from '@solana/rpc-subscriptions-api';
+import { createSubscriptionRpc, RpcSubscriptions } from '@solana/rpc-subscriptions-spec';
+
+import {
+    createDefaultRpcSubscriptionsTransport,
+    createDefaultSolanaRpcSubscriptionsChannelCreator,
+    createSolanaRpcSubscriptionsApi,
+} from '..';
+
+function createLocalhostSolanaRpcSubscriptions(): RpcSubscriptions<SolanaRpcSubscriptionsApi> {
+    return createSubscriptionRpc({
+        api: createSolanaRpcSubscriptionsApi(),
+        transport: createDefaultRpcSubscriptionsTransport({
+            createChannel: createDefaultSolanaRpcSubscriptionsChannelCreator({ url: 'ws://localhost:8900' }),
+        }),
+    });
+}
+
+describe('accountNotifications', () => {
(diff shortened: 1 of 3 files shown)


mcintyre94:
* **#3507** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3507?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3506** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3506?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

steveluscher (approved):
I guess `slotNotifications` would actually be a good one to start with, because it doesn't require writes, but this is fine too!

Added via Giphy

packages/rpc-subscriptions/src/__tests__/rpc-subscriptions-functional-test.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+describe('accountNotifications', () => {
+    let rpcSubscriptions: RpcSubscriptions<AccountNotificationsApi>;
+    beforeEach(() => {
+        rpcSubscriptions = createLocalhostSolanaRpcSubscriptions();
+    });
+
+    // eslint-disable-next-line jest/no-disabled-tests
+    it.skip('can subscribe to account notifications', async () => {
```
steveluscher, line 27: I think you can actually write this as `it.failing` instead of skipping it.

mcintyre94: I tried that, but because it fails by timing out it doesn't seem to work.

lorisleiva (approved):

steveluscher:
### Merge activity

* **Oct 31, 10:05 AM PDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3506).

PR #3518, opened 2024-11-01
Add failing tests for bigints in RPC errors

lorisleiva:
### Problem

I've encountered a new issue with the latest RC when updating program clients.

Functions like `isProgramError` that check the custom error code for the program are no longer working because the relevant `SolanaError` context now contains `bigint` error codes or instruction indices instead of the `number` type defined in their context.

Unfortunately, we cannot rely on numeric keypaths for these errors because these only transform successful responses.

### Solution

Since we already have error factories that, given some RPC error message returns a proper `SolanaError`, we can adjust the logic of these functions such that they return a consistent output no matter if they receive `bigint` or `number` codes or indices.

These methods are:
- `getSolanaErrorFromJsonRpcError`: The main RPC error entry point. Which may delegate to:
- `getSolanaErrorFromTransactionError`: The function that resolves transaction errors. Which may delegate to:
- `getSolanaErrorFromInstructionError`: The function that resolves instruction errors.

The benefits of doing this is, these methods are used in both RPC calls and RPC subscriptions. Additionally, we can clean up [some custom logic](https://github.com/solana-labs/solana-web3.js/blob/33286743f6818e809a39d6332b435b56548b7907/packages/rpc-transport-http/src/http-transport-for-solana-rpc.ts#L21-L24) we're doing for RPC calls only.

### Plan

- First, in this PR, we add some failing tests to show that the current RPC error factories are not able to handle error messages that contain `bigint` error code or indices.
- We then make these tests pass in the next PR by adjusting the RPC error factories. That PR also adds a changeset.

### DIFF
--- a/packages/errors/src/__tests__/instruction-error-test.ts
+++ b/packages/errors/src/__tests__/instruction-error-test.ts
@@ -7,66 +7,87 @@ import {
 import { SolanaError } from '../error';
 import { getSolanaErrorFromInstructionError } from '../instruction-error';
 
+const EXPECTED_ERROR_CODES = [
+    ['GenericError', 4615001],
+    ['InvalidArgument', 4615002],
+    ['InvalidInstructionData', 4615003],
+    ['InvalidAccountData', 4615004],
+    ['AccountDataTooSmall', 4615005],
+    ['InsufficientFunds', 4615006],
+    ['IncorrectProgramId', 4615007],
+    ['MissingRequiredSignature', 4615008],
+    ['AccountAlreadyInitialized', 4615009],
+    ['UninitializedAccount', 4615010],
+    ['UnbalancedInstruction', 4615011],
+    ['ModifiedProgramId', 4615012],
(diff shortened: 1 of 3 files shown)


lorisleiva:
* **#3519** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3519?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3518** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3518?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

mcintyre94 (approved):

lorisleiva:
### Merge activity

* **Nov 5, 3:30 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3518).

PR #3552, opened 2024-11-11
manifest fixes 

leantOnSol:
- removes unnecesary + misleading `engine` field in all package manifests (this field should serve no purpose and differs from the actual `engines` field specified in all of these manifests)
- remove `engineStrict` field from root manifest (outdated + overwritten by `engine-strict = true` in .npmrc regardless)
- update root manifest `engines.node` to `>=20.18.0`

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,9 +53,8 @@
         "typedoc-plugin-missing-exports": "^3.0.0",
         "typescript": "^5.6.3"
     },
-    "engineStrict": true,
     "engines": {
-        "node": ">=14.0.0",
+        "node": ">=20.18.0",
         "npm": "please-use-pnpm",
         "pnpm": "^9",
         "yarn": "please-use-pnpm"
--- a/.changeset/empty-wombats-guess.md
+++ b/.changeset/empty-wombats-guess.md
@@ -1,0 +1,20 @@
+---
+"@solana/addresses": patch
+"@solana/assertions": patch
+"@solana/codecs-core": patch
+"@solana/codecs-data-structures": patch
+"@solana/codecs-numbers": patch
+"@solana/codecs-strings": patch
+"@solana/compat": patch
+"@solana/errors": patch
+"@solana/keys": patch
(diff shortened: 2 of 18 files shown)


steveluscher (commented):
Oh, interesting. We officially only support Node >=LTS but I remember Node 17.4 being set there because that's the first version where Ed25519 is supported. I think the engines field there was so that you couldn't run the subpackage tests on anything lower than Node 17.4.

Your node tests probably failed because Node didn't auto-inject the crypto global until v19.

Aren't the `engines` configs still necessary in each subpackage, because that's what gets published to npm? What if someone _just_ installs `@solana/keys` in their project? An `engines` config in the workspace root won't get published to npm, right?

leantOnSol:
> Oh, interesting. We officially only support Node >=LTS but I remember Node 17.4 being set there because that's the first version where Ed25519 is supported. I think the engines field there was so that you couldn't run the subpackage tests on anything lower than Node 17.4.
> 
> Your node tests probably failed because Node didn't auto-inject the crypto global until v19.
> 
> Aren't the `engines` configs still necessary in each subpackage, because that's what gets published to npm? What if someone _just_ installs `@solana/keys` in their project? An `engines` config in the workspace root won't get published to npm, right?

Yes, I left the `engines` fields in all subpackages untouched, just removed the additional `engine` fields that were there as well.
So all subpackages still have a valid `engines.node` field set to `>=20.18.0`!

steveluscher:
Oh, ha, my bad. Got it! I’ll ship this tomorrow.

package.json
```diff
@@ -53,9 +53,8 @@
         "typedoc-plugin-missing-exports": "^3.0.0",
         "typescript": "^5.6.3"
     },
-    "engineStrict": true,
     "engines": {
-        "node": ">=14.0.0",
+        "node": ">=20.18.0",
```
steveluscher, line 57: Note for future visitors. At the time this was landed, this was two LTS versions ago.

PR #3582, opened 2024-11-17
Enable no unsafe argument

Dozie2001:
tackles [Issue 3465](https://github.com/solana-labs/solana-web3.js/issues/3465)

### DIFF
--- a/examples/react-app/eslint.config.mjs
+++ b/examples/react-app/eslint.config.mjs
@@ -25,7 +25,7 @@ export default [
         },
         rules: {
             '@typescript-eslint/no-misused-promises': 'off',
-            '@typescript-eslint/no-unsafe-argument': 'off',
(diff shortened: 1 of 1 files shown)


steveluscher (changes requested):
Thanks for the submission, @Dozie2001!

The idea is to turn the lint rules on, one at a time, to _run_ the linter (`pnpm turbo test:lint`), look at the errors that result, and then to fix the actual errors in the code until the linter runs clean.

Also, `on` is not a valid option for ESLint. The idea is to change it from `off` to `error`.

Dozie2001:
> Thanks for the submission, @Dozie2001!

I will get to it @steveluscher

steveluscher:
Feel free to reopen these if ever you find time to revisit them!

PR #3601, opened 2024-11-18
Publish GitHub Pages previews on every PR

steveluscher:


steveluscher:
* **#3601** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3601?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/solana-labs/solana-web3.js/3601?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`

<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

.github/workflows/actions/publish-gh-pages/action.yml
```diff
@@ -0,0 +1,29 @@
...
+    - uses: actions/download-artifact@v4
+      with:
+        path: .ghpages-deploy
+        pattern: ghpages-deploy-artifacts*
+        merge-multiple: true
+
+    - name: Deploy to Github Pages
+      uses: peaceiris/actions-gh-pages@v3
+      with:
+        destination_dir: ${{ inputs.pr-number }}/
```
steveluscher, line 26: In the preview scenario, where there is a `destination_dir` set, I'm not sure if I need to set `keep_files: true` or not (so that multiple PRs can participate in the same `gh-pages` branch).

.github/workflows/actions/compile-gh-pages/action.yml
```diff
@@ -0,0 +1,32 @@
...
+        mkdir -p .ghpages-deploy
+        mv ./examples/react-app/dist/ .ghpages-deploy/example/
+        mv ./docs/* .ghpages-deploy/
+
+    - uses: actions/upload-artifact@v4
+      with:
+        include-hidden-files: true
+        name: ghpages-deploy-artifacts
+        path: .ghpages-deploy
+        retention-days: 1
```
mcintyre94, line 32: Is this going to delete the preview if it's not reviewed within 1 day? I guess we could just run the job again, but is there a downside to making this a bit longer?

.github/workflows/actions/compile-gh-pages/action.yml
```diff
@@ -0,0 +1,32 @@
+name: Compile GitHub Pages
+description: Compiles content for GitHub Pages and uploads it to the artifact cache
+
+inputs:
+  pr-number:
+    type: number
+  repository:
+    default: 'solana-web3.js'
+    type: string
```
mcintyre94, line 9: I think it'd be useful to add a comment here that the inputs are only used for the `REACT_EXAMPLE_APP_BASE_PATH` and therefore for new web3js

It took me a while to figure out what's going on when the legacy job calls this without any inputs

.github/workflows/actions/compile-gh-pages/action.yml
```diff
@@ -0,0 +1,32 @@
...
+      run: |
+        mkdir -p .ghpages-deploy
+        mv ./examples/react-app/dist/ .ghpages-deploy/example/
+        mv ./docs/* .ghpages-deploy/
+
+    - uses: actions/upload-artifact@v4
+      with:
+        include-hidden-files: true
+        name: ghpages-deploy-artifacts
+        path: .ghpages-deploy
```
mcintyre94, line 31: Probably just me not knowing stuff, but I'd find a comment here useful that this is uploading for a later deploy to github-pages. Rather than `actions/upload-artifact` actually being an upload to github pages which is what I first thought was going on.

mcintyre94 (approved):
Looks good! This took a while to understand, but makes sense

The only thing I'm not sure about is whether we want to build this `/v1.x` path on every preview, but I figure not doing that probably just makes things more complicated for little gain?

.github/workflows/actions/compile-gh-pages/action.yml
steveluscher: Because all of these CI workflows are stateless, the way that they assemble the GitHub Pages site is by uploading them to a temporary ‘artifacts cache’ that only needs to live for the duration of the CI run. 1 day is actually too long.

PR #3614, opened 2024-11-22
Add failover RPC transport example

kox:
Adding a basic fail-over example to abstract a list of RPC transport providers. It contains a delay in between fail-overs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,7 +304,43 @@ const rpc = createSolanaRpcFromTransport(retryingTransport);
 Support for handling network failures can be implemented in the transport itself. Here’s an example of some failover logic integrated into a transport:
 
 ```ts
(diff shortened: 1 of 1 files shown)


steveluscher:
Love it! Why the sleep though?

steveluscher (approved):

kox:
> Love it! Why the sleep though?

You are right. Considering that it's a failover, the user likely doesn't want to wait at all. 

Should I include this change in the PR?

steveluscher:
Yeah, sure; tear it out and we'll get this merged.

kox:
> Yeah, sure; tear it out and we'll get this merged.

Brilliant.! I sent it. 

Thanks in advance

Issue #1969, opened 2023-12-14
Connection.getBlock return types are missing the blockHeight field


jeff-delonge:
## Overview

In 1.87.6, the return types of  `Connection.getBlock` such as `VersionedBlockResponse` and `BlockResponse` omit the `blockHeight` field. If `Connection.getBlock` returns a non-null value, the returned object includes this field.

The types should be updated to include the `blockHeight` field.

## Steps to reproduce

```ts
    const connection = new Connection("https://api.mainnet-beta.solana.com", "finalized");
    const latestBlockhashResult = await connection.getLatestBlockhashAndContext("finalized");
    const block = await connection.getBlock(latestBlockhashResult.context.slot, {
        commitment: "finalized",
        maxSupportedTransactionVersion: undefined,
        rewards: false,
        transactionDetails: "none",
    });
    console.log("latestBlockhashResult", latestBlockhashResult);
    console.log("block", block);
```

## Description of bug

The above logs something like the following. Note that the `block` includes the `blockHeight` field.
```
latestBlockhashResult {
  context: { apiVersion: '1.16.18', slot: 235912008 },
  value: {
    blockhash: 'D7Qr3rwEbYKh9cYvV25wmXCcvzJqHxixnZs91tL1Rvsa',
    lastValidBlockHeight: 217323342
  }
}
block {
  blockHeight: 217323042,
  blockTime: 1702587441,
  blockhash: 'D7Qr3rwEbYKh9cYvV25wmXCcvzJqHxixnZs91tL1Rvsa',
  parentSlot: 235912007,
  previousBlockhash: '7vXphDU9znbs6SNBqiyPSEkvJtiZE5D5g55mxw5NKSZk',
  rewards: undefined
}
```

If instead I try to build the following code (note the `const blockHeight = block?.blockHeight`), TypeScript will complain because the `blockHeight` field is missing from the `VersionedBlockResponse` type in @solana/web3.js:
```ts
    const connection = new Connection("https://api.mainnet-beta.solana.com", "finalized");
    const latestBlockhashResult = await connection.getLatestBlockhashAndContext("finalized");
    const block = await connection.getBlock(latestBlockhashResult.context.slot, {
        commitment: "finalized",
        maxSupportedTransactionVersion: undefined,
        rewards: false,
        transactionDetails: "none",
    });
    const blockHeight = block?.blockHeight;
    console.log("latestBlockhashResult", latestBlockhashResult);
    console.log("block", block);
```

[...]

PR #3592, opened 2024-11-18
Added programId field in TokenBalance type

shivaji43:
Tackles the issue (#3091) 
Screenshot 2024-11-19 234833
Screenshot 2024-11-19 233507
Screenshot 2024-11-19 233442

Done proper linting , built and complied properly , also tested to see if programId returns as expected ,
also added the screenshots

Please tell me if I am missing something and Thank you for being patient and helping me through this.

PR #3625, opened 2024-11-25
added blockheight to BlockResponse and VersionedBlockResponse

shivaji43:
Screenshot 2024-11-25 172002
Screenshot 2024-11-25 172045
Screenshot 2024-11-25 172100

added blockheight to BlockResponse and VersionedBlockResponse
linted and tested , the error is now gone because the property is added in the return type , solving #1969


shivaji43:
@steveluscher please take a look

steveluscher (changes requested):
This change will not cause the library to throw if the RPC sends a `blockHeight` that is neither `null`, nor `undefined`, nor `number`. You have to also implement the `superstruct` validations, just like you did in #3592.

PR #3632, opened 2024-11-26
Change IInstruction to use ReadonlyUint8Array for data

mcintyre94:
This PR modifies the `data` field of `IInstruction` to use `ReadonlyUint8Array`. This means that you can create an `IInstruction` using bytes output from an encoder as input.

### DIFF
--- a/.changeset/spicy-parrots-raise.md
+++ b/.changeset/spicy-parrots-raise.md
@@ -1,0 +1,7 @@
+---
+'@solana/transaction-messages': patch
+'@solana/instructions': patch
+'@solana/errors': patch
+---
+
+Change `data` field of transaction message instructions to use `ReadonlyUint8Array`
--- a/examples/deserialize-transaction/src/example.ts
+++ b/examples/deserialize-transaction/src/example.ts
@@ -381,7 +381,13 @@ if (identifiedInstruction === SystemInstruction.TransferSol) {
     log.info('[step 4] The first instruction calls the TransferSol instruction of the system program');
     // Narrow the type again, the instruction must have accounts to be parsed as a transfer SOL instruction
     assertIsInstructionWithAccounts(firstInstruction);
-    const parsedFirstInstruction = parseTransferSolInstruction(firstInstruction);
+
+    // TODO: This can just be `parseTransferSolInstruction(firstInstruction)` when the client is updated
+    // with the `@solana/web3.js` version that changes the instruction data type to `ReadonlyUint8Array`
+    const parsedFirstInstruction = parseTransferSolInstruction({
+        ...firstInstruction,
+        data: firstInstruction.data as unknown as Uint8Array,
+    });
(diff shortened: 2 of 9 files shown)


examples/deserialize-transaction/src/example.ts
```diff
@@ -381,7 +381,13 @@ if (identifiedInstruction === SystemInstruction.TransferSol) {
...
     // Narrow the type again, the instruction must have accounts to be parsed as a transfer SOL instruction
     assertIsInstructionWithAccounts(firstInstruction);
-    const parsedFirstInstruction = parseTransferSolInstruction(firstInstruction);
+
+    // TODO: This can just be `parseTransferSolInstruction(firstInstruction)` when the client is updated
+    // with the `@solana/web3.js` version that changes the instruction data type to `ReadonlyUint8Array`
+    const parsedFirstInstruction = parseTransferSolInstruction({
+        ...firstInstruction,
+        data: firstInstruction.data as unknown as Uint8Array,
+    });
```
mcintyre94, line 390: We'll want to create a new build of the Codama JS clients with this change to make this go away.

This was actually where I first hit this issue. I was using an API that returned instruction data as base58, and then trying to convert its output into an `IInstruction` that I could parse with Codama. Since instruction data is `Uint8Array` this required me to cast my `ReadonlyUint8Array` from encoding the API's base58 data to a `Uint8Array` like this. I figured it makes sense to just widen the type on `IInstruction` as in this PR

.changeset/spicy-parrots-raise.md
```diff
@@ -0,0 +1,7 @@
+---
+'@solana/transaction-messages': patch
+'@solana/instructions': patch
+'@solana/errors': patch
```
mcintyre94, line 4: Patch is debatable. I'm widening the type of `IInstruction`, so any code that creates one of those is unaffected, but any code that treats it as mutable is not. In our code I needed to update some type params. Downstream consumers like Codama clients need to be updated to avoid any code changes for their dependencies. Happy to change this to whatever we think is best.

steveluscher (approved):
Added via Giphy

mcintyre94:
I've pushed a commit that refactors `IInstructionWithData` to have an unrestricted data type, and not extend `IInstruction`. This means that you can define a type like `IInstruction & IInstructionWithData<MyParsedInstructionData>` which might be useful

I noticed the type felt too restrictive because as-is we can't change Codama to use `IInstructionWithData<ReadonlyUint8Array>` before merging this PR - since it's tied to `Uint8Array`. This PR would already loosen it to `ReadonlyUint8Array` but it might be useful to be able to loosen the type of data further

It also better matches what we do in eg `TransactionMessageWithBlockhashLifetime` which doesn't extend `TransactionMessage`

Happy to revert if this is too much of a breaking change to justify though!

lorisleiva (approved):
Thanks! As we discussed on Slack, let's merge this first and then update the Codama renderer before coordinating the release of the clients.

Issue #3465, opened 2024-10-26
Re-enable `@typescript-eslint` rules one at a time


steveluscher:
Create one PR for each of these rules re-enabled:

https://github.com/solana-labs/solana-web3.js/blob/612a4d8688fc2a481d24ec7f415701c6aa92c707/eslint.config.mjs#L9-L20

https://github.com/solana-labs/solana-web3.js/blob/612a4d8688fc2a481d24ec7f415701c6aa92c707/eslint.config.mjs#L9-L20

Chirag-S-Kotian:
assign this to me.

steveluscher:
Hi @Chirag-S-Kotian – thanks for wanting to contribute! Feel free to start sending PRs; there's no need to have a single assignee here because there's not enough competition for this work. We appreciate you!

PR #3691, opened 2024-12-06
Enable rule @typescript-eslint/restrict-template-expressions and fix the only error

kox:
This PR intends to fix a small part of issue #3465 

Specifically, it changes the rule @typescript-eslint/restrict-template-expressions, which is currently set to off, to error.


steveluscher (approved):
Very nice find. `getErrorMessage` returns a `ReactNode` ergo isn't safe to interpolate into a string.
nice_catch-1838825837

kox:
> Very nice find. `getErrorMessage` returns a `ReactNode` ergo isn't safe to interpolate into a string.

Thanks for merging, that was fast!

PR #3694, opened 2024-12-08
Enable eslint restrict-template-expressions plugin has a few extra linting errors

kox:
While I was adding an extra eslint rule, I realized that there were a few extra linting errors from restrict-template-expressions rule. 

One of them I had to comment it because when it gets stringified, it leaks relevant information. Probably it should be sanitized first.

### DIFF
--- a/examples/deserialize-transaction/src/example.ts
+++ b/examples/deserialize-transaction/src/example.ts
@@ -310,7 +310,7 @@ const decompiledTransactionMessage = await decompileTransactionMessageFetchingLo
 // This is the same data structure that was created before we first signed the transaction
 
 // We can see the fee payer:
-log.info(`[step 3] The transaction fee payer is ${decompiledTransactionMessage.feePayer}`);
+log.info(`[step 3] The transaction fee payer is ${decompiledTransactionMessage.feePayer.address}`);
 
 // And the lifetime constraint:
 log.info(decompiledTransactionMessage.lifetimeConstraint, '[step 3] The transaction lifetime constraint');
--- a/packages/errors/src/message-formatter.ts
+++ b/packages/errors/src/message-formatter.ts
@@ -28,6 +28,7 @@ export function getHumanReadableErrorMessage<TErrorCode extends SolanaErrorCode>
             const variableName = messageFormatString.slice(state[START_INDEX] + 1, endIndex);
 
             fragments.push(
+                // eslint-disable-next-line @typescript-eslint/restrict-template-expressions 
                 variableName in context ? `${context[variableName as keyof typeof context]}` : `$${variableName}`,
             );
         } else if (state[TYPE] === StateType.Text) {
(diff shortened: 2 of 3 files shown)


examples/deserialize-transaction/src/example.ts
```diff
@@ -310,7 +310,7 @@ const decompiledTransactionMessage = await decompileTransactionMessageFetchingLo
 // This is the same data structure that was created before we first signed the transaction
 
 // We can see the fee payer:
-log.info(`[step 3] The transaction fee payer is ${decompiledTransactionMessage.feePayer}`);
+log.info(`[step 3] The transaction fee payer is ${decompiledTransactionMessage.feePayer.address}`);
```
steveluscher, line 313: Nice.

packages/rpc-transport-http/src/__tests__/http-transport-test.ts
```diff
@@ -71,6 +71,9 @@ describe('createHttpTransport', () => {
                 thrownError = e as SolanaError<typeof SOLANA_ERROR__RPC__TRANSPORT_HTTP_ERROR>;
             }
             expect(thrownError).toBeDefined();
+            
+            // Note: Stringifying this variable, it will leak header values and the test will fail
+            // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
```
steveluscher, line 76: > One of them I had to comment it because when it gets stringified, it leaks relevant information

That's the right thing to do though, because this is a test that tests exactly that, right?

packages/errors/src/message-formatter.ts
```diff
@@ -28,7 +28,7 @@ export function getHumanReadableErrorMessage<TErrorCode extends SolanaErrorCode>
             const variableName = messageFormatString.slice(state[START_INDEX] + 1, endIndex);
 
             fragments.push(
-                variableName in context ? `${context[variableName as keyof typeof context]}` : `$${variableName}`,
+                variableName in context ? `${context[variableName] as string}` : `$${variableName}`,
```
steveluscher, line 31: I'm not sure this is true though, the value at `context[variableName]` might not be a string, as it is not here:

https://github.com/solana-labs/solana-web3.js/blob/master/packages/errors/src/context.ts#L493-L495

I think this is an appropriate place to suppress lint again. The suppression expresses ‘yeah, we know some of these will come up `[object Object]` but that's OK.’

steveluscher (changes requested):

packages/rpc-transport-http/src/__tests__/http-transport-test.ts
kox: Ok, I guess it makes sense. I was a bit confused because a proper stringify would render the protected data.

packages/errors/src/message-formatter.ts
kox: Thanks for the clarification, I didn't realize that use case. I rolled the changes and I just added the suppression.

mcintyre94 (approved):

steveluscher (approved):

kox:
thanks for your help

PR #3695, opened 2024-12-09
feat: add warning message for loading module from CDN directly

loatheb:
For web3 projects to load modules, it is very dangerous to load modules from CDN directly using tags without using build tools. 

Such problems have occurred in the past with ledger and `solana/web3.js`.As a result, once the package is tampered with, even if the project is not updated, it will be affected immediately. 

Therefore, mark it clearly in the document and do not use tag to load modules directly from CDN.


steveluscher:

PR #3703, opened 2024-12-10
Issue 2978: Verify signature for address helper

LunarFang416:
Closes solana-labs/web3.js-issue-conveyer#11 

Added documentation for the verifySignatureForAddress helper function, which simplifies the process of verifying digital signatures for a given address. The function internally converts the address to a public key and verifies the signature for the associated data.

### DIFF
--- a/packages/keys/README.md
+++ b/packages/keys/README.md
@@ -179,3 +179,21 @@ if (!(await verifySignature(publicKey, signature, data))) {
     throw new Error('The data were *not* signed by the private key associated with `publicKey`');
 }
 ```
+
+### `verifySignatureForAddress()`
+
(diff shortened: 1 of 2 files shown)


mcintyre94 (changes requested):
Thanks for taking a look at this, but I agree with Steve's comment on the issue here: https://github.com/solana-labs/solana-web3.js/issues/2978#issuecomment-2436528156

I don't think we want to implement this with this signature, because constructing the public key is relatively expensive and we don't want to encourage code that does that repeatedly.

Could you change this to just take `address: string` and return `Promise<CryptoKey>`? The actual verification would live outside. So this could just be `createPublicKeyForAddress` or something like that.

In addition to land this PR we would need unit tests for the new function

It'd also be nice to update the `deserialize-transaction` example to use this function since that motivated the issue and it'd demonstrate its usage nicely.

steveluscher:

PR #3710, opened 2024-12-11
Enable eslint rule and disable errors due on data flexibility

kox:
Following solana-labs/web3.js-issue-conveyer#4 , this PR intends to activate the eslint rule @typescript-eslint/restrict-plus-operands.

The few discovered errors have been disabled to support possible data values different than strings.


steveluscher:

PR #3719, opened 2025-01-07
Fix rent epoch type

backmeupplz:
So, for some reason rpc's return `2 ^ 64 - 1` as `rentEpoch`. Obviously, this is larger than `Number` in JS, so I switched the types to `bigint` and it all works now.

I'm a newby in Solana and have no idea why this works.

See https://github.com/solana-labs/solana-program-library/issues/7635 for the reference.

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -10,6 +10,7 @@ import {Agent as NodeHttpsAgent} from 'https';
 import {
   type as pick,
   number,
+  bigint,
   string,
   array,
   boolean,
@@ -1930,7 +1931,7 @@ const AccountInfoResult = pick({
   owner: PublicKeyFromString,
   lamports: number(),
   data: BufferFromRawAccountData,
-  rentEpoch: number(),
+  rentEpoch: bigint(),
 });
 
 /**


backmeupplz:
https://github.com/solana-labs/solana-web3.js/pull/3133 is relevant to this discussion too

backmeupplz:
I can see that you guys are working on the new version of the package, but when the mainnet RPC returns a `bigint` for one of the account fields, it breaks everything and you can't even use `getOrCreateAssociatedTokenAccount` 💀

backmeupplz:
Maybe other `number` checks for `rentEpoch` should also be changed to `bigint` though

buffalojoec (commented):
@backmeupplz good spot. This is because the default value for `rentEpoch` is now being set to Rust's `u64::MAX`, as the concept of rent-paying accounts is being phased out.

This change causes actual valid JS numbers to break the struct validation, though. I think what we really want is this:

```js
coerce(bigint(), number(), (value) => BigInt(value))
```

cc @steveluscher for review and to determine release process. Changing the type is a breaking change afaict, so can we sneak this into the next minor?

backmeupplz:
@buffalojoec imo whether it breaks the implementation or not depends on the types of values passed to `rentEpoch`

from what i've seen it's only ever been an int, so switching it all to bigint might be ok

however, again, this is debatable as the feature is being phased out

the thing that bugs me is that like... am i the first one to notice? 🥲 this literally breaks types in `spl-token` package making it virtually unusable (as `getAccountInfo` is broken)... is no one else using it? 😰

PR #3730, opened 2025-04-22
fix: update `base-x` to 3.0.11

steveluscher:


steveluscher:
This is dead in the water. Overrides do not cascade transitively. They would have to be employed by the root project (ie. the people who depend on web3.js).

PR #3754, opened 2025-10-26
Bump bs58 to latest to resolve unmet peer dep issues in other solana libs

BytesOfMan:
Because other sol libs have upgraded their bs58 dep to 6.0.0, when they are combined with this lib it creates dependency tree issues.

Bump to the latest version to match what other so libs use.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,17986 @@
+{
+  "name": "@solana/web3.js",
+  "version": "1.0.0-maintenance",
+  "lockfileVersion": 3,
+  "requires": true,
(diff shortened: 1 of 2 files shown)


BytesOfMan:
#3755

BytesOfMan:
still an issue

BytesOfMan:
is there some alt way to get this fixed? lots of use cases require multiple sol libs, they should all use the same bs58 dep

BytesOfMan:
sol-dep-fix-pls

BytesOfMan:
...

PR #3806, opened 2026-05-01
V3 RC0

dhl:


src/connection.ts
```diff
@@ -1516,297 +1659,281 @@ export type BlockSignatures = {
...
-      });
-    });
+function createKitRpcClient(url: string, config: RpcTransportConfig) {
+  const typedTransport = createRpcTransport(url, config);
+  return {
+    typedRpc: createRpc({
+      api: createSolanaRpcApi({
+        ...DEFAULT_RPC_CONFIG,
+        // Match Kit's client-side default commitment when unspecified.
+        defaultCommitment: 'confirmed',
```
amilz, line 1860: I think `defaultCommitment needs to come from the `config` rather than hard-coded?
if not, shouldn't we just leverage `DEFAULT_RPC_CONFIG` which already includes `defaultCommitment: 'confirmed'`?

src/connection.ts
```diff
@@ -3888,22 +3574,57 @@ export class Connection {
     });
   }
 
+  private _resolveCommitment(requestedCommitment?: Commitment): Commitment {
+    return requestedCommitment ?? this._commitment ?? 'confirmed';
```
amilz, line 3584: I think this probably should return `undefined` when the caller didn't specify and the connection has no default. Then Kit's defaults do the work.

src/connection.ts
```diff
@@ -1849,825 +1966,347 @@ export type TokenAccountBalancePair = {
...
-  /** Currently processing slot */
-  slot: number;
-  /** Parent of the current slot */
-  parent: number;
-  /** The root block of the current slot's fork */
-  root: number;
-};
+function getLegacyTransactionConfirmationTimeoutMs(
+  initialTimeoutMs: number | undefined,
+  commitment: Commitment | undefined,
```
amilz:
```suggestion
  commitment?: Commitment,
```

src/connection.ts
```diff
@@ -1849,825 +1966,347 @@ export type TokenAccountBalancePair = {
...
+): number {
+  switch (commitment) {
+    case 'processed':
+    case 'confirmed':
+      return initialTimeoutMs || 30 * 1000;
+    case 'finalized':
+    case undefined:
+      return initialTimeoutMs || 60 * 1000;
```
amilz, line 2308: some of our other fucntions force undefined to 'confirmed', so i dont' think we'd ever hit this which might trigger a slight regression.

src/connection.ts
```diff
@@ -4532,178 +4438,125 @@ export class Connection {
...
+          : this._typedRpc.getMinimumBalanceForRentExemption(rpcDataLength, {
+              commitment: rpcCommitment,
+            })
+      ).send();
+    } catch (_error) {
       console.warn('Unable to fetch minimum balance for rent exemption');
-      return 0;
+      return 0n as ReturnType<
+        GetMinimumBalanceForRentExemptionApi['getMinimumBalanceForRentExemption']
+      >;
```
amilz, line 4483: Looks like this is just maintaining status quo, but is there any good reason to keep this?
IMO better to just do

```ts
`throwSolanaRpcErrorIfNeeded(_error, 'failed to get minimum balance for rent exemption')`
```

src/address.ts
amilz, line 1: Should we leverage more functions from kit's `packages/addresses/src/program-derived-address.ts` ?
- getProgramDerivedAddress
- createProgramDerivedAddress 
etc

src/rpc-subscriptions/controller.ts
```diff
@@ -0,0 +1,388 @@
...
+                hash,
+                subscriptionHandle.serverSubscriptionId,
+              );
+            } catch (error) {
+              if (shouldAbortSubscriptionUpdate()) {
+                return;
+              }
+              this._subscriptionRegistry.setSubscription(hash, {
+                ...subscription,
+                state: 'pending',
```
amilz: Since we don't offer any additional logic on 'pending' state -- are there risks of inifinite loop on certain error types?

src/rpc-subscriptions/registry.ts
```diff
@@ -0,0 +1,385 @@
...
+    const stateChangeCallbacks =
+      this._stateChangeCallbacksByHash.get(hash) ?? new Set();
+    this._stateChangeCallbacksByHash.set(hash, stateChangeCallbacks);
+    stateChangeCallbacks.add(callback);
+    const currentState = this._subscriptionsByHash.get(hash)?.state;
+    if (currentState !== undefined) {
+      try {
+        callback(currentState);
+      } catch {
+        // Ignore observer errors so registry updates continue.
```
amilz: should we at least `console.error` here and 377 in case useful issue to user?

src/connection.ts
```diff
@@ -1,215 +1,463 @@
...
+ */
+export type Logs = {
+  err: TransactionError | null;
+  logs: string[];
+  signature: string;
+};
+
+function toKitAddress(address: Address): KitAddress {
+  return address.toBase58() as unknown as KitAddress;
+}
```
amilz: same as `src/kit-adapters/address.ts:5`

src/address.ts
```diff
@@ -0,0 +1,396 @@
...
+  'Invalid seeds, address must fall off the curve';
+const ERROR__FAILED_TO_FIND_VIABLE_PROGRAM_ADDRESS_NONCE =
+  'Unable to find a viable program address nonce';
+const ADDRESS_CODEC = getAddressCodec();
+const PDA_MARKER_BYTES = new TextEncoder().encode('ProgramDerivedAddress');
+
+/**
+ * A Solana address
+ */
+export class Address {
```
amilz, line 57: I think rather than having `KitAddress` and `Address` (web3.js Address, a new class), it'd make more sense to just have `Address` (for Kit which it already is) and introduce a new `AddressClass` or `Web3Address` or something like that. 
That way `Address` means the same thing to everybody.

src/connection.ts
```diff
@@ -4735,167 +4606,166 @@ export class Connection {
...
-    const args = this._buildArgsAtLeastConfirmed(
-      [slot],
-      commitment as Finality,
-      undefined /* encoding */,
+    const finality = this._resolveSupportedFinality(commitment);
+    const rpcSlot = coerceNumericToBigInt(slot, 'slot');
+    const fullConfig = buildTypedFullBlockConfig(
+      finality,
       config,
+      rawConfig != null,
```
amilz, line 4765: When callers omit the config object, `rawConfig != null` is false and `fullConfig` becomes `undefined`; Kit then applies its own `confirmed` default. A `Connection` constructed with default `finalized` therefore sends `confirmed` for `getBlock(slot)` unless the caller passes any config object.

src/connection.ts
```diff
@@ -5768,155 +5574,207 @@ export class Connection {
...
-    }
-
-    if (
-      configOrSigners &&
-      typeof configOrSigners === 'object' &&
-      'innerInstructions' in configOrSigners
-    ) {
-      config.innerInstructions = configOrSigners.innerInstructions;
-    }
+    } catch (error) {
```
amilz, line 5758: check out `unwrapSimulationError`
https://github.com/anza-xyz/kit/pull/1230

src/kit-adapters/response.ts
```diff
@@ -0,0 +1,993 @@
...
+
+function mapRpcParsedInstruction(
+  instruction:
+    | RawParsedMessageInstruction
+    | RawParsedInnerInstruction
+    | ParsedInstruction
+    | PartiallyDecodedInstruction,
+): ParsedInstruction | PartiallyDecodedInstruction {
+  if ('parsed' in instruction) {
+    return {
```
amilz, line 651: is `stackHeight` available here? if so should we include when present?

src/transaction/legacy.ts
amilz, line 388: Can we include `KitInstructions` in the `add()` function. Should be fairly straightforward: 

```ts
else if (isKitInstruction(item)) {
        this.instructions.push(fromKitInstruction(item));
```

This would allow teams to leverage existing Codama client renderers immediately. 


Some instruction assertions in Kit at `packages/instructions/src/instruction.ts` or a more generic one in my original commit below.

Ref commit: b42e7fa

src/keypair.ts
amilz, line 1: WDYT about actually migrating this to a `KeyPairSigner` in kit so we can get signing rails to be the same. Already 90% there. Here's an example where I experimented w/ this before:
https://github.com/amilz/solana-kit/blob/feat/add-legacy/packages/legacy/src/keypair.ts

package.json
```diff
@@ -53,78 +51,77 @@
...
+    "@eslint/js": "^9.39.2",
+    "@rollup/plugin-node-resolve": "^16.0.3",
+    "@rollup/plugin-replace": "^6.0.3",
+    "@rollup/plugin-terser": "^1.0.0",
+    "@solana-program/address-lookup-table": "^0.11.0",
+    "@solana-program/compute-budget": "^0.15.0",
+    "@solana-program/stake": "^0.6.0",
+    "@solana-program/system": "^0.12.0",
     "@types/chai-as-promised": "^8.0.0",
     "@types/express-serve-static-core": "^5.0.0",
```
amilz: `@types/express-serve-static-core` and `tslib` are not used, i don't think

src/kit-adapters/instruction.ts
amilz, line 1: I didn't see new unit tests for these helpers. If we don't already have them, plz add (e.g. 137b177)

package.json
```diff
@@ -1,13 +1,13 @@
 {
   "name": "@solana/web3.js",
-  "version": "1.0.0-maintenance",
+  "version": "3.0.0-rc.0",
   "description": "Solana Javascript API",
   "keywords": [
     "api",
```
amilz, line 8:
```suggestion
    "solana",
    "web3",
    "api",
```

amilz (commented):
Looking good. Here's a partial reiview. Will dig in again tomorrow.

package.json
dhl: Resolved in 057329becb4db4564dc9d35b50d826eba1fb4451

src/kit-adapters/instruction.ts
dhl: Resolved in f4a6e15598b5580470d93bbbe789588cd03ede94 and af85201ee9e1714f1ec0461edf6b90266d77605a.

The helpers are now treated as internal adapters, and should not be published as a separate package/bundle.

src/keypair.ts
dhl: It was considered and tried. The idea was abandoned as there is significant friction between KeyPairSigner and Keypair.

The current Keypair implementation is implemented only with functions from @solana/keys, and is simpler and shorter.

src/transaction/legacy.ts
dhl: Added in a17d997b0944d58bec9d8df15ebec6c7eb097373!

The ref commit could not be used, as it does always correctly identify a Kit transaction.

src/address.ts
dhl: The PublicKey rename follows what happened on Rust side.

The name clash already exists today: AccountMeta, Transaction, etc. Context should make things clear.

amilz: > The PublicKey rename follows what happened on Rust side.

To clarify, I'm not pushing back on `PublicKey` going away ‚ only on which thing gets named `Address`.

Rust's `Address` is a struct with methods, so the web3js class is the closer Rust analog, agreed. But @solana/address diverged: in TS, `Address` is a branded `string`, not a class. Since this lib/PR is the bridge into kit, I'd weight TS-ecosystem alignment over Rust parity. We have helpers, codecs, etc. that use this name: `address()`, `isAddress()`, `assertIsAddress()`, `getAddressCodec()`, etc.

The longer-term framing matters more to me though: if the goal is moving folks to kit, the end state is `Address` = kit's branded string. If the v3 class takes that name now, users will ultimately have migrated Publickey->Address then Address-> other Address . IMO that second jump just risks additional confusion we could avoid here.
With a distinct class name (`Web3Address` / `AddressClass` / `ExpandedAddress`), it's a one-way rename ‚ codemod-friendly, no in-flight collisions, and the migration ends with one canonical `Address` everywhere. Framed another way: the class is a kit `Address` plus methods, so a wrapper-style name reflects the actual relationship.

> The name clash already exists today: AccountMeta, Transaction, etc.

Fair, but those clash on similar shapes and not being renamed in this PR. `Address` is a sharper fork (class-with-methods vs. branded string), and we're already renaming it in this PR.

dhl: I appreciate you raising the name collision concern.

After testing how TypeScript handles both assignment directions and considering where these collisions are likely to happen, I'm confident that it's manageable — and the evidence shows the issue is smaller than it first appears.

Here's a concrete example:

```typescript
import { Keypair, Address as Web3Address } from "@solana/web3.js";
import { address, type Address as KitAddress } from "@solana/kit";

const keypair = await Keypair.generate();
const web3Address = new Web3Address(keypair.publicKey);
const kitAddress = address(web3Address.toBase58());

printWeb3Address(kitAddress);   // Error! Passing Kit Address to Web3.js param
printKitAddress(web3Address);   // Error! Passing Web3.js Address to Kit param
```

When passing a Web3.js `Address` where a Kit `Address` is expected, TypeScript already produces a clear error:

```typescript
Argument of type 'Address<string>' is not assignable to parameter of type 'Address'.
  Type 'NominalType<"brand", "Address"> & NominalType<"stringEncoding", "base58"> & String' 
  is missing the following properties from type 'Address': _publicKeyBytes, equals, toBase58, toJSON, and 4 more.
```

We can make this even clearer by adding a marker to the Web3.js `Address` class:

```typescript
Argument of type 'Address<string>' is not assignable to parameter of type 'Address'.
  Type 'NominalType<"brand", "Address"> & ... & String' 
  is missing the following properties from type 'Address': 
    "Expected Web3.js Address; convert Kit address with new Address(kitAddress).", 
    _publicKeyBytes, equals, toBase58, and 5 more.
```

When passing a Kit `Address` where a Web3.js `Address` is expected, the error is even clearer — it explicitly shows the full import paths and the missing Kit brand:

```typescript
Argument of type 'import("/[fs path]/node_modules/@solana/web3.js/lib/index").Address' 
is not assignable to parameter of type 'import("/[fs path]/node_modules/@solana/addresses/dist/types/address").Address'.
  Property '"__brand:@solana/kit"' is missing in type 'Address' but required in type 'NominalType<"brand", "Address">'.
```

In practice, only Web3.js users are likely to encounter collisions, and the context of where these types are used also makes it very clear which `Address` is intended.

We're going with the rename to `Address` because:

- It aligns with the canonical name chosen by the protocol team.
- It gives existing Web3.js users the clean, expected one-step rename they deserve.
- It plays to the Web3.js' class-based strength. The address validation is preserved in runtime for Web3.js `Address`, but erased for Kit `Address`. Type check for the class-based approach stronger and order-of-magnitude faster.

To make the experience excellent despite the shared name, we're adding clearer runtime output by setting `Symbol.toStringTag` of Web3.js `Address` to "Address (Web3.js)".

I believe this is the right call for Web3.js users and the broader ecosystem, where we are united by the shared identity in `Address`, and embrace the diversity in behavior that fits each SDK's design and strength.

src/address.ts
dhl: Let's do that. I tried not to modify the original logic unnecessarily to preserve errors and reduce churn, but perhaps that is not so bad.

createProgramDerivedAddress is not exposed publicly in Kit, however. Will open a PR in Kit for that.

We could migrate over to createProgramDerivedAddress if it's actually exposed.

dhl: Resolved in 1bce566dd1fc8e0d3dda830fa32ca214b5b94c67

src/connection.ts
dhl: Resolved in 0a3836cd50bc174be225b9ae234593fb181b55f6

src/keypair.ts
amilz: WDYT about adding a `toKitSigner` method that calls `createSignerFromKeyPair(this.#keypair)` from `@solana/signers` ?

src/address.ts
```diff
@@ -0,0 +1,405 @@
...
+
+  /**
+   * Check that a pubkey is on the ed25519 curve.
+   */
+  static isOnCurve(addressData: AddressInitData): boolean {
+    const address = new Address(addressData);
+    return isOnCurve(address.toBytes());
+  }
+}
+
```
amilz, line 309:
```suggestion

/**
 * @deprecated Use {@link Address} instead. `PublicKey` is kept as an alias for
 * backwards compatibility and will be removed in a future major version.
 */
export const PublicKey = Address;

/**
 * @deprecated Use {@link Address} instead.
 */
export type PublicKey = Address;

/**
 * @deprecated Use {@link AddressInitData} instead.
 */
 
```

src/rpc-subscriptions/registry.ts
dhl: Resolved in 5314b318852062978d335ddeab6407314743919e

src/address.ts
dhl: The alias is defined in `src/publickey.ts` https://github.com/blueshift-gg/solana-web3.js/blob/0a3836cd50bc174be225b9ae234593fb181b55f6/src/publickey.ts

src/keypair.ts
```diff
@@ -1,102 +1,117 @@
...
   /**
    * The public key for this keypair
    *
-   * @returns {PublicKey} PublicKey
+   * @returns {Address} Address
    */
-  get publicKey(): PublicKey {
-    return new PublicKey(this._keypair.publicKey);
+  get publicKey(): Address {
+    return new Address(this.#publicKeyBytes);
```
amilz, line 90: since we are using the `Address` naming, wdyt about adding a deprecate warning here and adding a `get address(): Address` getter so folks can call (`(await Keypair.generate()).address`).

Also, WDYT about adding a `toKitSigner` method that calls `createSignerFromKeyPair(this.#keypair)` from `@solana/signers` ?

src/rpc-subscriptions/controller.ts
dhl: Good catch. We preserved that behavior from v1, but it's worth changing.

Resolved in 5fd4d4973c2240a411a645504d80d52787b43afb.

src/keypair.ts
dhl: I think that's a great idea.

I'm wondering if it should be done now or we can minimize churn and do later.

Keypair.address does sound better tbh

dhl: `address` getter resolved in 195e7fb6bac5b7b0cb6d0f1a2cfb581c72019dee

Instead of adding to the classic interface, perhaps `toKitSigner` should be a utility function. If there is enough demand for it, maybe we could then move that into the class interface at some point.

src/address.ts
```diff
@@ -0,0 +1,405 @@
...
+   * Return the base-58 representation of the public key
+   */
+  toBase58(): string {
+    return ADDRESS_CODEC.decode(this._publicKeyBytes);
+  }
+
+  toJSON(): string {
+    return this.toBase58();
+  }
+
```
amilz, line 116: Since we are using `ADDRESS_CODEC.decode`, we are already getting Kit type benefits here, but we lose it by setting our return as a `string`.  Can we bump return type to `KitAddress`? Benefit here is cleaner interop w/ Kit clients/ix builders, e.g.:

```ts
// Before
  import { address } from '@solana/kit';
  const kitIx = getTransferSolInstruction({
    amount: 0.02 * LAMPORTS_PER_SOL,
    destination: address(recipientB.publicKey.toBase58()),
    source
  });

// After
  const kitIx = getTransferSolInstruction({
    amount: 0.02 * LAMPORTS_PER_SOL,
    destination: recipientB.publicKey.toBase58(),
    source
  });
```

```suggestion
  /**
   * Return the base-58 representation of the public key
   */
  toBase58(): KitAddress {
    return ADDRESS_CODEC.decode(this._publicKeyBytes);
  }

  toJSON(): KitAddress {
    return this.toBase58();
  }
```

After doing this, we can probably remove the cast `to KitAddress` in our kit-adapter.

src/address.ts
```diff
@@ -0,0 +1,405 @@
...
+  get [Symbol.toStringTag](): string {
+    return `Address(${this.toString()})`;
+  }
+
+  /**
+   * Return the base-58 representation of the public key
+   */
+  toString(): string {
```
amilz, line 187: same
```suggestion
  toString(): KitAddress {
```

amilz:
Should we just use this in src/blockhash.ts

```ts
export type { Blockhash } from "@solana/rpc-types";
```

src/message/legacy.ts
```diff
@@ -41,15 +78,15 @@ export type MessageArgs = {
...
   /** The hash of a recent ledger block */
   recentBlockhash: Blockhash;
   /** Instructions that will be executed in sequence and committed in one atomic transaction if all succeed. */
   instructions: CompiledInstruction[];
 };
 
 export type CompileLegacyArgs = {
-  payerKey: PublicKey;
+  payerKey: Address;
   instructions: Array<TransactionInstruction>;
```
amilz: I'm loving this functionality! `const tx = new Transaction().add(legacyIx, kitIx);`

Missed this earlier, but can we expand message and v0 txn to include the same support for `KitInstruction`? Can [PR this](https://github.com/blueshift-gg/solana-web3.js/compare/v3...amilz:solana-web3.js:chore/add-more-kit-ix-support) if you want.

src/nonce-account.ts
```diff
@@ -1,58 +1,59 @@
...
->([
-  BufferLayout.u32('version'),
-  BufferLayout.u32('state'),
-  Layout.publicKey('authorizedPubkey'),
-  Layout.publicKey('nonce'),
-  BufferLayout.struct<Readonly<{lamportsPerSignature: number}>>(
-    [FeeCalculatorLayout],
-    'feeCalculator',
-  ),
+const NONCE_ACCOUNT_DECODER = getStructDecoder([
```
amilz: Can leverage system client since we already have it. 

```ts
import { getNonceDecoder, getNonceSize } from '@solana-program/system';

const NONCE_ACCOUNT_DECODER = getNonceDecoder();

export const NONCE_ACCOUNT_LENGTH = getNonceSize();
```

src/address.ts
dhl: Agreed with @amilz we will preserve the conventional semantics of the toString return type here as string.

src/address.ts
dhl: Agreed with @amilz to keep toJSON(), to support JSON.stringify. Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify

Resolved in 4e1ef63db33d592ea41443738650a33b811c8cf4

src/nonce-account.ts
dhl: Resolved in f42d6020c91655db6e0b83e53301e7dfa9b91fca

dhl:
> Should we just use this in src/blockhash.ts
> 
> ```ts
> export type { Blockhash } from "@solana/rpc-types";
> ```

Resolved in 5c392f608bcc1e2b11db055ebb3645a64b71c91c

src/message/legacy.ts
dhl: Resolved in 647cfc705663d568893464a094c9269254eca390

PR #3819, opened 2026-05-21
refactor(Keypair): add Kit Signer Support

amilz:
Makes Kit signers first-class citizens in v3 while preserving the v1 `Keypair` surface and method signatures.

## What changed
- `Keypair` structurally implements Kit's `KeyPairSigner<KitAddress>`. `isKeyPairSigner(keypair)` returns `true`, and `keypair` can be passed directly to any Kit API that accepts a `KeyPairSigner`, `MessagePartialSigner`, or `TransactionPartialSigner`.
- `Transaction.sign(...)`, `Transaction.partialSign(...)`, `VersionedTransaction.sign(...)`, `Connection.sendTransaction(...)`, `Connection.simulateTransaction(...)`, and `sendAndConfirmTransaction(...)` now accept `Signer`, a union of the legacy v1 shape (renamed `Web3Signer`) plus Kit's `MessagePartialSigner` and `TransactionPartialSigner`. The internal dispatcher picks the right strategy per signer.
- Bridge code lives in `src/kit-adapters/signing.ts` (dispatcher + strategy) and `src/kit-adapters/brand.ts` (centralized `Blockhash → Nonce` and `TransactionMessageBytes` brand casts).

## Impact
Signing Interoperability with Web3js, Kit, Keychain:

```ts
  const web3Keypair = await Keypair.generate();
  const recipient = await Keypair.generate();
  const keychainSigner = await createKeychainSigner({});
  const kitSigner = await generateKeyPairSigner();
  // ...
  const tx = new Transaction()
    .add(
      getTransferSolInstruction({
        amount: TRANSFER_LAMPORTS,
        destination: recipientAddr.toBase58(),
        source: web3Keypair
      }),
    )
    .add(
      SystemProgram.transfer({
        fromPubkey: kitSignerPubkey,
        toPubkey: recipient.publicKey,
        lamports: TRANSFER_LAMPORTS,
      }),
    )
    .add(...)
    );
  // ...
  const signature = await sendAndConfirmTransaction(connection, tx, [
    web3Keypair,
    kitSigner,
    keychainSigner,
  ]);
```

Kind of annoying but to do this, `keypair.address` is now `KitAddress` (a branded base58 string), not an `Address` instance, to satisfy Kit's `KeyPairSigner`; use legacy `keypair.publicKey` whenever you need `Address` class methods. 

## v1 user impact

Same as `rc.0`:

```diff
- const kp = Keypair.generate();
+ const kp = await Keypair.generate();
```

## Test plan
- [x] `pnpm test:typecheck`
- [x] `pnpm test:lint`
- [x] `pnpm test:unit` (790 passing)

Closes DEV-601

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -21,6 +21,7 @@ These notes summarize the user-facing changes that landed since 1.98.4.
 - `Address.unique()` was removed.
 - Legacy BN-era `PublicKey` constructor inputs such as `PublicKeyData` or `{ _bn: BN }` are no longer accepted.
 - `Keypair.generate()`, `Keypair.fromSecretKey(...)`, and `Keypair.fromSeed(...)` are now async. Tests and stories that previously used sync key generation often need to become async or switch to app-local dummy public-key helpers.
+- `Keypair.address` is now a Kit branded base58 address string for signer API compatibility. Use `Keypair.publicKey` when you need the web3.js `Address` class methods such as `.toBytes()`, `.equals(...)`, or `.toBase58()`.
 - `Transaction.verifySignatures()` is now async and returns a promise.
 - `Transaction.serialize()` is now async and must be awaited.
 - Remaining sync signing helpers that existed earlier in the range were removed in favor of the async signing surface.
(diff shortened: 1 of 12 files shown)


dhl:
We have mixed opinions about this PR. It would be great if we could align our opinions here. From our perspective:

Adding Kit types to function inputs for seamless compat – Not always necessary, but hard to argue against if the use case is good! Creates a smoother interop API and reduces pain points for consuming Kit in Web3.js.

Returning Kit types in Web3.js, especially at this spot, is hard to defend. It may be possible to slowly coax the API more towards full interop over time, but for now, this is a red line we absolutely don't want to cross. It is confusing and too big of a change to make right now.

I hope this is an acceptable stance to take for now. It's not no forever, just not right now.

amilz:
Closing in favor of #3821

PR #3821, opened 2026-05-26
refactor(Keypair): Migrate Keypair to KeypairSigner

amilz:
- Leverages Kit's @solana/signer package for signing functions 
- Makes `Keypair` an extension of Kit's `KeyPairSigner` class while maintaining existing functionality
- Deprecates legacy `Signer` interface in lieu of `const newSigner = await Keypair.fromSecretKey(oldSigner.secretKey)`

## What changed
- `Keypair` structurally implements Kit's `KeyPairSigner<KitAddress>`. `isKeyPairSigner(keypair)` returns `true`, and `keypair` can be passed directly to any Kit API that accepts a `KeyPairSigner`, `MessagePartialSigner`, or `TransactionPartialSigner`.
- `Transaction.sign(...)`, `Transaction.partialSign(...)`, `VersionedTransaction.sign(...)`, `Connection.sendTransaction(...)`, `Connection.simulateTransaction(...)`, and `sendAndConfirmTransaction(...)` now only accept `Signer`, a union of Kit's `MessagePartialSigner` and `TransactionPartialSigner`. The internal dispatcher picks the right strategy per signer.
- Bridge code lives in `src/kit-adapters/signing.ts` (dispatcher + strategy) and `src/kit-adapters/brand.ts` (centralized `Blockhash → Nonce` and `TransactionMessageBytes` brand casts).

## Impact
Signing Interoperability with Web3js, Kit, Keychain:

```ts
  const web3Keypair = await Keypair.generate();
  const recipient = await Keypair.generate();
  const keychainSigner = await createKeychainSigner({});
  const kitSigner = await generateKeyPairSigner();
  // ...
  const tx = new Transaction()
    .add(
      getTransferSolInstruction({
        amount: TRANSFER_LAMPORTS,
        destination: recipientAddr.toBase58(),
        source: web3Keypair
      }),
    )
    .add(
      SystemProgram.transfer({
        fromPubkey: kitSignerPubkey,
        toPubkey: recipient.publicKey,
        lamports: TRANSFER_LAMPORTS,
      }),
    )
    .add(...)
    );
  // ...
  const signature = await sendAndConfirmTransaction(connection, tx, [
    web3Keypair,
    kitSigner,
    keychainSigner,
  ]);
```

Kind of annoying but to do this, `keypair.address` is now `KitAddress` (a branded base58 string), not an `Address` instance, to satisfy Kit's `KeyPairSigner`; use legacy `keypair.publicKey` whenever you need `Address` class methods. 

## v1 user impact

Same as `rc.0`:

```diff
- const kp = new Keypair();
- const kp = Keypair.generate();
+ const kp = await Keypair.generate();
```

## Change from #3819 

Drops the legacy Signer interface:

```diff
- export interface Signer {
-   address: Address;
- 
-   /** @deprecated Use `address` instead. */
-   publicKey: Address;
-   secretKey?: Uint8Array;
-   signBytes(message: Uint8Array): Promise<Uint8Array>;
- }
```

1-line breaking change for users of legacy `Signer`:

```ts
const newSigner = await Keypair.fromSecretKey(oldSigner.secretKey);
```

## Comparing options

image

## Test plan
- [x] `pnpm test:typecheck`
- [x] `pnpm test:lint`
- [x] `pnpm test:unit` (790 passing)

Closes DEV-601

### DIFF
--- a/src/kit-adapters/signing.ts
+++ b/src/kit-adapters/signing.ts
@@ -1,0 +1,189 @@
+import {
+  assertIsTransactionWithinSizeLimit,
+  createSignableMessage,
+  type Address as KitAddress,
+  isMessagePartialSigner,
+  isTransactionPartialSigner,
+  type MessagePartialSigner,
+  signatureBytes,
+  type Transaction as KitTransaction,
+  type TransactionPartialSigner,
+  type TransactionWithLifetime,
+} from '@solana/kit';
+
+import {Address} from '../address';
+import type {Signer, Web3Signer} from '../keypair';
+import {SIGNATURE_LENGTH_IN_BYTES} from '../transaction/constants';
+import {sign} from '../utils/ed25519';
+import {toPackedUint8Array} from '../utils/typed-array';
+import {asTransactionMessageBytes} from './brand';
+
+type SignableTransaction = Parameters<
+  TransactionPartialSigner['signTransactions']
+>[0][number];
+
+type SignaturePair = Readonly<{
+  publicKey: Address;
+  signature?: Uint8Array | null;
+}>;
+
+type KitSignerCandidate = {
+  readonly [key: string]: unknown;
+  readonly address: KitAddress;
+};
+
+type SigningStrategy =
+  | {
+      kind: 'kit-tx';
+      signer: TransactionPartialSigner;
+      address: KitAddress;
+      lifetime: TransactionWithLifetime['lifetimeConstraint'];
+    }
+  | {
+      kind: 'kit-msg';
+      signer: MessagePartialSigner;
+      address: KitAddress;
+    }
+  | {
+      kind: 'secret-bytes';
+      secretKey: Uint8Array;
+    };
+
+/** @internal */
+export function getSignerPublicKey(signer: Signer): Address {
+  if ('publicKey' in signer) {
+    return signer.publicKey;
+  }
+  return new Address(signer.address);
+}
+
+/**
+ * Sign the serialized bytes of a legacy or versioned transaction message,
+ * dispatching to whichever signing mechanism the input signer provides.
+ *
+ * Strategy precedence (see {@link pickSigningStrategy}):
+ *   1. Kit `TransactionPartialSigner` + lifetime → `signTransactions`
+ *   2. Kit `MessagePartialSigner`                → `signMessages`
+ *   3. v1 `Web3Signer` (`secretKey` bytes)       → ed25519 fallback
+ * A Kit signer with only `signTransactions` and no lifetime is rejected.
+ *
+ * @internal
+ */
+export async function signTransactionMessageBytes(
+  signer: Signer,
+  messageBytes: Uint8Array,
+  requiredSignerPublicKeys: readonly Address[],
+  signatures: readonly SignaturePair[] = [],
+  lifetimeConstraint?: TransactionWithLifetime['lifetimeConstraint'],
+): Promise<Uint8Array | undefined> {
+  const strategy = pickSigningStrategy(signer, lifetimeConstraint);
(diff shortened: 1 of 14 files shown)


deanmlittle:
> keypair.address is now KitAddress (a branded base58 string), not an Address instance, to satisfy Kit's KeyPairSigner; use legacy keypair.publicKey whenever you need Address class methods.

This is unreasonable for v3 imo. The entire point of deprecating Pubkey in lieu of Address the way we have is to give users a few release cycles to upgrade their syntax so we don't recreate the exact kind of hard landing kit itself was for them. It is far too aggressive to force this on users this early.

src/kit-adapters/signing.ts
```diff
@@ -0,0 +1,153 @@
...
+ * @internal
+ */
+export async function signTransactionMessageBytes(
+  signer: Signer,
+  messageBytes: Uint8Array,
+  requiredSignerPublicKeys: readonly Address[],
+  signatures: readonly SignaturePair[] = [],
+  lifetimeConstraint?: TransactionWithLifetime['lifetimeConstraint'],
+): Promise<Uint8Array | undefined> {
+  const strategy = pickSigningStrategy(signer, lifetimeConstraint);
```
dhl, line 62: Minor observation: the Kit signer rejection flow is not immediately obvious.

Would have been better if the `pickSigningStrategy` returns a null, e.g.

```typescript
function pickSigningStrategy(
  signer: Signer,
  lifetime: TransactionWithLifetime['lifetimeConstraint'] | undefined,
): SigningStrategy | null {
  const candidate = signer as Signer & {readonly [key: string]: unknown};
  const hasTransactionPartial = isTransactionPartialSigner(candidate);
  const hasMessagePartial = isMessagePartialSigner(candidate);

  if (hasTransactionPartial && lifetime != null) {
    return {
      kind: 'kit-tx',
      signer: candidate,
      address: candidate.address,
      lifetime,
    };
  }
  if (hasMessagePartial) {
    return {
      kind: 'kit-msg',
      signer: candidate,
      address: candidate.address,
    };
  }
  return null;
}
```

and let `signTransactionMessageBytes` throws.

```typescript
export async function signTransactionMessageBytes(
  signer: Signer,
  messageBytes: Uint8Array,
  requiredSignerPublicKeys: readonly Address[],
  signatures: readonly SignaturePair[] = [],
  lifetimeConstraint?: TransactionWithLifetime['lifetimeConstraint'],
): Promise<Uint8Array | undefined> {
  const strategy = pickSigningStrategy(signer, lifetimeConstraint);
  if (!strategy) {
    throw new Error(
      'TransactionPartialSigner support requires transaction lifetime information. Use a MessagePartialSigner-compatible signer or provide a transaction with a blockhash lifetime or nonce lifetime.',
    );
  }

  switch (strategy.kind) {
    case 'kit-tx': {
      const [dict] = await strategy.signer.signTransactions([
        buildSignableTransaction(
          messageBytes,
          requiredSignerPublicKeys,
          signatures,
          strategy.lifetime,
        ),
      ]);
      return dict[strategy.address];
    }
    case 'kit-msg': {
      const [dict] = await strategy.signer.signMessages([
        createSignableMessage(toPackedUint8Array(messageBytes)),
      ]);
      return dict[strategy.address];
    }
  }
}
```

dhl (approved):
LGTM. Universal signing achieved 🚀

src/kit-adapters/signing.ts
amilz: cleaned up in [8acb73e](https://github.com/solana-foundation/solana-web3.js/pull/3821/commits/8acb73ea3f8966a017a609d7108a7b0666892a3a)

PR #3823, opened 2026-05-27
feat: support InstructionPlan inputs in Transaction.add and Message compilers

amilz:
## Summary
- `Transaction.add()`, `Message.compile()`, `MessageV0.compile()`, and the `TransactionMessage` constructor all now accept `@solana/kit` `InstructionPlan` inputs in addition to legacy/`Kit` instructions.
- Plans are flattened in order via kit's `flattenInstructionPlan`; `MessagePackerInstructionPlan` leaves are rejected (cannot be honored inside a single transaction/message).
- Shared helper `expandInstructionPlans()` and shared `InstructionInput` type live in `src/kit-adapters/instruction-plan.ts` and are used by all four entry points.
- `Transaction.add()` body cleaned up: duck-typed `any` branches replaced with `instanceof Transaction` / `instanceof TransactionInstruction`.

This lets callers pass codama-client plan builders (e.g. `getCreateMintInstructionPlan` from `@solana-program/token`) directly into any of the entry points:

```ts
import { getCreateMintInstructionPlan } from '@solana-program/token';

const tx = new Transaction().add(
  getCreateMintInstructionPlan({
    decimals: 6,
    mintAuthority: payer.address,
    newMint,
    payer,
  }),
);
```

## Test plan
- [x] `pnpm test:typecheck`
- [x] `pnpm test:lint`
- [x] `pnpm test:prettier`
- [x] `pnpm test:unit` — 793 passing (7 new tests across `Transaction.add`, `Message.compile`, `MessageV0.compile`, `TransactionMessage`, plus a real-builder test using `@solana-program/token`'s `getCreateMintInstructionPlan`)

### DIFF
--- a/src/transaction/legacy.ts
+++ b/src/transaction/legacy.ts
@@ -399,11 +399,12 @@ export class Transaction {
       Transaction | TransactionInstructionCtorFields | InstructionInput
     >
   ): Transaction {
-    if (items.length === 0) {
+    const expanded = expandInstructionPlans(items);
+    if (expanded.length === 0) {
       throw new Error('No instructions');
     }
 
-    expandInstructionPlans(items).forEach(item => {
+    expanded.forEach(item => {
       if (item instanceof Transaction) {
         this.instructions = this.instructions.concat(item.instructions);
       } else if (isKitInstruction(item)) {
--- a/src/transaction/legacy.ts
+++ b/src/transaction/legacy.ts
@@ -1,7 +1,6 @@
 import {
   type Blockhash,
   fixDecoderSize,
-  flattenInstructionPlan,
   getArrayDecoder,
   getBase58Codec,
   getBytesDecoder,
@@ -10,8 +9,6 @@ import {
   getStructDecoder,
   type Instruction as KitInstruction,
   type InstructionPlan,
-  isInstructionPlan,
-  isSingleInstructionPlan,
 } from '@solana/kit';
 
 import {PACKET_DATA_SIZE, SIGNATURE_LENGTH_IN_BYTES} from './constants';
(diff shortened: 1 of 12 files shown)


src/transaction/legacy.ts
```diff
@@ -381,30 +384,34 @@ export class Transaction {
...
-    if (items.length === 0) {
+    const expanded = expandInstructionPlans(items);
+    if (expanded.length === 0) {
       throw new Error('No instructions');
     }
 
-    items.forEach((item: any) => {
-      if ('instructions' in item) {
+    expanded.forEach(item => {
+      if (item instanceof Transaction) {
```
dhl, line 408: These were actually changed a few years ago by prior devs from `item instanceof Transaction` to `'instructions' in item`. I suppose it is to support loose, "Transaction-like" object being passed in during runtime.

I had left these alone on purpose during the V1 -> V3 migration.

I am supportive of the change, but this is potentially a runtime-level breaking change that is worth flagging in release notes.

Well-behaved, TypeScript applications won't be affected by this change.

src/transaction/legacy.ts
```diff
@@ -381,30 +384,34 @@ export class Transaction {
...
         this.instructions = this.instructions.concat(item.instructions);
       } else if (isKitInstruction(item)) {
         this.instructions.push(
           new TransactionInstruction(toLegacyInstructionFields(item)),
         );
-      } else if ('data' in item && 'programId' in item && 'keys' in item) {
+      } else if (item instanceof TransactionInstruction) {
```
dhl, line 414: Ditto

src/transaction/legacy.ts
dhl: Let's update the release note so the change in behavior is documented.

amilz: good catch — documented the runtime behavior change in RELEASE_NOTES.md under Breaking Changes in b481cc6e51

amilz: done — added a bullet under Breaking Changes in b481cc6e51

dhl (approved):
LGTM 🚀

PR #3828, opened 2026-05-29
fix: avoid unsubscribe retry loop on closed websocket

### LINKED ISSUE #3761: [URGENT] Infinite loop in Connection._updateSubscriptions()

sueun-dev:
#### Problem

When an unsubscribe RPC call fails because the underlying websocket is already closing or closed, `_updateSubscriptions()` currently treats it like a retryable unsubscribe failure. It restores the subscription to `subscribed` and recurses, which can keep retrying the same unsubscribe call on the dead socket.

Fixes #3761.

#### Summary of Changes

- Preserve websocket `readyState` on the local RPC websocket error.
- Treat unsubscribe failures from closing/closed sockets as locally unsubscribed instead of retrying them.
- Add subscription regression coverage for the closing-socket unsubscribe path.

#### Test plan

- `pnpm exec cross-env NODE_ENV=test NODE_OPTIONS='--import tsx' mocha './test/connection-subscriptions.test.ts' --grep 'socket is closing'`
- `pnpm run test:typecheck`
- `pnpm run test:prettier`
- `PATH="$PWD/node_modules/.pnpm/node_modules/.bin:$PATH" pnpm run test:lint`
- `pnpm run test:unit`
- `pnpm run compile:js`
- `git diff --check`

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -35,7 +35,7 @@ import fetchImpl from './fetch-impl';
 import {DurableNonce, NonceAccount} from './nonce-account';
 import {PublicKey} from './publickey';
 import {Signer} from './keypair';
-import RpcWebSocketClient from './rpc-websocket';
+import RpcWebSocketClient, {RpcWebSocketConnectionError} from './rpc-websocket';
 import {MS_PER_SLOT} from './timing';
 import {
   Transaction,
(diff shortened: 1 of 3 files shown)


dhl:
Thanks for the PR @sueun-dev.

We are now focusing our effort on the [3.x line](https://github.com/solana-foundation/solana-web3.js/tree/v3.x) of Web3.js.

Only security patches will be considered for v1 :pray:

PR #3829, opened 2026-06-01
docs(skill): add @solana/spl-token → @solana-program/token migration …

amilz:
…guide

Adds @solana/spl-token → @solana-program/token migration coverage to the v1→v3 upgrade path:

- docs/web3js-spl-token-migration.md: human-facing migration guide
- skills/web3js-v1-to-v3-migration/reference/spl-token.md: execution-oriented token surface, loaded on demand when spl-token usage is present
- web3js-v1-to-v3-migration SKILL/doc: extend description, Fast Triage, and trigger phrases to cover the token surface and point at the reference

Covers instruction-builder swaps, async ATA derivation, plan helpers, Keypair-as-TransactionSigner, and bigint/Option decoding.

### DIFF
--- a/skills/web3js-v1-to-v3-migration/reference/spl-token.md
+++ b/skills/web3js-v1-to-v3-migration/reference/spl-token.md
@@ -123,7 +123,8 @@ const createAtaIx = getCreateAssociatedTokenIdempotentInstruction({
   owner,   // kit Address (the wallet)
   mint,    // kit Address
 });
-new Transaction().add(createAtaIx, /* transfer/mintTo */);
+const tx = new Transaction().add(createAtaIx, /* transfer/mintTo */);
+await sendAndConfirmTransaction(connection, tx, [payer]);
 ```
 
 If the app's flow needs the decoded `Account` object after sending, fetch it with `connection.getAccountInfo(new Address(ata))` and decode with `getTokenDecoder().decode(data)`.
--- a/skills/web3js-v1-to-v3-migration/reference/spl-token.md
+++ b/skills/web3js-v1-to-v3-migration/reference/spl-token.md
@@ -1,0 +1,313 @@
+# `@solana/spl-token` → `@solana-program/token` (token surface)
+
+Reference companion to the v1→v3 migration skill, loaded when the migration touches token-program usage. For human-facing background, see [`docs/web3js-spl-token-migration.md`](../../../docs/web3js-spl-token-migration.md).
+
+## When this applies
+- The v1→v3 migration also touches `@solana/spl-token`: token instructions, ATAs, mint/account fetches, or token-account decoding.
+- The codebase still imports `createMintToInstruction`, `createTransferCheckedInstruction`, `getOrCreateAssociatedTokenAccount`, `getAssociatedTokenAddressSync`, `getMint`, `getAccount`, `MintLayout`, `AccountLayout`, `TOKEN_PROGRAM_ID`, or `ASSOCIATED_TOKEN_PROGRAM_ID`.
+
+If the mint is owned by Token-2022 (extensions, transfer hooks, etc.), use `@solana-program/token-2022` instead — the naming, instruction surface, and steps below are otherwise identical; only the package and the `tokenProgram` argument on ATA derivation differ.
+
+## Goal
+Migrate token-program usage from `@solana/spl-token` to `@solana-program/token` while staying on v3 `@solana/web3.js`'s `Connection` / `Transaction` / `Keypair` / `sendAndConfirmTransaction` surface. Do **not** rewrite the surrounding code to kit transaction messages, plugin clients, or a kit RPC — v3 web3.js bridges to `@solana-program/token` natively and that's the minimal-churn migration path.
+
+## Operating Model
+- Work in narrow, behavior-scoped slices: imports → PDA derivation → instruction builders → high-level helpers → account reads.
+- The bulk of churn is mechanical: `createXInstruction(...)` → `getXInstruction(...)`, `PublicKey` → `Address` (call `.toBase58()` on v3 `Address` instances at builder boundaries), sync ATA derivation → async `findAssociatedTokenPda`, `getMint`/`getAccount` → `connection.getAccountInfo(...)` + `getMintDecoder().decode(data)` / `getTokenDecoder().decode(data)`.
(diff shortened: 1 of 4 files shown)


skills/web3js-v1-to-v3-migration/reference/spl-token.md
```diff
@@ -0,0 +1,313 @@
...
+- Keep unchecked variants only when migrating code that intentionally avoids the decimals check and you've confirmed that's still correct.
+
+### Classic Token vs Token-2022
+- If the mint is owned by the classic Token program, use `@solana-program/token`.
+- If the mint is owned by Token-2022 (extensions, transfer hooks, etc.), use `@solana-program/token-2022`. The workflow is identical; only the package and the `tokenProgram` argument on ATA derivation differ.
+- If the code path supports both, pass the owning program address through explicitly — do not hardcode `TOKEN_PROGRAM_ADDRESS`.
+
+### `PublicKey` at the boundary
+- Inside the migrated module, use the v3 `Address` class for everything that touches `SystemProgram` / `Connection`, and call `.toBase58()` to produce the kit-branded string at every `@solana-program/token` builder boundary.
+- At third-party boundaries that still hand back v1 `PublicKey`, convert once (`new Address(pk.toBase58())`) and keep the v3 `Address` downstream. Do not let v1 `PublicKey` propagate.
```
dhl: In having a GPT 5.4 powered agent trying out this guide over a codebase that had been migrated from Web3.js v1 -> v3, the agent seemed to have been confused by this instruction here and took the v3 PublicKey (which is already an Address), and got an Address out by `new Address(publicKey.toBase58())`.

I suspect a some kind of guardrail here to tell agent not to do this wrapping if the project is already using Web3.js v3 or that `PublicKey` is already an alias of `Address`.

amilz: ahh--good catch. wdyt about [1f37703](https://github.com/solana-foundation/solana-web3.js/pull/3829/commits/1f37703b9e5ccf210795420fd2b52664f6704088) to mitigate this spinning?

skills/web3js-v1-to-v3-migration/reference/spl-token.md
```diff
@@ -0,0 +1,324 @@
...
+
+// after
+const [ata] = await findAssociatedTokenPda({ mint, owner, tokenProgram: TOKEN_PROGRAM_ADDRESS });
+const createAtaIx = getCreateAssociatedTokenIdempotentInstruction({
+  payer,   // a v3 Keypair — it's a TransactionSigner (see step 6)
+  ata,     // kit Address
+  owner,   // kit Address (the wallet)
+  mint,    // kit Address
+});
+new Transaction().add(createAtaIx, /* transfer/mintTo */);
```
amilz:
```suggestion
const tx = new Transaction().add(createAtaIx, /* transfer/mintTo */);
await sendAndConfirmTransaction(connection, tx, [payer]);
```

skills/web3js-v1-to-v3-migration/reference/spl-token.md
dhl: Looks like that will do, I'll try to re-run the migration with this change to see if the problem is addressed. In any case, despite the unnecessary rewrapping, it was harmless and functionally correct

dhl (commented):
I was able to have an agent successfully migrate a simple @solana/spl-token app to @solana-program/token using this guide. That’s a great real-world validation of the interop work in v3.

That said, the createMint before/after example highlights a noticeable DX regression:

**v1 + @solana/spl-token**
```typescript
const mint = await createMint(connection, payer, payer.publicKey, null, 6);
```

**v3 + @solana-program/token**
```typescript
const mint = Keypair.generate();
const tx = new Transaction().add(
  getCreateMintInstructionPlan({
    payer,
    newMint: mint,
    decimals: 6,
    mintAuthority: payer.publicKey.toBase58(),
    freezeAuthority: null,
  })
);
await sendAndConfirmTransaction(connection, tx, [payer, mint]);
```

The new pattern is more explicit and aligned with the Kit client and philosophy (which I personally prefer and use when teaching). 

However, it removes the convenient, familiar one-liner that many developers — especially new comers — appreciate for basic operations. It is worth noting that this may be a step backward in ergonomics for those scenarios.

To make the migration path more compelling and reduce friction for a broader set of users, would it be worth also providing thin optional convenience helpers on top of @solana-program/token? Something that restores the "do it all" helpers experience where desired, while still exposing the explicit primitives for those who want them. This would give developers a choice of abstraction level without compromising the underlying design.

Once again, appreciate the effort on this! The interop story is already strong, and a bit more ergonomic sugar could help keep Web3.js accessible for both simple and more complex use cases.

amilz:
> To make the migration path more compelling and reduce friction for a broader set of users, would it be worth also providing thin optional convenience helpers on top of @solana-program/token? Something that restores the "do it all" helpers experience where desired, while still exposing the explicit primitives for those who want them. This would give developers a choice of abstraction level without compromising the underlying design.

Heard--yes makes sense to have something more familiar. Until we have a cleaner solution, wdyt about merging this PR so that skill has a migration path (even if it's not quite our desired end state).

That said...just want to highlight that the screenshot is a little distorted. It's really 1 line of code vs 3 (txn assembly, ix building, and send/confirm). We can look at the solana-program package but also think we might be able to find an elegant solution in our `Connection` class. Or since web3js already has program classes like `SystemProgram`, it might make sense to just add a `TokenProgram` and build our own helpers internally. IMO It's ubiquitous enough to consider.

dhl (approved):

dhl:
Agreed this is a good stop-gap.

IMHO the action helpers allow intents to be stated directly without exposing the underlying mechanics, and it just makes the API a little bit easier to use and learn for simple cases.

Under the hood, it is doing exactly what the spelled out form is doing:

```typescript
export async function createMint(
    connection: Connection,
    payer: Signer,
    mintAuthority: PublicKey,
    freezeAuthority: PublicKey | null,
    decimals: number,
    keypair = Keypair.generate(),
    confirmOptions?: ConfirmOptions,
    programId = TOKEN_PROGRAM_ID,
): Promise<PublicKey> {
    const lamports = await getMinimumBalanceForRentExemptMint(connection);

    const transaction = new Transaction().add(
        SystemProgram.createAccount({
            fromPubkey: payer.publicKey,
            newAccountPubkey: keypair.publicKey,
            space: MINT_SIZE,
            lamports,
            programId,
        }),
        createInitializeMint2Instruction(keypair.publicKey, decimals, mintAuthority, freezeAuthority, programId),
    );

    await sendAndConfirmTransaction(connection, transaction, [payer, keypair], confirmOptions);

    return keypair.publicKey;
}
```

Those action helpers are worth keeping, and we don't even need to introduce a new TokenProgram construct; @solana/spl-token already fulfills the role.

PR #3836, opened 2026-06-11
feat: add loader v3 program class

ChiefWoods:
A wrapper of `solana-program/loader-v3`, just like the `ComputeBudgetProgram` class

### DIFF
--- a/src/programs/loader-v3.ts
+++ b/src/programs/loader-v3.ts
@@ -1,0 +1,575 @@
+import {createNoopSigner, type ReadonlyUint8Array} from '@solana/kit';
+import {
+  getCloseInstruction,
+  getDeployWithMaxDataLenInstruction,
+  getExtendProgramInstruction,
+  getInitializeBufferInstruction,
+  getSetAuthorityCheckedInstruction,
+  getSetAuthorityInstruction,
+  getUpgradeInstruction,
+  getWriteInstruction,
+  identifyLoaderV3Instruction,
+  LOADER_V3_PROGRAM_ADDRESS,
+  LoaderV3Instruction as GeneratedLoaderV3Instruction,
+  parseLoaderV3Instruction,
+  type ParsedLoaderV3Instruction,
+} from '@solana-program/loader-v3';
+
+import {Address} from '../address';
+import {fromKitAddress, toKitAddress} from '../kit-adapters/address';
+import {
+  fromKitInstruction,
+  toKitInstruction,
+} from '../kit-adapters/instruction';
+import {TransactionInstruction} from '../transaction';
+
+const LOADER_V3_PROGRAM_ID = new Address(LOADER_V3_PROGRAM_ADDRESS);
+
+/**
+ * An enumeration of valid LoaderV3InstructionType's
+ */
+export type LoaderV3InstructionType =
+  | 'InitializeBuffer'
+  | 'Write'
+  | 'DeployWithMaxDataLen'
+  | 'Upgrade'
+  | 'SetAuthority'
+  | 'Close'
+  | 'ExtendProgram'
+  | 'SetAuthorityChecked';
+
+export type InitializeBufferParams = {
+  /** Source account to initialize. */
+  sourceAccount: Address;
+  /** Buffer authority. */
+  bufferAuthority: Address;
+};
+
+export type WriteParams = {
+  /** Buffer account. */
+  bufferAccount: Address;
+  /** Buffer authority. */
+  bufferAuthority: Address;
+  /** Offset into the buffer to write. */
+  offset: number;
+  /** Bytes to write. */
+  bytes: ReadonlyUint8Array | Uint8Array;
+};
+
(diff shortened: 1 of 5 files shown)


test/program-tests/loader-v3.test.ts
```diff
@@ -0,0 +1,234 @@
...
+    };
+
+    const instruction = LoaderV3Program.setAuthorityChecked(params);
+
+    expect(LoaderV3Instruction.decodeInstructionType(instruction)).to.eq(
+      'SetAuthorityChecked',
+    );
+    expect(LoaderV3Instruction.decodeSetAuthorityChecked(instruction)).to.eql(
+      params,
+    );
```
amilz, line 180: Should we differentiate this from the `setAuthority` tests and verify that `newAuthority` is a signer?

test/program-tests/loader-v3.test.ts
```diff
@@ -0,0 +1,234 @@
...
+      bufferOrProgramDataAccount,
+      currentAuthority,
+    };
+
+    const instruction = LoaderV3Program.setAuthority(params);
+
+    expect(LoaderV3Instruction.decodeInstructionType(instruction)).to.eq(
+      'SetAuthority',
+    );
+    expect(LoaderV3Instruction.decodeSetAuthority(instruction)).to.eql(params);
```
amilz, line 160: do we want to add an assertion for `newAuthority` is undefined?

src/programs/loader-v3.ts
```diff
@@ -0,0 +1,575 @@
...
+type ParsedAnyLoaderV3Instruction = ParsedLoaderV3Instruction<string>;
+
+type ParsedInstructionOfType<
+  TInstructionType extends GeneratedLoaderV3Instruction,
+> = Extract<ParsedAnyLoaderV3Instruction, {instructionType: TInstructionType}>;
+
+function getInstructionType(
+  instruction: TransactionInstruction,
+): LoaderV3InstructionType {
+  checkProgramId(instruction.programId);
```
amilz, line 159: add test coverage for passing w/ a prong program id

amilz (commented):
LGTM @dhl want to take a look? 
Couple of test nits.

test/program-tests/loader-v3.test.ts
ChiefWoods: There's already a 'setAuthority without newAuthority' test path

test/program-tests/loader-v3.test.ts
ChiefWoods: Didn't add this test because @solana-program/loader-v3 internally does not check whether new authority is a signer either. Currently none of the programs file checks for writable/signer in their decode<Instruction> functions.

dhl:
Thanks for the contribution once again @ChiefWoods :pray: :heart:

@amilz @ChiefWoods I know we are following the pattern from other classes we inherited from V1 here, but given that there are no state changes in anything we do here, I would if it is better to just export the adapted functions for `Address` remapping without scoping them under a class as static methods?

What I am thinking about is to be able to programmatically adapt the codama clients so we don't have to hand code these going forward.

Thoughts?

src/programs/loader-v3.ts
```diff
@@ -0,0 +1,575 @@
...
+      bytes: Uint8Array.from(parsedInstruction.data.bytes),
+    };
+  }
+
+  /**
+   * Decode a deploy with max data len instruction and retrieve the instruction params.
+   */
+  static decodeDeployWithMaxDataLen(
+    instruction: TransactionInstruction,
+  ): DeployWithMaxDataLenParams {
```
dhl, line 252: maxDataLen in `DeployWithMaxDataLenParams` is too wide to be used here.

In v3, we tend to take numerical inputs as `number | bigint` to make it easier to work with, but for outputs, we use bigint only to match on-chain data type.

The codama clients also tend to do the same:

```typescript
export type DeployWithMaxDataLenInstructionData = { discriminator: number; maxDataLen: bigint };

export type DeployWithMaxDataLenInstructionDataArgs = { maxDataLen: number | bigint };
```

(See https://github.com/solana-program/loader-v3/blob/5185fe3534065725c1014e369a03b415ad3a1954/clients/js/src/generated/instructions/deployWithMaxDataLen.ts#L78-L80)

We should narrow the type here.

dhl:
LGTM overall. Just one type narrowing issue we should sort out.

I think the class vs module exports question is beyond the scope of 3.0 release - what @ChiefWoods did here is consistent with the rest of the program clients.

src/programs/loader-v3.ts
ChiefWoods: Fair point, amended

dhl (approved):

amilz:
@ChiefWoods can you resolve the CI issue (I think likely just lint/format)

amilz (approved):

amilz:
Thank you @ChiefWoods

PR #3845, opened 2026-07-02
feat: re-export @solana/codecs

### LINKED ISSUE #3843: Web3.js v3 should re-export @solana/codecs
## Motivation

Given the central role that @solana/codecs play in Web3.js v3, codecs should be re-exported the same way it is done in Kit.

HkSolDev:
### Description
This PR re-exports `@solana/codecs` from `@solana/web3.js` to align with `@solana/kit` and provide standard access to codec helpers in Web3.js v3.

Fixes #3843

### Changes
* **package.json**: Added `@solana/codecs` to direct dependencies.
* **src/index.ts**: Added `export * from '@solana/codecs';`.
* **test/exports.test.ts**: Added a unit test validating the exports are functional.

### Verification
Ran unit tests locally matching the new exports test suite:
`pnpm test:unit -g "web3.js exports"` -> **1 passing (4ms)**

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,8 @@
     "@solana/fast-stable-stringify": "^6.8.0",
     "@solana/kit": "^6.8.0",
     "@solana/rpc-spec-types": "^6.8.0",
-    "@solana/signers": "^6.8.0"
+    "@solana/signers": "^6.8.0",
+    "@solana/codecs": "^6.8.0"
   },
   "devDependencies": {
     "@commitlint/config-conventional": "^19.2.2",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -32,6 +32,9 @@ importers:
       '@solana/assertions':
         specifier: ^6.8.0
         version: 6.8.0(typescript@5.9.3)
+      '@solana/codecs':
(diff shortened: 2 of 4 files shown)


dhl:
Hey @HkSolDev!

Thanks for the PR.

The motivation for #3845 is to explore how we could take advantage of @solana/codecs that we are already downloading transitively because of our dependency on Kit.

Declaring dependency on @solana/codecs directly as in this PR would certainly work, but that is not solving the problem set out in the issue, and it adds maintenance burden on our end.

I think short of having Kit exposing codecs directly, so we could re-export from `@solana/kit/codecs`, the extra maintenance burden is hard to justify.

HkSolDev:
Hi @dhl 

That make perfectly sense to avoid the direct dependency here if the solana/kit already have to, I have opened an [issue](https://github.com/anza-xyz/kit/issues/1834) and a [PR](https://github.com/anza-xyz/kit/pull/1835) on [Kit](https://github.com/anza-xyz/kit) to expose @solana/codecs via a suggest way @solana/kit/codecs.

Once the PR will merge, I will update the PR to import and re-export the codecs directly form the @solana/kit/codecs as suggested

PR #3848, opened 2026-07-14
feat: add ZK Elgamal Proof program class 

ChiefWoods:
A wrapper of `solana-program/zk-elgamal-proof`, just like the LoaderV3 class


amilz:
Closing for now. Will reopen if we decide to go a different approach on new program classes

PR #3861, opened 2026-08-14
feat: add v1 transaction support to v3.x

amilz:
## Summary
- Bump `@solana/kit` and supporting `@solana/*` / `@solana-program/*` packages to their kit-v7 majors (v1 codecs and types live in kit 7)
- Add `MessageV1` (SIMD-0296) mirroring `MessageV0`: no address lookup tables, plus a message-level `TransactionConfig` (`computeUnitLimit`, `heapSize`, `loadedAccountsDataSizeLimit`, `priorityFeeLamports` — total lamports, not micro-lamports/CU)
- `VersionedTransaction` handles the v1 message-first wire envelope on serialize/deserialize; `TransactionMessage` gains `compileToV1Message(config?)` and preserves `transactionConfig` through decompile
- `getTransaction`, `getBlock`, and `getParsedTransaction` with `maxSupportedTransactionVersion: 1` now materialize `MessageV1` (previously v1 responses were silently decoded as legacy `Message`); response typings widen `0` → `0 | 1` and `ParsedMessage` exposes `transactionConfig`; `blockSubscribe` accepts `maxSupportedTransactionVersion: 1`

V1 Context: https://solana.com/upgrades/larger-transaction-sizes

## Notes
- Kit does not export its v1 config mask/values helpers or `V1TransactionConfig`, so `src/message/transaction-config.ts` reimplements the (small) mask/values/decompile logic against the same wire encoding, cross-checked in tests via kit's `getCompiledTransactionMessageDecoder`/`getTransactionDecoder`
- The RPC JSON field for the v1 priority fee is `priorityFee` (per Agave's `UiTransactionConfig`); it maps to `priorityFeeLamports: bigint` on the client type
- `getSignaturesForAddress` has no version parameter in kit's rpc-api; out of scope here

## Test Plan
- `pnpm test:lint && pnpm test:typecheck && pnpm test:prettier && pnpm test:unit` (838 passing)
- `pnpm compile:js && pnpm compile:typedefs` pass
- New coverage: `test/message-tests/v1.test.ts`, v1 envelope suite in `test/transaction.test.ts`, decompile/config tests in `test/transaction-tests/message.test.ts`, mocked v1 RPC responses in `test/connection.test.ts`, blockSubscribe config passthrough in `test/connection-subscriptions.test.ts`
- have run some manual tests on `solana-local-validator` 4.2 and surfpool 1.5 (working)

Closes DEV-895


src/message/v1.ts
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  /** v1 messages do not support address lookup tables */
+  get addressTableLookups(): Array<MessageAddressTableLookup> {
+    return [];
+  }
+
+  get numAccountKeysFromLookups(): number {
+    return 0;
+  }
```
dhl: IMHO we should drop these. There is very little reason to keep the ATL methods, especially when the concepts don't exist on V1 and there is no interface for the message types to keep.

`numAccountKeysFromLookups` only exists on V1, but is absent on Legacy, and so isn't present on the VersionedMessage union type.

More importantly, the cost of carrying a defunct concept in V1 message type isn't worth keeping it.

amilz: agreed.  dd88ddc0a7728a8cc6186cb6708cc89751869abf

src/connection.ts
```diff
@@ -1070,23 +1076,26 @@ type RpcParsedInstruction = Extract<
...
+type RpcParsedMessageAccount = TransactionForAccounts<
+  0 | 1
+>['transaction']['accountKeys'][number];
 
 type RpcParsedAddressTableLookup = NonNullable<
   NonNullable<
-    TransactionForFullJson<0>['transaction']['message']['addressTableLookups']
+    TransactionForFullJson<
+      0 | 1
+    >['transaction']['message']['addressTableLookups']
```
dhl: ATL not valid for v1 message

src/message/transaction-config.ts
dhl, line 1: Nit: It would be better if `TransactionConfig` is scoped under `src/message/v1.ts`. This would allow future transaction messages to have their own TransactionConfig without needing restructuring.

src/message/transaction-config.ts
```diff
@@ -0,0 +1,129 @@
...
+/** A single config value as encoded in a compiled v1 message. */
+export type CompiledTransactionConfigValue =
+  | {kind: 'u32'; value: number}
+  | {kind: 'u64'; value: bigint};
+
+// The priority fee occupies bits 0 and 1; both must be set or both unset.
+const PRIORITY_FEE_LAMPORTS_BIT_MASK = 3;
+const COMPUTE_UNIT_LIMIT_BIT_MASK = 4;
+const LOADED_ACCOUNTS_DATA_SIZE_LIMIT_BIT_MASK = 8;
+const HEAP_SIZE_BIT_MASK = 16;
```
dhl: Nit: Consider having a more direct representation of the bit masks to match the SIMD descriptions more closely.


```suggestion
const PRIORITY_FEE_LAMPORTS_BIT_MASK = 0b11;
const COMPUTE_UNIT_LIMIT_BIT_MASK = 0b100;
const LOADED_ACCOUNTS_DATA_SIZE_LIMIT_BIT_MASK = 0b1000;
const HEAP_SIZE_BIT_MASK = 0b10000;
```

src/message/transaction-config.ts
```diff
@@ -0,0 +1,129 @@
...
+   */
+  priorityFeeLamports?: bigint | number;
+};
+
+/** A single config value as encoded in a compiled v1 message. */
+export type CompiledTransactionConfigValue =
+  | {kind: 'u32'; value: number}
+  | {kind: 'u64'; value: bigint};
+
+// The priority fee occupies bits 0 and 1; both must be set or both unset.
```
dhl:
```suggestion
// Reference: https://github.com/solana-foundation/solana-improvement-documents/blob/fc519fb3d1ef0f7624b6232bda958438feba09ce/proposals/0385-transaction-v1.md#transactionconfigmask-1
```

src/message/transaction-config.ts
amilz: kit v8 exports this. killed the file and imported from kit. rexport from src/message/v1.ts

[7248efa](https://github.com/solana-foundation/solana-web3.js/pull/3861/commits/7248efa3167069dfa9863c0d0ad61e69f845fd1e)

src/message/transaction-config.ts
amilz: kit v8 exports this. killed the constants and imported from kit. [7248efa](https://github.com/solana-foundation/solana-web3.js/pull/3861/commits/7248efa3167069dfa9863c0d0ad61e69f845fd1e)

src/message/transaction-config.ts
amilz: since we now import from kit, did not include link

src/message/v1.ts
```diff
@@ -0,0 +1,307 @@
...
+        `Expected versioned message with version 1 but found version ${maskedPrefix}`,
+      );
+    }
+    const decoded = MESSAGE_DECODER.decode(toUint8ArrayView(serializedMessage));
+    if (decoded.version !== 1) {
+      throw new Error(
+        `Expected versioned message with version 1 but found version ${decoded.version}`,
+      );
+    }
+    return MessageV1.fromCompiledMessage(decoded);
```
dhl:
```suggestion
    const decoded = MESSAGE_DECODER.decode(toUint8ArrayView(serializedMessage));                                                                                                                                                                          
     if (decoded.version !== 1) {                                                                                                                                                                                                                          
       throw new Error(                                                                                                                                                                                                                                    
         decoded.version === 'legacy'                                                                                                                                                                                                                      
           ? 'Expected versioned message but received legacy message'                                                                                                                                                                                      
           : `Expected versioned message with version 1 but found version ${decoded.version}`,                                                                                                                                                             
       );                                                                                                                                                                                                                                                  
     }                                                                                                                                                                                                                                                     
     return MessageV1.fromCompiledMessage(decoded);
```

src/message/v1.ts
```diff
@@ -0,0 +1,307 @@
...
+      const unsignedAccountIndex = index - numSignedAccounts;
+      const numUnsignedAccounts = numStaticAccountKeys - numSignedAccounts;
+      const numWritableUnsignedAccounts =
+        numUnsignedAccounts - this.header.numReadonlyUnsignedAccounts;
+      return unsignedAccountIndex < numWritableUnsignedAccounts;
+    } else {
+      const numWritableSignedAccounts =
+        numSignedAccounts - this.header.numReadonlySignedAccounts;
+      return index < numWritableSignedAccounts;
+    }
```
dhl: Consider going for a simpler calculation based on the accounts ordering in the Addresses byte arrays.

```suggestion
    const numSigners = this.header.numRequiredSignatures;                                                                                                                                                                                                 
    const numWritableSignedAccounts =                                                                                                                                                                                                                     
      numSigners - this.header.numReadonlySignedAccounts;                                                                                                                                                                                                 
    const numWritableUnsignedAccounts =                                                                                                                                                                                                                   
      this.staticAccountKeys.length -                                                                                                                                                                                                                     
      numSigners -                                                                                                                                                                                                                                        
      this.header.numReadonlyUnsignedAccounts;                                                                                                                                                                                                            
    return (                                                                                                                                                                                                                                              
      index < numWritableSignedAccounts ||                                                                                                                                                                                                                
      (index >= numSigners && index < numSigners + numWritableUnsignedAccounts)                                                                                                                                                                           
    );
```

src/message/v1.ts
```diff
@@ -0,0 +1,307 @@
...
+
+export type CompileV1Args = {
+  payerKey: Address;
+  instructions: Array<InstructionInput>;
+  recentBlockhash: Blockhash;
+  transactionConfig?: V1TransactionConfig;
+};
+
+/**
+ * A v1 transaction message (SIMD-0296).
```
dhl: SIMD should be SIMD-0385

src/transaction/message.ts
```diff
@@ -148,4 +163,23 @@ export class TransactionMessage {
       addressLookupTableAccounts,
     });
   }
+
+  /**
+   * Compile to a v1 message (SIMD-0296).
```
dhl: SIMD should be SIMD-0385

src/message/v1.ts
dhl: Suggestion:

If we take advantage of the MESSAGE_DECODER earlier, we should be able to simplify the code a lot more. 

For malformed non-V1 messages, MESSAGE_DECODER.decode would fail early, but that is okay, because a malformed V1 message would fail in a similar way and have consistency in behavior.

We may need to adjust some tests as they were truncated messages with just the version byte.

```suggestion
    const decoded = MESSAGE_DECODER.decode(toUint8ArrayView(serializedMessage));                                                                                                   
     if (decoded.version !== 1) {                                                                                                                                                   
       throw new Error(                                                                                                                                                             
         decoded.version === 'legacy'                                                                                                                                               
           ? 'Expected versioned message but received legacy message'                                                                                                               
           : `Expected versioned message with version 1 but found version ${decoded.version}`,                                                                                      
       );                                                                                                                                                                           
     }                                                                                                                                                                              
     return MessageV1.fromCompiledMessage(decoded);
```

src/transaction/versioned.ts
```diff
@@ -94,6 +112,44 @@ export class VersionedTransaction {
...
+    // shortU16 signature count, which never has its high bit set for any
+    // transaction that fits in a packet.
+    const prefix = serializedTransaction[0];
+    if ((prefix & ~VERSION_PREFIX_MASK) !== 0) {
+      const version = prefix & VERSION_PREFIX_MASK;
+      if (version === 0) {
+        throw new Error(
+          'Version 0 transactions must be encoded with signatures first',
+        );
+      }
```
dhl: Not sure if this check makes sense.

The prudent action to take here is just to error out on the malformed transaction instead of assuming we are working with a version 0 transaction without encoded signatures.

[`solana-transaction`](https://github.com/anza-xyz/solana-sdk/blob/06b8acf9dfc505da943e25274454a95a5bdaa42a/transaction/src/versioned/mod.rs#L397-L463) shows one pretty clean approach to handle this.

test/message-tests/v1.test.ts
```diff
@@ -0,0 +1,343 @@
...
+  });
+
+  it('serializes with the version 1 prefix byte', () => {
+    const message = MessageV1.compile({
+      payerKey: getUniqueAddress(),
+      recentBlockhash: TEST_RECENT_BLOCKHASH,
+      instructions: [],
+    });
+    const serialized = message.serialize();
+    expect(serialized[0]).to.eq((1 << 7) + 1);
```
dhl: Better to compare against the v1 discriminator directly instead of computing it https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0385-transaction-v1.md#versionbyte

```suggestion
    expect(serialized[0]).to.eq(0x81);
```

test/transaction.test.ts
```diff
@@ -1671,6 +1673,82 @@ describe('VersionedTransaction', () => {
...
+          priorityFeeLamports: 5_000n,
+        },
+      }).compileToV1Message();
+      const transaction = new VersionedTransaction(message);
+
+      const serialized = transaction.serialize();
+      const serializedMessage = message.serialize();
+      // message-first envelope: version prefix byte first, then the message,
+      // then one 64-byte signature per required signer with no count prefix
+      expect(serialized[0]).to.eq((1 << 7) + 1);
```
dhl:
```suggestion
      expect(serialized[0]).to.eq(0x81);
```

dhl (approved):
Other than a couple of nits, LGTM.

Thanks @amilz!

src/transaction/versioned.ts
amilz: i can change it but fwiw [kit does this](https://github.com/anza-xyz/kit/blob/4daa5d16ac3c3ff83721dd8149c04d924c4daf59/packages/transactions/src/codecs/transaction-codec.ts#L66-L71)

Issue #3773, opened 2026-01-13
Wrong return type for `connection.getParsedBlock()` when `transactionDetails` is `full`


pedrobranco:
Hello team,

When using the method `connection.getParsedBlock()` with the option `rawConfig` with property `transactionDetails: 'full'`, the return type definition is `ParsedAccountsModeBlockResponse`, which should be the full object `ParsedBlockResponse` according to the JSON RPC documentation, the following comment [here](https://github.com/solana-foundation/solana-web3.js/blob/maintenance/v1.x/src/connection.ts#L563-L570) and the function body [here](https://github.com/solana-foundation/solana-web3.js/blob/maintenance/v1.x/src/connection.ts#L4958):

```
  /**
   * Level of transaction detail to return, either "full", "accounts", "signatures", or "none". If
   * parameter not provided, the default detail level is "full". If "accounts" are requested,
   * transaction details only include signatures and an annotated list of accounts in each
   * transaction. Transaction metadata is limited to only: fee, err, pre_balances, post_balances,
   * pre_token_balances, and post_token_balances.
   */
  transactionDetails?: 'accounts' | 'full' | 'none' | 'signatures';
```

I can open a PR with the fix replacing the default return value to `Promise<ParsedBlockResponse>`.

amilz:
We're moving to [v3](https://github.com/solana-foundation/solana-web3.js/tree/v3.x) and the dep issue is moot, so closing this. If you see this issue on v3 branch please raise again. Thanks.

PR #3866, opened 2026-08-20
fix: add support for reading v1 transactions to v1.x

amilz:
[V1 Transactions are shipping to in Q3.](https://solana.com/upgrades/larger-transaction-sizes). This PR adds support for reading v1 transactions:

- Accept `version: 1` in `getTransaction(s)`/`getBlock` and parsed variants when the
  caller passes `maxSupportedTransactionVersion: 1`
- Add read-only `MessageV1` and surface the v1 `transactionConfig` on json and jsonParsed paths.
- Deserialize v1 wire bytes in `VersionedMessage.deserialize` / `VersionedTransaction.deserialize`
  (version byte first, signatures at the tail).
- Added/updated tests for `MessageV1` and `transactionConfig` and mocked oversized transaction

Closes DEV-930


src/message/v1.ts
```diff
@@ -0,0 +1,233 @@
...
+  return bytes[0] + bytes[1] * 2 ** 8 + bytes[2] * 2 ** 16 + bytes[3] * 2 ** 24;
+}
+
+function decodeU64(byteArray: Array<number>): number {
+  const bytes = guardedSplice(byteArray, 0, 8);
+  let value = BigInt(0);
+  for (let i = bytes.length - 1; i >= 0; i--) {
+    value = (value << BigInt(8)) | BigInt(bytes[i]);
+  }
+  return Number(value);
```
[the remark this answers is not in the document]
amilz: Intentionally consistent w/ this codebase — every u64 read path in v1 is a `number` (`lamports`, `fee`, `prioritizationFee`, `blockTime`)

test/message-tests/v1.test.ts
```diff
@@ -0,0 +1,147 @@
+import {expect} from 'chai';
+
+import {MessageV1} from '../../src/message';
+import {PublicKey} from '../../src/publickey';
+
+function createTestKeys(count: number): Array<PublicKey> {
+  return new Array(count).fill(0).map(() => PublicKey.unique());
+}
+
+describe('MessageV1', () => {
```
mcintyre94, line 35: In case any of it is useful, Kit has a pretty big test suite for this:  https://github.com/anza-xyz/kit/tree/main/packages/transaction-messages/src/codecs/v1/__tests__

amilz: added some coverage [fade28e](https://github.com/solana-foundation/solana-web3.js/pull/3866/commits/fade28ebe93c8ddc554ba116d792bceebfa61912)

src/connection.ts
```diff
@@ -483,10 +490,24 @@ function versionedMessageFromResponse(
       compiledInstructions: response.instructions.map(ix => ({
         programIdIndex: ix.programIdIndex,
         accountKeyIndexes: ix.accounts,
-        data: bs58.decode(ix.data),
+        data: new Uint8Array(bs58.decode(ix.data)),
```
serejke: This is the **v0** branch.

This branch pins `bs58@^4.0.1`, whose `decode` returns a Node `Buffer`. Wrapping it produces a plain `Uint8Array` with copied memory, so any consumer calling `Buffer` methods on `compiledInstructions[].data` breaks **at runtime with no type error** — the declared type was already `Uint8Array`, only the runtime object was ever a `Buffer`. It also adds an allocation and a copy per instruction on the block-decode path.

Line 507 (the v1 branch) needs this. 493 doesn't — suggest reverting it to `bs58.decode(ix.data)`.

src/connection.ts
```diff
@@ -4907,19 +4944,19 @@ export class Connection {
   async getParsedBlock(
     slot: number,
     rawConfig?: GetVersionedBlockConfig,
-  ): Promise<ParsedAccountsModeBlockResponse>;
+  ): Promise<ParsedBlockResponse | null>;
```
serejke: These overloads gaining `| null` — and this one's return type changing from `ParsedAccountsModeBlockResponse` to `ParsedBlockResponse` — are correct. The implementation signature already declares `| null` and returns a `nullable` RPC result, so the published overloads were simply wrong.

My ask is scoping, not the fix. This is #3773, not v1 support, and it breaks every `strictNullChecks` consumer at compile time — which is invisible from the PR title. Please split it into its own PR so it lands with its own changelog entry.

src/message/v1.ts
serejke, line 40: `priorityFee` is a u64 on the wire and `Number()` is lossy above 2^53: `18446744073709551615` decodes as `18446744073709552000`.

The `deserialize transactionConfig (large priority fee)` test asserts exactly `Number.MAX_SAFE_INTEGER`, so it documents the ceiling rather than catching it. #3861 already uses `bigint` for this field.

Fixing this line alone is not enough — the value is mangled twice more upstream on the RPC path: `priorityFee: nullable(number())` in the response struct, and `jayson/lib/client/browser` calling `JSON.parse(responseText, this.options.reviver)` with no reviver passed from `connection.ts`, so precision is already gone before either check runs.

src/message/v1.ts
```diff
@@ -0,0 +1,233 @@
...
+  /** The compute budget configuration of this transaction */
+  transactionConfig?: TransactionConfig | null;
+};
+
+export class MessageV1 {
+  header: MessageHeader;
+  staticAccountKeys: Array<PublicKey>;
+  recentBlockhash: Blockhash;
+  compiledInstructions: Array<MessageCompiledInstruction>;
+  transactionConfig: TransactionConfig | null;
```
serejke: Every v1 message carries a config on the wire, so `| null` creates two representations of "declared nothing": constructed -> `null`, decoded with mask `0` -> an object of four nulls.

That distinction is load-bearing in v1 specifically — unset means **zero** compute units and **zero** loadable account data, i.e. a transaction that cannot execute — so a consumer checking whether a budget was declared needs one unambiguous shape to test.

Suggest always materialising the object and dropping `| null`.

src/message/v1.ts
```diff
@@ -0,0 +1,233 @@
...
+  }
+  return Number(value);
+}
+
+/**
+ * The transaction config of a v1 transaction message, which expresses the
+ * compute budget directly in the message instead of through ComputeBudget
+ * program instructions. Fields the transaction did not set are `null`.
+ */
+export type TransactionConfig = {
```
serejke, line 48: `TransactionConfig` is a broad name for a top-level export sitting beside `GetTransactionConfig`, `GetVersionedTransactionConfig` and `SimulateTransactionConfig`, none of which are related to it.

It also diverges from #3861 on four axes for the same object: field name (`priorityFee` vs `priorityFeeLamports`), type (`number` vs `bigint`), unset marker (`null` vs `undefined`), and all-unset handling (preserved vs collapsed to `undefined`). Worth settling across both PRs before either merges.

src/transaction/constants.ts
```diff
@@ -10,3 +10,5 @@ export const PACKET_DATA_SIZE = 1280 - 40 - 8;
 export const VERSION_PREFIX_MASK = 0x7f;
 
 export const SIGNATURE_LENGTH_IN_BYTES = 64;
+
+export const VERSION_1_MESSAGE_PREFIX = 0x81;
```
serejke, line 14: No v1 size constant lands with this: `PACKET_DATA_SIZE` is still `1232` with no 4096 sibling.

Related, at `src/transaction/versioned.ts:66`, `VersionedTransaction.serialize()` still encodes into a fixed `new Uint8Array(2048)`. Harmless today because `MessageV1.serialize()` throws first — but this PR adds a **3976-byte** v1 fixture, so whoever implements v1 serialization lands on it.

Correcting myself: `@solana/buffer-layout` fails loudly there rather than truncating — `BufferLayout.struct([blob(3000)]).encode(payload, new Uint8Array(2048))` throws `RangeError: encoding overruns Uint8Array`. So the risk is a confusing error that names neither v1 nor the buffer, not silent corruption. Still worth sizing that buffer from a named constant rather than a literal while the format ceiling is moving.

serejke (changes requested):
Read this against SIMD-0385 and cross-checked the decoder against kit's own v1 codec on the PR's 3976-byte fixture — the wire decoding is correct field for field, and the trailing-bytes assertion at the end of `MessageV1.deserialize` is a genuinely good call (#3861 drops it and has a real bug as a result).

Requesting changes on seven things, four of which I'd consider blocking:

- Two unrelated **breaking** changes are bundled in under a v1 title: the `Buffer` -> `Uint8Array` switch on the **v0** instruction-data path (runtime break, no type error) and the `getParsedBlock` overload retyping (compile break for every `strictNullChecks` consumer).
- `priorityFee` is decoded as a `number` from a u64, and the test asserting `Number.MAX_SAFE_INTEGER` documents the cliff rather than catching it.
- `transactionConfig: ... | null` gives "declared nothing" two shapes, which matters more in v1 than elsewhere because unset means zero CU and zero loadable account data.
- `VersionedTransaction.serialize()` still writes into a fixed 2048-byte buffer while this PR adds a 3976-byte v1 fixture.

Separately: this type diverges from the one in #3861 on four axes for the same object. Worth settling across both PRs before either merges, since 1.x users will write against this shape first.

src/message/v1.ts
amilz: 3861 is for v3 branch which includes wholesale migration to bigint. v1.x uses numbers so leaving this to be consistent w/ the rest of the branch.

src/message/v1.ts
amilz: leaving this one as-is

src/transaction/constants.ts
amilz: can add a constant.

src/message/v1.ts
```diff
@@ -0,0 +1,242 @@
...
+function decodeU64(byteArray: Array<number>): number {
+  const bytes = guardedSplice(byteArray, 0, 8);
+  let value = BigInt(0);
+  for (let i = bytes.length - 1; i >= 0; i--) {
+    value = (value << BigInt(8)) | BigInt(bytes[i]);
+  }
+  assert(
+    value <= BigInt(Number.MAX_SAFE_INTEGER),
+    'Expected u64 value to be within the safe integer range',
+  );
```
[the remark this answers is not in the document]
amilz: It's an intentional trade-off to use numbers instead of bigint (as we've already discussed). Since we use numbers, the assert protects against presenting wrong data. This field is for lamports, so this becomes an issue at 2^53 lamports (`Number.MAX_SAFE_INTEGER`), or 9M SOL as a priority fee. I think this is an unreachable edge case, so please @greptile-apps re-review.

src/transaction/constants.ts
serejke: Thanks for adding it — but as it stands the constant is unreferenced. `V1_TRANSACTION_SIZE_LIMIT` occurs exactly once in the repo, on its own declaration line, and since `src/transaction/index.ts` does `export * from './constants'` it ships as public API that documents a limit nothing enforces.

Worth comparing against how `PACKET_DATA_SIZE` earns its place: the v0 serializer sizes its buffers from it (`src/message/v0.ts:290`, `:313`, `:365`) and `src/transaction/legacy.ts:861` enforces it. The v1 equivalent — `src/transaction/versioned.ts:66` — is still `new Uint8Array(2048)`, a literal that is *below* the 4096 the new constant declares.

Nothing breaks today, since this PR can't serialize v1 at all. But shipping the constant unwired is arguably worse than not shipping it: the next person to touch this sees a 4096 limit in the public API and a 2048 literal in the serializer, and has to work out which one is real. Either size that buffer from it, or drop it and leave the sizing to the PR that adds v1 serialization.

src/connection.ts
serejke: Thanks. One leftover: the same wrap is still on the v1 branch at line 507, so `v0` now hands back a `Buffer` and `v1` a `Uint8Array` for the same declared field. Dropping it there too makes them match:

```diff
-        data: new Uint8Array(bs58.decode(ix.data)),
+        data: bs58.decode(ix.data),
```

amilz: thank you @serejke !

jacobcreech (approved):

PR #3883, opened 2026-09-06
fix: honor Connection commitment as Kit client default commitment

amilz:
Threads the Connection's constructor commitment into the internal Kit HTTP client (`createKitRpcClient`) and WS subscriptions runtime so both default to the Connection's commitment instead of always `confirmed`. The subscriptions runtime now assembles its clients from `createSolanaRpcSubscriptionsApi` + transport, since Kit's `createSolanaRpcSubscriptions` factory does not accept a `defaultCommitment`.

Closes DEV-1031


test/kit-commitment-defaults.test.ts
dhl, line 1: Tests should ideally follow an arrange-act-assert format. What we have here is difficult to follow

dhl (changes requested):
Implementation looks fine. Tests could use some work to make it easier for humans to maintain (probably fine for agents).

test/kit-commitment-defaults.test.ts
amilz: refactored to use existing connection/subscription test files following AAA. [bea4354](https://github.com/solana-foundation/solana-web3.js/pull/3883/commits/bea43548a64a68a7507671188f0b0a4369d19b09)

Woody4618 (approved):

PR #3884, opened 2026-09-08
feat: add @solana/wallet-adapter

L0STE:
This adds the wallet-adapter package for web3.js v3: the providers, hooks, UI and error classes people know from
`@solana/wallet-adapter-react`, rebuilt on top of `@solana/kit-plugin-wallet`.

The idea is simple. Kit already does the hard part: it discovers Wallet Standard wallets, handles authorization,
accounts, persistence, reconnect and signing. So this package doesn't do any of that. It only fills the gap
between Kit and a v1 application:

- It converts between web3.js transactions and Kit transactions through wire bytes, and carries across the
  client-side fields wire bytes can't hold (`lastValidBlockHeight`, `nonceInfo`, `minNonceContextSlot`) when the
  wallet leaves the lifetime untouched. It completes a legacy transaction's payer and blockhash before signing
  like v1 did, and forwards the send options Kit's sending signer doesn't.
- It exposes the v1 snapshot from `useWallet()`, keeps the picker selection separate from the active connection,
  and turns every failure into the v1 `WalletError` subclass with the wallet's own message in `cause`.
- It owns the React lifecycle: Kit's browser client is created after commit and disposed on unmount or config
  change, children render on the server and hydrate cleanly, and operation identities stay stable so effects can
  depend on them.
- It ships the v1 UI, rebuilt on a native `<dialog>` and real buttons.

Everything else is Kit. Where Wallet Standard has no equivalent for a v1 concept, we dropped it instead of
imitating it; the list is below, because we'd like your read on it.

The package sits alone under `packages/wallet-adapter` and installs on its own, so it doesn't touch the root
package or collide with #3872. Published dependencies only. One commit on `v3.x`.

## What's inside

`WalletProvider`, `WalletContext`, `useWallet`, `ConnectionProvider`, `useConnection`, `useAnchorWallet`,
`useLocalStorage`. `WalletMultiButton`, `WalletModal`, the connect, disconnect and modal buttons with their `Base*`
variants and headless hooks. Ten `WalletError` subclasses. A `/core` entry exposing `createWalletController`
for non-React consumers, with the same snapshot and operations.

18 source files, about 1,600 lines, plus 1,200 lines of tests. Roughly 330 of the source lines are the v1 UI and 109 an SVG. Two runtime dependencies,
both Wallet Standard helpers.

## Migrating from v1

Imports:

| v1 | v3 |
| --- | --- |
| `@solana/wallet-adapter-react`, `-react-ui`, `-base-ui` | `@solana/wallet-adapter` (UI also at `/ui`) |
| `@solana/wallet-adapter-react-ui/styles.css` | `@solana/wallet-adapter/styles.css` |
| `@solana/wallet-adapter-base` | `@solana/wallet-adapter/core` |
| `@solana/wallet-adapter-wallets` and wallet constructor packages | gone; wallets register themselves through Wallet Standard |

`WalletProvider` props:

| v1 | v3 |
| --- | --- |
| `wallets={[new PhantomWalletAdapter()]}` | discovered; optional `filter` predicate |
| network chosen per adapter | `chain="solana:mainnet"`, required |
| `autoConnect` off by default, could be a callback | boolean, on by default; call `signIn()` yourself |
| `localStorageKey` (`walletName`, stored the picked name) | `storageKey` (`kit-wallet`, stores the authorized account); old value isn't migrated |
| `onError(error, adapter?)` | same; `adapter` carries `name` and `icon` |

Changing `chain`, `filter` or the storage options rebuilds the client and disconnects, so keep them stable.

What behaves differently day to day:

- `select(name)` just records the choice and never throws. It no longer connects; `connect()` does, and rejects
  with `WalletNotSelectedError` or `WalletNotReadyError`. The packaged modal selects and connects in one click.
- `wallet` and `publicKey` describe the live connection; `selectedWallet` is the picker's choice. If a switch is
  rejected, the previous connection stays.
- `signTransaction`, `signAllTransactions` and `signMessage` are `undefined` when the account can't do them;
  `signIn` when the selected wallet can't. Sign-in hands back the Standard output for you to verify.
- Everything returns a promise. Hooks throw `WalletConfigError` when their provider is missing.
- A `WalletError` thrown inside an operation comes out unchanged, so `instanceof` still works. A request that a
  newer one superseded rejects with Kit's `AbortError` and isn't reported anywhere.
- `sendTransaction` fills a legacy transaction's payer and blockhash, applies `options.signers`, checks the
  wallet's supported versions, and passes `skipPreflight`, `preflightCommitment`, `maxRetries` and
  `minContextSlot` to whichever side sends. Extra signers on a `VersionedTransaction` need Kit message signing;
  v3 `Keypair` has it.
- The UI keeps its names and CSS classes. The modal is a native `<dialog>` with no overlay element, so
  `div.wallet-adapter-modal` and `.wallet-adapter-modal-overlay` selectors become `.wallet-adapter-modal`. Menu
  actions are buttons. The trigger does nothing while connecting or disconnecting. A consumer `onClick` runs
  first and `preventDefault` cancels the built-in action.
- React 19, because Kit's plugin requires it.

## What we dropped from v1, and a question for you

These v1 things are not in this package. Each was a deliberate call on our side, but some of them may be
things you'd rather keep, and some may have been dropped upstream on purpose already. We'd like to know which
is which before this lands.

- Per-wallet adapter objects with `adapter.on('connect' | 'disconnect' | 'error' | 'readyStateChange')`. Every
  Standard wallet shares one implementation now, so the snapshot is the only event source.
- `adapter.url` and the install links v1 UIs built from it. Wallet Standard has no install URL; keeping this
  would mean maintaining a name-to-URL list, which is the kind of code this package is trying not to own.
- `readyState` beyond `Installed`, and the modal's "More options" grouping. A discovered wallet is by definition
  installed.
- `supportedTransactionVersions` on the wallet object. We check it before signing, we just don't expose it.
- `select()` auto-connecting when `autoConnect` is set. The modal connects explicitly instead; a custom picker
  calls `connect()`.
- The `walletName` localStorage cache and its migration. Kit stores the authorized account instead.
- Adapter type aliases (`Adapter`, `WalletAdapter*`, `Signer*WalletAdapter*`, `WalletName`,
  `isWalletAdapterCompatibleStandardWallet`) and the error classes only adapter constructors threw
  (`WalletLoadError`, `WalletTimeoutError`, `WalletWindow*Error`, `WalletAccountError`, `WalletPublicKeyError`,
  `WalletKeypairError`, `WalletDisconnectedError`).
- `Base*WalletAdapter` classes, `EventEmitter`, wallet constructors and the aggregate wallets package.
- Automatic Mobile Wallet Adapter and WalletConnect setup. Both have maintained Wallet Standard integrations
  that apps register once; after that they're discovered like any other wallet.
- React 18.

`useAnchorWallet` is kept with Anchor's shape, but published Anchor still depends on web3.js v1, so it will only
work once Anchor ships a v3 build. We didn't want to claim more than that.

## How we checked it

Typecheck (sources and tests), build, 33 unit tests and prettier with the repo config, all in CI. The tests
register real Wallet Standard wallets and drive Kit's real store; nothing in Kit is mocked. CI also packs the
package and installs it with npm into a fresh app at the declared peer ranges, since that's where the previous
attempt broke.

We also ran it in a real Chromium (Vite plus Playwright, not part of the PR) with an in-page Wallet Standard
wallet that rejects any request not carrying its own account object, the way reference wallets do. That covered
the modal's focus and Escape handling, a wallet rejection showing its own message and reaching `onError` once,
connect, `signMessage` with the identity check, legacy preparation with every send option reaching the wallet,
and the account menu's focus, Escape, mouse ordering and outside-click behaviour, with no console errors.

We have not clicked through Phantom or Solflare as extensions, and we have not run it in Safari.

## Kit bugs we found on the way

None of these affect this package as shipped, but they'll bite anyone using the plugin directly.

- `kit-plugin-wallet` keys wallet state (subscriptions, the disconnect marker, availability) by `wallet.name`.
  The registry dedups by object identity, so two wallets with the same name are legal, and then disconnecting
  one disconnects the other.
- `kit-plugin-wallet`'s `signMessage` passes the `UiWalletAccount` handle to the wallet's feature instead of
  the wallet's own account object. Wallets that compare the account by identity, which the reference wrapper
  does, throw `invalid account`. `signIn` and the account signer already convert it.
- `wallet-account-signer` reuses an input `lifetimeConstraint` whenever the token string matches, ignoring the
  constraint kind and the nonce account, so a wallet that keeps the nonce value but changes the account returns
  a stale constraint.
- The sending signer forwards only `minContextSlot`. Wallet Standard also accepts `skipPreflight`,
  `preflightCommitment` and `maxRetries`, and v1 sent them; this package calls the feature directly for that.
- A superseded connect or sign-in is signalled only as a `DOMException` whose message says "superseded". There's
  no way to tell it from a wallet's own `AbortError` without matching the text.

When #3872 lands, moving this package into the workspace is a lockfile deletion and a `workspace:` range.
Closes DEV-1043


packages/wallet-adapter/src/wallet-controller.ts
```diff
@@ -0,0 +1,462 @@
...
+
+const base58 = getBase58Decoder();
+
+/** v1 passed its own errors through unchanged and wrapped everything else. */
+function wrap<T extends WalletError>(
+  error: unknown,
+  Wrapper: new (message: string, cause: unknown) => T,
+  message: string,
+): WalletError {
+  if (error instanceof WalletError) return error;
```
amilz, line 76: I think you could narrow this a bit for known errors from @solana/error (in kit). if adopted update tests.

```ts
isSolanaError(error, SOLANA_ERROR__WALLET__NOT_CONNECTED)
    ? WalletNotConnectedError
    : isSolanaError(error, SOLANA_ERROR__WALLET__NO_SIGNER_CONNECTED) ||
        isSolanaError(error, SOLANA_ERROR__WALLET__SIGNER_NOT_AVAILABLE)
      ? WalletNotReadyError
      : Wrapper;
```

packages/wallet-adapter/package.json
```diff
@@ -0,0 +1,96 @@
...
+    "compile:js": "tsc --build",
+    "clean": "rm -rf dist *.tsbuildinfo",
+    "prepack": "pnpm run compile:js",
+    "test:unit": "vitest run",
+    "test:typecheck": "tsc --noEmit && tsc --noEmit -p tsconfig.test.json",
+    "test:install": "./scripts/check-install.sh"
+  },
+  "peerDependencies": {
+    "@solana/kit": "^8.2.0",
+    "@solana/kit-plugin-wallet": "^0.19.0",
```
amilz: wdyt about putting `@solana/kit-plugin-wallet` as a dep rather than peer? 
i suspect generally it's not going to be installed seperately so version conflicts are unlikely. as long as our package stays up to date with kit...we should not diverge from kit's version req and the associated plugin. just would be nice to not have to ask folks to install this.

packages/wallet-adapter/src/ui/WalletModal.tsx
```diff
@@ -0,0 +1,127 @@
...
+      >
+        <div className="wallet-adapter-modal-container">
+          <div className="wallet-adapter-modal-wrapper">
+            <button
+              type="button"
+              aria-label="Close wallet selection"
+              onClick={() => setVisible(false)}
+              className="wallet-adapter-modal-button-close"
+            >
+              <svg width="14" height="14" aria-hidden="true">
```
amilz, line 93: can we make this a const so it's just easier to know what it is

packages/wallet-adapter/src/useAnchorWallet.ts
```diff
@@ -0,0 +1,30 @@
...
+  signAllTransactions<T extends Transaction | VersionedTransaction>(
+    transactions: T[],
+  ): Promise<T[]>;
+  signTransaction<T extends Transaction | VersionedTransaction>(
+    transaction: T,
+  ): Promise<T>;
+}
+
+/** The connected wallet as Anchor's `Wallet` interface, or `undefined` when it cannot sign. */
+export function useAnchorWallet(): AnchorWallet | undefined {
```
amilz, line 30: Per your comment in the PR--this going to be problematic until we have anchor v2. WDYT about a deprecated warning just to get the strikethrough effect in ppl's code's and we can have the tooltip explain they users will have to manually implement for now if they want to use this.

packages/wallet-adapter/package.json
```diff
@@ -0,0 +1,96 @@
+{
+  "name": "@solana/wallet-adapter",
+  "version": "0.1.0",
```
amilz: IMO let's mirror web3.js and we'll bump them together going forward (like how kit does it). 
This will also help differentiate from the legacy @solana/wallet-adapter-* packages.

```suggestion
  "version": "3.0.0",
```

packages/wallet-adapter/src/ui/WalletModal.tsx
```diff
@@ -0,0 +1,130 @@
...
+  );
+  const {wallets, connecting, selectedWallet, select, connect} = useWallet();
+  const {setVisible} = useWalletModal();
+  const [portal, setPortal] = useState<Element | null>(null);
+  const [error, setError] = useState<string | null>(null);
+  useLayoutEffect(
+    () => setPortal(document.querySelector(container)),
+    [container],
+  );
+  useLayoutEffect(() => {
```
mcintyre94, line 39: I'm not sure how well it works with react, but this would be enormously simpler if you could use the native dialog open/close commands: https://developer.mozilla.org/en-US/docs/Web/API/Invoker_Commands_API 

I think it should work since this is self contained and no outside react components need to access its state etc

packages/wallet-adapter/src/ui/WalletMultiButton.tsx
```diff
@@ -0,0 +1,178 @@
...
+  } = useWalletMultiButton({onSelectWallet: () => setVisible(true)});
+  const [menuOpen, setMenuOpen] = useState(false);
+  const [copied, setCopied] = useState(false);
+  const [error, setError] = useState<string | null>(null);
+  const ref = useRef<HTMLDivElement>(null);
+  const address = publicKey?.toBase58();
+  // A request superseded by a newer one rejects with AbortError; nothing failed.
+  const showError = (error: Error) => {
+    if (error.name !== 'AbortError') setError(error.message);
+  };
```
mcintyre94, line 43: Optional/FYI: we have `isAbortError` in Kit for this purpose

packages/wallet-adapter/src/types.ts
```diff
@@ -0,0 +1,110 @@
...
+import type {
+  SolanaSignOffchainMessageInput,
+  SolanaSignOffchainMessageOutput,
+} from '@solana/wallet-standard-features';
+
+/** A wallet discovered through Wallet Standard. */
+export type UiWallet = WalletState['wallets'][number];
+
+/** An account belonging to a {@link UiWallet}. */
+export type UiWalletAccount = NonNullable<WalletState['connected']>['account'];
```
mcintyre94: Optional: these types are defined in `@wallet-standard/ui`

packages/wallet-adapter/src/types.ts
```diff
@@ -0,0 +1,110 @@
...
+
+/** Input accepted by Sign In With Solana. */
+export type SignInInput = Parameters<WalletNamespace['signIn']>[1];
+
+/** Output returned by Sign In With Solana. */
+export type SignInOutput = Awaited<ReturnType<WalletNamespace['signIn']>>;
```
mcintyre94: Optional: feature related types are available from `@solana/wallet-standard-features`

packages/wallet-adapter/src/wallet-controller.ts
```diff
@@ -0,0 +1,523 @@
...
+}
+
+/** Kit rejects a request superseded by a newer connect or sign-in; nothing failed. */
+function superseded(error: unknown): boolean {
+  // Kit's only signal is this DOMException; a wallet's own AbortError is a failure and is reported.
+  return (
+    error instanceof DOMException &&
+    error.name === 'AbortError' &&
+    error.message.includes('superseded')
+  );
```
mcintyre94, line 94: Again you could use `isAbortError` from Kit for this if you like

packages/wallet-adapter/src/wallet-controller.ts
```diff
@@ -0,0 +1,523 @@
...
+        connected?.wallet,
+      );
+    }
+  }
+  const canSignOffchainMessages = () =>
+    active()?.account.features.includes('solana:signOffchainMessage') ?? false;
+  async function signOffchainMessage(
+    message: string,
+    options: SignOffchainMessageOptions = {},
+  ): Promise<SignOffchainMessageOutput> {
```
mcintyre94, line 262: FYI we'll be adding this to the Kit plugin, which should let you simplify this. https://github.com/anza-xyz/kit-plugins/pull/414

packages/wallet-adapter/src/wallet-controller.ts
```diff
@@ -0,0 +1,523 @@
...
+      if (!connected || !canSignMessages()) {
+        throw new WalletNotReadyError(
+          'The connected wallet cannot sign messages.',
+        );
+      }
+      // Wallets compare the account by identity, so pass their own account object, not Kit's UI handle.
+      const feature = getWalletAccountFeature(
+        connected.account,
+        'solana:signMessage',
+      ) as SolanaSignMessageFeature['solana:signMessage'];
```
mcintyre94: FYI this is fixed in the kit plugin: https://github.com/anza-xyz/kit-plugins/pull/415 (publishing as 0.20.0)

packages/wallet-adapter/src/wallet-controller.ts
```diff
@@ -0,0 +1,523 @@
...
+  // Kit's only signal is this DOMException; a wallet's own AbortError is a failure and is reported.
+  return (
+    error instanceof DOMException &&
+    error.name === 'AbortError' &&
+    error.message.includes('superseded')
+  );
+}
+
+/** v1 refused transaction versions the wallet does not advertise instead of letting the wallet fail. */
+function assertSupportedVersions(
```
mcintyre94, line 98: FYI as of 0.20.0 of the wallet plugin (publishing now), we have `connected.supportedTransactionVersions`, which gives the intersection of `supportedTransactionVersions` for transaction features for the connected account.

You may want to keep this function as-is, because it's selecting for the specific feature which is more precise if the supported versions differ between features

But I also noticed that in your description you mentioned not exposing `supportedTransactionVersions`. I'd suggest exposing the kit-plugins intersection on your adapter, because wallet support for v1 transactions is going to be rolling out. See https://github.com/solana-foundation/solana-com/pull/2081 - Foundation plan to advise devs to check this field. So I think a new wallet-adapter-like library should probably include it.

packages/wallet-adapter/README.md
```diff
@@ -0,0 +1,112 @@
+# `@solana/wallet-adapter`
+
+Wallet-adapter hooks and UI for [`@solana/web3.js`](https://github.com/solana-foundation/solana-web3.js) v3 applications.
+
+The familiar v1 providers, hooks, components and error classes, implemented on [`@solana/kit-plugin-wallet`](https://github.com/anza-xyz/kit-plugins). Wallets are discovered through Wallet Standard registration in the browser; there are no per-wallet adapter constructors. Coming from `@solana/wallet-adapter-react`? The migration notes are in the pull request that added this package.
```
mcintyre94: I'd suggest you need to put the migration notes somewhere stable in the repo/this readme, or at the very least link to this PR so people don't have to use git blame/search to find it

packages/wallet-adapter/src/wallet-controller.ts
```diff
@@ -0,0 +1,523 @@
...
+            );
+          }
+          await transaction.sign(messageSigners);
+        } else {
+          await transaction.partialSign(...signers);
+        }
+      }
+      if (sending) {
+        // Kit's sending signer forwards only minContextSlot; v1 forwarded every send option, so call the feature.
+        const {maxRetries, minContextSlot, ...standardOptions} = sendOptions;
```
mcintyre94, line 389: I'll look into whether we can align the Kit signer with the feature here

packages/wallet-adapter/src/wallet-controller.ts
amilz: @L0STE +1 that point. As tx v1 is rolling out I think apps that design around v1 will want to screen wallets before having the user connect (e.g. to prompt user to dl a working wallet or upgrade an existing wallet to latest version). 
Example of how this could be implemented before a connect: https://github.com/solana-foundation/transaction-v1-examples/blob/main/ts/wallet-table/src/wallets.ts

packages/wallet-adapter/README.md
amilz: I can add a doc like we have in web3.js

packages/wallet-adapter/src/types.ts
amilz: [45f119d](https://github.com/solana-foundation/solana-web3.js/pull/3884/commits/45f119d2d29fa8d1c69c23ca319cb24c10d44f23)

packages/wallet-adapter/src/types.ts
amilz: [45f119d](https://github.com/solana-foundation/solana-web3.js/pull/3884/commits/45f119d2d29fa8d1c69c23ca319cb24c10d44f23)

packages/wallet-adapter/src/wallet-controller.ts
amilz: [45f119d](https://github.com/solana-foundation/solana-web3.js/pull/3884/commits/45f119d2d29fa8d1c69c23ca319cb24c10d44f23)

packages/wallet-adapter/src/wallet-controller.ts
amilz: added in [45f119d](https://github.com/solana-foundation/solana-web3.js/pull/3884/commits/45f119d2d29fa8d1c69c23ca319cb24c10d44f23)

packages/wallet-adapter/README.md
amilz: [45f119d](https://github.com/solana-foundation/solana-web3.js/pull/3884/commits/45f119d2d29fa8d1c69c23ca319cb24c10d44f23)

packages/wallet-adapter/src/ui/WalletMultiButton.tsx
amilz: vitest gave some issues so i left it out. I'm fine either way.

packages/wallet-adapter/src/ui/WalletModal.tsx
amilz: passing for now. can reevaluate in future

packages/wallet-adapter/src/wallet-controller.ts
```diff
@@ -0,0 +1,514 @@
...
+    active()?.account.features.includes('solana:signMessage') ?? false;
+  async function signMessage(message: Uint8Array): Promise<Uint8Array> {
+    const connected = active();
+    try {
+      if (!connected || !canSignMessages()) {
+        throw new WalletNotReadyError(
+          'The connected wallet cannot sign messages.',
+        );
+      }
+      return await namespace.signMessage(message);
```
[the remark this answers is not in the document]
amilz: See https://github.com/anza-xyz/kit-plugins/pull/415

justfile
```diff
@@ -106,11 +106,39 @@ test-live:
...
+    cd packages/web3.js
+    pkill -f solana-test-validator || true
+    rm -rf test-ledger
+    ./scripts/start-shared-test-validator.sh &
+    validator_script_pid=$!
+    trap 'pkill -f solana-test-validator || true; kill $validator_script_pid 2>/dev/null || true' EXIT
+    cd ../..
+    until [ "$(curl -sf -m 2 http://127.0.0.1:8899/health || true)" = "ok" ]; do sleep 1; done
+    slot() { curl -sf -m 2 -X POST -H 'content-type: application/json' -d '{"jsonrpc":"2.0","id":1,"method":"getSlot"}' http://127.0.0.1:8899 | jq -r .result; }
+    until [ "$(slot || echo 0)" -ge 32 ]; do sleep 1; done
```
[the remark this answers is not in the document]
amilz: [68aed2a](https://github.com/solana-foundation/solana-web3.js/pull/3884/commits/68aed2afb729d462c6a7e5c323d1c90bcde02af9)

packages/wallet-adapter/src/useLocalStorage.ts
```diff
@@ -0,0 +1,37 @@
...
+  const value = state[0];
+  const initial = useRef<{key: string; value: T} | null>({key, value});
+  useEffect(() => {
+    // Mounting (including StrictMode) must not overwrite storage with a fallback.
+    if (initial.current?.key === key && Object.is(initial.current.value, value))
+      return;
+    initial.current = null;
+    try {
+      if (value === null) globalThis.localStorage?.removeItem(key);
+      else globalThis.localStorage?.setItem(key, JSON.stringify(value));
```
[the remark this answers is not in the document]
amilz: [68aed2a](https://github.com/solana-foundation/solana-web3.js/pull/3884/commits/68aed2afb729d462c6a7e5c323d1c90bcde02af9)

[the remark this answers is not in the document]
amilz: [68aed2a](https://github.com/solana-foundation/solana-web3.js/pull/3884/commits/68aed2afb729d462c6a7e5c323d1c90bcde02af9)

packages/wallet-adapter/src/wallet-controller.ts
```diff
@@ -0,0 +1,514 @@
...
+  function resolve(name = selectedName): UiWallet | undefined {
+    const state = namespace.getState();
+    if (name === undefined) {
+      const account = state.connected?.account ?? state.reconnectingTo;
+      return account
+        ? state.wallets.find(wallet => wallet.accounts.includes(account))
+        : undefined;
+    }
+    const matches = state.wallets.filter(wallet => wallet.name === name);
+    return matches.length === 1 ? matches[0] : undefined;
```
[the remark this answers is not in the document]
amilz: [68aed2a](https://github.com/solana-foundation/solana-web3.js/pull/3884/commits/68aed2afb729d462c6a7e5c323d1c90bcde02af9)

PR #3888, opened 2026-09-10
refactor: sign transactions with transaction signers

amilz:
- `Transaction.sign` / `partialSign` and `VersionedTransaction.sign` delegate to Kit's `partiallySignTransactionWithSigners` via one shared helper in `kit-adapters/signing.ts`. Kit handles signer dedupe and signature merging.
- `Signer` is now `MessagePartialSigner & TransactionPartialSigner`, matching what v1's secret-key `Signer` could do. 
- Now passes `TransactionPartialSigner` into each transaction signing method instead of `Signer` or `MessagePartialSigner`

Alternative that Supersedes #3886.


packages/web3.js/src/transaction/legacy.ts
```diff
@@ -753,63 +751,40 @@ export class Transaction {
...
+    for (const {publicKey} of signers) {
+      const signature = signatures[publicKey.toBase58()];
+      if (signature != null) {
+        this._addSignature(publicKey, signature);
       }
-      return {
-        nonce: blockhashAsNonce(this.nonceInfo.nonce),
-        nonceAccountAddress: nonceAccountAddress.toBase58(),
-      };
     }
```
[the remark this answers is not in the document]
amilz: [b42583c](https://github.com/solana-foundation/solana-web3.js/pull/3888/commits/b42583c74c1baf76def58f9f9c67ed94d6b72c7e)

packages/web3.js/src/kit-adapters/signing.ts
```diff
@@ -1,148 +1,123 @@
...
-  const signatureMap: KitTransaction['signatures'] = {};
-  for (const publicKey of requiredSignerPublicKeys) {
-    signatureMap[publicKey.toBase58()] = null;
+  const signed = await partiallySignTransactionWithSigners(
+    dedupedSigners,
+    transaction,
+  );
+
+  const result: Record<string, Uint8Array> = {};
+  for (const [address, signature] of Object.entries(signed.signatures)) {
```
dhl: I'm a bit concerned about using `partiallySignTransactionWithSigners` here.

My understanding is that if one of the signers implement `TransactionModifyingSigner`, then the signers will be signing over the modified transaction, producing a signature for a different transaction, breaking the API here.

IMO the correct way to do this is is to call `signTransactions` directly:

```suggestion
   const signatureDictionaries = await Promise.all(
    dedupedSigners.map(async signer => {
      const [dictionary] = await signer.signTransactions([transaction]);
      return dictionary;
    }),
  );
  const signedSignatures = signatureDictionaries.reduce<
    KitTransaction['signatures']
  >(
    (merged, dictionary) => ({...merged, ...dictionary}),
    transaction.signatures,
  );

  const result: Record<string, Uint8Array> = {};
  for (const [address, signature] of Object.entries(signedSignatures)) {
```

@lorisleiva @mcintyre94 could you weigh in here?

packages/web3.js/src/keypair.ts
```diff
@@ -14,11 +14,9 @@ import {PublicKey} from './publickey';
 import {toPackedUint8Array} from './utils/typed-array';
 
 /**
- * Union of signer shapes accepted by web3.js transaction signing APIs.
- * Dispatch is documented on `signTransactionMessageBytes` in
- * `src/kit-adapters/signing.ts`.
+ * A Kit signer that can sign both transactions and arbitrary messages.
  */
-export type Signer = MessagePartialSigner | TransactionPartialSigner;
+export type Signer = MessagePartialSigner & TransactionPartialSigner;
```
dhl, line 19: Nice fix. Signer should signal the Kit-compatible signing capabilities we require. The union was definitely incorrect.

packages/web3.js/src/loader.ts
dhl, line 1: Both Loader and BpfLoader should take the narrower TransactionPartialSigner

packages/web3.js/src/kit-adapters/signing.ts
lorisleiva: 100% agree. If this API is only dealing with `TransactionPartialSigners` then it must only use the signature dictionaries otherwise you'd end up with signatures for the wrong (modified) message.

packages/web3.js/src/loader.ts
amilz: [f41a5bb](https://github.com/solana-foundation/solana-web3.js/pull/3888/commits/f41a5bba5c43451f27d8cbed9e101ad1e6474c77)

packages/web3.js/src/kit-adapters/signing.ts
amilz: thanks guys. updated in [bbaf734](https://github.com/solana-foundation/solana-web3.js/pull/3888/commits/bbaf7342c691b21cf0274f1d8be36af5b85205b9)

PR #3892, opened 2026-09-12
fix(web3.js): reconcile signature status before reporting blockheight expiry (SOLA8-99)

amilz:
Match how we handle nonce confirmation and do a final `getSignatureStatus` check after blockheight expiration for late landing transactions.

Closes APE-254


dev-jodee (approved):
Reconcile signature status before reporting blockheight expiry (SOLA8-99). CI green.

PR #3893, opened 2026-09-12
fix(web3.js): copy message header at construction so caller mutation cannot change signed bytes (SOLA8-46)

amilz:
Copy & freeze message header to avoid mutation.

Closes APE-258


dev-jodee (approved):
Copy message header at construction (SOLA8-46). CI green.

PR #3894, opened 2026-09-12
fix(web3.js): reject instructions carrying both Kit and legacy fields (SOLA8-86)

amilz:
Reject if malformed instruction contains legacy and kit fields. 

Closes APE-252
Closes APE-253


dev-jodee (approved):
Reject instructions with both Kit and legacy fields (SOLA8-86). CI green.

PR #3895, opened 2026-09-12
chore(wallet-adapter): add assertion for endpoint and chain

amilz:
Ensures endpoint and chain identifier do not blatantly mismatch to avoid confusion of sending to wrong cluster.
Edit: update to only check against public endpoints. We can make this more restrictive in the future if requested.

Closes APE-261


dev-jodee (approved):
Wallet-adapter assertion for endpoint/chain validation. CI green.

PR #3896, opened 2026-09-12
fix(wallet-adapter): skip redundant wallet reconnects

amilz:
- Return early from `connect()` when the selected wallet is already active.
- Restore v1 behavior, where connecting an already-connected adapter was a no-op.
- Add regression coverage ensuring redundant connects do not invoke the wallet again. 

Closes APE-262, APE-263


dev-jodee (approved):
Skip redundant wallet reconnects. CI green.

PR #3897, opened 2026-09-12
fix(wallet-adapter): preserve concurrent transaction lifetime

amilz:
avoid overwriting a legacy transaction lifetime if the caller completes it
while `sendTransaction` is awaiting `getLatestBlockhash`.

Add regression coverage for a caller-provided blockhash and block height.

Refs: SOLA8-2
Closes: APE-264


amilz:
@greptile-apps try again

dev-jodee (approved):
Preserve concurrent transaction lifetime in wallet-adapter. CI green.

Issue #3849, opened 2026-07-16
sendAndConfirmRawTransaction silently drops maxRetries


MicroMilo:
## Summary

ConfirmOptions explicitly accepts maxRetries, and Connection.sendEncodedTransaction translates a received maxRetries into the RPC config. sendAndConfirmRawTransaction reconstructs SendOptions but omits only maxRetries, whereas sendAndConfirmTransaction forwards it.

Impact: Calling sendAndConfirmRawTransaction with maxRetries set produces a sendOptions object without that property. Connection.sendRawTransaction therefore cannot place maxRetries in the sendTransaction RPC configuration, and the node uses its default retry policy instead of the caller's requested value.

## Code path

Checked against current `maintenance/v1.x` at commit `5b4e63daed1561ce58585a639041732c04aa354a`.

- `src/connection.ts:254-265`
- `src/utils/send-and-confirm-raw-transaction.ts:79-88`
- `src/utils/send-and-confirm-transaction.ts:29-40`
- `src/connection.ts:6020-6045`

## Steps to reproduce

Validation level: current HEAD supplemental dynamic witness reproduced.

```bash
node -e "const fs=require('fs');const s=fs.readFileSync('src/utils/send-and-confirm-raw-transaction.ts','utf8');const b=s.match(/const sendOptions = options && \{([\s\S]*?)\n  \};/)[1];const forwarded=[...b.matchAll(/^\s+(\w+):/gm)].map(m=>m[1]);console.log(JSON.stringify({helper:'sendAndConfirmRawTransaction',accepted:'maxRetries',forwarded,containsMaxRetries:forwarded.includes('maxRetries')}));if(!forwarded.includes('maxRetries'))process.exitCode=1;"
```

Observed output:

```text
{"helper":"sendAndConfirmRawTransaction","accepted":"maxRetries","forwarded":["skipPreflight","preflightCommitment","minContextSlot"],"containsMaxRetries":false}
```

Neighboring control:

```bash
[...]

PR #3901, opened 2026-09-14
fix: forward maxRetries from sendAndConfirmRawTransaction

### LINKED ISSUE #3902: sendAndConfirmRawTransaction still drops maxRetries on v3.x
Same bug as #3849, still present on `v3.x`.

`ConfirmOptions` includes `maxRetries`, and `sendAndConfirmTransaction` already puts it on `SendOptions`. `sendAndConfirmRawTransaction` rebuilds that object without `maxRetries`, so `Connection.sendRawTransaction` never sends the caller's retry count. The node falls back to its default retry policy.

On current `v3.x`:

```ts
const sendOptions = options && {
  skipPreflight: options.skipPreflight,
  preflightCommitment: options.preflightCommitment || options.commitment,
  minContextSlot: options.minContextSlot,
};
```

`sendAndConfirmTransaction` on the same branch has `maxRetries: options.maxRetries` in that object.

#3849 was closed as v1.x maintenance-only. I don't have permission to reopen it, so this is the v3.x report. Fix is in #3901.

o-mid:
#### Problem

`sendAndConfirmRawTransaction` accepts `ConfirmOptions.maxRetries` but rebuilds `SendOptions` without it. `Connection.sendRawTransaction` therefore never puts the caller's retry count on the RPC, and the node uses its default retry policy. `sendAndConfirmTransaction` already forwards `maxRetries`.

This is the same drop as #3849, still present on `v3.x`. Filed as #3902 because #3849 was closed as v1.x maintenance-only.

#### Summary of Changes

Forward `options.maxRetries` the same way `sendAndConfirmTransaction` does. Added a stub test that `maxRetries: 7n` reaches `sendRawTransaction`.

Fixes #3902


koriyoshi2041 (approved):
Verified at `0adf3aaf`. The raw helper now mirrors the signed helper by forwarding `maxRetries`, and the stub regression covers the caller-supplied value reaching `sendRawTransaction`.

Local checks: 3 focused tests on Node 22, package typecheck, ESLint, Prettier, and diff hygiene all pass.

PR #3903, opened 2026-09-14
fix: report raw confirmation failures as send errors

o-mid:
#### Problem

`sendAndConfirmRawTransaction` throws `SendTransactionError` after `confirmTransaction` returns `status.err`. That means the transaction was already sent. The helper still set `action` to `simulate` whenever `skipPreflight` was false, so the error said `Simulation failed` and hid the signature. `sendAndConfirmTransaction` always uses `action: 'send'` here.

#### Summary of Changes

Always report a confirmation failure as a send error. Added a stub test with `skipPreflight: false` that expects the send-error message.


amilz (approved):

amilz:
Thanks @o-mid

PR #3905, opened 2026-09-14
fix: require transaction signers for versioned extra signers

o-mid:
#### Problem

`sendTransaction` filters extra signers on a `VersionedTransaction` for `signMessages`. `VersionedTransaction.sign` calls `signTransactions`. A transaction-only Kit signer is rejected. A message-only object passes the check and then fails inside `sign`.

#### Summary of Changes

Require `signTransactions` and pass those signers through. Updated the release note. Added a unit test with a transaction-only extra signer.


amilz:
@o-mid rebased and updated the filter to leverage kit's `isTransactionPartialSigner` check ([src](https://github.com/anza-xyz/kit/blob/1b931d7420947f6e0c29ac1107a3673599733b67/packages/signers/src/transaction-partial-signer.ts#L74-L78)):         

```ts
const transactionSigners = signers.filter(isTransactionPartialSigner);
```

thank you!

amilz (approved):

PR #3907, opened 2026-09-14
fix(web3.js): confirm inherited confirmation strategies in raw send-and-confirm

amilz:
`sendAndConfirmRawTransaction()` discriminated confirmation strategies with own-property checks, so a strategy whose properties come from its prototype was treated as ConfirmOptions and downgraded to legacy timeout confirmation. Use `in` checks, matching Connection.confirmTransaction().

Also add coverage for a plus an explicit `maxRetries: 0n`.

Closes APE-280
SOLA8-17
SOLA8-23


dev-jodee (approved):
confirm inherited confirmation strategies; CI green

PR #3912, opened 2026-09-15
fix(web3.js): validate discriminator in decodeExtendLookupTable

amilz:
`decodeExtendLookupTable` decoded instruction data without checking the instruction discriminator, so a `CreateLookupTable` instruction was accepted and returned as an extension with an empty address list. 
web3.js v1 rejected this via `decodeData`'s instruction index check.

SOLA8-55
Closes APE-284


dev-jodee (approved):
Adds discriminator validation to decodeExtendLookupTable. Prevents instruction type confusion with test coverage.

PR #3913, opened 2026-09-15
fix(web3.js): reject v1 config masks with unsupported bits

amilz:
The v1 message decoder ignored unrecognized config-mask bits, consuming no bytes for the values those bits describe and shifting instruction parsing. Crafted bytes therefore deserialized into a different instruction layout and reserialized to different bytes. web3.js v1 rejects unknown mask bits.

SOLA8-83
Closes DEV-1085


dev-jodee (approved):
Validates transaction config mask to reject unsupported bits. Covers edge case with test.

dev-jodee (approved):
Config mask validation fix. SOLA8-83 closed, CI green.

PR #3915, opened 2026-09-15
fix(web3.js)/block subscribe fields

amilz:
- Report `loadedAddresses` only when the RPC supplies it instead of fabricating empty readonly/writable arrays for encodings that omit lookup metadata.
- The `blockSubscribe` wire field is `showRewards`; emitting `rewards` left the option unset so the server always included block rewards.

Closes DEV-1089 DEV-1088


dev-jodee (approved):
Block subscribe field fixes (DEV-1089, DEV-1088). Omit fabricated loadedAddresses, send rewards as showRewards wire field. Tests added, CI green.

PR #3899, opened 2026-09-13
fix(web3.js): improve websocket subscription recovery

amilz:
- Share one WebSocket pool across subscription types
- Restore Kit’s default subscription sharding
- Clean up failed subscriptions so retries work
- Add focused WebSocket integration tests

Closes APE-270 APE-269 APE-268 APE-267

PR #3917, opened 2026-09-16
fix(web3.js): release local subscription handles on auto-disposal

amilz:
- When the RPC auto-closes a signature subscription after the final status notification, we skipped local cleanup and left the abort controller, Kit iterator, and channel-pool slot alive. Every `confirmTransaction` on a reused `Connection` leaked one handle. We now abort the local handle on that path. (SOLA8-28)
- Adds a regression test proving a rejected subscription open releases its local handle. The leak itself was already fixed in #3899. (SOLA8-50)

Closes DEV-1095 DEV-1094


dev-jodee (approved):
Fix subscription handle leak on auto-disposal; CI green.

PR #3918, opened 2026-09-16
fix(web3.js): honor abort signal during durable-nonce reconciliation

amilz:
Race the post-advance signature status lookup and its retry delay against the cancellation promise so an aborted confirmation rejects with the abort reason instead of settling as a nonce expiry.

Closes DEV-1096


dev-jodee (approved):
Honor abort signal in nonce reconciliation; CI green.

PR #3919, opened 2026-09-16
chore(web3.js): compiled instruction data into the decompiled instruction

amilz:
Copy compiled instruction data into the decompiled instruction, and store a copy of the caller's message in Transaction.populate.

Closes DEV-1066 DEV-1097


dev-jodee (approved):
Shallow copy fix for message aliasing. Decompile instruction data (SOLA8-61) and populate message accounts (SOLA8-75) now safe from mutations. Tests cover both paths. CI green.

PR #3920, opened 2026-09-16
fix(wallet-adapter): validate send options before send txn

amilz:
Send options bound for the wallet were checked only after the caller's legacy transaction had been given a fee payer, a lifetime and any additional signatures, so a refused option left those fields committed on an object the caller can reuse. The check now runs before any of that work.

SOLA8-88
Closes DEV-1100


dev-jodee (approved):
Validate send options before transaction state commit. Prevents caller reuse of incompletely-configured txns. CI green.

PR #3921, opened 2026-09-17
fix(wallet-adapter): ignore a connect request superseded by a new selection

amilz:
The picker closed on any connect request it still owned, so a request started for one wallet could close the modal (or report its failure) after a selection made outside the picker had already moved on to another wallet. Completion now also requires the selection to still be the one the request was made for.

Closes APE-295 SOLA8-89


dev-jodee (approved):
Wallet adapter: picker only closes on its own connect request, not superseded ones. Prevents modal closure race when selection moves to another wallet. CI green.

PR #3922, opened 2026-09-17
fix(web3.js): reject ambiguous token and program account filters

amilz:
Reject ambiguous account filters instead of silently narrowing them. 
Agave rejects dual-key filter objects at deserialization ([`RpcFilterType`](https://github.com/anza-xyz/agave/blob/master/rpc-client-types/src/filter.rs#L12-L18), [`RpcTokenAccountsFilter`](https://github.com/anza-xyz/agave/blob/master/rpc-client-types/src/config.rs#L192-L197)), so this fails fast client-side rather than changing valid behavior (Verified against local validator).

- Add `getTokenAccountsRpcFilter`, used by `getTokenAccountsByOwner`, `getTokenAccountsByDelegate`, and `getParsedTokenAccountsByOwner`.
- `getProgramAccountsRpcFilters` now requires exactly one of `memcmp`/`dataSize` per element. Covers `getProgramAccounts`, `getParsedProgramAccounts`, and `onProgramAccountChange`.

Closes APE-296 (Apex SOLA8-76)


dev-jodee (approved):
Account filter validation: rejects ambiguous dual-key filters (mint+programId, memcmp+dataSize) client-side, matching Agave behavior. Tests cover all cases. CI green.

PR #3924, opened 2026-09-17
fix(web3.js): dispose signature listener even when the callback throws

amilz:
A throwing `onSignature` / `onSignatureWithOptions` callback skipped local cleanup after the terminal status arrived. The server subscription was already auto-disposed, but the registry kept the callback set and subscribed state, so a later identical listener was multiplexed onto a dead subscription and never received the terminal event.

Run local disposal in a finally block around the caller callback.

Fixes DEV-1112 (SOLA8-21)


dev-jodee (approved):
fix(web3.js): dispose signature listener even when callback throws — ensures cleanup in finally block. Handles terminal status + listener cleanup. SOLA8-21, DEV-1112. CI ✓

PR #3925, opened 2026-09-17
fix(web3.js): reject unsupported RPC transaction versions

amilz:
Adds a check to make sure our adapter only allows v 0 or 1 and not some random #. 

Fixes DEV-1114 (Apex SOLA8-45).


dev-jodee (approved):
Transaction version validation. Rejects v2+ to prevent silent field loss with legacy Message. Tests added.

PR #3926, opened 2026-09-17
fix(web3.js): keep the subscription channel open when a notification cannot be decoded

amilz:
A throw while decoding one notification was handled like a socket failure, so the RPC's documented `error: data too large for bs58 encoding` sentinel closed the shared channel and tore down every co-subscription. The runtime now logs and drops only that notification, and the account adapter recognizes the sentinel with an error that points at base64.

Closes DEV-1110 SOLA8-80


dev-jodee (approved):
Notification decode isolation fix. Prevents channel teardown on single decode errors. Tests added.

PR #3928, opened 2026-09-18
fix(web3.js): prefer the nonce confirmation strategy when nonceInfo

amilz:
`sendAndConfirmTransaction` checked `recentBlockhash/lastValidBlockHeight` before `nonceInfo`, so a reused Transaction carrying stale blockheight fields alongside a durable nonce was confirmed against the expired blockhash lifetime and could report `TransactionExpiredBlockheightExceededError` after the nonce-backed transaction had already landed. The compiled and sent message uses the nonce as its lifetime, so confirmation now does too.

Closes DEV-1113 (SOLA8-94)


dev-jodee (approved):
Prefer nonce lifetime in confirmation logic. CI green.

PR #3929, opened 2026-09-18
fix(web3.js): reject non-lookup-table accounts in getAddressLookupTable

amilz:
Check the owner against the Address Lookup Table program before decoding, and reject state whose discriminator is not the initialized LookupTable variant in `AddressLookupTableAccount.deserialize`.

Fixes DEV-1109 (SOLA8-72)


dev-jodee (approved):
Validate LookupTable owner + discriminator. CI green.

PR #3933, opened 2026-09-18
fix(web3.js): reject dataSlice in parsed account queries

amilz:
Kit's `jsonParsed` overloads exclude `dataSlice`, b/c it's excluded in [agave](https://github.com/anza-xyz/agave/blob/73aea15fbaae40aca6ee5ab08c6b2dc020990d93/account-decoder/src/lib.rs#L46). `getParsedAccountInfo` and `getMultipleParsedAccounts` now take config types that omit dataSlice and throw if one is supplied, instead of silently dropping it.


dev-jodee (approved):
reject dataSlice in parsed account queries to match agave behavior. CI green.

PR #3934, opened 2026-09-18
fix(web3.js): populate SendTransactionError logs from preflight errors

amilz:
`Connection.simulateTransaction` read preflight logs from `error.data.logs`, which Kit no longer populates, so `SendTransactionError.logs` and `transactionError.logs` were always undefined. Route the legacy simulation catch path through the shared Kit-aware error detail extractor.

Fixes DEV-1140


dev-jodee (approved):
populate SendTransactionError logs from Kit preflight errors. Fixes DEV-1140. CI green.

PR #3935, opened 2026-09-18
fix(web3.js): resolve send-and-confirm commitment properly

amilz:
first from options, then connection, then finalized

Closes DEV-1128 DEV-1118


dev-jodee (approved):
Commitment fallback chain (options → connection → finalized) prevents undefined behavior. Tests verify call-time capture + mutation safety. ✅

PR #3936, opened 2026-09-18
fix(web3.js): keep sig-status fallback errs from escaping as unhandled rej

amilz:
`confirmTransaction` makes a detached `getSignatureStatus` call after subscription setup as an early-settlement optimization. It had no rejection handler, so a transient transport or JSON-RPC failure surfaced as an unhandled promise rejection, which terminates modern Node processes by default.

The call is only an optimization; the signature subscription and the expiry strategy's final status check already settle the confirmation. Its failure is now caught and ignored.

Test: first `getSignatureStatuses` rejects, subscription never responds, block height passes `lastValidBlockHeight`. Asserts no `unhandledRejection` fires and the transaction still confirms via the expiry path.

Fixes DEV-1132


dev-jodee (approved):
Error handling: catch transient sig-status fallback errors to prevent unhandled rejections. CI green.

PR #3937, opened 2026-09-18
fix(web3.js)!: throw when getMinimumBalanceForRentExemption fails

amilz:
`Connection.getMinimumBalanceForRentExemption` caught RPC and transport failures, logged a warning, and resolved `0n`. `0n` isn't the right min balance, so return error instead.

Fixes DEV-1134


dev-jodee (approved):
Error handling fix: propagate rent-exemption RPC failures instead of returning 0n. CI green.

PR #3944, opened 2026-09-21
ci: split pull request jobs and add a publish dist-tag selector

amilz:
- Run the web3.js suite and the wallet-adapter suite as two parallel jobs so the Playwright and surfpool setup no longer serialises behind the validator tests. 
- Add a dist-tag input to the publish workflow so a stable version can ship under a tag other than latest, e.g. next.

Closes DEV-1081


dev-jodee (approved):
CI: split parallel jobs + dist-tag selector. All checks green.

